Issue #53, opened 2017-12-15
Add debug mode


rubensworks:
* [ ] Add profiler (at bus-level?) so that the execution time of each run/test per actor can be seen.
* [ ] Allow query plan to be dumped. (Probably only after execution, as the `left-deep-smallest` actor results in dynamic plans)
* [x] Make debug mode not disable stack traces.

PR #54, opened 2017-12-15
[WIP] Actor Profiling

rubensworks:
WIP: do not merge.

This adds a simple profiling implementation that measures the total active time of each actor.

The problem is that the _total time_ for each actor is measures, not the _self time_.
For instance, actor1 calling actor2 will also include the time of actor2.
If we want to measure this, we would somehow have to pass the actor stack between all mediator calls, so that we know when each actor is actually active, and when it is not. This would however require us to change the signature the `Mediator.mediate` and `Actor.run` methods to include this stack, not sure if there is a cleaner way...

#53


rubensworks:
Closing this PR, but keeping the branch just in case.

PR #100, opened 2018-03-02
Added stats serializer

mielvds:
This PR adds a serializer for evaluation or debugging purposes. 

TODO: implement the per-result request count, when the right method way to obtain that number is found.

### DIFF
--- a/packages/actor-sparql-serialize-stats/lib/ActorSparqlSerializeStats.ts
+++ b/packages/actor-sparql-serialize-stats/lib/ActorSparqlSerializeStats.ts
@@ -9,14 +9,8 @@ import {Readable} from "stream";
  */
 export class ActorSparqlSerializeStats extends ActorSparqlSerializeFixedMediaTypes {
 
-  private startTime: [number, number] = process.hrtime();
-  private result: number = 1;
-
-  constructor(args: IActorSparqlSerializeStatsArgs) {
+  constructor(args: IActorSparqlSerializeFixedMediaTypesArgs) {
     super(args);
-    if (args.delay) {
-      this.delay = args.delay;
-    }
   }
 
   public async testHandleChecked(action: IActionSparqlSerialize) {
@@ -27,19 +21,19 @@ export class ActorSparqlSerializeStats extends ActorSparqlSerializeFixedMediaTyp
   }
 
   public pushHeader(data: Readable) {
-    const header: string = ['Result', 'Delay (ms)', // 'Requests (#)'
+    const header: string = ['Result', 'Delay (ms)',
     ].join(',');
     data.push(header + '\n');
   }
 
-  public pushStat(data: Readable) {
-    const row: string = [this.result++, this.delay(), // this._fragmentsClient.getRequestCount()
+  public pushStat(data: Readable, startTime: [number, number], result: number) {
+    const row: string = [result, this.delay(startTime),
     ].join(',');
     data.push(row + '\n');
   }
 
-  public pushFooter(data: Readable) {
-    const footer: string = ['TOTAL', this.delay(), // this._fragmentsClient.getRequestCount()
+  public pushFooter(data: Readable, startTime: [number, number]) {
+    const footer: string = ['TOTAL', this.delay(startTime),
     ].join(',');
     data.push(footer + '\n');
     data.push(null);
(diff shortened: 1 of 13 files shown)


packages/actor-sparql-serialize-stats/lib/ActorSparqlSerializeStats.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  public async testHandleChecked(action: IActionSparqlSerialize) {
+    if (['bindings', 'quads'].indexOf(action.type) < 0) {
+      throw new Error('This actor can only handle bindings streams or quad streams.');
+    }
+    return true;
+  }
+
+  public pushHeader(data: Readable) {
+    const header: string = ['Result', 'Delay (ms)', // 'Requests (#)'
```
rubensworks: I guess these commented parts will go once that TODO from the PR description is resolved?

packages/actor-sparql-serialize-stats/test/ActorSparqlSerializeStats-test.ts
```diff
@@ -0,0 +1,103 @@
...
+  });
+
+  describe('The ActorSparqlSerializeStats module', () => {
+    it('should be a function', () => {
+      expect(ActorSparqlSerializeStats).toBeInstanceOf(Function);
+    });
+
+    it('should be a ActorSparqlSerializeStats constructor', () => {
+      expect(new (<any> ActorSparqlSerializeStats)({ name: 'actor', bus })).toBeInstanceOf(ActorSparqlSerializeStats);
+      // expect(new (<any> ActorSparqlSerializeStats)({ name: 'actor', bus })).toBeInstanceOf(ActorSparqlSerialize);
```
rubensworks: This line can go.

packages/actor-sparql-serialize-stats/lib/ActorSparqlSerializeStats.ts
```diff
@@ -0,0 +1,73 @@
...
+import {ActorSparqlSerializeFixedMediaTypes, IActionSparqlSerialize,
+  IActorSparqlSerializeFixedMediaTypesArgs, IActorSparqlSerializeOutput} from "@comunica/bus-sparql-serialize";
+import {Readable} from "stream";
+
+/**
+ * Serializes SPARQL results for testing and debugging.
+ */
+export class ActorSparqlSerializeStats extends ActorSparqlSerializeFixedMediaTypes {
+
+  private startTime: [number, number] = process.hrtime();
```
rubensworks: Not sure what kind of behaviour we want here.

Now, the start time will be the point in time at which the comunica engine is initialized. Every query that is executed in this engine will use this same start time.

Perhaps what we want is the start time at which a query is initialized? In this case, this start time would have to be initialized in the `runHandle ` method.

packages/actor-sparql-serialize-stats/lib/ActorSparqlSerializeStats.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  private delay = function()  {
+    const time: [number, number] = process.hrtime(this.startTime);
+    return time[0] * 1000 + (time[1] / 1000000);
+  };
+
+}
+
+export interface IActorSparqlSerializeStatsArgs extends IActorSparqlSerializeFixedMediaTypesArgs {
```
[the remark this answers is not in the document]
mielvds: If you run a unit test with timings, you need to be able to supply a fake time function, because you cannot predict timings. If any other way, let me know!

mielvds:
@rubensworks I rebased to master, but now `packages/actor-rdf-serialize-jsonld/test/ActorRdfSerializeJsonLd-test.ts`fails

packages/actor-sparql-serialize-stats/lib/ActorSparqlSerializeStats.ts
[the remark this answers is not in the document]
rubensworks: You could make `delay` an actual class method, and mock the class with a custom delay function.
This could be as simple as writing something like `actor.delay = () => ...` in your `beforeEach`.

rubensworks:
@mielvds What is the exact error? Did you try running install in the root again?

The tests should be fine in any case, Travis seems to run just fine.

mielvds:
` /Users/mielvandersande/Desktop/Werk/Technologies/Comunica/comunica/packages/actor-rdf-serialize-jsonld/node_modules/jsonld/lib/jsonld.js:1023
      ...defaults
      ^^^
    
    SyntaxError: Unexpected token ...
`

mielvds:
I'm on 8.2.1, what version should it be?

rubensworks:
Not sure what version it was introduced in, but I'm on 8.5.0, and that works.

mielvds:
Now on 8.9.4 and that works !

rubensworks:
@mielvds Could you add a simple test for that `delay` function as well? It causes coverage to drop. A simple is-number check should be fine I guess.

rubensworks:
@mielvds Would you have time to finish this today? I'd like release 1.0.0 to npm tomorrow and merge this PR before that.

mielvds:
there you go. All set?

rubensworks:
Yep, looks good! Thanks!

PR #106, opened 2018-03-09
Make default BGP actor sort all patterns to reduce HTTP requests

rubensworks:
This should improve overall performance. This should only be merged _after_ the watdiv benchmarks are run again, and results show no performance decrease.


rubensworks:
Moved this to a new actor in #111

PR #111, opened 2018-03-13
Add BGP left-deep-smallest-sort Query Operation Actor

rubensworks:
As originally introduced in #106, and similar to #102, this is a variant of the left-deep-smallest actor with the only difference that it also sorts the remaining patterns by increasing count.

Results show that this actor only in some cases outperforms the default actor, so we should keep the other actor as default. In future work, we should add a mediator that intelligently identifies the cases in which the other actor is sub-optimal, and chooses another actor.

### DIFF
--- a/packages/actor-query-operation-bgp-left-deep-smallest-sort/lib/ActorQueryOperationBgpLeftDeepSmallestSort.ts
+++ b/packages/actor-query-operation-bgp-left-deep-smallest-sort/lib/ActorQueryOperationBgpLeftDeepSmallestSort.ts
@@ -1,0 +1,230 @@
+import {
+  ActorQueryOperation,
+  ActorQueryOperationTypedMediated,
+  Bindings,
+  BindingsStream,
+  IActorQueryOperationOutput,
+  IActorQueryOperationOutputBindings,
+  IActorQueryOperationTypedMediatedArgs,
+} from "@comunica/bus-query-operation";
+import {IActorTest} from "@comunica/core";
+import {EmptyIterator, MultiTransformIterator} from "asynciterator";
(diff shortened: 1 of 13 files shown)


rubensworks:
@joachimvh Nothing has changed since last PR btw

packages/actor-query-operation-bgp-left-deep-smallest-sort/lib/ActorQueryOperationBgpLeftDeepSmallestSort.ts
```diff
@@ -0,0 +1,230 @@
...
+   * @param {{[p: string]: any}[]} otherPatterns The array of optional metadata for the other patterns.
+   * @return {number} The estimated number of total items.
+   */
+  public static estimateCombinedTotalItems(smallestPattern: {[id: string]: any},
+                                           otherPatterns: {[id: string]: any}[]): number {
+    const smallestCount: number = ActorQueryOperationBgpLeftDeepSmallestSort.getTotalItems(smallestPattern);
+    return otherPatterns
+      .map((otherPattern) => smallestCount * ActorQueryOperationBgpLeftDeepSmallestSort.getTotalItems(
+        otherPattern))
+      .reduce((sum, element) => sum + element, 0);
```
joachimvh: This reminds me I also saw this in the other actor. What is the reason for using a sum here?

rubensworks: It's just a heuristic to estimate the upper bound of the total number of items for the whole BGP.

PR #126, opened 2018-05-02
Add support for expressions

wschella:


wschella:
This PR is final now.

rubensworks:
Some preliminary comments on Sparqlee itself:
* *.ts, tsconfig, tslint, test files, .vscode, gulpfile should not be published to npm, have a look at the comunica package.json and .npmignore files.
* It looks like everything in your dist folder is the compiled result, which is fine, but then your lib, util, and index should not be published.
* Remove bluebird as dependency, we removed it from comunica as it significantly slows down stuff.
* Remove rdf-js as dependency, and add @types/rdf-js instead as dev dependency.
* We should probably look into moving the repo to the comunica org here for coherence.

rubensworks:
It looks like queries that produce intermediary type errors crash the engine.

For example, this query fails when `?v5` is a named node:
```
PREFIX  dbp-cat: <http://dbpedia.org/resource/Category/>
PREFIX  dbp-yago: <http://dbpedia.org/class/yago/>
PREFIX  dbp-owl: <http://dbpedia.org/ontology/>
PREFIX  georss: <http://www.georss.org/georss/>
PREFIX  dbpedia: <http://dbpedia.org/resource/>
PREFIX  rdfs: <http://www.w3.org/2000/01/rdf-schema#>
PREFIX  dbp-prop: <http://dbpedia.org/property/>
PREFIX  foaf: <http://xmlns.com/foaf/0.1/>
PREFIX  cwork: <http://www.bbc.co.uk/ontologies/creativework/>

SELECT DISTINCT  ?creativeWork ?v7 ?v4 ?v6 ?v5
FROM <http://graph.version.0>
FROM <http://graph.version.4>
WHERE
  { ?creativeWork
              cwork:mentions  dbpedia:Rutube .
    dbpedia:Rutube
              ?v4             ?v5
    OPTIONAL
      { ?v5  rdfs:label  ?v6
        FILTER ( langMatches(lang(?v6), "en") || ( ! langMatches(lang(?v6), "*") ) )
      }
    FILTER ( langMatches(lang(?v5), "en") || ( ! langMatches(lang(?v5), "*") ) )
    OPTIONAL
      { ?v4  rdfs:label  ?v7 }
  }
```

Comunica (with sparqlee plugged in) errors with:
```
[500] Server error in results: Error: Argument types not valid for operator.
```

This is what the spec says:

> If the evaluation of the expression produces an error, the variable remains unbound for that solution but the query evaluation continues.

This means that for each element in the bindings stream, the expression errors should be caught, and not propagated upwards to the engine.

wschella:
I believe you are referring to this section about the errors?  https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#assignment
If I'm interpreting correctly, this only refers to expressions in SELECT(extend), BIND, and GROUP BY. The errors here are thrown by the FILTER, which doesn't have control over the assignment, although it _could_ replace the original term it is bound to with a 'null', but the FILTER evaluation [spec](https://www.w3.org/TR/rdf-sparql-query/#evaluation) says  'to throw a type error'.

wschella:
- Regarding the packaged files, definitely an error on my part.
- Regarding the extra typescript publishing, again, error on my part (although it is nice for debugging).
- Regarding bluebird, i'll move to native Promises for consistency, but i'd like to benchmark first, cause i'm reading lot's of conflicting empirical evidence.
- Regarding rdfjs, nice find :)
- Regarding moving here: we would get biggest benefits from moving it to the communica repo itself  (for quicker testing and shorter feedback loop), but that seems inappropriate. We should discuss this in person some time.

wschella:
@rubensworks So what is the desired behaviour? I could set all bindings in the mapping to null if it throws an error, but that deviates from the spec (as FILTER never modifies bindings), and the nulls will propagate and cause errors somewhere else.

rubensworks:
I don't seem to find any clear answer in the spec regarding what a _type error_ means.

In [this article](https://www.cray.com/blog/equality-inequality-sparql/), they say that a filter with a type error evaluates to _false_ for that binding. This seems to be consistent with how Virtuoso handles type errors.

So unless anyone is able to find any specific information in the spec on how to handle type errors in `FILTER`s, I propose to let them evaluate to `false`.

wschella:
@rubensworks or @joachimvh, could one of you review to behaviour for the leftjoin? I'm not quite sure i'm interpreting the spec correctly. Errors probably need to be ignored. 
Spec: (https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#defn_algLeftJoin)
Code: https://github.com/comunica/comunica/pull/126/files#diff-2138333d624deed7c04ecb29cba56048R45

rubensworks:
As I understand the spec, errors should indeed be ignored and cause the errorred bindings not being applied, which will lead to incomplete bindings.

wschella:
The wanted behaviour should be implemented now.
Coverage has decreased somewhat, since [this test](https://github.com/comunica/comunica/pull/126/files#diff-2a926ed2484291b1f38c1d9577d35a7cR138) fails to get picked up by coverage report (however i could verify it's execution with logs).

rubensworks:
> this test fails

Try adding a `return` before the last `expect `.

rubensworks:
I will try to review it tomorrow.

One thing I can already comment on:
Now that the development of the filter actor looks about finished and working, it might be good to move this to a standalone actor, something like `actor-query-operation-filter-sparqlee`. This way, we will still be able to use the old filter actor, should we ever need it.

wschella:
That's definitely a good idea. One thing i'm doubting about being _fully_ working is the leftjoin actor. It will probably need to be rewritten and use next() on the right hand stream as well, cause it might be consumed before the promise resolves / rejects.

packages/actor-query-operation-extend/lib/ActorQueryOperationExtend.ts
```diff
@@ -0,0 +1,66 @@
...
+    const evaluator = new AsyncEvaluator(expression);
+    const throwIfHardError = (err: any) => {
+      if (!(err instanceof ExpressionError)) {
+        bindingsStream.emit('error', err);
+        throw err;
+      }
+    };
+
+    const transform = (item: Bindings, next: any) => {
+      evaluator.evaluate(item)
```
rubensworks: I would use `await` and `try/catch` syntax for this block, as this will make it easier to read.

packages/actor-query-operation-extend/lib/ActorQueryOperationExtend.ts
```diff
@@ -0,0 +1,66 @@
...
+          throwIfHardError(err);
+          bindingsStream._push(Bindings({}));
+        })
+        .then(next);
+    };
+
+    const variables = [bindingKey];
+    const bindingsStream = output.bindingsStream.transform<Bindings>({ transform });
+
+    // TODO check if metadata can be kept
```
rubensworks: Sure, shouldn't be a problem.

packages/actor-query-operation-filter-direct/lib/ActorQueryOperationFilterDirect.ts
```diff
@@ -13,30 +15,36 @@ export class ActorQueryOperationFilterDirect extends ActorQueryOperationTypedMed
...
+        bindingsStream.emit('error', err);
+        throw err;
       }
     };
-    const bindingsStream = output.bindingsStream.filter(filter);
 
+    const transform = (item: Bindings, next: any) => {
+      evaluator.evaluateAsEBV(item)
+        // Push the binding if it evaluates true, else don't
+        .then((result) => { if (result) { bindingsStream._push(item); } })
```
rubensworks: Same as above, simplify with syntactic sugar.

packages/actor-query-operation-leftjoin-nestedloop/lib/ActorQueryOperationLeftJoinNestedLoop.ts
```diff
@@ -33,12 +48,14 @@ export class ActorQueryOperationLeftJoinNestedLoop extends ActorQueryOperationTy
         rightStream.on('data', (rightItem) => {
           const join = ActorRdfJoin.join(leftItem, rightItem);
           if (join) {
-            // if (pattern.expression) {
-            //   // TODO: do this once expressions are implemented
-            //   // Values of both streams can be used for this expression
-            // }
-
-            bindingsStream._push(join);
+            // If we want to keep orderening, next() should be used here.
```
rubensworks: Ordering is not required, so this shouldn't be a problem.

But the `next()` call can indeed cause problems. I suggest implementing a CountDownLatch for calling `next()`.

rubensworks:
Closing this in favor of #299.

PR #129, opened 2018-05-07
Convert blank nodes when solving SPARQL patterns

### LINKED ISSUE #104: Convert bnodes to variables internally
At the moment, we have to check _if variable or blank node_ in a lot of places. We could simplify a lot of code by simply converting all blank nodes to variables after the query was parsed.

joachimvh:
Fixes #104

### DIFF
--- a/packages/actor-rdf-resolve-quad-pattern-sparql-json/lib/ActorRdfResolveQuadPatternSparqlJson.ts
+++ b/packages/actor-rdf-resolve-quad-pattern-sparql-json/lib/ActorRdfResolveQuadPatternSparqlJson.ts
@@ -5,7 +5,7 @@ import {ActorRdfResolveQuadPattern, IActionRdfResolveQuadPattern,
 import {Actor, IActorArgs, IActorTest, Mediator} from "@comunica/core";
 import {AsyncIterator, BufferedIterator} from "asynciterator";
 import {PromiseProxyIterator} from "asynciterator-promiseproxy";
-import {blankNode, literal, namedNode, variable} from "rdf-data-model";
+import {blankNode, literal, namedNode, quad, variable} from "rdf-data-model";
 import * as RDF from "rdf-js";
 import {getTerms, getVariables, mapTerms} from "rdf-terms";
 import {Algebra, Factory, toSparql} from "sparqlalgebrajs";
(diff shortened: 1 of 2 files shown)


packages/actor-rdf-resolve-quad-pattern-sparql-json/lib/ActorRdfResolveQuadPatternSparqlJson.ts
```diff
@@ -26,6 +26,51 @@ export class ActorRdfResolveQuadPatternSparqlJson
...
+   */
+  public static replaceBlankNodes(pattern: RDF.Quad): RDF.Quad {
+    const variableNames: string[] = getVariables(getTerms(pattern)).map((v) => v.value);
+    const result: { [id: string]: RDF.Term } = {};
+    // track the names the blank nodes get mapped to (required if the name has to change)
+    const blankMap: { [id: string]: string } = {};
+    let changed = false;
+
+    // for every position, convert to a variable if there is a blank node
+    for (const pos of QUAD_TERM_NAMES) {
```
rubensworks: I would suggest simplifying this loop using [`mapQuadTerms`](https://www.npmjs.com/package/rdf-terms#map-quad-terms), this would also allow us to get rid of the `<any>` cast on the next line.

packages/actor-rdf-resolve-quad-pattern-sparql-json/lib/ActorRdfResolveQuadPatternSparqlJson.ts
```diff
@@ -26,6 +26,51 @@ export class ActorRdfResolveQuadPatternSparqlJson
...
+      const term: RDF.Term = (<any> pattern)[pos];
+      if (term.termType === 'BlankNode') {
+        let name = term.value;
+        if (blankMap[name]) {
+          name = blankMap[name];
+        } else {
+          if (variableNames.indexOf(name) >= 0) {
+            // increase index added to name until we find one that is availabe (2 loops at most)
```
rubensworks: Typo in `availabe `

packages/actor-rdf-resolve-quad-pattern-sparql-json/lib/ActorRdfResolveQuadPatternSparqlJson.ts
```diff
@@ -26,6 +26,51 @@ export class ActorRdfResolveQuadPatternSparqlJson
...
+
+    // for every position, convert to a variable if there is a blank node
+    for (const pos of QUAD_TERM_NAMES) {
+      const term: RDF.Term = (<any> pattern)[pos];
+      if (term.termType === 'BlankNode') {
+        let name = term.value;
+        if (blankMap[name]) {
+          name = blankMap[name];
+        } else {
+          if (variableNames.indexOf(name) >= 0) {
```
rubensworks: Using a hash instead of an array here would make this slightly faster.

packages/actor-rdf-resolve-quad-pattern-sparql-json/lib/ActorRdfResolveQuadPatternSparqlJson.ts
joachimvh: actually didn't know that existed :D

packages/actor-rdf-resolve-quad-pattern-sparql-json/lib/ActorRdfResolveQuadPatternSparqlJson.ts
joachimvh: I'm actually wondering know if there might be a reason why the array could be faster (since it can contain at most 2 values) :D

rubensworks: Oh, you're right. Probably doesn't matter then.

packages/actor-rdf-resolve-quad-pattern-sparql-json/lib/ActorRdfResolveQuadPatternSparqlJson.ts
joachimvh: As far as I can see that function only exists in the documentation but not actually in the library

rubensworks: README seems to be incorrect. The actual name is `mapTerms` (will fix the README right away!)

joachimvh:
Fixed issues

PR #131, opened 2018-05-24
Add GraphQL query parser and tree-based result serializer

rubensworks:
~~Still requires tests and code cleanup, do not review or merge yet.~~

Depends on #173.

This allows GraphQL queries such as the following to be defined:
(Get all bands MJ has written a song for)
```
{
  label
  writer(label_en: \"Michael Jackson\")
  artist {
    label
  }
}
```

With the following output:
```json
[
  {
    "artist": {
      "label": "Barry Gibb"
    },
    "writer": "http://dbpedia.org/resource/Michael_Jackson",
    "label": "All in Your Name"
  }
  {
    "label": "A Place with No Name",
    "artist": {
      "label": "Michael Jackson"
    },
    "writer": "http://dbpedia.org/resource/Michael_Jackson"
  },
  ...
]
```


rubensworks:
Ok, this is done and can be reviewed.

Depends on #173.

packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
```diff
@@ -94,7 +96,8 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
     const combinationPromise = this.mediatorContextPreprocess.mediate({ context });
 
     // Parse query
-    let operation: Algebra.Operation = (await this.mediatorSparqlParse.mediate({ query })).operation;
+    let operation: Algebra.Operation = (await this.mediatorSparqlParse.mediate(
+      { query, queryFormat: context && context.queryFormat ? context.queryFormat : 'sparql', context })).operation;
```
joachimvh: Would be interesting if the query format could be automatically detected. I.e., the test function of both sparql and graphql return whether it can parse (or both just try)

Issue #125, opened 2018-04-16
Source type discovery


rubensworks:
We should have a way to autodiscovery the source type of a URI. Because currently we need to provide the source type during query input. Ideally, this should not be necessary anymore.

* bus-context-preprocesing
* actor-context-preprocesing-rdf-source-identifier
* bus-rdf-source-identifier
* actor-rdf-source-identifier-sparql: Has /sparql
* actor-rdf-source-identifier-hypermedia-qpf: Has QPF
* actor-rdf-source-identifier-file: Fallback

PR #134, opened 2018-05-29
Feature/source identification

rubensworks:
As required by #125, this allows sources to be tagged with the `auto` source type, after which the type will be (heuristically) determined.
As this adds a small overhead, it is still possible to tag the sources to a specific type.

This also makes it possible to execute the following command (federated query over TPF, SPARQL, file) more easily:
```
https://fragments.linkedsoftwaredependencies.org/npm http://dbpedia.restdesc.org/ https://ruben.verborgh.org/profile/#me "SELECT * WHERE { ?s a ?o } LIMIT 100"
```


packages/actor-rdf-source-identifier-sparql/lib/ActorRdfSourceIdentifierSparql.ts
```diff
@@ -0,0 +1,43 @@
...
+  public async test(action: IActionRdfSourceIdentifier): Promise<IActorTest> {
+    if (!action.sourceValue.startsWith('http')) {
+      throw new Error(`Actor ${this.name} can only detect SPARQL endpoints hosted via HTTP(S).`);
+    }
+    const url: string = action.sourceValue + '?query=' + encodeURIComponent('ASK { ?s a ?o }');
+    const headers: Headers = new Headers();
+    headers.append('Accept', 'application/sparql-results+json');
+    const httpAction: IActionHttp = { input: url, init: { headers, method: 'HEAD' } };
+    const httpResponse: IActorHttpOutput = await this.mediatorHttp.mediate(httpAction);
+    if (!httpResponse.ok || httpResponse.headers.get('Content-Type').indexOf('application/sparql-results+json') < 0) {
```
joachimvh: Do all sparql endpoints support sparql json?

packages/actor-init-sparql/config/config-default.json
```diff
@@ -87,6 +94,11 @@
             "@type": "MediatorCombineUnion",
             "cc:Mediator/bus": { "@id": "cbss:Bus/SparqlSerialize" },
             "cmcu:Mediator/CombineUnion/field": "mediaTypes"
+          },
+          "mediatorContextPreprocess": {
```
joachimvh: Is this not supposed to be assigned to an actor somewhere? Or are there some defaults I'm missing that do this?

packages/mediator-combine-pipeline/README.md
```diff
@@ -0,0 +1,18 @@
+# Comunica Mediator Sequential
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Fmediator-sequential.svg)](https://www.npmjs.com/package/@comunica/mediator-sequential)
+
+A comunica mediator that goes over all actors in sequence and forwards I/O.
+This required the action input and the actor output to be of the same type.
```
joachimvh: typo: requires

packages/actor-init-sparql/config/config-default.json
```diff
@@ -87,6 +94,11 @@
...
+            "@id": "ex:mediatorContextPreprocess",
+            "@type": "MediatorCombinePipeline",
```
joachimvh: Why is this a pipeline mediator? Doesn't each of the preprocess actors simply handle a single source and only returns true on the test if it's the applicable actor? This actor would actually make more sense to me if the thing they have in their `test` function now would be in their `run` function. What am I missing?

packages/actor-http-native/lib/ActorHttpNative.ts
```diff
@@ -53,9 +53,15 @@ export class ActorHttpNative extends ActorHttp {
           httpResponse = null;
           reject(e);
         });
+        // Avoid memory leak on HEAD requests
+        if (options.method === 'HEAD') {
+          httpResponse.destroy();
```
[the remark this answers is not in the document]
rubensworks: Yep, HEAD requests caused the application to hang indefinitely after everything was done.

packages/actor-rdf-source-identifier-file-content-type/lib/ActorRdfSourceIdentifierFileContentType.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  public async test(action: IActionRdfSourceIdentifier): Promise<IActorTest> {
+    if (!action.sourceValue.startsWith('http')) {
+      throw new Error(`Actor ${this.name} can only detect files hosted via HTTP(S).`);
+    }
+    const headers: Headers = new Headers();
+    headers.append('Accept', this.allowedMediaTypes.join(';'));
+    const httpAction: IActionHttp = { input: action.sourceValue, init: { headers, method: 'HEAD' } };
+    const httpResponse: IActorHttpOutput = await this.mediatorHttp.mediate(httpAction);
```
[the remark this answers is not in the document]
rubensworks: I thought so as well initially. Most of that code used to be in `run`, but I changed it because of the following reasons:

1. In a way, you could see the real action of this actor being the labelling of the source. Where the `test` is then checking if the actor is able to label the input, and reject otherwise. If all of this logic would be inside the `run`, then we would have to allow `run` to reject as well, which is something that the mediator is not able to handle (by default).
2. It was difficult to come up with a mediator for this. We would have to create a new mediator that `runs` all actors, and only takes the output of the one that does not reject. (Which makes the behaviour of `run` closely resemble the intention of `test`)

packages/actor-rdf-source-identifier-hypermedia-qpf/lib/ActorRdfSourceIdentifierHypermediaQpf.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    const httpAction: IActionHttp = { input: action.sourceValue, init: { headers } };
+    const httpResponse: IActorHttpOutput = await this.mediatorHttp.mediate(httpAction);
+    if (httpResponse.ok) {
+      const body = (await require('stream-to-string')(httpResponse.body));
+
+      // Check if body contains all required things
+      let valid = true;
+      for (const line of this.toContain) {
+        if (body.indexOf(line) < 0) {
```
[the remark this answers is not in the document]
rubensworks: It's meant to be quick and heuristic. I'd say, we leave it like this for now, and improve if this causes any problems in the future. I'll update the READMEs to clarify that these are heuristical.

packages/actor-rdf-source-identifier-sparql/lib/ActorRdfSourceIdentifierSparql.ts
rubensworks: This is not a requirement in the spec. But in practise that's the go-to format. (And again, it's heuristic :-) )

packages/actor-init-sparql/config/config-default.json
rubensworks: Yes, this is the SPARQL init actor.

packages/actor-init-sparql/config/config-default.json
rubensworks: You're probably confusing things. Here's an ASCII figure explaining the structure:

```
Actor: SPARQL init
  |
  \ Mediator (pipeline): Context Preprocess
    | Actor: Context Preprocess - RDF Source Identifier
    \ Mediator (race): RDF Source Identifier
      | Actor: RDF Source Identifier: File
      | Actor: RDF Source Identifier: SPARQL
      | Actor: RDF Source Identifier: Hypermedia
```

PR #135, opened 2018-05-30
Use order for source identifiers

joachimvh:
This makes sure that sources that match both files and hypermedia
are always recognized as hypermedia


packages/actor-rdf-source-identifier-file-content-type/lib/ActorRdfSourceIdentifierFileContentType.ts
```diff
@@ -28,7 +28,7 @@ export class ActorRdfSourceIdentifierFileContentType extends ActorRdfSourceIdent
         .get('Content-Type').indexOf(mediaType) >= 0)) {
       throw new Error(`${action.sourceValue} is not an RDF file of valid content type: ${this.allowedMediaTypes}`);
     }
-    return true;
+    return { order: 2 };
```
rubensworks: I would not hardcode these orders, but parameterize them, so that they can be changed in the config file. (I would add them as default values though)

Furthermore, in order to remain consistent with the parsers and serializers, it might be better to use floats in the `[0,1]` range.

joachimvh:
@rubensworks added the mediator type

PR #139, opened 2018-06-05
Feature/file

joachimvh:
Supports having a local file as a source


packages/actor-rdf-dereference-file/components/Actor/RdfDereference/File.jsonld
```diff
@@ -0,0 +1,64 @@
...
+          "comment": "The RDF parse mediator",
+          "required": true,
+          "unique": true
+        },
+        {
+          "@id": "cardf:mediaMappingEntry",
+          "comment": "Maps an extension to a file type",
+          "required": true,
+          "default": [
+            { "cardf:extension": ".ttl",      "cardf:mediaType": "text/turtle" },
```
rubensworks: Was there a particular reason to include the dots in the extension? If not, I'd remove them.

packages/actor-rdf-dereference-file/components/Actor/RdfDereference/File.jsonld
```diff
@@ -0,0 +1,64 @@
...
+            { "cardf:extension": ".turtle",   "cardf:mediaType": "text/turtle" },
+            { "cardf:extension": ".nt",       "cardf:mediaType": "text/turtle" },
```
rubensworks: Media type of this and all below are wrong. (I know that N3 can handle all of them anyways, but this is not necessarily like this for all possible parsers)

packages/actor-rdf-dereference-file/lib/ActorRdfDereferenceFile.ts
```diff
@@ -0,0 +1,78 @@
...
+  public readonly mediaMappings: { [id: string]: string };
+
+  constructor(args: IActorRdfDereferenceFileArgs) {
+    super(args);
+  }
+
+  public async test(action: IActionRdfDereference): Promise<IActorTest> {
+    if (!action.url.startsWith("file:///")) {
+      // check if the url contains a (relative) path
+      const error = await new Promise((resolve, reject) => fs.access(action.url, fs.constants.F_OK, (err) => {
```
rubensworks: This block can be made cleaner using `Util.promisify`.

packages/actor-rdf-dereference-file/lib/ActorRdfDereferenceFile.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+    return true;
+  }
+
+  public async run(action: IActionRdfDereference): Promise<IActorRdfDereferenceOutput> {
+    let mediaType = action.mediaType;
+
+    // deduce media type from file extension if possible
+    if (!mediaType) {
+      mediaType = this.mediaMappings[path.extname(action.url)];
```
rubensworks: Should we throw an error if the media type is still undefined after this?

packages/actor-rdf-dereference-file/components/Actor/RdfDereference/File.jsonld
joachimvh: Because `path.extname` returns the dot, so to be consistent with that. But I don't really care either way.

packages/actor-rdf-dereference-file/lib/ActorRdfDereferenceFile.ts
joachimvh: handleMediaType is optional for IActionRootRdfParse, which is why I didn't throw an error if there is none (although I'm assuming it will probably fail further down the stream).

packages/actor-rdf-dereference-file/components/Actor/RdfDereference/File.jsonld
rubensworks: They apparently do that for distinguishing between cases where there is an extension, and where there is none. (To avoid `undefined` I guess)

In any case, this is not a concern in our case, so we can remove the dot I guess. Saves us some bytes :p

packages/actor-rdf-dereference-file/lib/ActorRdfDereferenceFile.ts
rubensworks: As long as the later error is sufficiently descriptive, then I'm fine with it :-)

joachimvh:
@rubensworks some changes

packages/actor-rdf-dereference-file/lib/ActorRdfDereferenceFile.ts
```diff
@@ -0,0 +1,80 @@
...
+  public readonly mediaMappings: { [id: string]: string };
+
+  constructor(args: IActorRdfDereferenceFileArgs) {
+    super(args);
+  }
+
+  public async test(action: IActionRdfDereference): Promise<IActorTest> {
+    if (!action.url.startsWith("file:///")) {
+      // check if the url contains a (relative) path
+      return promisify(fs.access)(action.url, fs.constants.F_OK)
```
rubensworks: You're already in an async context, why not just use `await`, and `try-catch` for handling the error?

Issue #161, opened 2018-06-28
Prepared statements


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request

____
#### Description:

We should provide prepared statement functionality.

Similar as in ARQ, we can allow _initial bindings_ to be passed via the context file, such that certain variables in the query can be bound immediately.

PR #165, opened 2018-07-02
Allow initialBindings to be passed to query context

rubensworks:
This adds support for prepared statements (#161).

### DIFF
--- a/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
+++ b/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
@@ -61,9 +61,9 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
         copiedOperation[key] = operation[key];
       }
 
-      if (operation.type === 'pattern' || operation.type === 'path') {
+      if (operation.type === Algebra.types.PATTERN || operation.type === Algebra.types.PATH) {
         for (const quadTerm of QUAD_TERM_NAMES) {
-          if (!(operation.type === 'path' && quadTerm === 'predicate')) {
+          if (!(operation.type === Algebra.types.PATH && quadTerm === 'predicate')) {
             const term: RDF.Term = operation[quadTerm];
             if (term.termType === 'Variable') {
               const termString: string = termToString(term);
--- a/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
+++ b/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
@@ -1,17 +1,22 @@
 import {IActionContextPreprocess, IActorContextPreprocessOutput} from "@comunica/bus-context-preprocess";
 import {ActorInit, IActionInit, IActorOutputInit} from "@comunica/bus-init";
-import {IActionQueryOperation, IActorQueryOperationOutput} from "@comunica/bus-query-operation";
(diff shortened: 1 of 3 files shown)


packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
```diff
@@ -31,6 +36,49 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
...
+  public static applyInitialBindings(operation: Algebra.Operation, initialBindings: Bindings): Algebra.Operation {
+    const copiedOperation: Algebra.Operation = <any> {};
+    for (const key of Object.keys(operation)) {
+      if (Array.isArray(operation[key])) {
+        if (key === 'variables') {
+          copiedOperation[key] = operation[key].filter(
+            (variable: RDF.Variable) => initialBindings.keySeq().indexOf(termToString(variable)) < 0);
+        } else {
+          copiedOperation[key] = operation[key].map(
+            (subOperation: Algebra.Operation) => ActorInitSparql.applyInitialBindings(subOperation, initialBindings));
```
joachimvh: It says `subOperation: Algebra.Operation` but this is not always the case though. There are also operations that have an array of terms as value. Not sure if any of those operations can actually appear at this point though (operations as Describe, From, Nps (part of path expressions) and Values)

packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
```diff
@@ -31,6 +36,49 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
...
+        }
+      } else if (ActorInitSparql.ALGEBRA_TYPES.indexOf(operation[key].type) >= 0) {
+        copiedOperation[key] = ActorInitSparql.applyInitialBindings(operation[key], initialBindings);
+      } else {
+        copiedOperation[key] = operation[key];
+      }
+
+      if (operation.type === 'pattern' || operation.type === 'path') {
```
joachimvh: You should totally use Algebra.types.PATTERN and .PATH to make it look cooler (and more future proof) ;)

packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
```diff
@@ -31,6 +36,49 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
...
+   * @param {Bindings} initialBindings Bindings to apply.
+   * @return {Operation} A copy of the given operation where all given bindings are applied.
+   */
+  public static applyInitialBindings(operation: Algebra.Operation, initialBindings: Bindings): Algebra.Operation {
+    const copiedOperation: Algebra.Operation = <any> {};
+    for (const key of Object.keys(operation)) {
+      if (Array.isArray(operation[key])) {
+        if (key === 'variables') {
+          copiedOperation[key] = operation[key].filter(
+            (variable: RDF.Variable) => initialBindings.keySeq().indexOf(termToString(variable)) < 0);
```
joachimvh: This can change the meaning of a query in case of Group By and Values queries though.

rubensworks: Indeed, that was intentional.
But I'm open to other suggestions.
Do you think it would make sense to either not remove the variables in those cases, or throw an error?

PR #180, opened 2018-07-20
Enhancement/hoist

### LINKED ISSUE #72: Deduplicate webpacked dependencies
Webpack support was added in #64.

Webpack will by default deduplicate dependencies, but because using a (Lerna) monorepo, Webpack can sometimes still duplicate dependencies (https://github.com/webpack/webpack/issues/5593).

There _is_ a plugin that is supposed to fix this (https://github.com/RoboBurned/dedup-resolve-webpack-plugin), but it can cause issues in dependency resolution when packaged in a monorepo.
(_Note: line 44 in the plugin should be replaced with `if (fs.realpathSync(request.path) !== fs.realpathSync(cacheEntry.path)) {`, and the failing modules should be blacklisted via the plugin._

In practise, when creating a web bundle in a regular installation (non-monorepo), this issue shouldn't occur, so let's test this when we get there.

rubensworks:
Depends on #173.

This enables Lerna's [hoisting](https://github.com/lerna/lerna/blob/master/doc/hoist.md) feature.
This will move all shared dependencies to the root of the monorepo.
This resolves the issue where webpacked builds could contain several identical copies of the same module.

This should also make WebStorm lag slightly less.

Closes #72

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,12 @@
 sudo: required
 services:
  - docker
+addons:
+  apt:
+    sources:
+     - ubuntu-toolchain-r-test
+    packages:
+     - g++-4.9
 language: node_js
 node_js:
  - "8"
@@ -8,6 +14,7 @@ node_js:
  - "10"
  - "node"
 before_install: yarn global add greenkeeper-lockfile@1
+install: CC=gcc-4.9 CXX=g++-4.9 yarn install --pure-lockfile
 before_script: greenkeeper-lockfile-update
 after_script: greenkeeper-lockfile-upload
 script:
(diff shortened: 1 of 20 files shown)


rubensworks:
> actor-init-rdf-parse requires a dev dependency on n3 for its test

Actually, it doesn't. (And it will be found anyways thanks to hoisting) The test includes it to fix (in a hacky way) an issue with imports.

This is however fixed in the newest n3 version, so once we update to that (see feature/rdfjs-n3 branch), this require can be removed.

joachimvh:
My tests failed there until I added the dependency so there is still some sort of problem then (until the new version is there I guess)

joachimvh:
This seems to work when using yarn (besides the whole hdt thing)

rubensworks:
For reference: HDT should be moved to a separate repo: #186.

Issue #109, opened 2018-03-12
Providing input data to Actors


joachimvh:
Some context: Miel wants to make a Memento actor. All HTTP calls that get executed should be sent to that actor, who then forwards the request to an actual HTTP actor after doing some changes to the request. This itself is actually quite easy in Comunica: there just needs to be a Bus with this Memento actor in between actors that request an HTTP query and those that resolve one.

The problem is that this Memento actor needs some metadata (a date) that is provided at the entry level (e.g., through the command line input). The question now is how to best make sure that date arrives at the Memento actor, preferably in such a way that it arrives together with the actual HTTP requests to make sure the date isn't actually used for requests that should not need this.

Some random ideas:

1. The one that was discussed the most. Create a general (read-only) "context" object that gets passed to all actors in the call chain. The init actor that parses the command line arguments would then fill in this object. This would require a change to the input interface of the actors to also take the context as input (which is easy) and a change to how all actors currently call their mediator to make sure they pass the context object down the chain (which would be more work). A potential problem is that somehow the init actor needs to use the same keys when writing the context object as the memento actor will use to read out the metadata, so these need to be synced somehow.

2. (The other ideas are things that were less discussed and maybe just mentioned). Have some sort of shared memory that all actors can access. In this case the context object would be stored there instead of being passed to all the actors. One problem here is that the metadata doesn't arrive at the same time as the query. It might also decrease the independency of all actors?

3. Have some sort of Bus where this context gets sent on separately. That way all actors that are interested can subscribe to this Bus. This also has the disadvantage of not having the metadata arrive at the same time as the query.

4. ... (I'm sure there are still many potential solutions. Like overloading the run function of the HTTP actors and other weird stuff).

rubensworks:
Option 2 reminds me of the blackboard pattern, which is also often used in cases like these, so we could probably learn some things from that.

rubensworks:
It may also be worth to mention the other option of using the query operation context for the most part, and creating/adapting the deeper actor action interfaces that exist between the QPF actor and the HTTP actor. I'll look into this option a bit more in any case, it may be simpler than it sounds.

RubenVerborgh:
The context might conceptually solve it, but what obligation is there for actors to look at the context? For instance, it could be cheaper to ignore the context, so actors would explicitly need to acknowledge that they reckon with it.

I'm thinking of a kind of decorator pattern, where you have a Memento actor as follows:
0. receives a job X with constraint "datetime = t"
[...]

PR #183, opened 2018-07-24
[WIP] Feature/action context

rubensworks:
Implements the generalization of a shared context #109.

**Do not merge or review yet.**

Context keys are now namespaced, for example:
`{ '@comunica/bus-rdf-resolve-quad-pattern:sources': [{ type: 'hypermedia', value: ... }] }`

The problem with this is that it becomes less convenient for programmatic API users to define sources.

A few possible solutions:
1. Add shorthand aliases for config entries. Downside is that there can be clashing among aliases now.
2. Developers must import and use string constant from the modules that define these keys. (we already define these keys) Downside is that context can not be constructed inline anymore.
3. Add convenience constructors for contexts, such as `newSourcesContext([...])`. Downside is that it becomes more complicated when your context consists of several other stuff as well.
4. Leave it as-is, developers now have to use longer key strings.

@joachimvh What is your opinion on these solutions?.


joachimvh:
I prefer 2 or 4. But 2 prevents defining contexts in config files, so I would keep it at 4. The majority of the actors doesn't make use of the context anyway, and if someone creates an actor that makes a lot of use of the context they can define their own variables to make for easier development.

rubensworks:
Ok, just added backwards compatibility. Ready for review now.

packages/actor-abstract-mediatyped/lib/ActorAbstractMediaTyped.ts
```diff
@@ -46,32 +46,36 @@ export abstract class ActorAbstractMediaTyped<HI, HT, HO>
    *
    * @param {HI} action The handle action to test.
    * @param {string} mediaType The media type to test.
+   * @param {ActionContext} context An optional context.
    * @return {Promise<T>} A promise that resolves to the handle test result.
    */
-  public abstract async testHandle(action: HI, mediaType: string): Promise<HT>;
+  public abstract async testHandle(action: HI, mediaType: string, context?: ActionContext): Promise<HT>;
```
joachimvh: Should the context parameter be optional? This might encourage developers to not pass on the context when they receive it.

rubensworks: Yes, that makes sense. The context can still be null, but all remove the `?`everywhere to make it clear that the context must be passed along.

PR #188, opened 2018-07-30
Feature/sources stream

rubensworks:
As discussed offline, this makes it so that sources are represented in a stream-like fashion instead of an array. For this the [AsyncReiterable](https://www.npmjs.com/package/asyncreiterable) datastructure was introduced.

This is needed for adding additional sources at query-time.

More information can be found in the commits.

### DIFF
--- a/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
+++ b/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
@@ -109,7 +109,7 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
       delete context.queryFormat;
     }
     if (Array.isArray(context[KEY_CONTEXT_SOURCES])) {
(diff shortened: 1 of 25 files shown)


packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
```diff
@@ -107,6 +108,9 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
       context[KEY_CONTEXT_QUERYFORMAT] = context.queryFormat;
       delete context.queryFormat;
     }
+    if (Array.isArray(context[KEY_CONTEXT_SOURCES])) {
+      context[KEY_CONTEXT_SOURCES] = AsyncReiterableArray.forFixedData(context[KEY_CONTEXT_SOURCES]);
```
joachimvh: I think the standard is to use `from`, i.e., `fromFixedData`, when doing conversions from one datatype to another like this. (Node has `Array.from()` for example)

rubensworks: Sure, I can change that :-)

PR #189, opened 2018-07-31
Add ActionObserver

rubensworks:
This allows all Actor#run invocations to be passively observed.

### DIFF
--- a/packages/core/lib/Bus.ts
+++ b/packages/core/lib/Bus.ts
@@ -1,3 +1,4 @@
+import {ActionObserver} from "./ActionObserver";
 import {Actor, IAction, IActorOutput, IActorTest} from "./Actor";
 
 /**
@@ -20,6 +21,7 @@ export class Bus<A extends Actor<I, T, O>, I extends IAction, T extends IActorTe
 
   public readonly name: string;
   protected readonly actors: A[] = [];
+  protected readonly observers: ActionObserver<I, O>[] = [];
 
   /**
    * All enumerable properties from the `args` object are inherited to this bus.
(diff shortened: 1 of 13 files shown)


packages/core/lib/Bus.ts
```diff
@@ -77,6 +109,19 @@ export class Bus<A extends Actor<I, T, O>, I extends IAction, T extends IActorTe
...
+  /**
+   * Invoked when an action was run by an actor.
+   *
+   * @param actor               The action on which the {@link Actor#run} method was invoked.
+   * @param {I}          action The original action input.
+   * @param {Promise<O>} output A promise resolving to the final action output.
+   */
+  public onRun(actor: Actor<I, T, O>, action: I, output: Promise<O>): void {
+    for (const observer of this.observers) {
+      observer.onRun(actor, action, output);
```
joachimvh: It is possible that an observer has an async onRun function. In that case, is it not possible that a stream in the output object is already drained before it reaches the observer (should the calling actor receive it first)?

rubensworks:
> it should probably at least be noted somewhere that observers should not edit the output.

Sure, I can clarify that in the jsdoc. It's probably overkill to make action I/O immutable as well.

joachimvh:
Yea, making it immutable doesn't help since it could still contain objects that can be changed.

packages/core/lib/Bus.ts
rubensworks: In the case of streams, I would say that the observer implementor is responsible for making sure that it doesn't mess with the actor's handling. For non-streams, I don't think it's a problem if onRun is async.

joachimvh: I meant the reverse. The actor might have already read some things from the stream before the observer actually gets access to the stream, causing it to not see everything that was in the output

Issue #79, opened 2018-01-31
Memento support


rubensworks:
Just realized this, but we don't have Memento time conneg support planned yet.

Ideally, this should also be implemented before we release.

@mielvds Are you up for this?

rubensworks:
Depends on #109.

Issue #116, opened 2018-03-19
Add usage information in all package READMEs


rubensworks:
All comunica packages should have a basic usage guide, including details such as which JSON parameters can be configured, and which classes can be used/extended.

PR #195, opened 2018-08-03
Feature/memento

mielvds:
This implements the Memento protocol (#79) and adds an `-d` parameter to the `actor-init-sparql`.

### DIFF
--- a/packages/actor-http-memento/lib/ActorHttpMemento.ts
+++ b/packages/actor-http-memento/lib/ActorHttpMemento.ts
@@ -1,46 +1,62 @@
-import {ActorHttp, IActionHttp, IActorHttpOutput} from "@comunica/bus-http";
-import {IActorArgs, IActorTest, Mediator} from "@comunica/core";
-import {Request} from "node-fetch";
+import { ActorHttp, IActionHttp, IActorHttpOutput } from "@comunica/bus-http";
+import { ActionContext, IActorArgs, IActorTest, Mediator } from "@comunica/core";
+import "isomorphic-fetch";
+import * as parseLink from 'parse-link-header';
 
 /**
  * A comunica Memento Http Actor.
  */
 export class ActorHttpMemento extends ActorHttp {
 
-  public readonly mediator: Mediator<ActorHttp,
-  IActionHttp, IActorTest, IActorHttpOutput>;
+  public readonly mediatorHttp: Mediator<ActorHttp,
+    IActionHttp, IActorTest, IActorHttpOutput>;
 
-  constructor(args: IActorArgs<IActionHttp, IActorTest, IActorHttpOutput>) {
+  constructor(args: IActorHttpMementoArgs) {
     super(args);
   }
 
   public async test(action: IActionHttp): Promise<IActorTest> {
-    return true; // TODO implement
+    if (!(action.context && action.context.has('datetime') && action.context.get('datetime') instanceof Date)) {
+      throw new Error('This actor only handles request with a set valid datetime.');
+    }
+    return true;
   }
 
   public async run(action: IActionHttp): Promise<IActorHttpOutput> {
-
-    if ((<any> action.input).headers) {
-      options.url = (<Request> action.input).headers;
-      Object.assign(options, action.input);
-    } else {
-      options.url = action.input;
-    }
+    const datetime: Date = action.context && action.context.get('datetime');
 
     // 1. Create ActionHttp
-    const httpAction: IActionHttp = {
-      input: new Request()
-    };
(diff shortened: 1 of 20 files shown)


rubensworks:
Something seems to be wrong with the lockfile. Maybe try forcefully checking the lockfile out from master and re-running `yarn install` will fix it.

Also, be sure to sign the CLA.

packages/actor-http-memento/lib/ActorHttpMemento.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (action.init && new Headers(action.init.headers || {}).has('accept-datetime')) {
+      throw new Error('The request already has a set datetime.');
+    }
+    return true;
+  }
+
+  public async run(action: IActionHttp): Promise<IActorHttpOutput> {
+    const datetime: Date = action.context && action.context.get('datetime');
+
+    // 1. Create ActionHttp
```
rubensworks: I would remove these 1., 2., 3. 4., 5. comments, they don't really add any information next to the code IMO.

packages/actor-http-native/lib/ActorHttpNative.ts
```diff
@@ -63,13 +63,15 @@ export class ActorHttpNative extends ActorHttp {
...
             const result = <IActorHttpOutput> {
               body: httpResponse,
-              headers: new Headers(httpResponse.headers),
+              headers,
               ok: httpResponse.statusCode < 300,
               redirected: options.url !== httpResponse.responseUrl,
               status: httpResponse.statusCode,
-              url: httpResponse.responseUrl,
+              // when the content came from another resource because of conneg, let Content-Location deliver the url
+              url: headers.has('content-location') ? headers.get('content-location') : httpResponse.responseUrl,
```
rubensworks: Can you also add a dedicated unit test for this to the HTTP-native actor?

mielvds:
@rubensworks I've tried to cover the requested changes as much as possible, but for the unit tests it takes me to long to figure out the 100% coverage in `actor-init-sparql` and `actor-http-memento`. I suggest you have a look?

rubensworks:
According to the coveralls logs this is the line that is missed: https://coveralls.io/builds/18360561/source?filename=packages%2Factor-init-sparql%2Flib%2FActorInitSparql.ts#L105

And here is the coverage file for the memento actor: https://coveralls.io/builds/18360561/source?filename=packages%2Factor-http-memento%2Flib%2FActorHttpMemento.ts#L23

(You should be able to see the non-covered lines when running `yarn test` as well)

mielvds:
Yeah of course, I got that. Let me be more specific:
- how to cover the branches?
- I'm not sure how to inspect the context when testing `run`from `actor-init-sparql`. Is there a trick I can use or should I change something to the class?

rubensworks:
So for actor-init-sparql, you can add a test that executes the command with the `-d` options similar to this test: https://github.com/comunica/comunica/blob/master/packages/actor-init-sparql/test/ActorInitSparql-test.ts#L218-L227

For the memento actor, you should add a test that _has_ a context, but _does not have_ an `init` or `accept-datetime` header, and ensure that it rejects.

mielvds:
it should be fine now

packages/actor-init-sparql/lib/ActorInitSparql.ts
```diff
@@ -99,6 +100,11 @@ Options:
       context.queryFormat = args.i;
     }
 
+    // Define the datetime
+    if (args.d) {
+      context.datetime = new Date(args.d);
```
rubensworks: I just noticed this, but this has to be set via a namespaced context key, just like `KEY_CONTEXT_SOURCES` below. You could add something like `KEY_CONTEXT_DATETIME` to the memento actor class.

mielvds: fixed.

rubensworks: `ActorHttpMemento` also has to be updated to use this key, it will always fail otherwise, as well as the unit tests.

mielvds: oops, amateur move. Now fixed

pietercolpaert:
I wonder whether we should add more documentation in the README.md about how this actor could be invoked from different query languages, as a library or as a command line interface. Right now, only the pull request mentions a `-d` parameter

rubensworks:
@pietercolpaert I think this is a problem in general for all actors. At some point we should start working on some detailed usage documentation for each actor, see #116.

Issue #114, opened 2018-03-14
Logging


rubensworks:
We should support logging like in the old ldf-client, for logging things like the executed HTTP requests. This is required for https://github.com/comunica/jQuery-Widget.js/issues/2.

Potentially related to #109.

rubensworks:
While the comunica prototype initially uses a [winston](https://www.npmjs.com/package/winston) logger, it felt too limited, so I suggest not using it again.

I think a more configurable logging framework (like [log4js](https://www.npmjs.com/package/log4js)) would be ideal, as it supports different logging levels and categories.

One thing to keep in mind is performance. So we might want to scope logging calls behind certain flags.

rubensworks:
This might be a good lib we can use: https://github.com/guigrpa/storyboard

rubensworks:
Depends on #109.

rubensworks:
[Bunyan](https://github.com/trentm/node-bunyan) seems to be good and flexible enough for our case, and there also exists an isomorphic version (https://www.npmjs.com/package/iso-bunyan-stream).

#204 has to be fixed before this can be implemented, as the initialize methods seems to be the best way to pass around logger instances. (I want to avoid `static` solutions)

PR #208, opened 2018-08-13
Feature/logging

rubensworks:
This implements logging as required by #114.

Logging is available everywhere:

* By default, comunica will use a void logger.
* When on the command-line, it will use the pretty-print logger (unless overridden via the context) with a given configurable logging level.
* Via the API, the logger instance can be overridden by providing it via the context.

At the moment, only the HTTP actor and left-deep-smallest BGP actor will do some basic logging to resp. INFO and DEBUG, just like `ldf-client`.

For advanced logging purposes, a [bunyan](https://www.npmjs.com/package/bunyan)-based logger is available. This is currently not being used anywhere.

This will allow https://github.com/comunica/jQuery-Widget.js/issues/2 to be implemented.

### DIFF
--- a/packages/actor-init-sparql/README.md
+++ b/packages/actor-init-sparql/README.md
@@ -150,6 +150,27 @@ myEngine.query('{ label writer(label_en: \"Michael Jackson\") artist { label } }
   .then(console.log);
 ```
 
+_Logging_
+
+Optionally, a custom logger can be used inside Comunica.
+By default, [`@comunica/logger-void`](https://github.com/comunica/comunica/tree/master/packages/logger-void/) is used,
+which will simply void all log calls.
+(_This default can be changed in the [configuration file](https://github.com/comunica/comunica/blob/master/packages/actor-init-sparql/config/config-default.json)_)
+
+Alternatively, [`@comunica/logger-pretty`](https://github.com/comunica/comunica/tree/master/packages/logger-pretty/),
+[`@comunica/logger-bunyan`](https://github.com/comunica/comunica/tree/master/packages/logger-bunyan/),
+or a custom logger implementing the [`Logger`](https://github.com/comunica/comunica/blob/master/packages/core/lib/Logger.ts) interface can be used.
+
+These loggers can be configured through the context as follows:
+```javascript
+import {LoggerPretty} from "@comunica/logger-pretty";
+
+const context = {
+  log: new LoggerPretty({ level: 'warn' });
+};
+myEngine.query('...', context);
+```
+
 ### Usage within browser
 
 This engine can run in the browser using [Webpack](https://www.npmjs.com/package/webpack).
--- a/packages/actor-init-sparql/components/Actor/Init/Sparql.jsonld
+++ b/packages/actor-init-sparql/components/Actor/Init/Sparql.jsonld
@@ -1,6 +1,7 @@
 {
   "@context": [
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-init-sparql/^1.0.0/components/context.jsonld",
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/logger-void/^1.0.0/components/context.jsonld",
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/bus-init/^1.0.0/components/context.jsonld"
   ],
   "@id": "npmd:@comunica/actor-init-sparql",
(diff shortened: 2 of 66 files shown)


packages/logger-bunyan/lib/LoggerBunyan.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * A bunyan-based logger implementation.
+ */
+export class LoggerBunyan extends Logger {
+
+  private readonly bunyanLogger: BunyanLogger;
+
+  constructor(args: ILoggerBunyanArgs) {
+    super();
+    args = <ILoggerBunyanArgs> args;
```
joachimvh: Why is this required?

packages/logger-bunyan/lib/stream/BunyanStreamProviderFile.ts
```diff
@@ -0,0 +1,23 @@
...
+export class BunyanStreamProviderFile extends BunyanStreamProvider {
+
+  public readonly path: string;
+
+  constructor(args: IBunyanStreamProviderFileArgs) {
+    super(args);
+  }
+
+  public createStream(): Stream {
+    return { type: 'file', name: this.name, path: this.path.substr(8), level: this.level };
```
joachimvh: I'm assuming this is to remove the `file:///` part. Might be a problem for Windows (not sure yet since I don't see this used anywhere?). I would suggest keeping the `file:///` part and converting it to a URL. `fs` accepts URLs (with the file protocol) for all its functions.

packages/actor-init-sparql/components/Actor/Init/Sparql.jsonld
```diff
@@ -1,6 +1,7 @@
 {
   "@context": [
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-init-sparql/^1.0.0/components/context.jsonld",
```
joachimvh: Are all these 1.0.0 references correct?

packages/actor-init-sparql/README.md
```diff
@@ -150,6 +150,27 @@ myEngine.query('{ label writer(label_en: \"Michael Jackson\") artist { label } }
   .then(console.log);
 ```
 
+_Logging_
+
+Optionally, a custom logger can be used inside Comunica.
+By default, [`@comunica/logger-void`](https://github.com/comunica/comunica/tree/master/packages/logger-void/) is used,
```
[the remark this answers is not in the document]
rubensworks: Yes, the CLI will for instance always use logger-pretty on error (?) level.

packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
```diff
@@ -108,6 +109,13 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
       context[KEY_CONTEXT_QUERYFORMAT] = context.queryFormat;
       delete context.queryFormat;
     }
+    if (context.log) {
+      context[KEY_CONTEXT_LOG] = context.log;
+      delete context.log;
```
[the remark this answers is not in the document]
rubensworks: Namespacing of context keys. This part has been abstract already in the master branch though, just need to rebase.

packages/actor-init-sparql/README.md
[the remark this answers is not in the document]
joachimvh: Might also want to add that to the readme then probably

packages/core/lib/Actor.ts
```diff
@@ -94,6 +104,50 @@ export abstract class Actor<I extends IAction, T extends IActorTest, O extends I
     return true;
   }
 
+  /* Proxy methods for the (optional) logger that is defined in the context */
+
+  protected logTrace(context: ActionContext, message: string, data?: any): void {
```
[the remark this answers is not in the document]
rubensworks: Hmm, yes, that makes sense, as we consistently do this in all logging calls already.

packages/logger-bunyan/lib/LoggerBunyan.ts
rubensworks: Whoops, that's wrong indeed.

packages/logger-bunyan/lib/stream/BunyanStreamProviderFile.ts
rubensworks: 👍

packages/actor-init-sparql/components/Actor/Init/Sparql.jsonld
rubensworks: Yes, _all_ context use `^1.0.0` at the moment. This should only be updated on major releases.

joachimvh: There are also some non-context references to 1.0.0, e.g.: https://github.com/comunica/comunica/pull/208/files#diff-1fccd41cbc783b6aea9b344380c5a8bfR116 (might also be some others)

PR #212, opened 2018-08-17
Feature/path

joachimvh:
This adds support for all path expressions in SPARQL queries.

Not tested yet on actual SPARQL queries


packages/actor-query-operation-path-nps/lib/ActorQueryOperationPathNps.ts
```diff
@@ -0,0 +1,44 @@
...
+    const output = ActorQueryOperation.getSafeBindings(
+      await this.mediatorQueryOperation.mediate({ operation: pattern, context }));
+
+    // remove the generated blank nodes from the bindings
+    const bindingsStream = output.bindingsStream.transform<Bindings>({
+      filter: (bindings) => {
+        return !predicate.iris.some((iri) => iri.equals(bindings.get(blankName)));
+      },
+      transform: (item, next) => {
+        bindingsStream._push(item.delete(blankName));
```
rubensworks: I think you can call `next` directly with the item that needs to be pushed.

packages/actor-query-operation-path-alt/lib/ActorQueryOperationPathAlt.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  constructor(args: IActorQueryOperationTypedMediatedArgs) {
+    super(args, Algebra.types.ALT);
+  }
+
+  public async runOperation(path: Algebra.Path, context: ActionContext)
+    : Promise<IActorQueryOperationOutputBindings> {
+    const predicate = <Algebra.Alt> path.predicate;
+
+    const left = ActorQueryOperation.getSafeBindings(await this.mediatorQueryOperation.mediate({
```
rubensworks: We can optimize this by moving the `await`'s to a higher level, so that left and right can run in parallel.

packages/actor-query-operation-path-seq/lib/ActorQueryOperationPathSeq.ts
```diff
@@ -0,0 +1,54 @@
...
+    super(args, Algebra.types.SEQ);
+  }
+
+  public async runOperation(path: Algebra.Path, context: ActionContext)
+    : Promise<IActorQueryOperationOutputBindings> {
+    const predicate = <Algebra.Seq> path.predicate;
+    const blank = this.generateBlankNode(path);
+    const blankName = termToString(blank);
+
+    const left = ActorQueryOperation.getSafeBindings(await this.mediatorQueryOperation.mediate({
```
rubensworks: We can optimize this by moving the await's to a higher level, so that left and right can run in parallel.

packages/actor-query-operation-path-seq/lib/ActorQueryOperationPathSeq.ts
```diff
@@ -0,0 +1,54 @@
...
+    }));
+    const right = ActorQueryOperation.getSafeBindings(await this.mediatorQueryOperation.mediate({
+      context, operation: ActorAbstractPath.FACTORY.createPath(blank, predicate.right, path.object, path.graph),
+    }));
+
+    const join = ActorQueryOperation.getSafeBindings(await this.mediatorJoin.mediate({ entries: [left, right] }));
+    // remove the generated blank nodes from the bindings
+    const bindingsStream = join.bindingsStream.transform<Bindings>({
+      transform: (item, next) => {
+        bindingsStream._push(item.delete(blankName));
```
rubensworks: `next` can probably be used here as well.

packages/actor-query-operation-path-zero-or-more/lib/ActorQueryOperationPathZeroOrMore.ts
```diff
@@ -0,0 +1,52 @@
...
+          },
+        });
+      return { type: 'bindings', bindingsStream, variables: [] };
+    } else { // if (sVar || oVar)
+      const v = termToString(sVar ? path.subject : path.object);
+      const pred = sVar ? ActorAbstractPath.FACTORY.createInv(predicate.path) : predicate.path;
+      const bindingsStream = (await this.ALPeval(sVar ? path.object : path.subject, pred, context))
+        .transform<Bindings>({
+          transform: (item, next) => {
+            bindingsStream._push(Bindings({ [v]: item }));
```
rubensworks: Check if `next` can be used.

packages/actor-query-operation-path-nps/lib/ActorQueryOperationPathNps.ts
rubensworks: This doesn't appear to be the case, ignoring...

packages/actor-query-operation-path-seq/lib/ActorQueryOperationPathSeq.ts
rubensworks: This doesn't appear to be the case, ignoring...

packages/actor-query-operation-path-zero-or-more/lib/ActorQueryOperationPathZeroOrMore.ts
rubensworks: This doesn't appear to be the case, ignoring...

PR #266, opened 2018-09-08
Enable serialization in actor-init-sparql-rdfjs

jacoscaz:
While using `@comunica/actor-init-sparql-rdfjs` in [quadstore](https://github.com/beautifulinteractions/node-quadstore/tree/devel), I've noticed that calls to `engine.resultToString()` would always results in the following error 

    Error: No actors are able to reply to the message in the bus https://linkedsoftwaredependencies.org/bundles/npm/@comunica/bus-sparql-serialize/Bus/SparqlSerialize

This would happen no matter what format I went for. I looked around a little and found out that the parent package `actor-init-sparql` had additional entries in its `config/config-default.json` file related to serializers, which were absent from `actor-init-sparql-rdfjs`'s own config file.

As a developer, I'd rather have to build custom configurations in order to opt **out** of baked-in serialization support than having to do so to opt **in**. Having serialization enabled by default in the RDF/JS entry point to `comunica` might save a few headaches.


rubensworks:
The original intention of this package was to keep the dependencies as low as possible, to only support in-memory querying, so without serializers.

But now that I think of it, these serializers are added in any case because of the dependency on `@comunica/actor-init-sparql`. I've created a new issue to improve this abstraction in the future: #267.

But for now, we could indeed add the serializers to this package as well.

jacoscaz:
This should do it! As a general consideration, I think that packages that are most likely to represent entry points to `comunica` for developers should come with all features that one would normally expect when working with SPARQL. 

Perhaps a new class of actors might work best here: `actor-comunicafordummies-sparql-rdfjs`... :)

PR #270, opened 2018-09-13
Feature/minus sparql algebra

BertCommeine:
The minus operator works for the local tests and most queries, however, we found one case where there were some unfiltered results:

The following queries could be used as a test: First one gives all results for the lefter-part of the query, second one for the righter. The last one is the MINUS itself. We found that 2 results from the second query weren't filtered out. After closer inspection we encountered some very interesting and inconsistent behavior from both MINUS and normal queries. @joachimvh and me decided it was improbable the bugs were caused by the MINUS operator. 

```
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>PREFIX dbpedia-owl: <http://dbpedia.org/property/> SELECT DISTINCT ?software ?company WHERE { ?software dbpedia-owl:developer ?company. }
```

```
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>PREFIX dbpedia-owl: <http://dbpedia.org/property/> SELECT DISTINCT ?software ?company WHERE { ?software dbpedia-owl:developer ?company.  ?company dbpedia-owl:locationCountry [ rdfs:label "Belgium"@en ]. }
```

```
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>PREFIX dbpedia-owl: <http://dbpedia.org/property/> SELECT DISTINCT ?software ?company WHERE {{ ?software dbpedia-owl:developer ?company } MINUS { ?company dbpedia-owl:locationCountry [ rdfs:label "Belgium"@en ] }}
```

### DIFF
--- a/greenkeeper.json
+++ b/greenkeeper.json
@@ -51,6 +51,7 @@
         "packages/actor-query-operation-slice/package.json",
         "packages/actor-query-operation-sparql-endpoint/package.json",
         "packages/actor-query-operation-union/package.json",
+        "packages/actor-query-operation-minus/package.json",
         "packages/actor-query-operation-values/package.json",
         "packages/actor-rdf-dereference-file/package.json",
         "packages/actor-rdf-dereference-http-parse/package.json",
--- a/greenkeeper.json
+++ b/greenkeeper.json
@@ -8,6 +8,7 @@
     "allPackages": {
       "packages": [
         "packages/actor-abstract-mediatyped/package.json",
+        "packages/actor-abstract-bindings-hash/package.json",
         "packages/actor-abstract-path/package.json",
         "packages/actor-context-preprocess-rdf-source-identifier/package.json",
         "packages/actor-http-memento/package.json",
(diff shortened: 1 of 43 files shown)


rubensworks:
Looks like you forgot to rebase onto master again.

BertCommeine:
The

> Looks like you forgot to rebase onto master again.

I did forget it, my bad.

BertCommeine:
> > Looks like you forgot to rebase onto master again.
> 
> I did forget it, my bad.

When rebasing, I had the same issues as I did last time: The common commits on my branch and the master branch don't get recognised as identical and get applied twice on the branch. Merging seemed like the beste option. If needed I could make another branch, cherry-pick commits again and reopen a PR, but this seemed like an easier way to get the same result

rubensworks:
> If needed I could make another branch, cherry-pick commits again and reopen a PR

I suggest you do this. The merge you did here produces a messy commit history, which we want to avoid :-)

BertCommeine:
> I suggest you do this. The merge you did here produces a messy commit history, which we want to avoid :-)

As you wish. Do you have any idea how I can avoid this in the future?

rubensworks:
> Do you have any idea how I can avoid this in the future?

The easiest thing would be to just make sure that you create new feature branches from the latest HEAD of the master branch :-)

But the HEAD can always progress in parallel, so in those cases, doing a `rebase` after `fetch`ing the master branch should work fine.

The problem you had here was that you started off from a diverged branch (rewritten history), and in those cases equal commits are indeed seen as different, which is why you end up with massive merge conflicts.

BertCommeine:
I see, thanks!

PR #271, opened 2018-09-13
Feature/minus sparql algebra 2

BertCommeine:
The minus operator works for the local tests and most queries, however, we found one case where there were some unfiltered results:

The following queries could be used as a test: First one gives all results for the lefter-part of the query, second one for the righter. The last one is the MINUS itself. We found that 2 results from the second query weren't filtered out. After closer inspection we encountered some very interesting and inconsistent behavior from both MINUS and normal queries. @joachimvh and me decided it was improbable the bugs were caused by the MINUS operator. 

```
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>PREFIX dbpedia-owl: <http://dbpedia.org/property/> SELECT DISTINCT ?software ?company WHERE { ?software dbpedia-owl:developer ?company. }
```

```
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>PREFIX dbpedia-owl: <http://dbpedia.org/property/> SELECT DISTINCT ?software ?company WHERE { ?software dbpedia-owl:developer ?company.  ?company dbpedia-owl:locationCountry [ rdfs:label "Belgium"@en ]. }
```

```
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>PREFIX dbpedia-owl: <http://dbpedia.org/property/> SELECT DISTINCT ?software ?company WHERE {{ ?software dbpedia-owl:developer ?company } MINUS { ?company dbpedia-owl:locationCountry [ rdfs:label "Belgium"@en ] }}
```


packages/actor-query-operation-minus/lib/ActorQueryOperationMinus.ts
```diff
@@ -0,0 +1,64 @@
...
+import * as RDF from "rdf-js";
+import {Algebra} from "sparqlalgebrajs";
+
+/**
+ * A comunica Minus Query Operation Actor.
+ */
+export class ActorQueryOperationMinus extends AbstractBindingHash<Algebra.Minus> {
+
+  private hashes: {[id: string]: boolean} = {};
+  private commons: string[];
```
rubensworks: These two fields should not be part of this class, as we want actors to be stateless. 
Instead, these fields should be moved inside the closure of `newHashFilter` or `runOperation `.

packages/actor-query-operation-minus/lib/ActorQueryOperationMinus.ts
```diff
@@ -0,0 +1,64 @@
...
+          await this.mediatorQueryOperation.mediate({ operation: pattern.left, context }));
+
+    this.commons = buffer.variables.filter((n: string) => -1 !== output.variables.indexOf(n));
+    if (this.commons.length > 0) {
+      let bindingsStream: BindingsStream = null;
+      const prom = new Promise((resolve) => {
+        bindingsStream = buffer.bindingsStream;
+        bindingsStream.on('data', (data) => {
+          const hash = ActorQueryOperationMinus.hash(this.hashAlgorithm, this.digestAlgorithm,
+              data.filter((v: RDF.Term, k: string) => -1 !== this.commons.indexOf(k)));
```
rubensworks: I suggest to make `commons` a `{[variableName: string]: boolean}` instead of an array. This would allow constant-time lookups, instead of `O(N)` as it is now.

packages/actor-query-operation-minus/lib/ActorQueryOperationMinus.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    const buffer = ActorQueryOperation.getSafeBindings(
+          await this.mediatorQueryOperation.mediate({ operation: pattern.right, context }));
+    const output = ActorQueryOperation.getSafeBindings(
+          await this.mediatorQueryOperation.mediate({ operation: pattern.left, context }));
+
+    this.commons = buffer.variables.filter((n: string) => -1 !== output.variables.indexOf(n));
+    if (this.commons.length > 0) {
+      let bindingsStream: BindingsStream = null;
+      const prom = new Promise((resolve) => {
```
rubensworks: As far as I understand it, you use this promise to make the execution block until the left stream is finished.

I suggest you look into the usages of `PromiseProxyIterator` in this project, and solve this in a similar way.
The reason for this is that we want to reach the `return` statement of this method as soon as possible. The streams however, are allowed to block at any time.

packages/actor-query-operation-minus/package.json
```diff
@@ -0,0 +1,67 @@
...
+  "homepage": "https://github.com/comunica/comunica#readme",
+  "files": [
+    "components",
+    "lib/**/*.d.ts",
+    "lib/**/*.js",
+    "index.d.ts",
+    "index.js"
+  ],
+  "dependencies": {
+    "lodash.difference": "^4.5.0",
```
rubensworks: I don't see this being used anywhere, can this dependency go?

packages/actor-query-operation-minus/lib/ActorQueryOperationMinus.ts
joachimvh: So to specificy, since there was some confusion about this, the current problem is that the return statement of this function can not be reached before the entire right side (buffer) is drained. This would cause everything else to be stuck, waiting on this actor. The problem is the `await prom;`. This should not happen before the output object is returned. This means the bindingsStream should actually return a promise calculating all this stuff (and since its type is an iterator and not a promise, you should use a `PromiseProxyIterator` to encapsulate).

packages/actor-query-operation-minus/lib/ActorQueryOperationMinus.ts
```diff
@@ -0,0 +1,70 @@
...
+    const buffer = ActorQueryOperation.getSafeBindings(
+            await this.mediatorQueryOperation.mediate({ operation: pattern.right, context }));
+    const output = ActorQueryOperation.getSafeBindings(
+            await this.mediatorQueryOperation.mediate({ operation: pattern.left, context }));
+
+    const commons: {[variableName: string]: boolean} = this.getCommonVariables(buffer.variables, output.variables);
+    if (Object.keys(commons).length !== 0) {
+      const hashes: {[id: string]: boolean} = {};
+
+      const prom = new Promise((resolve) => {
```
rubensworks: Can you move this promise inside the body of your `PromiseProxyIterator `? This will make it so that the promise will be created lazily, so it won't start working if it's not needed.

rubensworks: Also, rename it to something more useful like `bufferBarrier` :-)

packages/actor-abstract-bindings-hash/components/Actor/AbstractFilterHash.jsonld
```diff
@@ -0,0 +1,46 @@
+{
+  "@context": [
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-abstract-filter-hash/^1.0.0/components/context.jsonld",
```
rubensworks: Wrong context.

Be sure to test with a full SPARQL query as well, because I assume this would not work at the moment.

packages/actor-query-operation-distinct-hash/components/Actor/QueryOperation/DistinctHash.jsonld
```diff
@@ -2,26 +2,27 @@
   "@context": [
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-query-operation-distinct-hash/^1.0.0/components/context.jsonld",
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-abstract-bindings-hash/^1.0.0/components/context.jsonld",
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-abstract-filter-hash/^1.0.0/components/context.jsonld",
```
rubensworks: Wrong import, this package does not exist anymore.

packages/actor-query-operation-distinct-hash/components/Actor/QueryOperation/DistinctHash.jsonld
```diff
@@ -2,26 +2,27 @@
...
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/bus-query-operation/^1.0.0/components/context.jsonld"
   ],
   "@id": "npmd:@comunica/actor-query-operation-distinct-hash",
   "components": [
     {
       "@id": "caqodh:Actor/QueryOperation/DistinctHash",
       "@type": "Class",
-      "extends": "caabh:Actor/AbstractBindingsHash",
+      "extends": "caafh:Actor/AbstractFilterHash",
```
rubensworks: All `caafh` prefixes are wrong.

packages/actor-query-operation-minus/package.json
```diff
@@ -0,0 +1,66 @@
...
+  },
+  "homepage": "https://github.com/comunica/comunica#readme",
+  "files": [
+    "components",
+    "lib/**/*.d.ts",
+    "lib/**/*.js",
+    "index.d.ts",
+    "index.js"
+  ],
+  "dependencies": {
```
rubensworks: Missing `asynciterator-promiseproxy` dependency.

packages/actor-query-operation-minus/lib/ActorQueryOperationMinus.ts
```diff
@@ -0,0 +1,72 @@
...
+    const buffer = ActorQueryOperation.getSafeBindings(
+            await this.mediatorQueryOperation.mediate({ operation: pattern.right, context }));
+    const output = ActorQueryOperation.getSafeBindings(
+            await this.mediatorQueryOperation.mediate({ operation: pattern.left, context }));
+
+    const commons: {[variableName: string]: boolean} = this.getCommonVariables(buffer.variables, output.variables);
+    if (Object.keys(commons).length !== 0) {
+      const hashes: {[id: string]: boolean} = {};
+      /**
+       * To assure we've filtered all B (`buffer) values from A (`output`) we wait until we've fetched all values of B.
```
rubensworks: Missing ` after buffer

packages/actor-query-operation-minus/lib/ActorQueryOperationMinus.ts
```diff
@@ -0,0 +1,72 @@
...
+       */
+      const bindingsStream = new PromiseProxyIterator(async () => {
+        await new Promise((resolve) => {
+          buffer.bindingsStream.on('data', (data) => {
+            const hash = ActorQueryOperationMinus.hash(this.hashAlgorithm, this.digestAlgorithm,
+                    data.filter((v: RDF.Term, k: string) => commons[k]));
+            hashes[hash] = true;
+          });
+          buffer.bindingsStream.on('end', () => {
+            resolve(hashes);
```
rubensworks: The output of this promise is not used anymore, so you can just do `resolve()`.

packages/actor-query-operation-reduced-hash/components/Actor/QueryOperation/ReducedHash.jsonld
```diff
@@ -2,6 +2,7 @@
   "@context": [
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-query-operation-reduced-hash/^1.0.0/components/context.jsonld",
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-abstract-bindings-hash/^1.0.0/components/context.jsonld",
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-abstract-filter-hash/^1.0.0/components/context.jsonld",
```
rubensworks: Wrong context import.

packages/actor-query-operation-reduced-hash/components/Actor/QueryOperation/ReducedHash.jsonld
```diff
@@ -14,14 +15,14 @@
       "comment": "A comunica Reduced Hash Query Operation Actor.",
       "parameters": [
         {
-          "@id": "caabh:Actor/AbstractBindingsHash/hashAlgorithm",
+          "@id": "caafh:Actor/AbstractFilterHash/hashAlgorithm",
```
rubensworks: Wrong prefixes.

PR #280, opened 2018-09-21
Feature/amf support [IN PROGRESS]

BertCommeine:
I was told to open a PR with my current progress so @rubensworks would find it easily. I think I've somewhat managed to extract the necessary metadata in my new actor, but I'm missing some test coverage. Then again, this is nothing but a start of supporting amf. More details about what I've done/discovered can be read in "remarks.md".


rubensworks:
Closing this in favor of https://github.com/comunica/comunica/tree/feature/amf

PR #288, opened 2018-10-02
Feature/http file extensions

### LINKED ISSUE #286: LD documents without content type not queryable.
#### Issue type:

- :bug: Bug

____
#### Description:

Documents without a Content-Type header value like http://wikiba.se/ontology#label are not queryable. We should provide a fallback-mechanism based on file extensions for such situations, just like we do for querying local files.

rubensworks:
This abstracts the fallback mechanism for determining media types based on file extensions, so that it can be used in the HTTP-based RDF dereferencer as well.

Closes #286


packages/actor-init-sparql-file/config/sets/rdf-source-identifiers.json
```diff
@@ -1,5 +1,6 @@
 {
   "@context": [
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-init-sparql/^1.0.0/components/context.jsonld",
```
joachimvh: Why was this added?

packages/actor-rdf-source-identifier-file-content-type/lib/ActorRdfSourceIdentifierFileContentType.ts
```diff
@@ -28,9 +28,9 @@ export class ActorRdfSourceIdentifierFileContentType extends ActorRdfSourceIdent
     const httpAction: IActionHttp = { context: action.context,
       init: { headers, method: 'HEAD' }, input: action.sourceValue };
     const httpResponse: IActorHttpOutput = await this.mediatorHttp.mediate(httpAction);
-    if (!httpResponse.ok || !httpResponse.headers.has('Content-Type')
-      || !this.allowedMediaTypes.find((mediaType: string) => httpResponse.headers
-        .get('Content-Type').indexOf(mediaType) >= 0)) {
+    if (!httpResponse.ok || (httpResponse.headers.has('Content-Type')
+      && !this.allowedMediaTypes.find((mediaType: string) => httpResponse.headers
+        .get('Content-Type').indexOf(mediaType) >= 0))) {
```
joachimvh: What about files with no content type and no valid file extension?

packages/actor-init-sparql-file/config/sets/rdf-source-identifiers.json
rubensworks: Because that's where the `config-sets` prefix is defined. This config did not work before apparently.

packages/actor-rdf-source-identifier-file-content-type/lib/ActorRdfSourceIdentifierFileContentType.ts
rubensworks: Those will simply fail :-)

We'll need this for that: https://github.com/comunica/comunica/issues/289

PR #299, opened 2018-10-11
Expressions

wschella:
This adds SPARQL expression support by use of the [sparqlee](https://github.com/comunica/sparqlee) library.


wschella:
The build will keep failing until I publish the new sparqlee version to npm.
Currently using `yarn link` to test with local build.

wschella:
@rubensworks this can be reviewed/merged I think.

rubensworks:
For reference, this PR increases number of passing SPARQL tests from _108 / 268_ to _114 / 268_. (Sparqlee isn't feature-complete yet, so this will increase in the future)

packages/actor-query-operation-extend/lib/ActorQueryOperationExtend.ts
```diff
@@ -0,0 +1,74 @@
...
+      await this.mediatorQueryOperation.mediate({ operation: input, context }));
+
+    const extendKey = termToString(variable);
+    const evaluator = new AsyncEvaluator(expression);
+
+    // Transform the stream by extending each Bindings with the expression result
+    const transform = async (bindings: Bindings, next: any) => {
+      try {
+        const result = await evaluator.evaluate(bindings);
+        const extended = bindings.set(extendKey, result); // Extend is undefined when the key exists.
```
rubensworks: Is this comment a TODO? Should this be _Extend is undefined when the **does not** key exists_?

packages/actor-query-operation-extend/lib/ActorQueryOperationExtend.ts
```diff
@@ -0,0 +1,74 @@
...
+        bindingsStream._push(extended);
+      } catch (err) {
+        if (this.isExpressionError(err)) {
+          // Errors silently don't actually extend according to the spec
+          // TODO: Do we try to emit a warning here?
```
rubensworks: You could use the logger (from the context) to emit a warning in this case.

packages/actor-query-operation-extend/lib/ActorQueryOperationExtend.ts
```diff
@@ -0,0 +1,74 @@
...
+          bindingsStream._push(bindings);
+        } else {
+          bindingsStream.emit('error', err);
+        }
+      }
+      next();
+    };
+
+    const variables = [extendKey]; // TODO: Can I access existing variables already to concat with?
```
rubensworks: > Can I access existing variables already to concat with?

👍 You can do this using the variables from `output`.

packages/actor-query-operation-filter-sparqlee/lib/ActorQueryOperationFilterSparqlee.ts
```diff
@@ -0,0 +1,61 @@
...
+export class ActorQueryOperationFilterSparqlee extends ActorQueryOperationTypedMediated<Algebra.Filter> {
+
+  constructor(args: IActorQueryOperationTypedMediatedArgs) {
+    super(args, 'filter');
+  }
+
+  public async testOperation(pattern: Algebra.Filter, context: ActionContext): Promise<IActorTest> {
+    // Will throw error for unsupported operators
+    // Coerce to boolean to check for falsy values
+    return !!new AsyncEvaluator(pattern.expression);
```
rubensworks: Is this an expensive operation? If so, perhaps we need to look into caching this somehow (not necessarily in this PR).

packages/actor-query-operation-filter-sparqlee/lib/ActorQueryOperationFilterSparqlee.ts
```diff
@@ -0,0 +1,61 @@
...
+    const outputRaw = await this.mediatorQueryOperation.mediate({ operation: pattern.input, context });
+    const output = ActorQueryOperation.getSafeBindings(outputRaw);
+    ActorQueryOperation.validateQueryOutput(output, 'bindings');
+    const { variables, metadata } = output;
+
+    const evaluator = new AsyncEvaluator(pattern.expression);
+    const transform = async (item: Bindings, next: any) => {
+      try {
+        const result = await evaluator.evaluateAsEBV(item);
+        if (result === true) { bindingsStream._push(item); }
```
rubensworks: I'm surprised the linter does not complain about this :-) Could you put this on separate lines?

Also, can you add a comment explaining the reason why you should explicitly compare to `true`, and not any thruthy value?

packages/actor-query-operation-leftjoin-nestedloop/lib/ActorQueryOperationLeftJoinNestedLoop.ts
```diff
@@ -20,37 +24,71 @@ export class ActorQueryOperationLeftJoinNestedLoop extends ActorQueryOperationTy
...
+          transform: async (innerItem: Bindings, nextInner: any) => {
+            const joinedBindings = ActorRdfJoin.join(outerItem, innerItem);
+            if (!joinedBindings) { nextInner(); return; }
+            if (!pattern.expression) {
+              joinedStream._push({ joinedBindings, result: true });
+              nextInner();
+              return;
+            }
+            try {
+              const evaluator = new AsyncEvaluator(pattern.expression);
```
rubensworks: Can you move this evaluator outside of this transformer function? I assume it can be reused for all invocations?

packages/actor-query-operation-leftjoin-nestedloop/lib/ActorQueryOperationLeftJoinNestedLoop.ts
```diff
@@ -20,37 +24,71 @@ export class ActorQueryOperationLeftJoinNestedLoop extends ActorQueryOperationTy
...
-      },
-    });
+      return joinedStream;
+    };
+
+    const leftJoinOuter = (leftItem: Bindings, nextLeft: any) => {
+      const innerStream = right.bindingsStream.clone(); // TODO: Why does this even work?
+      const joinedStream = leftJoinInner(leftItem, innerStream);
+
+      // TODO: Does this even work for large streams?
```
rubensworks: Good catch, we should look into this in the future.

packages/actor-query-operation-leftjoin-nestedloop/lib/ActorQueryOperationLeftJoinNestedLoop.ts
```diff
@@ -20,37 +24,71 @@ export class ActorQueryOperationLeftJoinNestedLoop extends ActorQueryOperationTy
...
     const metadata = () => Promise.all([pattern.left, pattern.right].map((entry) => entry.metadata))
       .then((metadatas) => metadatas.reduce((acc, val) => acc * val.totalItems, 1))
       .catch(() => Infinity)
-      .then((totalItems) =>  ({ totalItems }) );
+      .then((totalItems) => ({ totalItems }));
 
     return { type: 'bindings', bindingsStream, metadata, variables };
   }
 
+  // TODO Duplication with e.g. Extend
```
rubensworks: Maybe we can move this function to sparqlee?

packages/actor-query-operation-extend/lib/ActorQueryOperationExtend.ts
wschella: Actually not, when you extend with a variable name already present in the bindings, the result of extend is undefined by the spec. In this case, it'll just overwrite. I'll rephrase the comment.

packages/actor-query-operation-filter-sparqlee/lib/ActorQueryOperationFilterSparqlee.ts
wschella: It's not a simple if else check, but it's not performance that will matter when dealing with network data for example. Could prepend a very simple check first, but the full check is required to see if the operation is supported. Shouldn't matter too much in the end.

packages/actor-query-operation-filter-sparqlee/lib/ActorQueryOperationFilterSparqlee.ts
wschella: I'll put it on separate lines. If the result is not a boolean, something went very wrong (since evaluateAsEBV coerces any term into a boolean/ Effective Boolean Value). Since this API should be stable, I'll replace by a truthy check.

packages/actor-query-operation-leftjoin-nestedloop/lib/ActorQueryOperationLeftJoinNestedLoop.ts
wschella: The only reason it's a separate function is so that it can be easily mocked for tests, I'll have to check if that's still a an option. Another elegant solution would be to just provide this on the query operation abstract class, this way the documentation stays in house too.

rubensworks: > The only reason it's a separate function is so that it can be easily mocked for tests

It should be possible to mock specific functions from (external) modules with Jest.

> Another elegant solution would be to just provide this on the query operation abstract class

I would create a separate package for this then, as this would require a dependency on sparqlee, while this is only needed for some query operators.

wschella:
I don't think the coveralls decrease is an actual decrease, just some floating point errors.
I think all conflicts are resolved now, and this can be merged.

wschella:
@rubensworks I don't think the coveralls decrease is an actual decrease, just some floating point errors. I think all requested changes are resolved now, and this can be merged.

rubensworks:
@joachimvh Could you double-check this one before we merge this?

packages/actor-query-operation-orderby-direct/lib/ActorQueryOperationOrderByDirect.ts
```diff
@@ -1,10 +1,14 @@
...
-import {ActorQueryOperation, ActorQueryOperationTypedMediated,
-  IActorQueryOperationOutputBindings, IActorQueryOperationTypedMediatedArgs} from "@comunica/bus-query-operation";
-import {ActionContext, IActorTest} from "@comunica/core";
-import {termToString} from "rdf-string";
-import {Algebra} from "sparqlalgebrajs";
-import {SortIterator} from "./SortIterator";
+import { Term } from "rdf-js";
+import { termToString } from "rdf-string";
+import { Algebra } from "sparqlalgebrajs";
+import { AsyncEvaluator, isExpressionError } from 'sparqlee';
```
joachimvh: The reason this class is called orderby-*direct* is because it makes use of actor-query-operation-filter-*direct*. Might make more sense then to create a new orderby actor instead here?

wschella: OrderBySparqlee?

joachimvh: Yea that's consistent with the others

packages/actor-query-operation-leftjoin-nestedloop/lib/ActorQueryOperationLeftJoinNestedLoop.ts
```diff
@@ -20,37 +24,64 @@ export class ActorQueryOperationLeftJoinNestedLoop extends ActorQueryOperationTy
...
+            nextInner();
+          },
         });
-      },
-    });
+      return joinedStream;
+    };
+
+    const leftJoinOuter = (leftItem: Bindings, nextLeft: any) => {
+      const innerStream = right.bindingsStream.clone(); // TODO: Why does this even work?
```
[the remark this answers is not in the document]
wschella: Okay, then i just misinterpreted asynciterator docs, will remove.

wschella:
@joachimvh @rubensworks all suggested changes have been made.

PR #302, opened 2018-10-17
Feature/treeontology

### LINKED ISSUE #295: Add support for tree ontology
#### Issue type:

____
#### Description:

The ability to resolve queries over a dataset defined by the [Tree Ontology](https://github.com/pietercolpaert/TreeOntology).

This includes:
* Pagination over a tree
* Applying filters that follow certain branches

Could be done by including a new query operation tree filter actor that indicates that we are working with a tree ontology in the context and a RDF resolve quad pattern tree actor that solves the query over the tree.

Lamasaurus:
Implementation of the tree ontology. Fixes #295


packages/actor-query-operation-filter-direct/lib/ActorQueryOperationFilterDirect.ts
```diff
@@ -16,11 +19,20 @@ export class ActorQueryOperationFilterDirect extends ActorQueryOperationTypedMed
   public async testOperation(pattern: Algebra.Filter, context: ActionContext): Promise<IActorTest> {
     // will throw error for unsupported operators
     SparqlExpressionEvaluator.createEvaluator(pattern.expression);
+
+    // Check if the source is a hypermedia tree, because this filter can't handle that
+    const source = await ActorQueryOperationSparqlEndpoint.getSingleSource(context);
+    if (source && source.type === "hypermedia-tree") {
```
joachimvh: I'm actually not convinced that this actor does not work on sources of the type hypermedia tree. It can still pull in all triples from that source and then apply the filters locally. And only for a few select filters would this be a difference. (I would suggest looking into applying a mediatortype to return a different value in the test function for these specific instances)

packages/actor-query-operation-filter-tree/lib/SparqlExpressionEvaluator.ts
```diff
@@ -0,0 +1,241 @@
+import {Bindings} from "@comunica/bus-query-operation";
+import * as RDF from "rdf-js";
+import {stringToTerm, termToString} from "rdf-string";
+import {Algebra} from "sparqlalgebrajs";
+
+export class SparqlExpressionEvaluator {
```
joachimvh: This file is identical to the evaluator in the other filter actor, which is a bad thing. There also seems to be some overlap between the 2 main actor classes. If both classes share so much there needs to be some sort of inheritance somewhere to prevent this duplication (possibly with an abstract class).

Even better might be, since this class uses so much of the other actor, to actually use both actors. I.e., have this actor do some changes to the query and context, and then call the other filter actor through the mediator with that new query. This would remove all duplication that is currently required.

packages/actor-rdf-resolve-quad-pattern-tree/lib/ActorRdfResolveQuadPatternTree.ts
```diff
@@ -0,0 +1,147 @@
...
+      return this.sources[hypermedia];
+    }
+
+    // Cache and return
+    return await (this.sources[hypermedia] = this.createSource(context));
+  }
+
+  protected async getOutput(source: RDF.Source, pattern: RDF.Quad, context: ActionContext)
+  : Promise<IActorRdfResolveQuadPatternOutput> {
+    debugger;
```
joachimvh: debug code

packages/actor-rdf-resolve-quad-pattern-tree/lib/ActorRdfResolveQuadPatternTree.ts
```diff
@@ -0,0 +1,147 @@
...
+  IActorRdfResolveQuadPatternOutput, ILazyQuadSource} from "@comunica/bus-rdf-resolve-quad-pattern";
+import {ActionContext, Actor, IActorArgs, IActorTest, Mediator} from "@comunica/core";
+import * as RDF from "rdf-js";
+import {termToString} from "rdf-string";
+import {MediatedQuadSource} from "./MediatedQuadSource";
+
+/**
+ * A comunica tree RDF Resolve Quad Pattern Actor.
+ */
+export class ActorRdfResolveQuadPatternTree extends ActorRdfResolveQuadPatternSource
```
joachimvh: Same comment here. This class is identical to the qpf rdf resolver. (And also the same issue with the mediated quad source). Having duplicated code creates an unmaintainable code base so this will need to change.

packages/actor-rdf-source-identifier-hypermedia-tree/components/Actor/RdfSourceIdentifier/HypermediaQpf.jsonld
```diff
@@ -0,0 +1,72 @@
+{
```
joachimvh: I'm confused what this file is doing here. An artifact of older code?

Lamasaurus: This is indeed an artifact of older code. Deleted it.

Lamasaurus: This is indeed an artifact of older code. I deleted it.

packages/actor-query-operation-filter-direct/lib/ActorQueryOperationFilterDirect.ts
```diff
@@ -16,11 +19,20 @@ export class ActorQueryOperationFilterDirect extends ActorQueryOperationTypedMed
...
+      throw new Error(this.name
+        + ' type of the source may not be \'hypermedia-tree\'.');
+    }
+
     return true;
   }
 
   public async runOperation(pattern: Algebra.Filter, context: ActionContext)
-    : Promise<IActorQueryOperationOutputBindings> {
+  : Promise<IActorQueryOperationOutputBindings> {
```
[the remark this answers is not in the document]
Lamasaurus: Ok, will do!

packages/actor-rdf-resolve-quad-pattern-tree/lib/ActorRdfResolveQuadPatternTree.ts
Lamasaurus: Oops, I'll pay attention that this doesn't get in here anymore.

pietercolpaert:
@joachimvh @rubensworks  I’m guessing this is not something that @Lamasaurus can resolve, what to put where. Can you decide on a strategy between you two and come back to us if we need to put it in hypermedia or in hypermedia-tree?

rubensworks:
`hypermedia` would indeed be the way to go. Let's discuss this when I get back!

Issue #138, opened 2018-06-04
Add RDFa parsing support


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request

____
#### Description:

Comunica already supports the parsing of several RDF serializations. We should also add support for parsing RDF inside HTML documents using RDFa.

pietercolpaert:
I made this https://www.npmjs.com/package/rdfa-processor - it’s not perfect but does the job.

rubensworks:
Cool!

What's not perfect about that implementation? Is it something that would cause problems when used within comunica?

PRs are welcome :-)

rubensworks:
> Doesn’t yet follow the RDFJS Sink spec

This would definitely be a blocker.

> I was also thinking that at the same time we should support JSON-LD and N3 snippets in an HTML page at the same time.

Yes, definitely makes sense. We'd probably want to have a `text/html` parsing actor in comunica that delegates all HTML to the RDFa parser, but also identifies these RDF snippets in `<script>` tags and pushes them onto the bus again to parse, ignoring those types that can not be parsed.

I'm just wondering if it would make sense to have this `<script>` identification part of an RDFa parser, which would accept a callback for parsing other types of serializations.

rubensworks:
Additionally, `alternate` links should/could be followed, such as this one:
```
<link rel="alternate" type="text/turtle" href="profile.ttl" />
```

rubensworks:
So here's a plan on what to do after #310 is merged.

#310 registers to the rdf-parse bus as an HTML parser for extracting script tags.
However, we also need to be able to extract RDFa (and other things such as microdata?).
So I suggest the following:


1. Create html-rdf-extract bus.
2. Create rdf-parse-html actor. This actor simply parses an HTML doc, and throws it on the html-rdf-extract bus. It needs a special (new?) mediator (mediator-mediatyped-union?) that takes the union of action results, because multiple actors will be able to extract RDF from HTML.
3. Create rdf-parse-html-script actor, which does basically what the actor from #310 does.
4. Remove the old actor introduced in #310
5. Create rdf-parse-html-rdfa actor, for extracting RDFa.

This architecture enables straightforward extensibility, and things like extracting microdata can be plugged in easily later on.

rubensworks:
For reference, here's a fully spec-compliant parser, which follows the RDFJS interfaces: https://github.com/rubensworks/rdfa-streaming-parser.js

PR #310, opened 2018-10-31
Feature/rdf html parser

astverha:
The rdf html parser actor accepts text/html and extracts json-ld from between script tags with type="application/ld+json".


packages/actor-rdf-parse-html/components/Actor/RdfParse/Html.jsonld
```diff
@@ -0,0 +1,49 @@
...
+        {
+          "@id": "caam:Actor/AbstractMediaTypedFixed/mediaTypeEntry",
+          "defaultScoped": {
+            "defaultScope": "carph:Actor/RdfParse/Html",
+            "defaultScopedValue": [
+              { "mediaTypeKey": "text/html", "priorityValue": "1.0" },
+              { "mediaTypeKey": "(text/html; charset=utf-8)", "priorityValue": "0.8" },
+              { "mediaTypeKey": "(text/html; charset=utf-8)", "priorityValue": "0.9" },
+              { "mediaTypeKey": "text/html; charset=utf-8", "priorityValue": "1.0" },
+              { "mediaTypeKey": "hypermedia-html", "priorityValue": "0.7" }
```
rubensworks: I don't think this is a standardized media type (correct me if I'm wrong).

packages/actor-rdf-parse-html/components/Actor/RdfParse/Html.jsonld
```diff
@@ -0,0 +1,49 @@
...
+        },
+        {
+          "@id": "caam:Actor/AbstractMediaTypedFixed/mediaTypeEntry",
+          "defaultScoped": {
+            "defaultScope": "carph:Actor/RdfParse/Html",
+            "defaultScopedValue": [
+              { "mediaTypeKey": "text/html", "priorityValue": "1.0" },
+              { "mediaTypeKey": "(text/html; charset=utf-8)", "priorityValue": "0.8" },
+              { "mediaTypeKey": "(text/html; charset=utf-8)", "priorityValue": "0.9" },
+              { "mediaTypeKey": "text/html; charset=utf-8", "priorityValue": "1.0" },
```
rubensworks: It's not a good idea to list all possible media type strings here, `text/html` should be sufficient, your code should be able to handle all other variants.

packages/actor-rdf-parse-html/lib/ActorRdfParseHtml.ts
```diff
@@ -0,0 +1,60 @@
...
+    super(args);
+  }
+
+  public async runHandle(action: IActionRdfParse, mediaType: string, context: ActionContext):
+      Promise<IActorRdfParseOutput> {
+
+    const htmlString: string = await require('stream-to-string')(action.input);
+
+    // JSON-LD extraction
+    let jsonString: string = '';
```
rubensworks: Not sure if this code will work properly when there are multiple JSON-LD script tags available. Be sure to add a test for that.

packages/actor-rdf-parse-html/lib/ActorRdfParseHtml.ts
```diff
@@ -0,0 +1,60 @@
...
+    const DOMParser = require('xmldom').DOMParser;
+    const doc = new DOMParser().parseFromString(htmlString, 'text/html');
+    const scripts = doc.getElementsByTagName('script');
+    for (let i = 0; i < scripts.length; i++) {
+      if (scripts[i].getAttribute("type") === "application/ld+json") {
```
rubensworks: This is hardcoded to only support JSON-LD, while any other (supported) RDF serialization should be parseable as well (like Turtle, TriG, ...)

packages/actor-rdf-parse-html/lib/ActorRdfParseHtml.ts
```diff
@@ -0,0 +1,60 @@
...
+import {ActionContext, Actor, IActorTest, Mediator} from "@comunica/core";
+import {Readable} from "stream";
+
+/**
+ * A HTML RDF Parse actor that listens on the 'rdf-parse' bus.
+ *
+ * It is able to extract JSON-LD from HTML files and parse the JSON-LD based RDF serializations
+ * and announce the presence of them by media type.
+ */
+// Gebaseerd op JSON LD parser
```
rubensworks: No Dutch in this project :-)

packages/actor-rdf-parse-html/lib/ActorRdfParseHtml.ts
```diff
@@ -0,0 +1,60 @@
...
+export class ActorRdfParseHtml extends ActorRdfParseFixedMediaTypes {
```
rubensworks: As far as I understand, this parser _only_ extracts JSON-LD snippets.
I though the goal was to parse RDFa as well?

packages/actor-rdf-parse-jsonld/lib/ActorRdfParseJsonLd.ts
```diff
@@ -28,8 +28,8 @@ export class ActorRdfParseJsonLd extends ActorRdfParseFixedMediaTypes {
       return blankNode(term.value.substr(2)); // Remove the '_:' prefix.
     case 'Literal':
       return literal(term.value, term.language || term.datatype);
-    /*case 'Variable':
-      return variable(term.value);*/ // Variables can not occur in JSON-LD bodies
+      /*case 'Variable':
```
rubensworks: Unneeded changes

packages/actor-rdf-parse-html/components/Actor/RdfParse/Html.jsonld
astverha: I totally forgot about those. text/html should indeed be sufficient

packages/actor-rdf-parse-html/lib/ActorRdfParseHtml.ts
astverha: The assigned task here was to only parse the JSON-LD from between the script tags.

rubensworks: I thought this was the PR for #138.

@pietercolpaert Did I misunderstand this?

pietercolpaert: I first limited @astverha’s task to JSON-LD snippets, to then gradually expand to others as well! First, let’s see how this PR looks for JSON-LD. There should indeed be other content-types there!

packages/actor-rdf-parse-html/lib/ActorRdfParseHtml.ts
pietercolpaert: It is the goal, but so far, only the json-ld snippet is extracted (first thing we need for a project)

packages/actor-rdf-parse-html/lib/ActorRdfParseHtml.ts
rubensworks: Ok, then I'll wait for merging this until it's feature-complete :-)

rubensworks:
@astverha Ok, so for reference:
I suggest to rename your actor to `ActorRdfParseHtmlScript`, as its main goal will be to extract RDF from the script tags in HTML.
I would propose to re-run the generator, because this will also cause some changes in your components.js files. You should be able to copy-paste pretty much everything then. Just take into account that the JSON-LD prefix in your components.js files will be different after re-generation.

packages/actor-init-sparql/config/sets/rdf-parsers.json
```diff
@@ -6,10 +6,22 @@
...
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-rdf-parse-jsonld/^1.0.0/components/context.jsonld",
-    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-rdf-parse-rdfxml/^1.0.0/components/context.jsonld"
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-rdf-parse-rdfxml/^1.0.0/components/context.jsonld",
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-rdf-parse-html-script/^1.0.0/components/context.jsonld",
+
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/mediator-race/^1.0.0/components/context.jsonld"
   ],
   "@id": "urn:comunica:my",
   "actors": [
+    {
```
rubensworks: I suggest to move this HTML entry to the end of the array, _and_ give it a `"priorityScale` of `0.8`. This will make it so that this HTML will have the lowest priority of all parsers. (as other RDF serializations are likely to be more efficient)

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,92 @@
...
+  IActorRdfParseOutput,
+  IActorTestRootRdfParse,
+} from "@comunica/bus-rdf-parse";
+import {ActionContext, Actor, Mediator} from "@comunica/core";
+import {Readable} from "stream";
+
+/**
+ * A HTML RDF Parse actor that listens on the 'rdf-parse' bus.
+ *
+ * It is able to extract JSON-LD from HTML files and parse the JSON-LD based RDF serializations
```
rubensworks: Comment is outdated, it can extract any RDF serialization from script tags.

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,92 @@
...
+    IActorOutputRootRdfParse>, IActionRootRdfParse, IActorTestRootRdfParse, IActorOutputRootRdfParse>;
+
+  constructor(args: IActorRdfParseFixedMediaTypesArgs) {
+    super(args);
+  }
+
+  public async runHandle(action: IActionRdfParse, mediaType: string, context: ActionContext):
+    Promise<IActorRdfParseOutput> {
+
+    const myQuads = new Readable({objectMode: true});
```
rubensworks: Unclear variable name, I propose to change this to just `quads`. (then you can also abbreviate the object on line 90)

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    myQuads._read = async () => {
+      if (!initialized) {
+        initialized = true;
+
+        // Stringify HTML input
+        const htmlString: string = await require('stream-to-string')(action.input);
+
+        // Extract script tags
+        const DOMParser = require('xmldom').DOMParser;
```
rubensworks: It looks like `xmldom` is not aware of the browser environment yet: https://github.com/jindw/xmldom/pull/152
This means that when this is executed in the browser, that the `xmldom` code will be used, instead of the browsers's built-in DOM parser.

@pietercolpaert How do you handle this in ldfetch? Perhaps you can help out here?

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,92 @@
...
+        const doc = new DOMParser().parseFromString(htmlString, 'text/html');
+        const scripts = doc.getElementsByTagName('script');
+
+        const supportedTypes = [
```
rubensworks: Hardcoding things like this is the root of all evil ;-)

You can dynamically discover all available content types by doing something like this:
```javascript
Object.keys((await this.mediatorRdfParse.mediate({context, mediaTypes: true})).mediaTypes);
```

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,92 @@
...
+          "application/n-quads",
+          "application/n-triples",
+          "application/trig",
+          "text/n3",
+          "text/turtle",
+        ];
+
+        // Loop script elements
+        let count: number = 0;      // count quad streams
+        for (let i = 0; i < scripts.length; i++) {
```
rubensworks: You can probably simplify this loop with a [for-of loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of)

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
rubensworks: Also, `xmldom` is not present in your `package.json` file.

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,92 @@
...
+            const stream: Readable = new Readable({objectMode: true});
+            stream.push(scripts[i].textContent);
+            stream.push(null);
+
+            // Push stream on parse bus
+            const returned = (await this.mediatorRdfParse.mediate(
+              {context, handle: {input: stream}, handleMediaType: supportedTypes[index]})).handle;
+
+            // Push parsed quads on main stream
+            await returned.quads.on('data', async (chunk) => {
```
rubensworks: Unneeded await and async.

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,92 @@
...
+              await myQuads.push(chunk);
```
rubensworks: Unneeded await

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,92 @@
...
+            });
+
+            // End the main stream
+            await returned.quads.on('end', async () => {
```
rubensworks: Unneeded await and async

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,92 @@
...
+              if (count > 1) {
+                count--;
+              } else {
+                await myQuads.push(null);
```
rubensworks: Unneeded await

packages/actor-rdf-parse-html-script/test/ActorRdfParseHtmlScript-test.ts
```diff
@@ -0,0 +1,111 @@
...
+
+      it('should run', () => {
+        return actor.run({handle: { input }, handleMediaType: 'text/html'})
+          .then(async (output) => expect(await arrayifyStream(output.handle.quads)).toEqualRdfQuadArray([
+            quad('http://example.org/a', 'http://example.org/b', 'http://example.org/c'),
+            quad('http://example.org/a', 'http://example.org/d', 'http://example.org/e'),
+          ]));
+      });
+
+      it('should run with wrong script type', () => {
```
rubensworks: I would dedicate this test to the case with a _single_ script tag with type `text/plain`.

I would then also add the following tests:
* 2 JSON-LD tags
* 1 JSON-LD tag + 1 text/plain tag
* 1 JSON-LD + 1 N-Quads + 1 JSON-LD

packages/actor-rdf-parse-jsonld/lib/ActorRdfParseJsonLd.ts
```diff
@@ -29,7 +29,7 @@ export class ActorRdfParseJsonLd extends ActorRdfParseFixedMediaTypes {
     case 'Literal':
       return literal(term.value, term.language || term.datatype);
     /*case 'Variable':
-      return variable(term.value);*/ // Variables can not occur in JSON-LD bodies
+        return variable(term.value);*/ // Variables can not occur in JSON-LD bodies
```
rubensworks: Unneeded change

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
rubensworks: Also, make sure to write your objects like this `{ value: true }` instead of this `{value:true}`. This applies to multiple places in this PR.

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
astverha: Type Error: scripts is not iterable

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
astverha: application/json+ld is not included in these mediaTypes. Where do I add it? Also, shouldn't text/html also be added?

rubensworks: > application/json+ld is not included in these mediaTypes

It should be there though. Which ones _are_ included then?

> shouldn't text/html also be added?

No. You don't want script tags with type `text/html`.

astverha:
```
[ 'application/trig',
  'application/n-quads',
  'text/turtle',
  'application/n-triples',
  'text/n3' ]
```

astverha: Also, running my tests with `Object.keys(...` results in a null value.

rubensworks: Ah, right, that's because your mediator only picks a single actor.

You'll need another mediator over the same bus, but of type `mediator-combine-union`. You should be able to do this like in `ActorRdfDereferenceHttpParse`. (Can you double check if the JSON-LD content types are present in `ActorRdfDereferenceHttpParse`? Because I can't find usage of `mediator-combine-union` in the config file. Otherwise, it might be wrong there as well...)

rubensworks: > Also, running my tests with Object.keys(... results in a null value.

That's because your test environment doesn't have any parsing actors registered.

astverha: The JSON-LD content types are not in `ActorRdfDereferenceHttpParse` either.

rubensworks: Ok, thanks for checking, reported here: #318.

Feel free to look into it, as I have no idea if or when I'll be able to look into it myself.

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,99 @@
...
+        const htmlString: string = await require('stream-to-string')(action.input);
+
+        // Extract script tags
+        const DOMParser = require('xmldom').DOMParser;
+        const doc = new DOMParser().parseFromString(htmlString, 'text/html');
+        const scripts = doc.getElementsByTagName('script');
+
+        const supportedTypes: string[] = Object.keys((await this.mediatorRdfParse
+          .mediate({
+            context: ActionContext({ '@comunica/bus-rdf-parse:source': { type: 'mediaTypes' } }),
```
rubensworks: No new context is needed, just pass the provided context.

packages/actor-rdf-parse-html-script/test/ActorRdfParseHtmlScript-test.ts
```diff
@@ -0,0 +1,211 @@
...
+
+        return actor.run(
+          { context: parseAction.context, handle: { input: parseAction.input }, handleMediaType: 'text/html' })
+          .then(async (output) => expect(await arrayifyStream(output.handle.quads)).toEqualRdfQuadArray([
+            quad('http://example.org/a', 'http://example.org/b', 'http://example.org/c'),
+            quad('http://example.org/a', 'http://example.org/d', 'http://example.org/e'),
+          ]));
+      });
+
+      it('should run with two JSON-LD scripts and one N-Quads script', () => {
```
rubensworks: The expected output of this test is wrong, right? The JSON-LD output is duplicated, and the N-Quads output is missing.

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,99 @@
...
+            mediaTypes: true,
+          })).mediaTypes);
+
+        // Loop script elements
+        let count: number = 0;      // count streams
+        let amountOfWrongTypes: number = 0;
```
rubensworks: I don't think these two separate counters are needed. Furthermore, the logic behind these counters seems incorrect.

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
astverha: Passing the provided context works when running a query but not when running the tests. Therefore I pass a new context to get the mediaTypes.

rubensworks: Then something must be wrong with your tests :-)

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
astverha: Using htmlparser2 now (https://github.com/fb55/htmlparser2/).

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,105 @@
...
+  IActorRdfParseOutput,
+  IActorTestRootRdfParse,
+} from "@comunica/bus-rdf-parse";
+import {ActionContext, Actor, Mediator} from "@comunica/core";
+import {Readable} from "stream";
+
+/**
+ * A HTML script RDF Parse actor that listens on the 'rdf-parse' bus.
+ *
+ * It is able to extract and parse any RDF serialization from HTML files
```
rubensworks: "It is able to extract and parse any RDF serialization from script tags in HTML files"

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,105 @@
...
+    IActorOutputRootRdfParse>, IActionRootRdfParse, IActorTestRootRdfParse, IActorOutputRootRdfParse>;
+
+  constructor(args: IActorRdfParseFixedMediaTypesArgs) {
+    super(args);
+  }
+
+  public async runHandle(action: IActionRdfParse, mediaType: string, context: ActionContext):
+    Promise<IActorRdfParseOutput> {
+
+    const quads = new Readable({objectMode: true});
```
rubensworks: `{ objectMode: true }`

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,105 @@
...
+
+    quads._read = async () => {
+
+      // Stringify HTML input
+      const htmlString: string = await require('stream-to-string')(action.input);
```
rubensworks: This should not be needed.
`htmlparser2` accepts streams AFAIK, so you should be able to pipe `action.input` into the parser directly. This will lead to much better performance, as stuff can happen asynchronously.

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,105 @@
...
+          }
+        },
+
+        ontext: (text: string) => {
+          if (streamOpened) {
+            stream.push(text);
+          }
+        },
+      }, {decodeEntities: true});
+      await parser.write(htmlString);
```
rubensworks: `write` does not return a promise(?), so no `await` is needed.

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,105 @@
...
+      parser.end();
+
+      if (noRDFScriptTags) {
```
rubensworks: This can fail if the parser is slow, and this code is reached before the parser ends.

It would be better to listen to the end-event of the parser, and push `null` there iff no script tags were found.

packages/actor-rdf-parse-html-script/lib/ActorRdfParseHtmlScript.ts
```diff
@@ -0,0 +1,105 @@
...
+            stream = new Readable({ objectMode: true });
+          }
+        },
+
+        ontext: (text: string) => {
+          if (streamOpened) {
+            stream.push(text);
+          }
+        },
+      }, {decodeEntities: true});
```
rubensworks: `{ decodeEntities: true }`

packages/actor-rdf-parse-html-script/test/ActorRdfParseHtmlScript-test.ts
```diff
@@ -0,0 +1,220 @@
...
+            quad('http://example.org/a', 'http://example.org/b', 'http://example.org/c'),
+            quad('http://example.org/a', 'http://example.org/d', 'http://example.org/e'),
+          ]) }});
+        }
+
+      },
+    };
+
+    context = ActionContext({ '@comunica/bus-rdf-parse:source':
+          { type: 'handle', value: 'myValue' },
```
rubensworks: This source is invalid. Pretty sure you don't even need it though.

packages/actor-rdf-parse-html-script/test/ActorRdfParseHtmlScript-test.ts
```diff
@@ -0,0 +1,220 @@
...
+    it('should not be able to create new ActorRdfParseHtml objects without \'new\'', () => {
+      expect(() => {
+        (<any> ActorRdfParseHtmlScript)();
+      }).toThrow();
+    });
+  });
+
+  describe('An ActorRdfParseHtmlScript instance', () => {
+    let actor: ActorRdfParseHtmlScript;
+    const  input: Readable = stringToStream(
```
rubensworks: Is this ever used? It looks like this is always overwritten in beforeEach.

packages/actor-rdf-parse-html-script/test/ActorRdfParseHtmlScript-test.ts
```diff
@@ -0,0 +1,220 @@
...
+      it('should run with two JSON-LD scripts', () => {
+        parseAction.input = stringToStream(
+          `<script type="application/ld+json">
+          [{
+              "@id": "http://example.org/a",
+              "http://example.org/b": "http://example.org/c",
+              "http://example.org/d": "http://example.org/e"
+          }]
+          </script>
+          <script type="application/ld+json">
```
rubensworks: It would be better to have different contents in this second tag.

packages/actor-rdf-parse-html-script/test/ActorRdfParseHtmlScript-test.ts
```diff
@@ -0,0 +1,220 @@
...
+      it('should run with two JSON-LD scripts and one N-Quads script', () => {
+        parseAction.input = stringToStream(
+          `<script type="application/ld+json">
+          [{
+              "@id": "http://example.org/a",
+              "http://example.org/b": "http://example.org/c",
+              "http://example.org/d": "http://example.org/e"
+          }]
+          </script>
+          <script type="application/n-quads">
```
rubensworks: Again, have different contents for these next tags.

packages/actor-rdf-source-identifier-file-content-type/components/Actor/RdfSourceIdentifier/FileContentType.jsonld
```diff
@@ -31,7 +31,8 @@
             "text/n3",
             "application/ld+json",
             "application/json",
-            "application/rdf+xml"
+            "application/rdf+xml",
+            "text/html"
```
rubensworks: Can you also add `application/xhtml+xml`?

packages/actor-rdf-parse-html-script/components/Actor/RdfParse/HtmlScript.jsonld
```diff
@@ -0,0 +1,45 @@
...
+          "comment": "The RDF Parse mediator",
+          "required": true,
+          "unique": true
+        },
+        {
+          "@id": "caam:Actor/AbstractMediaTypedFixed/mediaTypeEntry",
+          "defaultScoped": {
+            "defaultScope": "carphs:Actor/RdfParse/HtmlScript",
+            "defaultScopedValue": [
+              { "mediaTypeKey": "text/html", "priorityValue": "1.0" }
```
rubensworks: Can you also add `application/xhtml+xml`? (with priority 0.9)

packages/actor-rdf-parse-html-script/test/ActorRdfParseHtmlScript-test.ts
```diff
@@ -0,0 +1,220 @@
...
+          .then(async (output) => expect(await arrayifyStream(output.handle.quads)).toEqualRdfQuadArray([
+            quad("http://example.org/a", "http://example.org/b", "http://example.org/c", ""),
+            quad("http://example.org/a", "http://example.org/d", "http://example.org/e", ""),
+            quad("http://example.org/a", "http://example.org/b", "http://example.org/c", ""),
+            quad("http://example.org/a", "http://example.org/d", "http://example.org/e", ""),
+            quad("http://example.org/a", "http://example.org/b", "http://example.org/c", ""),
+            quad("http://example.org/a", "http://example.org/d", "http://example.org/e", ""),
+          ]));
+      });
+
```
rubensworks: Could you also add a test with a script tag that has _no_ `type` attribute?

Also, a test with a script tag inside a realistic HTML document would be good as well. (with stuff like head, body, h1, ...)

pietercolpaert:
> type="application/json+ld"

should be `type="application/ld+json"`

rubensworks:
So two big todo's remaining for this:
* Call the parser in a streaming way
* Fix unit tests

Issue #313, opened 2018-11-02
When are we using Requester-browser.js and when ActorHttpNodeFetch?


RubenVerborgh:
#### Issue type:

- :question: Question

____
#### Question:

I was looking at the ActorHttpNodeFetch code, where `fetch` is used, but inside the browser `Requester-browser.js` with good old `XmlHttpRequest` is also used still. When is which used and why?

(This as part of my search process to create a module that uses `solid.auth.fetch` instead of regular `window.fetch`.)

RubenVerborgh:
`fetch` is slower? Interesting, I can't imagine why.

In any case, `Requester-browser.js` does more than just `fetch`, it also solves a Memento -related performance issue. But maybe it is worthwhile to replace `Requester` by `fetch`, and have a browser version of `fetch` doing the Memento stuff.

rubensworks:
> have a browser version of fetch doing the Memento stuff.

Indeed, there's some other stuff in there as well. I wasn't aware of that. We should indeed port this to the fetch version as well.

> But maybe it is worthwhile to replace Requester by fetch

I would keep `actor-http-native` fully `fetch`-free. I can imagine cases where people don't want any `fetch` usages at all, for example when old browsers are targeted. (or performance issues like we had before)

RubenVerborgh:
> for example when old browsers are targeted

I can't imagine us targeting browsers without `fetch`.

> (or performance issues like we had before)

That would be a problem.

I'll create a browser version of the fetch thing.

rubensworks:
> I can't imagine us targeting browsers without fetch.

Not us indeed, but our target audience should be bigger than just us :-)

> I'll create a browser version of the fetch thing.

👍 

Note that our usage of `isomorphic-fetch` automatically handles the usage of `fetch` for browsers when used outside of Node. So any additional logic should be added in `ActorHttpNodeFetch` directly.

RubenVerborgh:
I created #316 so we can see whether `fetch` is the way to go. If it is, I can port this stuff to `fetch`.

> So any additional logic should be added in ActorHttpNodeFetch directly.

Not really, because the Memento logic only applies to browsers.

PR #315, opened 2018-11-04
Replace XMLHttpRequest by fetch

RubenVerborgh:
By using the latest `fetch` interface, we can more easily swap in other libraries.

Related: #313.

Warning: untested, just ported on sight and with help from the TypeScript compiler. But I figured you guys could more easily test this in the browser than I can.


rubensworks:
Sorry, I lost track of #313, replied now.

This PR is confusing `actor-http-fetch` and `actor-http-native`.

You should be able to achieve your goal by using `actor-http-fetch` in your config file instead of `actor-http-native` (default).

RubenVerborgh:
@rubensworks There's still the Memento-related thing in there, and some header normalization.

RubenVerborgh:
Also aborting requests.

rubensworks:
Not needed anymore (#316).

PR #333, opened 2018-11-12
Split rdf resolve quad pattern qpf

Lamasaurus:
Split `actor-rdf-resolve-quad-pattern-qpf` into `actor-rdf-resolve-quad-pattern-hypermedia` and `actor-rdf-resolve-hypermedia` that are conected with a race mediator using the `bur-rdf-resolve-hypermedia` bus.

resolve_quad_pattern_split


rubensworks:
Thanks! The architecture figure looks good in any case, will try to review the code ASAP (no guarantee that I'll find time for it this week though)

Lamasaurus:
@rubensworks ok, I'll already work further on the tree ontology in the mean time.

packages/actor-abstract-mediatype-utilities/README.md
```diff
@@ -0,0 +1,15 @@
+# Comunica Abstract Binding Hash Actor
```
rubensworks: This readme seems inconsistent with the package name.

packages/actor-abstract-mediatype-utilities/lib/AbstractMediatypeUtilities.ts
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * Get the single source if the context contains just a single source.
+   * @param {ActionContext} context A context, can be null.
+   * @return {Promise<IDataSource>} A promise resolving to the single datasource or null.
+   */
+  public static async getSingleSource(context: ActionContext): Promise<IDataSource> {
+    if (context && context.has(KEY_CONTEXT_SOURCE)) {
+      // If the single source is set
+      return context.get(KEY_CONTEXT_SOURCE);
+    } else if (context && context.has(KEY_CONTEXT_SOURCES)) {
```
rubensworks: This block should not be needed, this will always be handled by the federated actor. You can assume that the context will always have a `KEY_CONTEXT_SOURCE` entry.

packages/actor-abstract-mediatype-utilities/README.md
rubensworks: Also, (why) is this package needed?

packages/actor-rdf-resolve-hypermedia-qpf/lib/ActorRdfResolveHypermediaQpf.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  /**
+   * Look for the search form
+   * @param {IActionRdfResolveHypermedia} the metadata to look for the form.
+   * @return {Promise<IActorRdfResolveHypermediaOutput>} A promise resolving to a hypermedia form.
+   */
+  public async run(action: IActionRdfResolveHypermedia):
+  Promise<IActorRdfResolveHypermediaOutput> {
+
+    if (action.metadata && action.metadata.searchForms) {
```
rubensworks: I would move this condition to your test method, as I think you might otherwise run into trouble when you register multiple actors onto this bus.

packages/actor-rdf-resolve-hypermedia-qpf/components/Actor/RdfResolveHypermedia/Qpf.jsonld
```diff
@@ -0,0 +1,68 @@
...
+  "components": [
+    {
+      "@id": "carrhq:Actor/RdfResolveHypermedia/Qpf",
+      "@type": "Class",
+      "extends": "cbrrh:Actor/RdfResolveHypermedia",
+      "requireElement": "ActorRdfResolveHypermediaQpf",
+      "comment": "A comunica QPF RDF Resolve Quad Pattern Actor.",
+      "parameters": [
+        {
+          "@id": "carrqph:Actor/RdfResolveQuadPattern/Hypermedia/subjectUri",
```
rubensworks: The `@id` of these parameters should have the prefix `carrhq:Actor/RdfResolveHypermedia/Qpf/` (because of the actor that defines these params)

packages/actor-abstract-mediatype-utilities/README.md
Lamasaurus: Because I found that the problem of looking up source types was one that reoccurred in a few different places. This way the implementation of looking for the source and comparing its type to a desired type is only implemented once. @joachimvh told me that I maybe could implement this behavior in an abstract module.

rubensworks: Indeed, abstraction is good. But I think it might be overkill in this case (when you remove support for `KEY_CONTEXT_SOURCES `), as the LOC will be very small then I think. But let's have a look at it next time you're at the office.

Lamasaurus:
@rubensworks does this branch still need changes?

rubensworks:
I'll have a look at it later today.

It looks like your rebase went wrong though, it includes commits that are already on master.

Lamasaurus:
> It looks like your rebase went wrong though, it includes commits that are already on master.
Oh yeah your right, I fixed it.

packages/actor-abstract-mediatype-utilities/lib/AbstractMediatypeUtilities.ts
```diff
@@ -0,0 +1,58 @@
...
+  IDataSource,
+  KEY_CONTEXT_SOURCE,
+  KEY_CONTEXT_SOURCES,
+} from "@comunica/bus-rdf-resolve-quad-pattern";
+import {ActionContext} from "@comunica/core";
+
+/**
+ * A comunica Hash Query Operation Actor.
+ */
+export abstract class AbstractMediatypeUtilities {
```
rubensworks: Since all methods in this class are static, and nothing will extend from it, it is not needed to have config files for this. So you can essentially remove the `components` folder in this package.

Also, I still don't really like this name of this class and package. What about `DataSourceUtils` (package name: `util-datasource`) ?

packages/actor-rdf-resolve-hypermedia-qpf/lib/ActorRdfResolveHypermediaQpf.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  public async test(action: IActionRdfResolveHypermedia): Promise<IActorTest> {
+    if (!(action.context.get(KEY_CONTEXT_SOURCE).type === "hypermedia")) {
+      throw new Error(this.name
+        + ' requires a single source with a Hypermedia \'hypermedia\' entrypoint to be present in the context.');
+    }
+
+    if (!action.metadata || !action.metadata.searchForms) {
+      throw new Error(`${this.name} requires metadata and searchForms to work on.`);
```
rubensworks: This error message can be improved. Something like this would be good: https://github.com/comunica/comunica/blob/master/packages/actor-rdf-resolve-quad-pattern-qpf/lib/ActorRdfResolveQuadPatternQpf.ts#L53-L54

packages/actor-rdf-resolve-hypermedia-qpf/lib/ActorRdfResolveHypermediaQpf.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    // TODO: in the future, a query-based search form getter should be used.
+    for (const searchForm of searchForms.values) {
+      if (this.graphUri
+        && this.subjectUri in searchForm.mappings
+        && this.predicateUri in searchForm.mappings
+        && this.objectUri in searchForm.mappings
+        && this.graphUri in searchForm.mappings
+        && Object.keys(searchForm.mappings).length === 4) {
+        return {searchForm};
```
rubensworks: `return { searchForm };`

packages/actor-rdf-resolve-hypermedia-qpf/lib/ActorRdfResolveHypermediaQpf.ts
```diff
@@ -0,0 +1,74 @@
...
+      }
+      if (this.subjectUri in searchForm.mappings
+        && this.predicateUri in searchForm.mappings
+        && this.objectUri in searchForm.mappings
+        && Object.keys(searchForm.mappings).length === 3) {
+        return {searchForm};
```
rubensworks: `return { searchForm };`

packages/actor-rdf-resolve-quad-pattern-hypermedia/package.json
```diff
@@ -35,20 +35,23 @@
     "index.js"
   ],
   "dependencies": {
```
rubensworks: No need to downgrade these dependency versions :-)

packages/actor-rdf-resolve-hypermedia-qpf/lib/ActorRdfResolveHypermediaQpf.ts
rubensworks: Actually, never mind, such an error is thrown in another actor.

rubensworks:
Oh, also: testing coverage for `ActorRdfResolveQuadPatternHypermedia.ts` is not at 100% yet according to Travis.

Issue #121, opened 2018-03-21
Metadata extraction does not seem to follow the spec


RubenVerborgh:
#### Issue type:

- :bug: Bug

#### Description:

It seems that #15 does not fully follow the TPF spec at https://www.hydra-cg.com/spec/latest/triple-pattern-fragments/.

What should happen is that the server looks for
```?dataset void:subset <CURRENTLOCATION>;
    hydra:search [
        hydra:template ?template;
        hydra:mapping  [ hydra:variable ?s; hydra:property rdf:subject ],
                       [ hydra:variable ?p; hydra:property rdf:predicate ],
                       [ hydra:variable ?o; hydra:property rdf:object ]
    ].
```

IF a quad-based format, then that pattern MUST be matched in a named graph
and that named graph SHOULD be related to the document.

Note that foaf:primaryTopic doesn’t play any role in this, except for SHOULD.

RubenVerborgh:
Bonus: implement this as a SPARQL query rather than as a hard-coded lookup algorithm.

rubensworks:
Ok, so `actor-rdf-metadata-primary-topic` is indeed too strict.
This can easily be fixed by removing that actor from the default config.
(*I won't do this yet until I can get a second opinion on this*)

We should however make a smarter mediator in the future that can still make use of that actor _if_ `foaf:primaryTopic` is present. And we should probably also add another actor that does the named graph match thingy as mentioned above.

However, @pietercolpaert's query still fails for me even if I disable this actor:
```
https://linked.open.gent/parking "SELECT * WHERE { ?s ?p ?o }"
```

I did some debugging, and it looks like the metadata of https://linked.open.gent/parking contains the following:
```
<https://linked.open.gent/parking#mappingS> hydra:variable "s".
<https://linked.open.gent/parking#mappingP> hydra:variable "p".
<https://linked.open.gent/parking#mappingO> hydra:variable "o".
<https://linked.open.gent/parking#mappingS> hydra:property "subject".
<https://linked.open.gent/parking#mappingP> hydra:property "property".
<https://linked.open.gent/parking#mappingO> hydra:property "object".
```
This is however incorrect, as `hydra:property` has a URI as range AFAIK, not a string, so this should be:
```
<https://linked.open.gent/parking#mappingS> hydra:variable "s".
<https://linked.open.gent/parking#mappingP> hydra:variable "p".
<https://linked.open.gent/parking#mappingO> hydra:variable "o".
[...]

PR #346, opened 2018-11-28
Add ResolveHypermediaNextPage actor

Lamasaurus:
This actor will return a `searchForm` that follows `hydra:next` links and just returns all quads until the `LIMIT` is reached.


packages/actor-rdf-resolve-hypermedia-next-page/package.json
```diff
@@ -0,0 +1,78 @@
...
+  },
+  "homepage": "https://github.com/comunica/comunica#readme",
+  "files": [
+    "components",
+    "lib/**/*.d.ts",
+    "lib/**/*.js",
+    "index.d.ts",
+    "index.js"
+  ],
+  "dependencies": {
```
rubensworks: I think all of these dependencies can be removed.

Lamasaurus:
This can just be tested on `http://fragments.dbpedia.org/2015-10/en` for example, but 'actor-rdf-resolve-hypermedia-qpf' should be disabled in some way.

rubensworks:
@Lamasaurus Can it be tested somewhere else? Because I want to test it with all modules enables, to see if they definitely don't conflict with each other.

Lamasaurus:
You can test it on this dataset: `https://lamasaurus.github.io/data/fr1.jsonld`. Its hydra:next will link to the same file again, so limit the amount of answers you want.

rubensworks:
@pietercolpaert Could you try this out and have a first look at the code? I probably won't have time for this until next week.

pietercolpaert:
> You can test it on this dataset: `https://lamasaurus.github.io/data/fr1.jsonld`. Its hydra:next will link to the same file again, so limit the amount of answers you want.

I confirm this works. However, @rubensworks, I guess we want to be cautious not to end up in cycles when querying the data... So probably in the scope of the next page actor we want to check whether a certain response url was already encountered when traversing this paged collection. What do you think?

Another dataset we can test on could be https://graph.irail.be/nmbs/connections, but due to a bug in the actor extracting TPF controls, this actor does not get triggered. (opening a new issue for this)

Yet another dataset could be https://hetarchief.be/nl/media/ons-vlaanderen/T2HCWUXLMRUVTOeJbUeCoWUS, but the `text/html` content-type is not supported yet. See #379 for the progress on this.

Any other datasets out there with hydra:next links we could test?

pietercolpaert:
Another thought: should we also support the Link header? Cfr [RFC 5988 ](https://tools.ietf.org/html/rfc5988)
```
   Link: </TheBook/chapter2>;
         rel="previous"; title*=UTF-8'de'letztes%20Kapitel,
         </TheBook/chapter4>;
         rel="next"
```

rubensworks:
> So probably in the scope of the next page actor we want to check whether a certain response url was already encountered when traversing this paged collection.

Personally, I don't think this is going to be a big problem in practise, so I would not implement something for this until we start noticing problems. Either way, implementing this should not be too complicated, we can keep state in the context

> Another thought: should we also support the Link header?

Sure, we should track this in a separate issue.

packages/actor-rdf-resolve-hypermedia-qpf/lib/ActorRdfResolveHypermediaQpf.ts
```diff
@@ -29,7 +30,26 @@ implements IActorRdfResolveHypermediaQpfArgs {
       throw new Error(`${this.name} requires metadata and searchForms to work on.`);
     }
 
-    return true;
+    // Find a quad pattern or triple pattern search form
+    const searchForms: ISearchForms = action.metadata.searchForms;
+    
+    // TODO: in the future, a query-based search form getter should be used.
```
pietercolpaert: Is this the reason for #415? Should we immediately do it right here?

packages/actor-rdf-resolve-quad-pattern-hypermedia/lib/ActorRdfResolveQuadPatternHypermedia.ts
```diff
@@ -54,7 +54,7 @@ export class ActorRdfResolveQuadPatternHypermedia extends ActorRdfResolveQuadPat
     }
     const metadata: {[id: string]: any} = await firstPageMetadata();
 
-    if (!metadata.searchForms || !metadata.searchForms.values.length) {
+    if (!metadata.next && (!metadata.searchForms || !metadata.searchForms.values.length)) {
```
pietercolpaert: Also datasets with a search form and with a next page should trigger this actor, but probably they will have to trigger it from their code, as sometimes fetching the next page is not the right thing to do during the query processing.

@rubensworks What do you think about this? This means of course that the QPF actor would be able to overrule this actor and say it is better at handling things with a search form with an spo in it, and that inside the QPF actor, you would call this actor when you want to get the next page.

packages/actor-rdf-resolve-hypermedia-qpf/lib/ActorRdfResolveHypermediaQpf.ts
rubensworks: I'm afraid this TODO is not that trivial to resolve, so I would leave it for a separate PR.

packages/actor-rdf-resolve-quad-pattern-hypermedia/lib/ActorRdfResolveQuadPatternHypermedia.ts
rubensworks: Yeah, this check is definitely too strict, I'd remove the `metadata.next` check. With this change,  Comunica will always fail when for example iterating over TPF pages when it reaches the last page.

Actors down the line should indeed invoke the next page themselves when needed.

I'll have to look into this PR in more detail, but I'm quite sure no changes should be needed in this file.

Lamasaurus: If I didn't add this to the check the next page actor was never invoked because there are no `searchForms` in the document. The next page actor makes up its own `searchForm` out of the `hydra:next` link while the `ActorRdfResolveHypermediaQpf` just returns the one that is present in the source. Maybe this check should not happen when it is possible that actors create their own `searchForms`?

packages/actor-rdf-resolve-hypermedia-qpf/lib/ActorRdfResolveHypermediaQpf.ts
```diff
@@ -14,6 +14,7 @@ implements IActorRdfResolveHypermediaQpfArgs {
   public readonly predicateUri: string;
   public readonly objectUri: string;
   public readonly graphUri?: string;
+  private searchForm: ISearchForm;
```
rubensworks: Actor should (almost) never be stateful. Maintaining a state like this can cause weird problems when test/run are run out-of-order.
Instead, you should re-determine the searchForm in both test and run.

packages/actor-rdf-resolve-hypermedia-qpf/lib/ActorRdfResolveHypermediaQpf.ts
```diff
@@ -29,7 +30,26 @@ implements IActorRdfResolveHypermediaQpfArgs {
...
+    
+    // TODO: in the future, a query-based search form getter should be used.
+    for (this.searchForm of searchForms.values) {
+      if ((this.graphUri
+        && this.subjectUri in this.searchForm.mappings
+        && this.predicateUri in this.searchForm.mappings
+        && this.objectUri in this.searchForm.mappings
+        && this.graphUri in this.searchForm.mappings
+        && Object.keys(this.searchForm.mappings).length === 4) || 
+        (this.subjectUri in this.searchForm.mappings
```
rubensworks: Indentation seems a bit off here.

packages/actor-rdf-resolve-hypermedia-qpf/test/ActorRdfResolveHypermediaQpf-test.ts
```diff
@@ -87,13 +87,25 @@ describe('ActorRdfResolveHypermediaQpf', () => {
...
   });
 
   describe('#run', () => {
     it('should return a searchForm', () => {
-      return expect(actor.run({metadata, context: ActionContext
+      const action = {metadata, context: ActionContext
         ({ '@comunica/bus-rdf-resolve-quad-pattern:source': { type: 'hypermedia', value: 'source' }}),
-      })).resolves.toEqual({ searchForm: metadata.searchForms.values[0] });
+      };
+      actor.test(action);
```
rubensworks: This should not be needed.

packages/actor-rdf-resolve-hypermedia-qpf/test/ActorRdfResolveHypermediaQpf-test.ts
```diff
@@ -110,18 +122,12 @@ describe('ActorRdfResolveHypermediaQpf', () => {
           },
         ]},
       };
-      return expect(actor.run({metadata, context: ActionContext
+      const action = {metadata, context: ActionContext
         ({ '@comunica/bus-rdf-resolve-quad-pattern:source': { type: 'hypermedia', value: 'source' }}),
-      })).resolves.toEqual({ searchForm: metadata.searchForms.values[0] });
+      };
+      actor.test(action);
```
rubensworks: This should not be needed.

packages/actor-rdf-resolve-quad-pattern-hypermedia/lib/ActorRdfResolveQuadPatternHypermedia.ts
rubensworks: I would suggest removing this check completely then.

packages/actor-init-sparql/config/sets/rdf-source-identifiers.json
```diff
@@ -18,6 +18,17 @@
...
+      "@id": "config-sets:rdf-source-identifiers.json#myActorRdfSourceIdentifierHypermediaNextPage",
+      "@type": "ActorRdfSourceIdentifierHypermediaQpf",
+      "carsihq:Actor/RdfSourceIdentifier/HypermediaQpf/mediatorHttp": {
+        "@id": "config-sets:http.json#mediatorHttp"
+      },
+      "carsihq:Actor/RdfSourceIdentifier/HypermediaQpf/toContain": [
+            "http://www.w3.org/ns/hydra/core#",
+            "next"
+      ]
+    },
```
rubensworks: Until https://github.com/comunica/comunica/issues/121#issuecomment-394976570 is fixed, this is indeed the only way to achieve the desired behaviour.

However, this will unfortunately require an additional HTTP request _for each query_ we do, which is not really desired.
We want to avoid bloating the default Comunica engine too much.

I would suggest not adding this next-page actor (and source identifier) to the default config, and instead create a dedicated (external) config for this. (For example: https://github.com/RubenVerborgh/LDflex-Comunica/commit/14624cfaf77402ee813533bcf566abb974d676c5)

@pietercolpaert Would creating a custom Comunica instance somewhere based on a separate config file make sense for your use case?

pietercolpaert: And also a separate one then for every new hydra:search form or hypermedia case we come up with? I think that would be detrimental to the idea of Comunica, although it could be a temporary fix.

Is it another option to prioritize https://github.com/comunica/comunica/issues/121#issuecomment-394976570 and fix this this week?

rubensworks: > Is it another option to prioritize #121 (comment) and fix this this week?

Not on my end no. But @Lamasaurus is welcome to have a go at it.

Lamasaurus:
@rubensworks I fixed the docker file, should I squash the commits?

rubensworks:
@Lamasaurus No need to squash, I can do that from GitHub.

pietercolpaert:
@rubensworks Can we now go ahead with this temporary fix until #121 is ok? Are the requested changes still valid?

rubensworks:
I assume this is not ready for review yet?

From a quick glance I see that there is a bunch of stuff in `actor-init-sparql-next-page` that was copy-pasted from `actor-init-sparql` that should not be needed. Instead, an abstraction like `actor-init-sparql-file` should be followed.

pietercolpaert:
This can be closed in favour of https://github.com/comunica/comunica/tree/refactor/metadata?

rubensworks:
Yep!

PR #362, opened 2018-12-19
Fix the Docker build for actor-init-sparql

vemonet:
Fixed the Dockerfile to implement new installation process `npm install -g @comunica/actor-init-sparql` . Now working properly

Build:
```shell
docker build -t comunica-sparql .
```

Run:
```shell
docker run -it comunica-sparql http://fragments.dbpedia.org/2015-10/en "CONSTRUCT WHERE { ?s ?p ?o } LIMIT 100"
```

### DIFF
--- a/packages/actor-init-sparql/Dockerfile
+++ b/packages/actor-init-sparql/Dockerfile
@@ -3,19 +3,12 @@ FROM node:10
 # Install location
 ENV dir /var/www/@comunica/actor-init-sparql/
 
-# Copy the engine files (generated from package.json!files)
-COPY components ${dir}/components/
-COPY config ${dir}/config/
-COPY lib/*.js ${dir}/lib/
-COPY bin/*.js ${dir}/bin/
-COPY index.js index-browser.js engine-default.js package.json ${dir}
-
-# Install the node module
-RUN cd ${dir} && npm install --only=production
-
 # Run base binary (generated from package.json!bin)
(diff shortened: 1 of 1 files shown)


rubensworks:
Thanks for the PR!

However, I won't be able to accept this in its current form.
Our dockerfiles are meant to be self-contained, i.e., they should be built from the source. Instead, you PR changes this to be built with the latest release from npm.
What was the reason for changing this, doesn't it work for you?

Note that we are automatically pushing Docker images upon every release, so you should be able to pull them in directly without building from here: https://hub.docker.com/r/comunica/actor-init-sparql/tags

vemonet:
OK, thanks! 

When I try to build actor-init-sparql with `docker build -t comunica-sparql ./packages/actor-init-sparql`, I got the following: 

```
Step 5/11 : COPY lib/*.js ${dir}/lib/
COPY failed: no source files were specified
```

And it seems normal since there is no .js file in the project, only ts. I guess a preprocessing step is needed here?

I made it work using `comunica/actor-init-sparql`, thanks a lot!

rubensworks:
> And it seems normal since there is no .js file in the project, only ts. I guess a preprocessing step is needed here?

Ah, indeed, you'll have to do a `yarn install` first if you want to build from the sources.

> I made it work using comunica/actor-init-sparql, thanks a lot!

Great! That's indeed a better solution :-)

PR #371, opened 2019-01-08
Check HTTP stream type when identifying source

joachimvh:
So the fetch problem in the browser was actually not related to the fetch library being outdated (although that could still have an impact in the command line version), but due to the fact that the stream was not converted properly when doing source type identification on QPF sources

### DIFF
--- a/packages/actor-rdf-source-identifier-hypermedia-qpf/lib/ActorRdfSourceIdentifierHypermediaQpf.ts
+++ b/packages/actor-rdf-source-identifier-hypermedia-qpf/lib/ActorRdfSourceIdentifierHypermediaQpf.ts
@@ -31,7 +31,9 @@ export class ActorRdfSourceIdentifierHypermediaQpf extends ActorRdfSourceIdentif
     const httpAction: IActionHttp = { context: action.context, input: action.sourceValue, init: { headers } };
     const httpResponse: IActorHttpOutput = await this.mediatorHttp.mediate(httpAction);
     if (httpResponse.ok) {
-      const body = (await require('stream-to-string')(httpResponse.body));
+      const stream = require('is-stream')(httpResponse.body)
+        ? httpResponse.body : require('node-web-streams').toNodeReadable(httpResponse.body);
+      const body = (await require('stream-to-string')(stream));
 
       // Check if body contains all required things
       let valid = true;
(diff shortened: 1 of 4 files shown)


packages/actor-rdf-source-identifier-hypermedia-qpf/lib/ActorRdfSourceIdentifierHypermediaQpf.ts
```diff
@@ -31,7 +31,9 @@ export class ActorRdfSourceIdentifierHypermediaQpf extends ActorRdfSourceIdentif
     const httpAction: IActionHttp = { context: action.context, input: action.sourceValue, init: { headers } };
     const httpResponse: IActorHttpOutput = await this.mediatorHttp.mediate(httpAction);
     if (httpResponse.ok) {
-      const body = (await require('stream-to-string')(httpResponse.body));
+      const stream = require('is-stream')(httpResponse.body)
+        ? httpResponse.body : require('node-web-streams').toNodeReadable(httpResponse.body);
+      const body = (await require('stream-to-string')(stream));
```
rubensworks: It looks like we're using this kind of logic in 4 different places now.
I'm wondering if we should either
* move this body stream conversion higher up into the HTTP actor(s)
* or abstract this code so that it can be converted with a single function.

joachimvh: Yea this isn't so clean at the moment. The problem is also that it's completely not obvious that you have to do this when you call the http actor (hence this bug). On the other hand, moving it to the http actor would require a change to the output type of the HTTP actor (which currently corresponds to the existing Response interface), which would make it a bit less prettier I guess, and might be a problem for actors that actually depend on the current behaviour (which I think only happens here? https://github.com/comunica/comunica/blob/master/packages/actor-http-memento/lib/ActorHttpMemento.ts#L48 )

rubensworks: Hmm, indeed.

In that case I would abstract this stream conversion pattern, so that we can reuse is in all relevant parts of the code.

joachimvh:
Added function to ActorHttp

PR #378, opened 2019-01-22
Also check decoded page URL with hydra controls

brechtvdv:


pietercolpaert:
I think this one can be closed in favor of https://github.com/comunica/comunica/tree/refactor/metadata?

rubensworks:
Not this one I think. @brechtvdv and I identified the root-cause of the problem, and it's actually a problem with the website of https://hetarchief.be/.

@brechtvdv Maybe you can share the details of the problem here before we close this PR? So that other people experiencing similar problems in the future know what is going on.

PR #380, opened 2019-01-22
Copy sources so the input array remains unchanged

### LINKED ISSUE #377: newEngineDynamic stops working after several reuses
#### Issue type:

- :bug: Bug

____
#### Description:

Following program:
```javascript

const newEngineDynamic = require('@comunica/actor-init-sparql').newEngineDynamic;

const query = `
SELECT ?name
{ 
    ?s dbpedia-owl:birthPlace <http://dbpedia.org/resource/Belgium>.
    ?s foaf:name ?name.
    ?s dbpedia-owl:birthPlace [ rdfs:label "Ghent"@en ].
}
`;
const sources = [ { type: 'hypermedia', value: 'http://fragments.dbpedia.org/2015/en' } ];

run(1);

function run (count) {
    console.log(`run ${count}`);
    newEngineDynamic().then(function (myEngine) {
        myEngine.query(query, { sources: sources })
            .then((result) => {
                console.log('got result');

                result.bindingsStream.on('data', data => {} );
                result.bindingsStream.on('end', () => {
                    console.log('ended');
                    run(count+1);
                });

            });
    }).catch(console.error);
}
```
The process stops at some point with the following output:
```
run 1
got result
ended
run 2
got result
ended
run 3
got result
ended
run 4
```

____
#### Environment:

1.5.2

joachimvh:
This fixes #377 .

The fix is quite easy, less so the reason for the fix. :D

The problem is that the context preprocessor pushed a `null` into the sources array after parsing the sources to indicate the end of the sources array ( https://github.com/comunica/comunica/blob/master/packages/actor-context-preprocess-rdf-source-identifier/lib/ActorContextPreprocessRdfSourceIdentifier.ts#L30 ). Problem is that since this was the same array, in the next run there would be 2 nulls at the end of the array, then 3, etc. Funnily enough this is actually not a problem until you reach the 4th iteration. Reason for that can actually be found in the ArrayIterator code. 

The ArrayIterator only closes the stream when the array has been fully read, no matter what is in it ( https://github.com/RubenVerborgh/AsyncIterator/blob/master/asynciterator.js#L574 ). The emitData function on the other hand, that is responsible for sending out data to data listeners, reads an element from the iterator, and emits it to the listeners **if it is not null**. ( https://github.com/RubenVerborgh/AsyncIterator/blob/master/asynciterator.js#L373 ).

So what happens is that the emitData function doesn't get called again since there's no new 'readable' trigger. But the ArrayIterator is not finished yet since there are still null elements in the array so it never fires its end event, causing the actor to never finish the preprocessing. Not sure why this still works in the first iterations that have multiple null elements. There are multiple emitData calls causing the nulls to still be cleared there, but at 4 there are no longer enough extra calls.

The reason for all this text is because I'm not sure if this is something that might need to be fixed in AsyncIterator or not :D

### DIFF
--- a/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
+++ b/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
@@ -116,7 +116,7 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
 
     // Ensure sources are an async re-iterable
     if (Array.isArray(context[KEY_CONTEXT_SOURCES])) {
-      context[KEY_CONTEXT_SOURCES] = AsyncReiterableArray.fromFixedData(context[KEY_CONTEXT_SOURCES]);
+      context[KEY_CONTEXT_SOURCES] = AsyncReiterableArray.fromFixedData(context[KEY_CONTEXT_SOURCES].slice());
     }
 
     context = ActionContext(context);


rubensworks:
Oh, good catch!

I actually think this is something that should be fixed in [AsyncReiterable](https://github.com/rubensworks/asyncreiterable.js/blob/master/lib/AsyncReiterableArray.ts) instead. We indeed seem to be modifying the original array there, which is incorrect.

joachimvh:
That fixes the problem of changing the array (which would already be good of course), but doesn't yet solve the problem if you have an array with multiple null elements in it though. So that would remain a problem

rubensworks:
> but doesn't yet solve the problem if you have an array with multiple null elements in it though.

I don't think this would be a valid state, right?

joachimvh:
What do you mean with a valid state? You can provide an array with multiple null elements as input to the iterator, or push multiple nulls to the iterator. I would still expect to receive an `end` trigger from the iterator once it reaches a null element

joachimvh:
Ah k. Should there then maybe be checks in place to make sure this can never occur? E.g., don't allow pushing a null if there already is one at the end of an array.

rubensworks:
> Should there then maybe be checks in place to make sure this can never occur? E.g., don't allow pushing a null if there already is one at the end of an array.

Pushing into an AsyncReiterable you mean?

Actually, that should already be caught: https://github.com/rubensworks/asyncreiterable.js/blob/master/lib/AsyncReiterableArray.ts#L68-L70
But it looks like there might be a bug in there, in case `this.iterators` is still empty at push-time.

joachimvh:
I just noticed that I was wrong in stating where the extra nulls come from actually. (I was surprised that safety code you linked would not work so had another look). The extra nulls actually get added here: https://github.com/rubensworks/asyncreiterable.js/blob/master/lib/AsyncReiterableArray.ts#L15-L17 which can keep adding them if they're already there.

rubensworks:
Indeed, so the only change needed there should be to change line 13 to `this.array = array.slice();`, right?

joachimvh:
Yes that would solve the duplication problem. But I would also add a check to the if statement on line 15 to only add the null if the last element isn't null already.

rubensworks:
Sure, that makes sense.

PR #396, opened 2019-02-20
Pass parent metadata to BGP and pattern actions

rubensworks:
This is a small addition that passes around (parent) metadata to (child) query operations. This is needed for our [AMF-aware actors](https://github.com/comunica/comunica-feature-amf).


rubensworks:
The two different keys are needed because they point to different things. The one is an array of hashes, and the other one is a single metadata hash.

You know, this just made me think of an improvement for our context entries: #398

PR #399, opened 2019-02-21
Abstract SPARQL HTTP service

### LINKED ISSUE #184: Abstract http bin
#### Issue type:

____
#### Description:

The HTTP script is currently hardcoded to work with actor-init-sparql, while this would be useful for our other init actors as well (such as actor-init-sparql-hdt). We should abstract this in the same way as our query bin scripts.

rubensworks:
This allows other init actors to easily expose themselves
as a SPARQL endpoint.

Closes #184

### DIFF
--- a/packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
+++ b/packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
@@ -1,0 +1,218 @@
+import * as http from "http";
+import * as querystring from "querystring";
+import {Writable} from "stream";
+import * as url from "url";
+import {newEngineDynamic} from "../index";
+import {ActorInitSparql} from "./ActorInitSparql";
+import {IQueryOptions} from "./QueryDynamic";
+import EventEmitter = NodeJS.EventEmitter;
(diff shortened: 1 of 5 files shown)


packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -0,0 +1,218 @@
...
+      context.log = new LoggerPretty({ level: args.l || 'warn' });
+    }
+
+    const options = {
+      configResourceUrl: process.env.COMUNICA_CONFIG
+        ? process.cwd() + '/' + process.env.COMUNICA_CONFIG : defaultConfigPath,
+      context,
+      mainModulePath: moduleRootPath,
+      port,
+      timeout,
```
joachimvh: Just wondering, why are these extra parameters now added to the options object?

rubensworks: Because `IHttpServiceSparqlEndpointArgs ` extends from `IQueryOptions `. I wanted to have only a single args object :-)

PR #421, opened 2019-03-11
Make some actors test function consistent and reflective of actual behaviour

wschella:
During the implementation of some actors I did, i had a wrong impression of how the testOperation function worked. These changes reflect the actual wanted behavior without misleading double checks that are already done in the abstract class.


wschella:
The travis fail is just the node / jest bug that should be resolved by merging.

PR #442, opened 2019-04-29
Add CORS headers for SPARQL interface

joachimvh:
Just discovered this problem while working on the tutorial and trying to query my own sparql endpoint through the widget.


rubensworks:
Good catch!

Issue #465, opened 2019-06-27
Add CORS proxy


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request

____
#### Description:

Since we've been having some problems with CORS (#439, #373), the jQuery widget is still failing on a lot of sources that don't set CORS headers, and there are problems with loading mixed schemes (HTTP/HTTPS), we should have a proper workaround.

We could do this by simply setting up a proxy server, such as https://corsproxy.github.io/.

It should be possible to pass custom proxies via the config file.

We may consider only using the CORS proxy if requests fail due to CORS.

RubenVerborgh:
> It should be possible to pass custom proxies via the config file.

People can list their preference for a proxy server in their WebID profile. One the one hand, this helps scalability. On the other hand, it helps privacy/trust from the perspective of the user.

rubensworks:
From offline discussion: the proxy should be passable via the context, instead of the config, so that late-binding is possible.

RubenVerborgh:
Maybe there's even different proxies for different sources, etc. I mean, we don't have to make this too complex, but on the interface level, supporting something like `getProxyFor(source, …)` might make sense.

rubensworks:
Hmm, true. Also needed because some sources may not even need proxying, for example when the CORS headers are properly set on the sources (but this is something that we can only know about _after_ the initial request).

PR #475, opened 2019-07-19
HTTP proxy actor

rubensworks:
This PR adds a generic HTTP proxy, which will be needed for handling CORS issues in browsers (#465).

### DIFF
--- a/packages/core/components/Actor.jsonld
+++ b/packages/core/components/Actor.jsonld
@@ -13,6 +13,11 @@
           "range": "cc:Bus",
           "unique": true,
           "required": true
+        },
+        {
+          "@id": "cc:Actor/beforeActor",
+          "comment": "An actor that must be registered in the bus before this actor.",
+          "range": "cc:Actor"
         }
       ],
       "constructorArguments": [
@@ -26,6 +31,10 @@
             {
               "keyRaw": "bus",
               "value": "cc:Actor/bus"
+            },
+            {
+              "keyRaw": "beforeActors",
+              "value": "cc:Actor/beforeActor"
             }
           ]
         }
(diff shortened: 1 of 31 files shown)


rubensworks:
For reference, this proxy can be tested with: http://proxy.linkeddatafragments.org/

packages/core/components/Actor.jsonld
```diff
@@ -13,6 +13,11 @@
           "range": "cc:Bus",
           "unique": true,
           "required": true
+        },
+        {
+          "@id": "cc:Actor/beforeActor",
```
joachimvh: why wasn't this called "dependency"?

packages/core/lib/Bus.ts
```diff
@@ -122,6 +124,42 @@ export class Bus<A extends Actor<I, T, O>, I extends IAction, T extends IActorTe
...
+    for (const dependency of dependencies) {
+      this.dependencyLinks.push({ dependent, dependency });
+    }
+    this.reorderForDependencies();
+  }
+
+  /**
+   * Reorder the bus based on all present dependencies.
+   */
+  public reorderForDependencies() {
```
joachimvh: This can give incorrect results if there are chained dependencies, depending on the order.
Example: 
  dependencyLinks: [ B depends on C, A depends on B ]
  this.actors = [ A, B, C ]
  After running this function the result will be [ B, A, C ], while C should be before B.

packages/core/components/Actor.jsonld
rubensworks: To make the direction of dependency a bit clearer. For example, `dependency` may be interpreted as `afterActor` as well.

packages/core/lib/Bus.ts
rubensworks: Yeah, I was a bit lazy in my initial implementation.
I've rewritten the whole thing with proper chained dependency support, and added a test for your case.

rubensworks:
@joachimvh Should you see any other problems with this, feel free to open an issue.

PR #483, opened 2019-08-05
Remove unnecessary code

ManuDeBuck:
Since every command line argument (and thus also `args.q`) will be a string this seems unnecessary code in `ActorInitSparql`.


rubensworks:
@ManuDeBuck Apparently there is a unit tests that fails now. This makes sense, as the test checks if execution fails when no further args are present after `-q`, and this does not fail anymore now.

It looks like `args.q` is converted to a boolean in that situation, so it seems to make sense for this `typeof` check to be there. So this change does not seem to be needed after all.

Perhaps your PR could be changed to simply add a comment that clarifies this?

PR #542, opened 2019-10-08
Support newest Sparqlalgebra.js version

joachimvh:
This should make Comunica support the new sparqlalgebra.js/sparql.js. There were two tests that were being annoying where I didn't find a better solution. Feel free to improve.

### DIFF
--- a/packages/actor-sparql-parse-algebra/test/ActorSparqlParseAlgebra-test.ts
+++ b/packages/actor-sparql-parse-algebra/test/ActorSparqlParseAlgebra-test.ts
@@ -43,8 +43,11 @@ describe('ActorSparqlParseAlgebra', () => {
       return expect(actor.test({ query: 'a', queryFormat: 'sparql' })).resolves.toBeTruthy();
     });
 
-    it('should run', () => {
-      return expect(actor.run({ query: "SELECT * WHERE { ?a a ?b }" })).resolves.toMatchObject(
+    it('should run', async () => {
+      const result = await actor.run({ query: "SELECT * WHERE { ?a a ?b }" });
+
+      // TODO: I am unable to find why the match only works like this
+      return expect(JSON.parse(JSON.stringify(result))).toMatchObject(
         {
           operation: {
             input: {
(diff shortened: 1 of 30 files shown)


packages/actor-sparql-parse-algebra/test/ActorSparqlParseAlgebra-test.ts
```diff
@@ -43,8 +43,11 @@ describe('ActorSparqlParseAlgebra', () => {
       return expect(actor.test({ query: 'a', queryFormat: 'sparql' })).resolves.toBeTruthy();
     });
 
-    it('should run', () => {
-      return expect(actor.run({ query: "SELECT * WHERE { ?a a ?b }" })).resolves.toMatchObject(
+    it('should run', async () => {
+      const result = await actor.run({ query: "SELECT * WHERE { ?a a ?b }" });
+
+      // TODO: I am unable to find why the match only works like this
```
rubensworks: Maybe because of the `equals` methods in the quads and terms?

It may actually be good to use a dedicated matcher for these tests, like here: https://github.com/RubenVerborgh/SPARQL.js/blob/master/test/matchers/toEqualParsedQuery.js
But it may require quite some effort, so not sure if this should be done for this PR...

(I'll look into more detail at this PR later)

rubensworks:
Note to self:

Libs to update when merging this:
* fetch-sparql-endpoint
* graphql-ld
* sparqlee
* graphql-to-sparql
* comunica-actor-rdf-resolve-quad-pattern-ostrich
* comunica-actor-rdf-resolve-quad-pattern-hdt

joachimvh:
How are you supposed to run those tests locally before pushing? Because both `lerna run integration` and running the integration script in one of the specific packages doesn't work

rubensworks:
> Because both lerna run integration and running the integration script in one of the specific packages doesn't work

Well, that should work :-)
What error are you getting?

joachimvh:
It was the "can't create a class without new" bug again, fixed by rebuilding.

The actual problem here seems to be because of the constructor of the SPARQL.js parser changing which wasn't caught by any test (because usually it's not relevant) so looking into that now.

rubensworks:
I also see that there a few spec tests started failing with this update:

* GROUP_CONCAT 2
* COUNT 7
* COUNT 6
* COUNT 5
* COUNT 4

@joachimvh Any idea if this is caused by sparqljs or sparqlalgebrajs?

(syn-bad-01 is fixed with this though 🎉 )

joachimvh:
All of them have a `COUNT(*)`. In the latest version of SPARQL.js the wildcard is now represented as a new Term (Wildcard), which SparqlAlgebra.js copied. This is also the main change of this pull request in Comunica.

The fact that these tests are failing means that there is something somewhere else that will also need to change for this. Not sure yet if it is in Comunica or Sparqlee.

rubensworks:
Ah yes, probably sparqlee then indeed.
We should update sparqlee as well before as part of this PR, to avoid breaking things.

PR #564, opened 2019-11-07
Add missing jsonld dependency to actor-rdf-serialize-jsonld

Vinnl:
It is `require`d in [ActorRdfSerializeJsonLd.ts:24](https://github.com/comunica/comunica/blob/master/packages/actor-rdf-serialize-jsonld/lib/ActorRdfSerializeJsonLd.ts#L24).


rubensworks:
Thanks for the PR.

However, the intention was to move away from the jsonld lib (f93753e93f2bc739fb3b475cafbbbd81665ee861). But it looks like the migration to the new serializer was not done properly (my bad...).

Instead of re-adding the jsonld dependency, I would suggest to just remove the `jsonLd` field, as it is unused at the moment.

Vinnl:
@rubensworks That makes sense, I've updated the PR.

rubensworks:
Thanks!

PR #587, opened 2019-11-27
Swap node-web-streams package to web-streams-node.

### LINKED ISSUE #275: Remove dependency on node-web-streams
The [node-web-streams](https://www.npmjs.com/package/node-web-streams) dependency is using a [git-based dependency](https://github.com/gwicke/node-web-streams/issues/1), and they don't seem to be keen on using a proper npm-based dependency.
So we should move to another package that does not use any git-based dependencies.

Git-based dependencies seem to cause issues for some people when installing, so we should avoid these kinds of dependencies.

mielvds:
Fixes #275 with a single package.json edit in `bus-http`

```
Test Suites: 147 passed, 147 total
Tests:       2071 passed, 2071 total
Snapshots:   0 total
Time:        18.17s, estimated 45s
Ran all test suites.
```

### DIFF
--- a/packages/actor-init-http/test/ActorInitHttp-test.ts
+++ b/packages/actor-init-http/test/ActorInitHttp-test.ts
@@ -57,7 +57,7 @@ describe('ActorInitHttp', () => {
       const actorFetch: ActorHttpNodeFetch = new ActorHttpNodeFetch({ name: 'actor-node-fetch', bus });
       (<any> actorFetch).run = (action) => Promise.resolve({
         body: action.input === 'https://www.google.com/noweb'
-          ? require('node-web-streams').toWebReadableStream(new PassThrough()) : new PassThrough(),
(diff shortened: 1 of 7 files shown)


mielvds:
> Weird that `node-web-streams` is still in the lockfile. Can you check if some other package still depend on it?
> 
> Also, this PR doesn't updated the `require('node-web-streams')` usages, so I suspect the Travis build will fail.

Wasn't done, on that now ;) No other dependencies AFAIK

rubensworks:
Just checked, and it's also depended upon from [`rdf-test-suite`](https://github.com/rubensworks/rdf-test-suite.js/blob/master/package.json#L97), so that's why it's still in the lockfile.

But since this is just a dev-dependency, it's not a problem for downstream users.

rubensworks:
Also, the one Travis failure is due to the same reason as #576, so we can ignore this here (but needs urgent fixing).

Issue #537, opened 2019-10-04
Random integration failure on rdfjs-information03


rubensworks:
#### Issue type:

- :bug: Bug

____
#### Description:

https://travis-ci.org/comunica/comunica/jobs/593534662#L1216

Not sure yet whether this is a Comunica bug or an rdf-test-suite-ldf bug.

rubensworks:
Happened again here: #549

rubensworks:
Also in #550 and #551.

rubensworks:
@joachimvh If you find some time, could you have a look at this one?

It looks like for some reason the N3Store in the (rdfjs) integration tests is sometimes empty.
Since the queries are similar other integration tests that don't break, I don't think Comunica is at fault here, so it must either be caused by N3 or rdf-test-suite-ldf.

\cc @RubenVerborgh

joachimvh:
Should be fixed in comunica/rdf-test-suite-ldf.js#28

I'm also wondering about [this](https://github.com/comunica/rdf-test-suite-ldf.js/blob/master/lib/testcase/ldf/LdfTestCaseEvaluationHandler.ts#L199) line. It always fails, since it checks if the test file exists in the local directory, while 1 line further it gets created in the temp directory using Path.join. But if the check happens with Path.join, the tests will fail if 2 of them use a data file with the same name but different contents (which is the case for the RDFJS test run). So not sure what the point is there.

Side note: in the same class I removed all references to the responseMocker causing the tests to run much faster. I assume it is needed in some cases (when the cache isn't filled/used?), but it would be nice to make sure it is only used when necessary because the performance difference is quite noticeable.

rubensworks:
@ManuDeBuck Do you think what @joachimvh describes could be a bug? Or it this intentional?

ManuDeBuck:
@joachimvh The first point you mention seems correct. This is indeed a mistake (don't know why I didn't notice this earlier on). It will indeed be problematic when using `Path.join`. It seems thus rather difficult to cache files via this way? I'll think about how we can fix this to make caching possible somehow.

[...]

PR #588, opened 2019-11-28
Only overwrite error function for tests

joachimvh:
This is related to #537 . I'm not gonna say it fixes it yet because there seem to be several problems there. This should fix the "Call retries were exceeded" error such as [here](https://travis-ci.org/comunica/comunica/jobs/617728041#L333) . Fortunately this error appears consistently when using the latest version of node. Not sure why though, maybe something changed to the console object and overwriting it entirely causes issues. Fun fact: the problem is also solved by removing a certain amount of tests from that file, no idea why.


joachimvh:
But as can be seen from the integration it doesn't fix the problem with the simple01 query yet :D (which unfortunately occurs more randomly it seems).

rubensworks:
Right. But it fixes #576. So that's already a great thing :-)

PR #589, opened 2019-12-05
Add actors for joining more than two streams pair-wise

rubensworks:
While current actors only allow joining 2 streams, this PR adds 2 different actors that join between more than 2 streams, by doing pairwise joins.

While this is not needed in Comunica itself at the moment, I needed it for an external actor, and I thought it would be useful to have this in the default Comunica actors directly.


packages/actor-rdf-join-multi-sequential/README.md
```diff
@@ -0,0 +1,18 @@
+# Comunica Multi Sequential RDF Join Actor
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-rdf-join-multi-sequential.svg)](https://www.npmjs.com/package/@comunica/actor-rdf-join-multi-sequential)
+
+A comunica Multi Sequential RDF Join Actor.
+It accepts 3 or streams at once, joins the first two, and joins the result with the remaining streams.
```
joachimvh: 3 or more

packages/actor-rdf-join-multi-sequential/lib/ActorRdfJoinMultiSequential.ts
```diff
@@ -0,0 +1,39 @@
...
+  constructor(args: IActorRdfJoinMultiSequentialArgs) {
+    super(args, 3, true);
+  }
+
+  protected async getOutput(action: IActionRdfJoin): Promise<IActorQueryOperationOutputBindings> {
+    // Join the two first streams, and then join the result with the remaining streams
+    const firstEntry: IActorQueryOperationOutputBindings = <IActorQueryOperationOutputBindings> await
+      this.mediatorJoin.mediate({ entries: [ action.entries[0], action.entries[1] ] });
+    const remainingEntries: IActorQueryOperationOutputBindings[] = action.entries.slice(2);
+    remainingEntries.unshift(firstEntry);
```
joachimvh: could also do a slice 1 and then replace the first element with the new entry to do less array operations

packages/actor-rdf-join-multi-smallest/README.md
```diff
@@ -0,0 +1,18 @@
+# Comunica Multi Smallest RDF Join Actor
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-rdf-join-multi-smallest.svg)](https://www.npmjs.com/package/@comunica/actor-rdf-join-multi-smallest)
+
+A comunica Multi Smallest RDF Join Actor.
+It accepts 3 or streams at once, joins the smallest two, and joins the result with the remaining streams.
```
joachimvh: or more

packages/actor-rdf-join-multi-sequential/components/Actor/RdfJoin/MultiSequential.jsonld
```diff
@@ -0,0 +1,35 @@
...
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/bus-rdf-join/^1.0.0/components/context.jsonld"
+  ],
+  "@id": "npmd:@comunica/actor-rdf-join-multi-sequential",
+  "components": [
+    {
+      "@id": "carjms:Actor/RdfJoin/MultiSequential",
+      "@type": "Class",
+      "extends": "cbrj:Actor/RdfJoin",
+      "requireElement": "ActorRdfJoinMultiSequential",
+      "comment": "A comunica Multi Sequential RDF Join Actor.",
```
joachimvh: Comunica

packages/actor-rdf-join-multi-smallest/components/Actor/RdfJoin/MultiSmallest.jsonld
```diff
@@ -0,0 +1,35 @@
...
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/bus-rdf-join/^1.0.0/components/context.jsonld"
+  ],
+  "@id": "npmd:@comunica/actor-rdf-join-multi-smallest",
+  "components": [
+    {
+      "@id": "carjms:Actor/RdfJoin/MultiSmallest",
+      "@type": "Class",
+      "extends": "cbrj:Actor/RdfJoin",
+      "requireElement": "ActorRdfJoinMultiSmallest",
+      "comment": "A comunica Multi Smallest RDF Join Actor.",
```
joachimvh: Comunica

packages/actor-rdf-join-multi-smallest/lib/ActorRdfJoinMultiSmallest.ts
```diff
@@ -0,0 +1,63 @@
...
+    return smallestId;
+  }
+
+  protected async getOutput(action: IActionRdfJoin): Promise<IActorQueryOperationOutputBindings> {
+    const entries: IActorQueryOperationOutputBindings[] = action.entries.slice();
+
+    // Determine the two smallest streams by estimated count
+    const entriesTotalItems = (await Promise.all(action.entries.map((entry) => entry.metadata())))
+      .map((metadata) => 'totalItems' in metadata ? metadata.totalItems : Infinity);
+    const smallestIndex1: number = ActorRdfJoinMultiSmallest.getSmallestPatternId(entriesTotalItems);
```
joachimvh: it's only going to be relevant once there are many many streams that need to be joined at the same time, but why not make the function getTwoSmallestPatternIds?

packages/actor-rdf-join-multi-smallest/lib/ActorRdfJoinMultiSmallest.ts
```diff
@@ -0,0 +1,63 @@
...
+    const smallestItem1 = entries.splice(smallestIndex1, 1)[0];
+    const smallestCount1 = entriesTotalItems.splice(smallestIndex1, 1);
+    const smallestIndex2: number = ActorRdfJoinMultiSmallest.getSmallestPatternId(entriesTotalItems);
+    const smallestItem2 = entries.splice(smallestIndex2, 1)[0];
+    const smallestCount2 = entriesTotalItems.splice(smallestIndex2, 1);
+
+    // Join the two selected streams, and then join the result with the remaining streams
+    const firstEntry: IActorQueryOperationOutputBindings = <IActorQueryOperationOutputBindings> await
+      this.mediatorJoin.mediate({ entries: [ smallestItem1, smallestItem2 ] });
+    entries.unshift(firstEntry);
```
joachimvh: why an unshift instead of a push since the order doesn't matter when this actor is used?

packages/bus-rdf-join/lib/ActorRdfJoin.ts
```diff
@@ -18,15 +18,22 @@ import {EmptyIterator} from "asynciterator";
...
+  protected limitEntries: number;
+  /**
+   * If true, the limitEntries field is a lower limit,
+   * otherwise, it is an upper limit.
+   */
+  protected limitEntriesMin: boolean;
 
   constructor(args: IActorArgs<IActionRdfJoin, IMediatorTypeIterations, IActorQueryOperationOutput>,
-              maxEntries?: number) {
+              limitEntries?: number, limitEntriesMin?: boolean) {
```
joachimvh: since a new field gets added, why not add a min field and a max field instead of 1 field and a boolean? Gives more flexibility in the future.

packages/actor-rdf-join-multi-sequential/lib/ActorRdfJoinMultiSequential.ts
rubensworks: Ah yes, good catch.

packages/actor-rdf-join-multi-smallest/lib/ActorRdfJoinMultiSmallest.ts
rubensworks: I think the implementation would become too complex and unreadable, but feel free to give it ago if you think you can do it in a consise way :-)

packages/actor-rdf-join-multi-smallest/lib/ActorRdfJoinMultiSmallest.ts
rubensworks: You're right, the order doesn't matter, a push will be more efficient.

packages/bus-rdf-join/lib/ActorRdfJoin.ts
rubensworks: True. Let's revisit this should the need for it arise.

packages/actor-rdf-join-multi-smallest/lib/ActorRdfJoinMultiSmallest.ts
joachimvh: I wrote it in notepad so there might be syntax errors, but something like this

```javascript
  public static getSmallestPatternIds(totalItems: number[], count) {
    const smallest = [];
    for (let j = 0; j < count; ++j) {
      smallest.push({ id: -1, count: Infinity });
    }
    
    for (let i = 0; i < totalItems.length; i++) {
      const count: number = totalItems[i];
      for (let k = count-1; k >= 0; --k) {
        // insert new entry if it is smaller
        if (count < smallest[k].count) {
            smallest[k].id = i;
            smallest[k].count = count;
        }
        // stop if the next entry would be larger (or we reached the end)
        if (count >= smallest[k].count || k === 0) {
          // shift larger entries to the right
          if (k < count-1) {
            smallest[k+1] = smallest[k];
          }
          break;
        }
      }
    }
    return smallest;
  }
```
I'll leave it up to you if it's concise enough :P

rubensworks: I've tried implementing it here: https://github.com/comunica/comunica/tree/feature/multi-join-abstracter
It doesn't seem to be working correctly though. Feel free to debug it further if you're up for it. I'll merge this PR in the meantime. We can still apply your improved algorithm at a later stage.

PR #624, opened 2020-03-16
Scope blank nodes to each federated source

### LINKED ISSUE #355: Scope blank nodes to original sources when federating
#### Issue type:

____
#### Description:

If we are federating over two sources (A and B), and a query over A produces a blank node, then this blank node should _not_ be matched in other sources (B), but only in itself (A).

Concretely, the federated actor should:
* skolemize all incoming blank nodes deterministically (the source identifier should be encoded in this as well),
* de-skolemize blank nodes when creating new sub-queries to sources.
* If a blank node appears in a pattern that originates from a different source than the target source, then an `EmptyIterator` should be returned, as this blank node should not be matched in the target source.

rubensworks:
When federating over multiple sources,
the outcoming blank nodes coming from each source
will receive a distinct blank node,
so that they can not be joined across different sources,
even if they have the same blank node label.

A reverse translation also takes place for incoming queries with blank nodes,
so that these blank nodes will only match if they come from that source.

Blank nodes coming from sources will receive a .skolemized field
containing a named node.
This named node can be queried again as an IRI, and this will
be interpreted by Comunica as a blank node corresponding to the proper
source, assuming that the array of sources remains the same.

Closes #355
Required for solid/query-ldflex#34

While the implementation is quite trivial, the impact is quite major, so this deserves a detailed review.

### DIFF
--- a/packages/actor-rdf-resolve-quad-pattern-federated/lib/BlankNodeScoped.ts
+++ b/packages/actor-rdf-resolve-quad-pattern-federated/lib/BlankNodeScoped.ts
@@ -1,6 +1,9 @@
 import * as RDF from "rdf-js";
 
-export class BlankNodeSkolemizable implements RDF.BlankNode {
+/**
+ * A blank node that is scoped to a certain source.
+ */
+export class BlankNodeScoped implements RDF.BlankNode {
   public readonly termType: 'BlankNode' = 'BlankNode';
   public readonly value: string;
   /**
--- a/packages/actor-rdf-resolve-quad-pattern-federated/lib/BlankNodeScoped.ts
+++ b/packages/actor-rdf-resolve-quad-pattern-federated/lib/BlankNodeScoped.ts
@@ -19,7 +19,7 @@ export class BlankNodeScoped implements RDF.BlankNode {
   }
 
   public equals(other: RDF.Term | null | undefined): boolean {
-    return other && other.termType === 'BlankNode' && other.value === this.value;
+    return !!other && other.termType === 'BlankNode' && other.value === this.value;
   }
 
 }
--- a/packages/actor-rdf-resolve-quad-pattern-federated/lib/BlankNodeSkolemizable.ts
+++ b/packages/actor-rdf-resolve-quad-pattern-federated/lib/BlankNodeSkolemizable.ts
@@ -1,0 +1,22 @@
+import * as RDF from "rdf-js";
+
+export class BlankNodeSkolemizable implements RDF.BlankNode {
+  public readonly termType: 'BlankNode' = 'BlankNode';
+  public readonly value: string;
+  /**
+   * This value can be obtained by consumers in query results,
+   * so that this can be passed into another query as an IRI,
+   * in order to obtain more results relating to this (blank) node.
+   */
+  public readonly skolemized: RDF.NamedNode;
+
+  constructor(value: string, skolemized: RDF.NamedNode) {
+    this.value = value;
+    this.skolemized = skolemized;
+  }
+
+  public equals(other: RDF.Term | null | undefined): boolean {
+    return other && other.termType === 'BlankNode' && other.value === this.value;
+  }
+
+}
--- a/packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
+++ b/packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
@@ -20,12 +20,14 @@ import {Quad} from "rdf-js";
  */
 export class FederatedQuadSource implements ILazyQuadSource {
 
+  private static readonly SKOLEM_PREFIX = 'urn:comunica_skolem:source_';
+
   protected readonly mediatorResolveQuadPattern: Mediator<Actor<IActionRdfResolveQuadPattern, IActorTest,
     IActorRdfResolveQuadPatternOutput>, IActionRdfResolveQuadPattern, IActorTest, IActorRdfResolveQuadPatternOutput>;
   protected readonly sources: DataSources;
   protected readonly contextDefault: ActionContext;
   protected readonly emptyPatterns: Map<IDataSource, RDF.BaseQuad[]>;
-  protected readonly sourceIds: Map<IDataSource, number>;
+  protected readonly sourceIds: Map<IDataSource, string>;
   protected readonly skipEmptyPatterns: boolean;
   protected readonly algebraFactory: Factory;
 
@@ -93,13 +95,13 @@ export class FederatedQuadSource implements ILazyQuadSource {
    * If the given term is a blank node, return a deterministic named node for it
    * based on the source id and the blank node value.
    * @param term Any RDF term.
-   * @param sourceId A numerical source identifier.
+   * @param sourceId A source identifier.
    * @return If the given term was a blank node, this will return a skolemized named node, otherwise the original term.
    */
-  public static skolemizeTerm(term: RDF.Term, sourceId: number): RDF.Term | BlankNodeScoped {
+  public static skolemizeTerm(term: RDF.Term, sourceId: string): RDF.Term | BlankNodeScoped {
     if (term.termType === 'BlankNode') {
       return new BlankNodeScoped(`bc_${sourceId}_${term.value}`,
-        DataFactory.namedNode(`urn:comunica_skolem:source_${sourceId}:${term.value}`));
+        DataFactory.namedNode(`${FederatedQuadSource.SKOLEM_PREFIX}${sourceId}:${term.value}`));
     }
     return term;
   }
@@ -107,10 +109,10 @@ export class FederatedQuadSource implements ILazyQuadSource {
   /**
    * Skolemize all terms in the given quad.
    * @param quad An RDF quad.
-   * @param sourceId A numerical source identifier.
+   * @param sourceId A source identifier.
    * @return The skolemized quad.
    */
-  public static skolemizeQuad<Q extends BaseQuad = Quad>(quad: Q, sourceId: number): Q {
+  public static skolemizeQuad<Q extends BaseQuad = Quad>(quad: Q, sourceId: string): Q {
     return mapTerms(quad, (term) => FederatedQuadSource.skolemizeTerm(term, sourceId));
   }
 
@@ -120,22 +122,24 @@ export class FederatedQuadSource implements ILazyQuadSource {
    * If the given term was a skolemized named node for another source, return false.
    * If the given term was not a skolemized named node, return the original term.
    * @param term Any RDF term.
-   * @param sourceId A numerical source identifier.
+   * @param sourceId A source identifier.
    */
-  public static deskolemizeTerm(term: RDF.Term, sourceId: number): RDF.Term | false {
+  public static deskolemizeTerm(term: RDF.Term, sourceId: string): RDF.Term | null {
     if (term.termType === 'BlankNode' && 'skolemized' in term) {
       term = (<BlankNodeScoped> term).skolemized;
     }
     if (term.termType === 'NamedNode') {
-      const match = /^urn:comunica_skolem:source_([0-9]+):(.+)$/.exec(term.value);
-      if (match) {
+      if (term.value.startsWith(FederatedQuadSource.SKOLEM_PREFIX)) {
+        const colonSeparator = term.value.indexOf(':', FederatedQuadSource.SKOLEM_PREFIX.length);
+        const termSourceId = term.value.substr(FederatedQuadSource.SKOLEM_PREFIX.length, colonSeparator - FederatedQuadSource.SKOLEM_PREFIX.length);
         // We had a skolemized term
-        if (parseInt(match[1], 10) === sourceId) {
+        if (termSourceId === sourceId) {
           // It can from the correct source
-          return DataFactory.blankNode(match[2]);
+          const termLabel = term.value.substr(colonSeparator + 1, term.value.length);
+          return DataFactory.blankNode(termLabel);
         } else {
           // It came from a different source
-          return false;
+          return null;
         }
       }
     }
@@ -172,12 +176,12 @@ export class FederatedQuadSource implements ILazyQuadSource {
   /**
    * Get the unique, deterministic id for the given source.
    * @param source A data source.
-   * @return The numerical id of the given source.
+   * @return The id of the given source.
    */
-  public getSourceId(source: IDataSource): number {
+  public getSourceId(source: IDataSource): string {
     let sourceId = this.sourceIds.get(source);
     if (sourceId === undefined) {
-      sourceId = this.sourceIds.size;
+      sourceId = `${this.sourceIds.size}`;
       this.sourceIds.set(source, sourceId);
     }
     return sourceId;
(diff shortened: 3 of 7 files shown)


rubensworks:
Note to self: test `BlankNodeSkolemizable#equals`

packages/actor-rdf-resolve-quad-pattern-federated/lib/BlankNodeSkolemizable.ts
```diff
@@ -0,0 +1,22 @@
+import * as RDF from "rdf-js";
+
+export class BlankNodeSkolemizable implements RDF.BlankNode {
```
RubenVerborgh: Perhaps a slight misnomer, given that all blank nodes can be skolemized.

How about something like `ScopedBlankNode` or `RestrictedBlankNode` or `ReusableBlankNode` or `TraceableBlankNode`?

packages/actor-rdf-resolve-quad-pattern-federated/lib/BlankNodeSkolemizable.ts
```diff
@@ -0,0 +1,22 @@
...
+   */
+  public readonly skolemized: RDF.NamedNode;
+
+  constructor(value: string, skolemized: RDF.NamedNode) {
+    this.value = value;
+    this.skolemized = skolemized;
+  }
+
+  public equals(other: RDF.Term | null | undefined): boolean {
+    return other && other.termType === 'BlankNode' && other.value === this.value;
```
RubenVerborgh: I expect this to be the default implementation?

That said, we need to think whether this is the right implementation. On the one hand, we don't want to mess with symmetry and transitivity. On the other hand, we likely want this method to give _more_ positives than for regular `BlankNode`s.

I expect a `BlankNodeSkolemizable` A to equal B iff both are `BlankNodeSkolemizable` instances and they share the same `skolemized` (regardless of `value`).

So we could define `equals` such that it fails for all non-`BlankNodeSkolemizable` instances (including any other `BlankNode` implementation). Then we'd need to ensure that the reverse is also true. We can accomplish that by ensuring that `value` is set to a unique value every time. (For instance, UUID, or just ensuring that whatever thing generates blank node `value`s never reuses. N3.js does that by default, for instance.) Which you seem to be doing, so fine.

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -83,6 +89,56 @@ export class FederatedQuadSource implements ILazyQuadSource {
...
+   * If the given term was not a skolemized named node, return the original term.
+   * @param term Any RDF term.
+   * @param sourceId A numerical source identifier.
+   */
+  public static deskolemizeTerm(term: RDF.Term, sourceId: number): RDF.Term | false {
+    if (term.termType === 'NamedNode' || term.termType === 'BlankNode') {
+      const match = /^urn:comunica_skolem:source_([0-9]+):(.+)$/.exec(term.value);
+      if (match) {
+        // We had a skolemized term
+        if (parseInt(match[1], 10) === sourceId) {
```
RubenVerborgh: We can skip the parsing if we give sources `string` IDs (which _could_ be stringified numbers, all fine).

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -83,6 +89,56 @@ export class FederatedQuadSource implements ILazyQuadSource {
...
+   * If a given term was a skolemized named node for the given source id,
+   * deskolemize it again to a blank node.
+   * If the given term was a skolemized named node for another source, return false.
+   * If the given term was not a skolemized named node, return the original term.
+   * @param term Any RDF term.
+   * @param sourceId A numerical source identifier.
+   */
+  public static deskolemizeTerm(term: RDF.Term, sourceId: number): RDF.Term | false {
+    if (term.termType === 'NamedNode' || term.termType === 'BlankNode') {
+      const match = /^urn:comunica_skolem:source_([0-9]+):(.+)$/.exec(term.value);
```
RubenVerborgh: If this thing is executed many times, we _might_ start getting a performance bottleneck. A `startsWith` could be a good initial check; then substringing will likely be faster.

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -83,6 +89,56 @@ export class FederatedQuadSource implements ILazyQuadSource {
...
+   * If the given term is a blank node, return a deterministic named node for it
+   * based on the source id and the blank node value.
+   * @param term Any RDF term.
+   * @param sourceId A numerical source identifier.
+   * @return If the given term was a blank node, this will return a skolemized named node, otherwise the original term.
+   */
+  public static skolemizeTerm(term: RDF.Term, sourceId: number): RDF.Term | BlankNodeSkolemizable {
+    if (term.termType === 'BlankNode') {
+      const skolemized = `urn:comunica_skolem:source_${sourceId}:${term.value}`;
+      return new BlankNodeSkolemizable(skolemized, DataFactory.namedNode(skolemized));
```
RubenVerborgh: If this is consistently the case, should we just define `value` as a getter in `BlankNodeSkolemizable` that returns `this.skolemized.value`? Or are you relying on polymorphism for making things fast in V8 (valid argument)?

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -83,6 +89,56 @@ export class FederatedQuadSource implements ILazyQuadSource {
...
+
+  /**
+   * If a given term was a skolemized named node for the given source id,
+   * deskolemize it again to a blank node.
+   * If the given term was a skolemized named node for another source, return false.
+   * If the given term was not a skolemized named node, return the original term.
+   * @param term Any RDF term.
+   * @param sourceId A numerical source identifier.
+   */
+  public static deskolemizeTerm(term: RDF.Term, sourceId: number): RDF.Term | false {
```
RubenVerborgh: Why `false` instead of `null`? Typing with `null` seems more logical.

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -110,6 +166,20 @@ export class FederatedQuadSource implements ILazyQuadSource {
...
 
+  /**
+   * Get the unique, deterministic id for the given source.
+   * @param source A data source.
+   * @return The numerical id of the given source.
+   */
+  public getSourceId(source: IDataSource): number {
+    let sourceId = this.sourceIds.get(source);
+    if (sourceId === undefined) {
+      sourceId = this.sourceIds.size;
```
RubenVerborgh: Actually, you _might_ want to give sources the opportunity to optionally define their own ID. Then, if a source is hidden behind a cache or something, it would still give the same ID.

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -149,25 +219,28 @@ export class FederatedQuadSource implements ILazyQuadSource {
...
-      // If we can predict that the given source will have no bindings for the given pattern,
-      // return an empty iterator.
-      const pattern: Algebra.Pattern = this.algebraFactory.createPattern(
-        FederatedQuadSource.nullToVariable(subject, 's'),
-        FederatedQuadSource.nullToVariable(predicate, 'p'),
-        FederatedQuadSource.nullToVariable(object, 'o'),
-        FederatedQuadSource.nullToVariable(graph, 'g'),
-      );
+      return new PromiseProxyIterator(async () => {
+        // Deskolemize terms, so we send the original blank nodes to each source.
```
RubenVerborgh: Hmm, but will that work for all sources? Clearly not for SPARQL endpoints.

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -149,25 +219,28 @@ export class FederatedQuadSource implements ILazyQuadSource {
...
+        const s = FederatedQuadSource.deskolemizeTerm(FederatedQuadSource.nullToVariable(subject, 's'), sourceId);
```
RubenVerborgh: No need to deskolemize thus for terms that are `null`, which seems like a common case. Perhaps `subject === null ? toVar(subject, 's') : deskolemizeTerm(subject)`?

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
RubenVerborgh: Alternative: wondering if we want `value` to be something more human-accessible (while still keeping in mind unicity). Like `_:bc_1_1` (for Blanknode Comunica sourceId value) or so.

packages/actor-rdf-resolve-quad-pattern-federated/lib/BlankNodeSkolemizable.ts
rubensworks: I initially had an `instanceof BlankNodeSkolemizable` check here, which I removed to ensure symmary as you've mentioned.

Since `equals` is defined on the term, such an `instanceof` check can only be used in 50% of the cases. And since we already make sure `value` is never reused, I don't see any benefit in adding such a check, just confusion and weird edge-cases.

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
rubensworks: I only did this here for type safety.
By default TypeScript will not produce compilation if the developer doesn't explicitly check for null, while it will for something like `false`.
(There is a compilation flag to require such null checks, which I want to enable in Comunica at some point, but this will require quite a bit of work (#628)).

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
rubensworks: Good point...

But I would keep this for another issue, since this also applied before this PR, and I don't think it's a trivial thing to fix.

packages/actor-rdf-resolve-quad-pattern-federated/lib/BlankNodeSkolemizable.ts
RubenVerborgh: But then I'm strongly wondering about the necessity of `skolemized`. Also as per my "human-accessible" comment below: https://github.com/comunica/comunica/pull/624#discussion_r393927250

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
RubenVerborgh: OK; be aware though that this is likely not optimal for V8 (`false` being a value type).

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
RubenVerborgh: It's unfixable in general; perhaps just good to add a note.

RubenVerborgh: It's unfixable in general; perhaps just good to add a note.

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
rubensworks: Good point. This function is part of the critical logic path, so optimizing this as much as possible is important.

rubensworks:
@RubenVerborgh Made some changes based on your comment.
Could you have another look at it?
It's important that we get this part right :-)

packages/actor-rdf-resolve-quad-pattern-federated/lib/BlankNodeScoped.ts
```diff
@@ -0,0 +1,25 @@
...
+   */
+  public readonly skolemized: RDF.NamedNode;
+
+  constructor(value: string, skolemized: RDF.NamedNode) {
+    this.value = value;
+    this.skolemized = skolemized;
+  }
+
+  public equals(other: RDF.Term | null | undefined): boolean {
+    return other && other.termType === 'BlankNode' && other.value === this.value;
```
RubenVerborgh: strictly `!!other && …`, but that will do

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -70,15 +78,56 @@ export class FederatedQuadSource implements ILazyQuadSource {
...
+    if (term.termType === 'BlankNode' && 'skolemized' in term) {
+      term = (<BlankNodeScoped> term).skolemized;
+    }
+    if (term.termType === 'NamedNode') {
+      if (term.value.startsWith(FederatedQuadSource.SKOLEM_PREFIX)) {
+        const colonSeparator = term.value.indexOf(':', FederatedQuadSource.SKOLEM_PREFIX.length);
+        const termSourceId = term.value.substr(FederatedQuadSource.SKOLEM_PREFIX.length, colonSeparator - FederatedQuadSource.SKOLEM_PREFIX.length);
+        // We had a skolemized term
+        if (termSourceId === sourceId) {
+          // It can from the correct source
```
RubenVerborgh: *came

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -149,25 +212,30 @@ export class FederatedQuadSource implements ILazyQuadSource {
...
-        FederatedQuadSource.nullToVariable(predicate, 'p'),
-        FederatedQuadSource.nullToVariable(object, 'o'),
-        FederatedQuadSource.nullToVariable(graph, 'g'),
-      );
+      return new PromiseProxyIterator(async () => {
+        // Convert falsy terms to variables, reverse operation of {@link ActorRdfResolveQuadPatternSource#variableToNull}.
+        // Deskolemize terms, so we send the original blank nodes to each source.
+        // Note that some sources may not match bnodes by label. SPARQL endpoints for example consider them variables.
+        const s = !subject   ? DataFactory.variable('vs') : FederatedQuadSource.deskolemizeTerm(subject,   sourceId);
+        const p = !predicate ? DataFactory.variable('vp') : FederatedQuadSource.deskolemizeTerm(predicate, sourceId);
```
RubenVerborgh: Predicates shouldn't be blanks I guess, but better to be safe than sorry.

rubensworks: True. By keeping it generic, we can in the future have support for [generalized RDF](https://www.w3.org/TR/rdf11-concepts/#section-generalized-rdf) should that ever be needed.

packages/actor-rdf-resolve-quad-pattern-federated/lib/BlankNodeScoped.ts
rubensworks: Right...
I should really enable those strict null checks in TS, because that would have detected this problem at compile time.

PR #644, opened 2020-04-01
Fix initialBindings not applied on full query

### LINKED ISSUE #640: "initialBindings" not applied when using functions
#### Issue type:

- :bug: Bug

____
#### Description:

I want to use the `initialBindings` option in a query. However, my pre-bound value is not processed when using e.g. a CONTAINS() function:

```javascript
import * as Comunica from '@comunica/actor-init-sparql';
import { LoggerPretty } from '@comunica/logger-pretty';
import { Bindings } from '@comunica/bus-query-operation';
import { literal } from '@rdfjs/data-model';

(async () => {
  const myEngine = Comunica.newEngine();

  const query = `PREFIX schema: <http://schema.org/>
    SELECT *
    WHERE {
      ?s a schema:Person .
      ?s schema:name ?name .
      FILTER(CONTAINS(?name, ?searchTerm))
    }`;

  const config = {
    log: new LoggerPretty({ level: 'debug' }),
    sources: [
      {
        type: 'sparql',
        value: 'https://api.data.netwerkdigitaalerfgoed.nl/datasets/rkd/rkdartists/services/rkdartists/sparql',
      },
    ],
    initialBindings: Bindings({ '?searchTerm': literal('Gogh') }),
  };

  await myEngine.query(query, config);
})();
```

The query generated by Comunica looks like this:

```sparql
SELECT ?s ?name WHERE {
  ?s <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Person>.
  ?s <http://schema.org/name> ?name.
  FILTER(CONTAINS(?name, ?searchTerm))
}
```

Expected result: `"Gogh"` instead of `?searchTerm`.

However, if I execute the query underneath...

```javascript
  const query = `PREFIX schema: <http://schema.org/>
    SELECT *
    WHERE {
      ?s a schema:Person .
      ?s schema:name ?searchTerm
    }`;
```

...then the generated query looks like this:

```sparql
SELECT ?s WHERE {
  ?s <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Person>.
  ?s <http://schema.org/name> "Gogh"^^<http://www.w3.org/2001/XMLSchema#string>.
}
```

____
#### Environment:

| Comunica Init Actor | 1.10.0
| node                | v12.0.0
| npm                 | 6.9.0
| yarn                | 1.15.2
| Operating System    | linux (Linux 4.9.184-linuxkit)

rubensworks:
This functionality was present in different places,
so this refactors things a bit so that only one
implementation exists, which is properly tested.

Closes #640


packages/bus-query-operation/lib/Bindings.ts
```diff
@@ -50,3 +52,195 @@ export function isBindings(maybeBindings: any): boolean {
...
+        if (strictTargetVariables) {
+          throw new Error(`Tried to bind variable ${termToString(op.variable)} in a BIND operator.`);
+        } else {
+          return {
+            recurse: true,
+            result: factory.createExtend(
+              op.input,
+              op.variable,
+              factory.createTermExpression(bindings.get(termToString(op.variable)))
+            ),
```
joachimvh, line 107: Since every instance of variable op.variable would have been overwritten so far it would make more sense to replace the above with `result: op.input` right? Or am I missing a case where this could be an issue?

packages/bus-query-operation/lib/Bindings.ts
```diff
@@ -50,3 +52,195 @@ export function isBindings(maybeBindings: any): boolean {
...
+      } else {
+        const variables = op.variables.filter((variable) => !bindings.has(termToString(variable)));
+        return {
+          recurse: true,
+          result: factory.createGroup(
+            op.input,
+            variables,
+            op.aggregates,
+          ),
+        };
```
joachimvh, line 139: This could create problems further down the line if variables is empty but I guess we don't care much if users break their own queries?

packages/bus-query-operation/lib/Bindings.ts
```diff
@@ -50,3 +52,195 @@ export function isBindings(maybeBindings: any): boolean {
...
+        // Materialize a bound aggregate operation.
+        // If strictTargetVariables is true, we throw if the expression target variable is attempted to be bound.
+        // Otherwise, we ignore this operation.
+        if (strictTargetVariables) {
+          throw new Error(`Tried to bind ${termToString(op.variable)} in a ${op.aggregator} aggregate.`);
+        } else {
+          return {
+            recurse: true,
+            result: op,
+          }
```
joachimvh, line 215: So in this case the variable will not be assigned the value of the bound variable (while in the case of an extend for example the variable does get bound to the new value). But I'll be honest I'm also not 100% sure what could break by doing this :D

packages/bus-query-operation/lib/Bindings.ts
rubensworks: Ah yes, you're right.

packages/bus-query-operation/lib/Bindings.ts
rubensworks: Yes indeed, I think that's the safest way to do it.

packages/bus-query-operation/lib/Bindings.ts
rubensworks: If I understand correctly, it doesn't make much sense to pre-bind a bound aggregate, so ignoring it feels like the safest way forward.

PR #657, opened 2020-05-04
Update to sparqlalgebra.js 2.2.2

### LINKED ISSUE #649: Invalid COUNT queries being sent to SPARQL endpoints
#### Issue type:

- :bug: Bug

____
#### Description:

The below query over an online TTL file and a Stardog triplestore return the right results (I think), but before that, Comunica sends about 504 unsuccesful "statistical" queries to Stardog that don't seem to make sense to me.

Sources:
```
{ value: 'http://localhost:5820/mydb/query' , type: 'sparql' }
{ value: 'http://myfile.org/file.ttl' , type: 'file' }
```
Query asked to Comunica engine:
```
SELECT * WHERE {
	{ 
                # only this named graph in triplestore + should be send in one piece => SERVICE
		SERVICE <http://localhost:5820/mydb/query> {
			GRAPH <virtual://myvirtgraph> {
				?iel rdf:type bot:Element , ?str .
				?str rdfs:label ?strLabel .
			}
		}
	} UNION {
                # also in TTL file
		?iel rdf:type bot:Element , ?str .
		?str rdfs:label ?strLabel .
	}
        # both in default graph of triplestore and TTL file
	?iel ex:hasRepair ?o .
	?o ?p2 ?o2 .
} 
```
Example GET request send by Comunica containing useless (?) query, since `?var0` can never have any results in this query:
```
http://localhost:5820/mydb/query?query=SELECT ?count WHERE {
  GRAPH <virtual://myvirtgraph> {
    ?s <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?o.
    BIND(?var0 AS ?count)
  }
}
```
Only after Comunica has send about 504 similar queries, I get the results related to the `SERVICE` part from my SPARQL endpoint
In fact, I just noticed this delay by looking at the errors from Stardog coming in my browser console. The virtual graph at the Stardog SPARQL endpoint seems pretty brittle when it comes to non-specific queries; all 504 "statistical" queries fail and return a 500 error.

____
#### Environment:
Comunica CDN latest (1.12.0)

joachimvh:
Should fix #649

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "rdf-quad": "^1.4.0",
     "rdf-test-suite": "^1.13.0",
     "rdf-test-suite-ldf": "^1.1.5",
-    "sparqlalgebrajs": "^2.2.1",
+    "sparqlalgebrajs": "^2.2.2",
     "stream-to-string": "^1.1.0",
     "streamify-array": "^1.0.0",
     "streamify-string": "^1.0.1",
--- a/packages/actor-abstract-path/package.json
+++ b/packages/actor-abstract-path/package.json
@@ -41,7 +41,7 @@
     "@rdfjs/data-model": "^1.1.1",
     "asynciterator": "^2.0.1",
     "rdf-string": "^1.3.1",
-    "sparqlalgebrajs": "^2.2.1"
+    "sparqlalgebrajs": "^2.2.2"
   },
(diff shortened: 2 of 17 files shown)


rubensworks:
Great, thanks!

PR #681, opened 2020-07-06
Fix orderby with non-string types, Closes #599

### LINKED ISSUE #599: ORDER BY doesn't take numeric type into account
#### Issue type:

- :bug: Bug

____
#### Description:

An `ORDER BY` clause in a SPARQL query doesn't sort `numeric`s numerically.

This might be related to #598 .

____
#### Example

**input.ttl:**
```turtle
@prefix p: <http://example.org#> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

p:one p:value "1"^^xsd:integer .
p:two p:value "2"^^xsd:integer .
p:eleven p:value "11"^^xsd:integer .
```

**query.sparql:**
```sparql
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
PREFIX p: <http://example.org#>

SELECT ?number ?value
WHERE {
  ?number p:value ?value .
} ORDER BY ?value
```
The query was executed as follows:
```
comunica-sparql-file input.ttl -f query.sparql
```

**Result:**
```json
[
{"?number":"http://example.org#one","?value":"\"1\"^^http://www.w3.org/2001/XMLSchema#integer"},
{"?number":"http://example.org#eleven","?value":"\"11\"^^http://www.w3.org/2001/XMLSchema#integer"},
{"?number":"http://example.org#two","?value":"\"2\"^^http://www.w3.org/2001/XMLSchema#integer"}
]
```

**Expected result:**
```json
[
{"?number":"http://example.org#one","?value":"\"1\"^^http://www.w3.org/2001/XMLSchema#integer"},
{"?number":"http://example.org#two","?value":"\"2\"^^http://www.w3.org/2001/XMLSchema#integer"},
{"?number":"http://example.org#eleven","?value":"\"11\"^^http://www.w3.org/2001/XMLSchema#integer"}
]

```

____
#### Environment:
| Comunica Init Actor | 1.10.0 
| node                | v10.16.3
| npm                 | 6.13.6
| Operating System    | linux (Linux 4.15.0-74-generic)

stephaniech97:
Fix orderby with non-string types, Closes #599

### DIFF
--- a/packages/actor-query-operation-orderby-sparqlee/lib/ActorQueryOperationOrderBySparqlee.ts
+++ b/packages/actor-query-operation-orderby-sparqlee/lib/ActorQueryOperationOrderBySparqlee.ts
@@ -2,7 +2,7 @@ import { Term } from "rdf-js";
 import { termToString } from "rdf-string";
 import { Algebra } from "sparqlalgebrajs";
 import { AsyncEvaluator, isExpressionError } from 'sparqlee';
-import { type } from 'sparqlee/dist/lib/util/Consts';
+import { orderTypes } from 'sparqlee/dist/lib/util/Consts';
 
 import {
   ActorQueryOperation, ActorQueryOperationTypedMediated,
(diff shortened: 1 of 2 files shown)


stephaniech97:
@rubensworks I read somewhere that it wouldn't be clean to have a function which returns many different types in typescript. So that is why I was reluctant to implementing it into sparqlee itself.

rubensworks:
I don't think this change would result in many different return types.
I would suggest moving the [comparator function](https://github.com/comunica/comunica/blob/e8dcd2f254c518d8f496d5dc5aaaf55b758bd8d5/packages/actor-query-operation-orderby-sparqlee/lib/ActorQueryOperationOrderBySparqlee.ts#L68-L102) (minus the SortIterator) to sparqlee.

Functions like `getLiteralType` and `literalValue` already exist there, so there would be no need to duplicate these anymore.

PR #682, opened 2020-07-08
Issue599 numeric orderby

stephaniech97:
Sorting function imported from sparqlee. Called by passing a term. Included tests.


rubensworks:
Is it possible you opened a PR for the wrong branch? It doesn't seem to be compatible with https://github.com/comunica/sparqlee/pull/69/.

PR #688, opened 2020-07-09
Add new fetch import to allow iteration of header, Closes #672

### LINKED ISSUE #672: Passed Headers object is not standards compatible (not iterable)
#### Issue type:

- :bug: Bug

____
#### Description:

It appears that we are sending a `Headers` object to the fetcher that is not standards-compatible, namely it is not iterable. See discussion at https://github.com/inrupt/solid-auth-fetcher/issues/176#issuecomment-642319240

My guess is that this is due to our usage of https://www.npmjs.com/package/isomorphic-fetch, which hasn't been updated in 4 years, and might therefore use an old version of the WHATWG fetch polyfill (although that functionality was also [added 4 years ago there](https://github.com/github/fetch/commit/a1b7674b6942d4265ea47b74760a486b2bf5e3da)).

We'd mainly need a test to check if indeed the object is iterable.
____
#### Environment:

current

stephaniech97:


packages/actor-http-memento/lib/ActorHttpMemento.ts
```diff
@@ -1,6 +1,7 @@
 import { ActorHttp, IActionHttp, IActorHttpOutput } from "@comunica/bus-http";
 import { ActionContext, IActorArgs, IActorTest, Mediator } from "@comunica/core";
-import "isomorphic-fetch";
+import "cross-fetch/polyfill";
```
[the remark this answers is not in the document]
stephaniech97: Initially when using polyfill I had an easier time fixing the tests as they tended to fail with "Request is not defined". Then on further working in it and looking in the node modules, only in the browser polyfill i found a clear implementation for the iteration of the headers. Which is why i chose polyfill instead.

packages/actor-http-native/lib/ActorHttpNative.ts
```diff
@@ -39,22 +39,21 @@ export class ActorHttpNative extends ActorHttp {
...
     } else {
-      options.headers = {};
-    }
-    if (!options.headers['user-agent']) {
-      options.headers['user-agent'] = this.userAgent;
+      options.headers = (<Request> action.input).headers;
     }
 
+    if (options.headers){
+      if (!options.headers.get('user-agent')) {
```
rubensworks, line 50:
```suggestion
      if (!options.headers.has('user-agent')) {
```

packages/actor-http-native/lib/ActorHttpNative.ts
```diff
@@ -39,22 +39,21 @@ export class ActorHttpNative extends ActorHttp {
...
-      options.headers = headers;
     } else {
-      options.headers = {};
-    }
-    if (!options.headers['user-agent']) {
-      options.headers['user-agent'] = this.userAgent;
+      options.headers = (<Request> action.input).headers;
     }
 
+    if (options.headers){
```
rubensworks, line 49: The following code can be simplified.
Right now you do the same append in two locations.
Instead, you can add an if-check to ensure headers is defined.
After that check, you can do the append.

package.json
```diff
@@ -14,6 +14,7 @@
     "@babel/polyfill": "^7.8.7",
     "@babel/preset-env": "^7.9.5",
     "@rdfjs/data-model": "^1.1.1",
+    "@strictsoftware/typedoc-plugin-monorepo": "^0.3.0",
```
rubensworks, line 17: Changes to this root package.json file should not be needed.

(reminder to check the changes files in a PR every now and then, to make sure that the diff makes sense)

packages/actor-http-memento/package.json
```diff
@@ -35,8 +35,7 @@
     "index.js"
   ],
   "dependencies": {
-    "@types/parse-link-header": "^1.0.0",
```
rubensworks, line 38: Aren't these types needed anymore?

stephaniech97:
> Looks great!
> 
> I assume header iteration now alsop works when building for the browser?

result.bindingsStream.on('data', function (data) {
            for (const el of data) {
                console.log(el);
            }
          console.log(data);
        });

I tried this as my script in the html file. Tested in the actor init sparql browser that all are utterable. Not sure if this is enough? Or should another script be used.

rubensworks:
> I tried this as my script in the html file. Tested in the actor init sparql browser that all are utterable. Not sure if this is enough? Or should another script be used.

You are testing iterability of the query results here, which is a different thing than the headers.

I would suggest adding a temporary header iteration in one of the HTTP actors with some dummy console.logs, building that for the browser, running it in a browser, and checking if you see the expected console.log's.

Side-note: I see that you removed one line from the root package.json. I meant that _all_ changes you made to this package.json file should not be needed.

stephaniech97:
> > I tried this as my script in the html file. Tested in the actor init sparql browser that all are utterable. Not sure if this is enough? Or should another script be used.
> 
> You are testing iterability of the query results here, which is a different thing than the headers.
> 
> I would suggest adding a temporary header iteration in one of the HTTP actors with some dummy console.logs, building that for the browser, running it in a browser, and checking if you see the expected console.log's.
> 
> Side-note: I see that you removed one line from the root package.json. I meant that _all_ changes you made to this package.json file should not be needed.

Now I have checked that it does work in the proper way. The root package.json is also reset to the previous version

rubensworks:
I see there are some merge conflicts though, could you resolve them so I can merge?

PR #689, opened 2020-07-09
Replace json-stable-stringify dependency, Closes #590

### LINKED ISSUE #590: Replace json-stable-stringify dependency
We currently make use of the [json-stable-stringify](https://github.com/substack/json-stable-stringify) for consistent hashing.

Since JSON-LD will start requiring a standard canonical JSON stringification in parsers, we could make use of https://github.com/erdtman/canonicalize, since this package will start being included by our JSON-LD dependencies in any case. It's also a bit smaller than the current json-stable-stringify dependency.

stephaniech97:


packages/actor-query-operation-distinct-hash/package.json
```diff
@@ -36,7 +36,7 @@
   ],
   "dependencies": {
     "@comunica/actor-abstract-bindings-hash": "^1.13.0",
-    "json-stable-stringify": "^1.0.1"
+    "canonicalize": "^1.0.1"
```
rubensworks, line 39: It looks like this dependency can even be removed here, since it appears it's not being used directly in this package.

Issue #662, opened 2020-05-18
HTTP endpoint uses stats writer as default


rubensworks:
#### Issue type:

- :bug: Bug

____
#### Description:

When using `bin/http.js`, the stats writer is used as default when visiting via browser.
We should use a more user-friendly format instead as default.
This default should be the same as the CLI of comunica, and should preferable not be hardcoded.

rubensworks:
Instead of #695, a cleaner solution would be to tweak the default priorities of serialization formats so that they correspond to the defaults of the CLI.

Should this not get us there, pre-filtering formats by result type should do the trick.

PR #695, opened 2020-07-15
Issue 662 http endpoint

stephaniech97:


stephaniech97:
> I would prefer to not have hardcoded media types in here, as this will make maintenance a lot harder.
> The only place that can hardcode media types is here (and should not be changed): https://github.com/comunica/comunica/blob/master/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts#L191-L203
> 
> Have you been able to detect why exactly the stats media type is picked when doing a browser request? This would be a good first step in finding a clean solution.

The reason why it is by default is that stats is always first mentioned in the variants. So you could also just swap them around. But it won't change the fact that we need to check the type of the query to decide what we will take (appl/json, simple or appl/trig)

stephaniech97:
> I would prefer to not have hardcoded media types in here, as this will make maintenance a lot harder.
> The only place that can hardcode media types is here (and should not be changed): https://github.com/comunica/comunica/blob/master/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts#L191-L203
> 
> Have you been able to detect why exactly the stats media type is picked when doing a browser request? This would be a good first step in finding a clean solution.

i have checked the reason why stats is always default. And the conclusion was that the sorting didn't remain its ordering when values were equal. So this new version checks which equal valued quality variant to take considering given ordering of variants.

packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
```diff
@@ -254,4 +254,4 @@ export interface IActorInitSparqlArgs extends IActorArgs<IActionInit, IActorTest
 export const KEY_CONTEXT_INITIALBINDINGS: string = '@comunica/actor-init-sparql:initialBindings';
 export const KEY_CONTEXT_QUERYFORMAT: string = '@comunica/actor-init-sparql:queryFormat';
 export const KEY_CONTEXT_GRAPHQL_SINGULARIZEVARIABLES: string = '@comunica/actor-init-sparql:singularizeVariables';
-export const KEY_CONTEXT_LENIENT: string = '@comunica/actor-init-sparql:lenient';
+export const KEY_CONTEXT_LENIENT: string = '@comunica/actor-init-sparql:lenient';
```
rubensworks, line 257: Unneeded blank line removal.

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -158,8 +159,48 @@ Options:
...
+
+      if (isValid && negotiation.length > 1) {
+        // The preferred negotiation
+        let finalNegotiation = negotiation[0];
+        // Best quality achievable
+        const quality = negotiation[0].q;
+        let found = false;
+        let k = 0;
+        // Iterate through all the variants in order
+        while (!found) {
```
rubensworks, line 180: I'm wondering if this algorithm can be made more efficient, as it's O(n^2) at the moment.
Wouldn't you just need to find the negotiation with the highest q?
If so, this could even be a one-liner.

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -158,8 +159,48 @@ Options:
...
+              }
+            }
+            l++;
+          }
+          k++;
+        }
+        mediaType = finalNegotiation.type;
+      }
+
+    } catch (exception) {
```
rubensworks, line 201: Why do you need a try-catch here?

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
stephaniech97: We have the negotiation with the highest quality, but it isn't necessarily unique, so finding the highest quality doesn't give the default preferred writer. we would for example have gotten a list from the negotiation as follows: [{type:stats: quality:1},{type:application/json: quality:1},{type:simple: quality:1}] However we prefer it as follows (given with the variants: [{type:application/json: quality:1},{type:simple: quality:1},{type:stats: quality:1}]. Then we would need to check if the application/json has the same quality given by the negotiator as the given negotiation[0]. And if not check for the second preferred variant. The main issue is that the negotiate library doesn't keep the ordering of same quality elements

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
stephaniech97: require('negotiate').choose(variants, request); Is not always possible since it throws an error so then i catch it but don't do anything with it since it will basically result into null and undefined of the previous declared values

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
rubensworks: Ah ok, got it.

Instead of having a double loop for this, wouldn't it be more efficient to just iterate over _all_ original variants (ordered by preference), where the first one that is found in the negotiated list is returned?

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
rubensworks: When does it throw an error?
We did not catch those errors before. Does this mean that we had a bug?

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
stephaniech97: the negotiated list gives back all qualities. So if it were to be 1 as best it would have a list [1,1,1,0.9,0.5,0.4]. Meaning that if contains wouldn't work. Wouldn't your suggestion also be a double loop internally? Contains is O(n) then n time in worst case would also be O(n^2)?

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
stephaniech97: The tests indeed never failed on it before. But somehow when I was writing it failed on the variant can not sort, so an issue in the negotiate package itself. I will recheck to look whether or not i did something fishy

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
rubensworks: I don't see why containment checking wouldn't work. The returned list _also_ contains the mediatype labels, right? So I would just compare by them.

And you're right about complexity, that would be the same. However, readability would be improved, since a lot less code would needed.

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -158,8 +159,31 @@ Options:
...
-    const mediaType: string = request.headers.accept && request.headers.accept !== '*/*'
-      ? require('negotiate').choose(variants, request)[0].type : null;
+
+    // Negotiate the best mediatype format
+    let mediaType = null;
+    const isValid = request.headers.accept && request.headers.accept !== '*/*';
+    let negotiation = isValid ? require('negotiate').choose(variants, request) : null;
+    mediaType = negotiation && negotiation.length > 0 ? negotiation[0].type : null;
+
+    if (negotiation && negotiation.length > 1) {
```
rubensworks, line 169: I still think this can be simplified significantly.

Do you think something like this would work?

```
// Assuming variants is sorted by q in decreasing order (pseudocode)
// Create negotiationsIndexed so that it is a hash containing the media types from 'negotiation' as keys.
for (variant of variants) {
  if (negotiationsIndexed[variant]) {
    mediaType = variant;
    break;
  }
}
```

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -156,12 +157,34 @@ Options:
...
+    const isValid = request.headers.accept && request.headers.accept !== '*/*';
+    const negotiation = isValid ? require('negotiate').choose(variants, request) : null;
+    let mediaType = negotiation && negotiation.length > 0 ? negotiation[0].type : null;
+
+    const negotiationsIndexed: any = {};
+    if (negotiation && negotiation.length > 1) {
+      const quality = negotiation[0].q;
+      // Assuming variants is sorted by q in decreasing order (pseudocode)
+      // Create negotiationsIndexed so that it is a hash containing the media types from 'negotiation' as keys.
+      negotiation.filter((element: any) => element.q === quality)
```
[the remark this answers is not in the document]
stephaniech97: negotiation is sorted in the highest quality being in the front. So the element.q will never be higher than quality

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -147,6 +148,40 @@ Options:
     stderr.write(`Server running on http://localhost:${this.port}/sparql\n`);
   }
 
+  /**
+   * Negotiatiate the best mediatype format
+   * @param {module:http.IncomingMessage} request Request object.
+   * @param {{type: string; quality: number}[]} variants Allowed variants sorted in preferred order.
+   */
+  public contentNegotiation(request: http.IncomingMessage,
```
rubensworks, line 156: `determineMediaType` would be a better method name.

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -147,6 +148,40 @@ Options:
...
+        }, {});
+
+      // Pick the first configured variant that has the desired quality.
+      for (const variant of variants) {
+        if (negotiationsIndexed[variant.type]) {
+          mediaType = variant.type;
+          break;
+        }
+      }
+      return mediaType;
```
rubensworks, line 180: This return can be removed, since going outside the if-block will return the exact same thing.

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -156,12 +191,12 @@ Options:
...
    * @param {module:http.ServerResponse} response Response object.
    */
+
   public async handleRequest(engine: ActorInitSparql, variants: { type: string; quality: number }[],
     stdout: Writable, stderr: Writable,
     request: http.IncomingMessage, response: http.ServerResponse): Promise<void> {
-    const mediaType: string = request.headers.accept && request.headers.accept !== '*/*' ?
-      // eslint-disable-next-line multiline-ternary
-      require('negotiate').choose(variants, request)[0].type : null;
+    // Negotiate the best mediatype format
```
rubensworks, line 198: Comment can be removed, since the method name already describes well what will happen.

packages/actor-init-sparql/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -761,3 +686,360 @@ describe('HttpServiceSparqlEndpoint', () => {
     });
   });
 });
+
+describe('A second HttpServiceSparqlEndpoint instance', () => {
```
rubensworks, line 690: Is there a reason why you decided to move these tests into a new describe block with a second instance?

packages/actor-init-sparql/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -761,3 +686,360 @@ describe('HttpServiceSparqlEndpoint', () => {
...
+    request.headers = { 'content-type': 'contenttypewhichdefinitelydoesnotexist',
+      accept: 'application/json,application/trig,simple,stats' };
+    request.method = 'GET';
+    return request;
+  }
+
+  it('should call writeQueryResult with the preferred default'
+    , async() => {
+      const choose = jest.fn(() => negotiation);
+      jest.doMock('negotiate', () => {
```
rubensworks, line 1020: Is there a need to mock? Testing-wise, it would be safer if we could call the original logic.

packages/actor-init-sparql/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -761,3 +686,360 @@ describe('HttpServiceSparqlEndpoint', () => {
...
+        return {
+          choose,
+        };
+      });
+      expect(instance.contentNegotiation(request, variants)).toEqual('application/trig');
+    });
+
+  it('should call writeQueryResult with null as mediatype'
```
rubensworks, line 1028: Test name does not really describe what is happening here.

packages/actor-init-sparql/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -761,3 +686,360 @@ describe('HttpServiceSparqlEndpoint', () => {
...
+  ql: 0.5,
+  qls: 0,
+  qc: 1,
+  qcs: 0,
+  qe: 1,
+  qes: 0,
+  qs: 0.2,
+  q: 0.08 }];
+
+describe('HandleNegotiation', () => {
```
rubensworks, line 987:
```suggestion
describe('content negotiation', () => {
```

packages/actor-init-sparql/test/HttpServiceSparqlEndpoint-test.ts
rubensworks: Also, I feel like we're missing many cases here with just two tests, so we should add a lot more of them.

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -14,6 +14,7 @@ import { newEngineDynamic } from '..';
 import { ActorInitSparql } from './ActorInitSparql';
 import { IQueryOptions } from './QueryDynamic';
 
+// Tslint:disable:no-var-requires
```
rubensworks, line 17: We're not using tslint anymore, so this should not be needed here.

packages/actor-init-sparql/test/MediaTypeSelection-test.ts
```diff
@@ -0,0 +1,105 @@
+import minimist = require('minimist');
```
rubensworks, line 1: Creating a separate test file doesn't feel like a good solution here. The convention is that each class has a corresponding test file. But `MediaTypeSelection` is not an existing class, so we shouldn't have this here.

Instead, I would just create a new describe block (`'determineMediaType'`) under the `'HttpServiceSparqlEndpoint'` describe block.
If you run into problems with conflicting mocks, I think you can scope them to specific describe blocks, but I'm sure the jest documentation can tell you more about that.

packages/actor-init-sparql/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -870,3 +871,76 @@ describe('HttpServiceSparqlEndpoint', () => {
...
+  { type: 'text/turtle', quality: 0.6 },
+  { type: 'application/n-triples', quality: 0.3 },
+  { type: 'text/n3', quality: 0.2 },
+  { type: 'application/ld+json', quality: 0.9 }];
+
+describe('determineMediaType', () => {
+  let instance: any;
+  let request: any;
+  let response: any;
+  let variants: any;
```
rubensworks, line 893: This field is always undefined?

packages/actor-init-sparql/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -511,6 +511,7 @@ describe('HttpServiceSparqlEndpoint', () => {
 
         expect(instance.writeQueryResult)
           .toHaveBeenCalledWith(engine, stdout, stderr, request, response, 'test_parseBody_result', chosen, false);
+        jest.dontMock('negotiate');
```
rubensworks, line 514: The end of a test feels like a strange place for this.
Do things still work properly when executing tests separately?
If not, then I would suggest moving this to the `beforeEach` of the `determineMediaType` block.

packages/actor-init-sparql/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -870,3 +871,76 @@ describe('HttpServiceSparqlEndpoint', () => {
...
+  { type: 'application/n-quads', quality: 0.7 },
+  { type: 'text/turtle', quality: 0.6 },
+  { type: 'application/n-triples', quality: 0.3 },
+  { type: 'text/n3', quality: 0.2 },
+  { type: 'application/ld+json', quality: 0.9 }];
+
+describe('determineMediaType', () => {
+  let instance: any;
+  let request: any;
+  let response: any;
```
rubensworks, line 892: This field is never used?

packages/actor-init-sparql/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -870,3 +871,76 @@ describe('HttpServiceSparqlEndpoint', () => {
...
+    instance = new HttpServiceSparqlEndpoint({});
+    instance.writeQueryResult = jest.fn();
+    request = makeRequest();
+    response = new ServerResponseMock();
+  });
+
+  function makeRequest() {
+    request = stringToStream('default_test_request_content');
+    request.url = 'url_sparql';
+    request.headers = { 'content-type': 'contenttypewhichdefinitelydoesnotexist',
```
rubensworks, line 905: A request should not have a content type. AFAIK, the negotiation code never checks this either?

packages/actor-init-sparql/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -870,3 +871,76 @@ describe('HttpServiceSparqlEndpoint', () => {
...
+
+  it('should choose second preferred mediatype if it is present in accept', async() => {
+    request.headers = { accept: 'application/trig,simple,stats' };
+    instance.parseBody = jest.fn(() => Promise.resolve('test_result'));
+    request.method = 'POST';
+    expect(instance.determineMediaType(request, variantsDefault)).toEqual('simple');
+  });
+
+  it('should choose the next preferred variant available in accept', async() => {
+    request.headers = { accept: 'application/trig,stats' };
```
rubensworks, line 941: Could you also add some test to check what happens with different q-values in the accept header?

packages/actor-init-sparql/test/HttpServiceSparqlEndpoint-test.ts
stephaniech97: They do work properly separately. The mock is only used in that one test. So i thought it would be more efficient to turn it off in the same test. But if preferred I can put it in the determineMediatype.

rubensworks: Ah, I missed the `doMock` earlier in that test. Never mind my comment 🙂

packages/actor-init-sparql/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -870,3 +871,84 @@ describe('HttpServiceSparqlEndpoint', () => {
...
+    instance.parseBody = jest.fn(() => Promise.resolve('test_result'));
+    request.method = 'POST';
+    expect(instance.determineMediaType(request, variantsDefault)).toEqual('stats');
+  });
+
+  it('should choose the preferred variant with different q values', async() => {
```
rubensworks, line 941: I suspect you forgot to add the actual q-values as specified here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept

stephaniech97: Variants are passed with their quality to the handleRequest function which is why i did not need to add them . But I'll add tests which start from run then to make it parse the mediatypes by itself instead of passing it to the function.

rubensworks: But these tests here don't go via `handleRequest`, right?
So those variants are not used here.

I would suggest having some tests with header in the form of `{ accept: 'application/n-quads;q=0.5,tree,application/n-triples;q=0.8,application/ld+json;q=0.2' }`

stephaniech97: handleRequest would call the determineMediaType with the variants passed as arguments. So it would be the same thing.

rubensworks: Sure, but in this block we are specifically testing `determineMediaType`, so we should test exactly that. From the point of view of `determineMediaType`, `handleRequest` should not even exist, so all relevant inputs that can be given to `determineMediaType` should be tested.

rubensworks:
Offline discussion made clear that this is not the proper way forward, see https://github.com/comunica/comunica/issues/662#issuecomment-671319431

PR #696, opened 2020-07-23
Feature improve definition of sources communica sparql http

### LINKED ISSUE #433: Improve definition of sources in comunica-sparql-http
#### Issue type:

____
#### Description:

Currently, sources have to defined in a verbose way when setting up a SPARQL endpoint:

```
comunica-sparql-http "{ \"sources\": [{ \"type\": \"hypermedia\", \"value\" : \"http://fragments.dbpedia.org/2015/en\" }]}"
```

We should allow a way to define sources in the same way as `comunica-sparql`:

```
comunica-sparql-http http://fragments.dbpedia.org/2015/en
```

The context-based option should still be possible by using the `-c` option.

FlorianFV:
Closes #433


packages/actor-init-sparql/lib/ActorInitSparql.ts
```diff
@@ -25,57 +25,128 @@ export {
...
-        const yarnVersion: string = await this.getScriptOutput('yarn -v', '_Yarn is unavailable_');
-        const os = `${OS.platform()} (${OS.type()} ${OS.release()})`;
-        return { stderr: require('streamify-string')(`| software            | version
+  public static isDevelopmentEnvironment(): boolean {
+    return existsSync(`${__dirname}/../test`);
+  }
+
+  public static async buildContext(args: minimist.ParsedArgs, queryOrHttp: number, helpMessage: string,
+    queryString?: string): Promise<any> {
+    const query: boolean = queryOrHttp === 0;
```
rubensworks, line 70: `queryOrHttp` is used nowhere except here. I would remove the param `queryOrHttp` in favour of a boolean `queryContext`, which would do the same as your current `query` variable.

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -25,18 +25,21 @@ export class HttpServiceSparqlEndpoint {
...
   public static readonly HELP_MESSAGE = `comunica-sparql-http exposes a Comunica engine as SPARQL endpoint
 
-context should be a JSON object or the path to such a JSON file.
-
 Usage:
-  comunica-sparql-http context.json [-p port] [-t timeout] [-l log-level] [-i] [--help]
-  comunica-sparql-http "{ \\"sources\\": [{ \\"type\\": \\"hypermedia\\", \\"value\\" : \\"http://fragments.dbpedia.org/2015/en\\" }]}" [-p port] [-t timeout] [-l log-level] [-i] [--help]
+  comunica-sparql-http -c context.json [-p port] [-t timeout] [-l log-level] [-i] [-b] [--help] [--version] [--lenient]
+  comunica-sparql-http -c "{ \\"sources\\": [{ \\"type\\": \\"hypermedia\\", \\"value\\" : \\"http://fragments.dbpedia.org/2015/en\\" }]}" [-p port] [-t timeout] [-l log-level] [-i] [-b] [--help] [--version] [--lenient]
+  comunica-sparql-http hypermedia@http://fragments.dbpedia.org/2015/en [-p port] [-t timeout] [-l log-level] [-i] [-b] [--help] [--version] [--lenient]
```
rubensworks, line 31: I would move this line to be the first example (so it's clear that this is the "proper" way of calling it).
I would also add an example with 2 sources (like the other help message).

packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -25,18 +25,21 @@ export class HttpServiceSparqlEndpoint {
 
   public static readonly HELP_MESSAGE = `comunica-sparql-http exposes a Comunica engine as SPARQL endpoint
 
-context should be a JSON object or the path to such a JSON file.
-
 Usage:
-  comunica-sparql-http context.json [-p port] [-t timeout] [-l log-level] [-i] [--help]
-  comunica-sparql-http "{ \\"sources\\": [{ \\"type\\": \\"hypermedia\\", \\"value\\" : \\"http://fragments.dbpedia.org/2015/en\\" }]}" [-p port] [-t timeout] [-l log-level] [-i] [--help]
+  comunica-sparql-http -c context.json [-p port] [-t timeout] [-l log-level] [-i] [-b] [--help] [--version] [--lenient]
```
rubensworks, line 29: No need to add all options here, it adds too much noise. They are described below in detail in any case.

packages/actor-init-sparql/lib/ActorInitSparql.ts
```diff
@@ -25,57 +25,128 @@ export {
...
-  -b            base IRI for the query (e.g., http://example.org/)
-  -l            sets the log level (e.g., debug, info, warn, ... defaults to warn)
-  -d            sets a datetime for querying Memento-enabled archives
-  -p            delegates all HTTP traffic through the given proxy (e.g. http://myproxy.org/?uri=)
-  --lenient     if failing requests and parsing errors should be logged instead of causing a hard crash
-  --help        print this help message
-  --listformats prints the supported MIME types
-  --version     prints version information
-`) };
+      return { stderr: helpMessage };
```
rubensworks, line 98: Instead of returning this as the context, I would throw an error here (reject the promise). Then we can put a try-catch around this when we call this function, and we can handle it accordingly.

packages/actor-init-sparql/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -122,14 +149,27 @@ describe('HttpServiceSparqlEndpoint', () => {
         moduleRootPath,
         env,
         defaultConfigPath,
-        exit);
-
-      expect(exit).toHaveBeenCalledWith(1);
-      expect(stderr.toString()).toBe(`${HttpServiceSparqlEndpoint.HELP_MESSAGE}Server running on http://localhost:3000/sparql\n`);
+        exit).then(res => {
+        expect(exit).toHaveBeenCalledWith(1);
+        expect(stderr.toString()).toBeDefined();
```
rubensworks, line 154: Where possible, I would check for an exact string in these unit tests. Because now it's still possible that the stderr value is defined, but is the wrong value.

packages/actor-init-sparql/lib/ActorInitSparql.ts
```diff
@@ -25,57 +25,128 @@ export {
  * A comunica SPARQL Init Actor.
  */
 export class ActorInitSparql extends ActorInitSparqlBrowser {
-  public constructor(args: IActorInitSparqlArgs) {
```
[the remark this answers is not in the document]
FlorianFV: The linter was complaining about an unnecessary constructor.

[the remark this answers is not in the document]
rubensworks: There's probably something going wrong on your end then, as Travis (and myself) is not having this linting error.
This constructor is definititely necessary though :-)

packages/actor-init-sparql/lib/ActorInitSparql.ts
```diff
@@ -25,57 +25,131 @@ export {
...
 | Comunica Init Actor | ${comunicaVersion} ${dev}
 | node                | ${nodeVersion}
 | npm                 | ${npmVersion}
 | yarn                | ${yarnVersion}
 | Operating System    | ${os}
-`) };
-      }
+`);
+
+      return Promise.reject(message);
```
rubensworks, line 92: `Promise.reject` is only allowed to take an Error, so you'll have to catch the error, and pass the message and streamify it higher up.

FlorianFV: The code I wrote did work, the help and version messages were written on the command line. When I change it to "new Error(message)" and I print the error, there is "Error:" before the help or version message and I don't think that is wanted?

packages/actor-init-sparql/lib/ActorInitSparql.ts
```diff
@@ -94,21 +166,18 @@ Options:
     } else if (args.f) {
       query = readFileSync(args.f, { encoding: 'utf8' });
     } else {
-      query = args._.pop();
+      query = args._[args._.length - 1];
```
rubensworks, line 169: Is there any reason for changing this?

rubensworks: Ah, nvm, just saw the commit message.

FlorianFV: This was because this piece of code used to run after the check if a help message was needed, but now that check is in the buildContext method, this piece of code runs before it, because the buildContext function consumes all the args to make it sources.
If no arguments are given, this index is actually -1, so I will add a check for that.

rubensworks:
Great work, thanks!

PR #699, opened 2020-07-24
Add bindings() option to collect bindings generated by bindingsStream/quadStream

### LINKED ISSUE #635: Add option to produce non-streaming results
#### Issue type:

____
#### Description:

For many use cases, the streaming capabilities of Comunica are irrelevant.
For instance, the following is a common pattern:
```javascript
const array = [];
result.bindingsStream.on('data', (data) => {
  array.push(data);
});
result.bindingsStream.on('end', () => {
  // handle array
});
```

We could additionally introduce something like this for convenience:

```
const array = await result.bindings();
```

This should also be done for quadStreams.

FlorianFV:
Closes #635

### DIFF
--- a/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
+++ b/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
@@ -6,16 +6,16 @@ import {
   IActionOptimizeQueryOperation,
   IActorOptimizeQueryOperationOutput,
 } from '@comunica/bus-optimize-query-operation';
-import {
-  ensureBindings,
+import { ensureBindings,
   IActionQueryOperation,
   IActorQueryOperationOutput,
   KEY_CONTEXT_BASEIRI,
   KEY_CONTEXT_QUERY_TIMESTAMP,
   materializeOperation,
-} from '@comunica/bus-query-operation';
+  IActorQueryOperationOutputBindings, IActorQueryOperationOutputQuads } from '@comunica/bus-query-operation';
 import { IDataSource, isDataSourceRawType, KEY_CONTEXT_SOURCES } from '@comunica/bus-rdf-resolve-quad-pattern';
 import { IActionSparqlParse, IActorSparqlParseOutput } from '@comunica/bus-sparql-parse';
+
 import {
 
   IActionSparqlSerialize,
@@ -169,6 +169,32 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
     // Execute query
     const resolve: IActionQueryOperation = { context, operation };
     const output = await this.mediatorQueryOperation.mediate(resolve);
+    // Set bindings
+    if ((<any> output).bindingsStream) {
+      (<IActorQueryOperationOutputBindings> output).bindings = () => {
+        const result: any[] = [];
+        (<IActorQueryOperationOutputBindings> output).bindingsStream.on('data', data => {
+          result.push(data.toObject());
+        });
+        return new Promise((resolveP, reject) => {
+          (<IActorQueryOperationOutputBindings> output).bindingsStream.on('end', () => {
+            resolveP(result);
+          });
+        });
+      };
+    } else if ((<any> output).quadStream) {
+      (<IActorQueryOperationOutputQuads> output).bindings = () => {
+        const result: any[] = [];
+        (<IActorQueryOperationOutputQuads> output).quadStream.on('data', data => {
+          result.push(data);
+        });
+        return new Promise((resolveP, reject) => {
+          (<IActorQueryOperationOutputQuads> output).quadStream.on('end', () => {
+            resolveP(result);
+          });
+        });
+      };
+    }
     output.context = context;
     return output;
   }
--- a/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
+++ b/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
@@ -87,7 +87,12 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
     super(args);
   }
 
-  public static enhanceQueryResults(results: IQueryResult): IQueryResult {
+  /**
+   * Add convenience methods to query results
+   * @param {IActorQueryOperationOutput} results Basic query results.
+   * @return {IQueryResult} Same query results with added fields.
+   */
+  public static enhanceQueryResults(results: IActorQueryOperationOutput): IQueryResult {
     // Set bindings
     if ((<IQueryResultBindings> results).bindingsStream) {
       (<IQueryResultBindings> results).bindings = () => {
@@ -116,7 +121,7 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
         });
       };
     }
-    return results;
+    return <IQueryResult> results;
   }
 
   public async test(action: IActionInit): Promise<IActorTest> {
--- a/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
+++ b/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
@@ -12,7 +12,9 @@ import { ensureBindings,
   KEY_CONTEXT_BASEIRI,
   KEY_CONTEXT_QUERY_TIMESTAMP,
   materializeOperation,
-  IActorQueryOperationOutputBindings, IActorQueryOperationOutputQuads } from '@comunica/bus-query-operation';
+  IActorQueryOperationOutputBindings,
+  IActorQueryOperationOutputQuads,
+  IActorQueryOperationOutputBoolean } from '@comunica/bus-query-operation';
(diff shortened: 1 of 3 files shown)


packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
```diff
@@ -169,6 +169,32 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
     // Execute query
     const resolve: IActionQueryOperation = { context, operation };
     const output = await this.mediatorQueryOperation.mediate(resolve);
+    // Set bindings
```
rubensworks, line 172: I would suggest moving all of this new code into a new static method (`enhanceQueryResults`?). This would also simplify testing.

packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
```diff
@@ -169,6 +169,32 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
...
+    if ((<any> output).bindingsStream) {
```
rubensworks, line 173: Casting to `any` should be avoided if possible. Does the compiler accept `'bindingsStream' in output` instead?

packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
```diff
@@ -169,6 +169,32 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
...
+      (<IActorQueryOperationOutputBindings> output).bindings = () => {
+        const result: any[] = [];
+        (<IActorQueryOperationOutputBindings> output).bindingsStream.on('data', data => {
+          result.push(data.toObject());
```
rubensworks, line 177: No need to convert to object.

packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
```diff
@@ -169,6 +169,32 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
     // Execute query
     const resolve: IActionQueryOperation = { context, operation };
     const output = await this.mediatorQueryOperation.mediate(resolve);
+    // Set bindings
+    if ((<any> output).bindingsStream) {
+      (<IActorQueryOperationOutputBindings> output).bindings = () => {
+        const result: any[] = [];
```
rubensworks, line 175: Result type can be made more concrete.

packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
```diff
@@ -169,6 +169,32 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
...
+        });
+        return new Promise((resolveP, reject) => {
+          (<IActorQueryOperationOutputBindings> output).bindingsStream.on('end', () => {
+            resolveP(result);
+          });
+        });
+      };
+    } else if ((<any> output).quadStream) {
+      (<IActorQueryOperationOutputQuads> output).bindings = () => {
+        const result: any[] = [];
```
rubensworks, line 187: Result type can be made more concrete.

packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
```diff
@@ -169,6 +169,32 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
...
     const resolve: IActionQueryOperation = { context, operation };
     const output = await this.mediatorQueryOperation.mediate(resolve);
+    // Set bindings
+    if ((<any> output).bindingsStream) {
+      (<IActorQueryOperationOutputBindings> output).bindings = () => {
+        const result: any[] = [];
+        (<IActorQueryOperationOutputBindings> output).bindingsStream.on('data', data => {
+          result.push(data.toObject());
+        });
+        return new Promise((resolveP, reject) => {
```
rubensworks, line 179:
```suggestion
        return new Promise((resolve, reject) => {
```

packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
```diff
@@ -169,6 +169,32 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
...
+          });
+        });
+      };
+    } else if ((<any> output).quadStream) {
+      (<IActorQueryOperationOutputQuads> output).bindings = () => {
+        const result: any[] = [];
+        (<IActorQueryOperationOutputQuads> output).quadStream.on('data', data => {
+          result.push(data);
+        });
+        return new Promise((resolveP, reject) => {
```
rubensworks, line 191:
```suggestion
        return new Promise((resolve, reject) => {
```

packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
rubensworks: Also, reject is never called. You should also listen to error events and forward those.

packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
rubensworks: Also, reject is never called. You should also listen to error events and forward those.

packages/bus-query-operation/lib/ActorQueryOperation.ts
```diff
@@ -249,6 +249,10 @@ export interface IActorQueryOperationOutputBindings extends IActorQueryOperation
    * The stream of bindings resulting from the given operation.
    */
   bindingsStream: BindingsStream;
+  /**
+   * The collection of bindings after an 'end' event occured.
+   */
+  bindings?: () => Promise<any[]>;
```
rubensworks, line 255: Result type should be more concrete.

packages/bus-query-operation/lib/ActorQueryOperation.ts
```diff
@@ -269,6 +273,10 @@ export interface IActorQueryOperationOutputQuads extends IActorQueryOperationOut
    * The stream of quads.
    */
   quadStream: RDF.Stream & AsyncIterator<RDF.Quad>;
+  /**
+   * The collection of bindings after an 'end' event occured.
+   */
+  bindings?: () => Promise<any[]>;
```
rubensworks, line 279: Result type should be more concrete.
Also, `quads`?

packages/bus-query-operation/lib/ActorQueryOperation.ts
```diff
@@ -249,6 +249,10 @@ export interface IActorQueryOperationOutputBindings extends IActorQueryOperation
    * The stream of bindings resulting from the given operation.
    */
   bindingsStream: BindingsStream;
+  /**
```
rubensworks, line 252: These new fields being optional makes them very inconvenient to use.

In the actor-init-sparql package, I would suggest creating the following instead:
* IQueryResultBindings: extends from IActorQueryOperationOutputBindings, and adds this new bindings method.
* IQueryResultQuads: extends from IActorQueryOperationOutputQuads, and adds this new quads method.
* IQueryResultBoolean: extends from IActorQueryOperationOutputBoolean, and adds nothing.
* IQueryResult: _union type_ over the three above.

The return type of `query` in ActorInitSparql can then be changed to return `IQueryResult`.
Then the "optional" any "any" stuff won't be needed anymore.

packages/actor-init-sparql/test/ActorInitSparql-test.ts
```diff
@@ -583,6 +584,13 @@ describe('ActorInitSparql', () => {
       });
     });
 
+    it('bindings() should collect all bindings until "end" event occurs on triples', async() => {
```
rubensworks, line 587: The tests look good, but will probably have to be changed a bit after my suggestions.

Could you also add some tests for empty results and erroring results?

packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
```diff
@@ -84,6 +87,38 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
     super(args);
   }
 
+  public static enhanceQueryResults(results: IQueryResult): IQueryResult {
```
rubensworks, line 90: Input here should be of type `IActorQueryOperationOutput`, output type is correct.

rubensworks, line 90: Add tsdoc comment above, such as _Add convenience methods to query results_, with a description of the param (and return type).

packages/bus-query-operation/lib/ActorQueryOperation.ts
```diff
@@ -237,7 +237,7 @@ export function getMetadata(actionOutput: IActorQueryOperationOutputStream): Pro
 }
 
 /**
- * Query operation output for a bindings stream.
+ * Query operation output for a bindings stream. (without bindings operator)
```
rubensworks, line 240: The comment changes in this file are not really needed, I think they make the description too complex.

PR #701, opened 2020-07-30
fix network request without headers, Closes #700

### LINKED ISSUE #700: Comunica makes network requests without headers on Node.js
#### Issue type:

- :bug: Bug

____
#### Description:

I was trying to use [@comunica/actor-init-sparql](https://github.com/comunica/comunica/tree/master/packages/actor-init-sparql) with the [Wikidata LDF service](https://query.wikidata.org/bigdata/ldf). Based on the README, this one-liner should more or less work:

```js
void require( '@comunica/actor-init-sparql' ).newEngine().query( 'SELECT * WHERE { <http://www.wikidata.org/entity/Q42> <http://www.wikidata.org/prop/direct/P31> ?class. }', { sources: [ 'https://query.wikidata.org/bigdata/ldf' ] } ).then( ( { bindingsStream } ) => bindingsStream.on( 'data', data => console.log( data.toObject() ) ), console.error )
```

(It’s not very nice code, but having it as a one-liner was convenient while investigating this issue. The `void` operator avoids printing the returned `Promise` in the REPL.) However, it fails because Wikidata responds with a 403 Forbidden error. I think this is because the request is missing a User-Agent header (which is [required on all Wikimedia sites](https://meta.wikimedia.org/wiki/User-Agent_policy)); when you add some `console.log()`s on request and response objects, it certainly *looks* like the request is including headers (including a default user agent), but buried deeply in Node internal objects you also find this:

```js
    _header: 'GET /bigdata/ldf HTTP/1.1\r\n' +                                                                                                                               
      'Host: query.wikidata.org\r\n' +                                                                                                                                       
      'Connection: close\r\n' +                                                                                                                                              
      '\r\n',
```

That seems to be the actual TCP payload for the HTTP request, and it doesn’t have any custom headers. I *think* this is because the `headers` object which Comunica passes down into Node’s request library ([http](https://nodejs.org/api/http.html)/[https](https://nodejs.org/api/https.html), wrapped in [follow-redirects](https://www.npmjs.com/package/follow-redirects), called in [@comunica/actor-http-native’s Requester](https://github.com/comunica/comunica/blob/d211aef87125a171b52ecc56ccd06a25b3a6a3b7/packages/actor-http-native/lib/Requester.ts)) is an instance of some `Headers` class (probably the one from [node-fetch](https://www.npmjs.com/package/node-fetch)), whereas Node expects a plain object as `headers`; since Node doesn’t recognize the `Headers` object, the request is made without any custom headers.

The following JS code produces the same issue:
```js
const https = require( 'https' );
const { Headers } = require( 'node-fetch' );

https.request(
	'https://query.wikidata.org/bigdata/ldf',
	{
		headers: new Headers( {
			accept: 'application/n-quads,application/trig;q=0.95,application/ld+json;q=0.9,application/n-triples;q=0.8,text/turtle;q=0.5,application/rdf+xml;q=0.5,application/json;q=0.45,text/n3;q=0.35,application/xml;q=0.3,text/xml;q=0.3,image/svg+xml;q=0.3,text/html;q=0.2,application/xhtml+xml;q=0.18',
			'user-agent': 'Comunica/actor-http-native (Node.js v14.5.0; linux)',
		} ),
	},
	( { statusCode, req: { _header } } ) => {
		console.log( statusCode );
		console.log( _header );
	},
).end();
```

____
#### Environment:

| Comunica Init Actor | 1.14.0 
| node                | v14.5.0
| npm                 | 6.14.5
| yarn                | _Yarn is unavailable_
| Operating System    | linux (Linux 5.4.0-40-generic)

stephaniech97:


packages/actor-http-native/lib/Requester.ts
```diff
@@ -56,14 +68,24 @@ export default class Requester {
         response.pipe(decoded);
         // Copy response properties
         decoded.statusCode = response.statusCode;
-        decoded.headers = response.headers;
+        // Wrap headers into an header object type
+        const responseHeaders: any = new Headers();
```
rubensworks, line 72: Duplicated code should be avoided at all cost (see code starting at line 84).
I would suggest abstracting this into a method such as `convertFetchHeadersToHash`.

packages/actor-http-native/lib/ActorHttpNative.ts
```diff
@@ -63,41 +63,56 @@ export class ActorHttpNative extends ActorHttp {
...
+      // Expose fetch cancel promise
+      httpResponse.cancel = () => {
+        httpResponse.destroy();
+        return Promise.resolve(undefined);
+      };
+      // Missing several of the required fetch fields
+
+      let headers = new Headers(httpResponse.headers);
+      if (httpResponse && httpResponse.input && httpResponse.input.headers) {
+        headers = httpResponse.input.headers;
```
rubensworks, line 93: Why is this needed here? Is the `httpResponse.headers` not sufficient?
Not sure, but I would think `httpResponse.input.headers` refers to the original request headers, which should not be in here, right?

packages/actor-http-native/lib/ActorHttpNative.ts
```diff
@@ -63,41 +63,56 @@ export class ActorHttpNative extends ActorHttp {
...
+      }
+
+      if (httpResponse && httpResponse.rawHeaders) {
+        for (let i = 0; i + 1 < httpResponse.rawHeaders.length; i += 2) {
+          headers.set(httpResponse.rawHeaders[i], httpResponse.rawHeaders[i + 1]);
```
rubensworks, line 98: Same question here, is `httpResponse.rawHeaders` needed if you already have all the entries from `httpResponse.headers`?

rubensworks: In any case, it's always good to add comments for non-trivialities like these.

packages/actor-http-native/test/ActorHttpNative-test.ts
```diff
@@ -3,11 +3,40 @@ import * as url from 'url';
...
 import { ActorHttp } from '@comunica/bus-http';
 import { Bus } from '@comunica/core';
-import { ActorHttpNative } from '../lib/ActorHttpNative';
+import { ActorHttpNative, processHttpResponse } from '../lib/ActorHttpNative';
 import Requester from '../lib/Requester';
 
+const http = require('http');
 const arrayifyStream = require('arrayify-stream');
 const mockSetup = require('./__mocks__/follow-redirects').mockSetup;
+const raw = [
```
rubensworks, line 12: Only one test seems to be using this field (and `options`), can we move them in the scope of this tests?

The test themselves look great!

packages/actor-http-native/lib/ActorHttpNative.ts
stephaniech97: In the specific case of the issue it was so that the headers did not contain the all the entries of the raw headers. This is why I forced them inside in this way. I'm not sure why that would ever happen

rubensworks: Which headers were not included specifically?
It's possible that this would be caused by some other bug elsewhere.

packages/actor-http-native/lib/Requester.ts
```diff
@@ -68,24 +68,25 @@ export default class Requester {
         response.pipe(decoded);
         // Copy response properties
         decoded.statusCode = response.statusCode;
-        // Wrap headers into an header object type
-        const responseHeaders: any = new Headers();
-        for (const header in response.headers) {
-          responseHeaders[header[0]] = header[1];
-        }
-        decoded.headers = responseHeaders;
+        decoded.headers = this.convertFetchHeadersToHash(response);
```
rubensworks, line 71: I would just pass `response.headers`. `convertFetchHeadersToHash` does not need to be aware of the full `response` object. Also simplifies testing.

stephaniech97: Zou je dan ook willen dat ik daar apart een test voor toevoeg?

rubensworks: Sure, but nothing to extensive though. Just an empty and non-empty conversion check should be fine.

stephaniech97: A test has been added. But as seen as in the new code there are now 2 elements passed from response. So now the entire response is again added. Unsure if thats still unclean. But there are passing issues if response.input.headers are passed into the function that's another reason as to why i'm passing the whole object now.

packages/actor-http-native/lib/Requester.ts
```diff
@@ -13,7 +13,7 @@ const { https } = require('follow-redirects');
 // Decode encoded streams with these decoders
 const DECODERS = { gzip: zlib.createGunzip, deflate: zlib.createInflate };
 
-export default class Requester {
+export class Requester {
```
rubensworks, line 16: This change will also have to be applied to the browser version: https://github.com/comunica/comunica/blob/master/packages/actor-http-native/lib/Requester-browser.ts#L14

These two always need to stay in sync.

rubensworks: Actually, instead of modifying this, why not just make `convertFetchHeadersToHash` a member method to this class?

packages/actor-http-native/lib/Requester.ts
rubensworks: Hrmm, this feels wrong. It's still inclear to me why `response.input.headers` is needed.

packages/actor-http-native/lib/Requester.ts
```diff
@@ -38,8 +38,8 @@ export default class Requester {
     // Unpacking headers object into a plain object
     const headersObject: any = {};
     if (settings.headers) {
-      for (const header of settings.headers) {
```
rubensworks: This change should not be needed.
If this is not iterable, then it means that `settings.headers` is invalid, and something is going wrong elsewhere.

packages/actor-http-native/lib/Requester.ts
```diff
@@ -57,6 +57,20 @@ export class Requester {
     return requestProxy;
   }
 
+  // Wrap headers into an header object type
+  public convertFetchHeadersToHash(response: any): any {
```
rubensworks, line 65: Avoid using `any`. Use specific types instead where possible.

packages/actor-http-native/lib/Requester.ts
```diff
@@ -37,7 +37,11 @@ export default class Requester {
 
     // Unpacking headers object into a plain object
     const headersObject: any = {};
-    if (settings.headers) {
+    if (settings.headers && settings.headers._headers) {
```
rubensworks, line 40: `_headers` is not part of the standard fetch API, so this should not be needed.
For what reason do you want this? If a test is failing, perhaps that test is wrong?

packages/actor-http-native/lib/Requester.ts
```diff
@@ -57,6 +57,20 @@ export class Requester {
     return requestProxy;
   }
 
+  // Wrap headers into an header object type
+  public convertFetchHeadersToHash(response: any): any {
+    const responseHeaders: Headers = new Headers();
+    if (response.input && response.input.headers) {
```
rubensworks, line 67: `response.input` is not part of the public `IncomingMessage` interface in the Node request API AFAIK. So this should not be needed.

packages/actor-http-native/test/ActorHttpNative-test.ts
```diff
@@ -49,9 +49,11 @@ describe('ActorHttpNative', () => {
 
   describe('An ActorHttpNative instance', () => {
     let actor: ActorHttpNative;
+    let mock;
```
rubensworks, line 52: Unused field?

packages/actor-http-native/lib/Requester.ts
rubensworks: Also, I guess the test mock may be wrong: https://github.com/comunica/comunica/blob/379553bfcf265a46cd91c8a69a4dccd2f0725f0d/packages/actor-http-native/test/__mocks__/follow-redirects.js#L15

packages/actor-http-native/lib/Requester.ts
FlorianFV: there is one test that fails in `yarn run integration`: DBPedia TPF & SPARQL (https://comunica.github.io/manifest-ldf-tests/sparql/sparql-manifest.ttl#simple03) with following errorstack:
https://hastebin.com/kuhufoceku.sql

rubensworks: Your hastebin is empty for me.

In any case, I assume _something_ is going wrong in another place in real requests, not corresponding to the test mocks. I would suggest debugging a bit to see why this difference exists.

FlorianFV:
@rubensworks

rubensworks:
@FlorianFV Build seems to be failing again due to the TPF & SPARQL integration query: https://travis-ci.org/github/comunica/comunica/jobs/717634850#L1303

packages/actor-http-native/lib/Requester.ts
```diff
@@ -46,6 +57,17 @@ export default class Requester {
     return requestProxy;
   }
 
+  // Wrap headers into an header object type
+  public convertFetchHeadersToHash(headers: IncomingHttpHeaders): Headers {
```
rubensworks, line 61: The method name is not really correct anymore, should be something like `convertRequestHeadersToFetchHeaders`

packages/actor-http-native/test/ActorHttpNative-test.ts
```diff
@@ -188,3 +198,29 @@ describe('Requester', () => {
...
+    });
+  });
+
+  it('works without headers', () => {
+    const requester = new Requester();
+    const req = requester.createRequest(url.parse('http://example.com/test'));
+    return new Promise(resolve => {
+      req.on('response', response => {
+        response.headers = {};
+        expect(requester.convertFetchHeadersToHash(response)).toEqual(new Headers({}));
```
rubensworks, line 221:
```suggestion
        expect(requester.convertFetchHeadersToHash(response.headers)).toEqual(new Headers({}));
```

rubensworks:
Thanks @stephaniech97 and @FlorianFV!

PR #702, opened 2020-07-31
Take value field instead of object for efficient hashfunction

### LINKED ISSUE #608: Inefficient hash function in symmetric hash join actor
#### Issue type:

- :bug: Bug

____
#### Description:

The hash function always returns the same value (I think).

I need to double check this (hence the issue), but pretty sure the hash function [here](https://github.com/comunica/comunica/blob/master/packages/actor-rdf-join-symmetrichash/lib/ActorRdfJoinSymmetricHash.ts#L24) always returns the same value (`"[Object object]"`).

FlorianFV:
Closes #608


FlorianFV:
A bit too simple according to Travis. A check if undefined and then return 'undefined'? Or is this a problem in the code that called it (wrong given variables)?

rubensworks:
Ah indeed, it can be possible that some variables are not present in the bindings. A `filter(variable => bindings.has(variable))` before the `.map` should do the trick.

Issue #287, opened 2018-10-02
SPARQL test suite compliance


rubensworks:
This issue lists all current failures from the SPARQL 1.1 test suite,
categorized by type.

## SPARQL Syntax

* [ ] syntax-oneof-03.rq (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_33)
* [ ] syntax-bindings-02a.rq with VALUES clause (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_35a)
* [ ] syntax-bindings-03a.rq with VALUES clause (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_36a)
* [ ] syntax-construct-where-02.rq (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_42)
* [ ] syn-bad-01.rq (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_43)
* [ ] syn-bad-02.rq (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_44)
* [ ] syn-bad-03.rq (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_45)
* [ ] syntax-BINDscope6.rq (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_60)
* [ ] syntax-BINDscope7.rq (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_61a)
* [ ] syntax-BINDscope8.rq (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_62a)
* [ ] syntax-SELECTscope2 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/syntax-query/manifest#test_65)

## Sub-queries

* [ ] Expression has variable that may be unbound (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/project-expression/manifest#projexp07)
* [ ] sq11 - Subquery limit per resource (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/subquery/manifest#subquery11)
[...]

PR #706, opened 2020-08-10
Property paths

FlorianFV:
#287

### DIFF
--- a/packages/actor-abstract-path/lib/ActorAbstractPath.ts
+++ b/packages/actor-abstract-path/lib/ActorAbstractPath.ts
@@ -109,42 +109,71 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
     });
   }
 
+  // Let the iterator `it` emit all bindings of size 2, with subjectStringVariable as value subjectVal
+  // and objectStringVariable as value all nodes reachable through predicate* beginning at objectVal
   public async ALPTwoVariables(subjectString: string, objectString: string, subjectVal: Term,
-    x: Term, predicate: Algebra.PropertyPathSymbol, context: ActionContext,
-    termHashes: {[id: string]: Term}, it: BufferedIterator<Bindings>, counter: any): Promise<void> {
-    const termString = termToString(subjectVal) + termToString(x);
-    if (termHashes[termString]) {
+    objectVal: Term, predicate: Algebra.PropertyPathSymbol, context: ActionContext,
+    termHashesGlobal: {[id: string]: Promise<Term[]>}, termHashesCurrentSubject: {[id: string]: boolean},
+    it: BufferedIterator<Bindings>, counter: any): Promise<void> {
+    const termString = termToString(objectVal);
+
+    // If this combination of subject and object already done, return nothing
+    if (termHashesCurrentSubject[termString]) {
       return;
     }
 
-    (<any> it)._push(Bindings({ [subjectString]: subjectVal, [objectString]: x }));
-    termHashes[termString] = x;
+    termHashesCurrentSubject[termString] = true;
+
+    (<any> it)._push(Bindings({ [subjectString]: subjectVal, [objectString]: objectVal }));
     counter.count++;
 
-    const thisVariable = this.generateVariable();
-    const vString = termToString(thisVariable);
-    const path = ActorAbstractPath.FACTORY.createPath(x, predicate, thisVariable);
-    const results = ActorQueryOperation.getSafeBindings(
-      await this.mediatorQueryOperation.mediate({ operation: path, context }),
-    );
-    results.bindingsStream.on('data', async bindings => {
-      const result = bindings.get(vString);
-      await this.ALPTwoVariables(
-        subjectString,
-        objectString,
-        subjectVal,
-        result,
-        predicate,
-        context,
-        termHashes,
-        it,
-        counter,
-      );
-    });
-    results.bindingsStream.on('end', () => {
-      if (--counter.count === 0) {
-        it.close();
+    // If every reachable node from object has already been calculated, use these for current subject too
+    if (termString in termHashesGlobal) {
+      const objects = await termHashesGlobal[termString];
+      for (const object of objects) {
+        if (!termHashesCurrentSubject[termToString(object)]) {
+          termHashesCurrentSubject[termToString(object)] = true;
+          (<any> it)._push(Bindings({ [subjectString]: subjectVal, [objectString]: object }));
+        }
       }
+    }
+
+    // Construct promise to calculate all reachable nodes from this object
+    const promise = new Promise<Term[]>(async(resolve, reject) => {
+      const objectsArray: Term[] = [];
+
+      const thisVariable = this.generateVariable();
+      const vString = termToString(thisVariable);
+      const path = ActorAbstractPath.FACTORY.createPath(objectVal, predicate, thisVariable);
+      const results = ActorQueryOperation.getSafeBindings(
+        await this.mediatorQueryOperation.mediate({ operation: path, context }),
+      );
+
+      results.bindingsStream.on('data', async bindings => {
+        const result = bindings.get(vString);
+        objectsArray.push(result);
+        await this.ALPTwoVariables(
+          subjectString,
+          objectString,
+          subjectVal,
+          result,
+          predicate,
+          context,
+          termHashesGlobal,
+          termHashesCurrentSubject,
+          it,
+          counter,
+        );
+      });
+      results.bindingsStream.on('end', () => {
+        if (--counter.count === 0) {
+          it.close();
+        }
+        resolve(objectsArray);
+      });
     });
+
+    // Set it in the termHashesGlobal when this object occurs again they can wait for this promise
+    termHashesGlobal[termString] = promise;
   }
 }
--- a/packages/actor-abstract-path/lib/ActorAbstractPath.ts
+++ b/packages/actor-abstract-path/lib/ActorAbstractPath.ts
@@ -2,11 +2,11 @@ import {
   ActorQueryOperation,
   ActorQueryOperationTypedMediated,
   IActorQueryOperationTypedMediatedArgs,
+  IActorQueryOperationOutputBindings,
 } from '@comunica/bus-query-operation';
 import { ActionContext, IActorTest } from '@comunica/core';
 import { variable } from '@rdfjs/data-model';
 import { AsyncIterator, BufferedIterator } from 'asynciterator';
-import { Map } from 'immutable';
 import { Term, Variable } from 'rdf-js';
 import { termToString } from 'rdf-string';
 import { Algebra, Factory } from 'sparqlalgebrajs';
@@ -20,6 +20,8 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
   protected static readonly FACTORY: Factory = new Factory();
 
   protected readonly predicateType: string;
+
+  public static isPathArbitraryLengthDistinctKey = 'isPathArbitraryLengthDistinct';
 
   protected constructor(args: IActorQueryOperationTypedMediatedArgs, predicateType: string) {
     super(args, 'path');
@@ -51,14 +53,21 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
   // Such connectivity matching does not introduce duplicates (it does not incorporate any count of the number
   // of ways the connection can be made) even if the repeated path itself would otherwise result in duplicates.
   // https://www.w3.org/TR/sparql11-query/#propertypaths
-  public isPathArbitraryLengthDistinct(context: ActionContext): ActionContext {
-    if (!context || !context.get('isPathArbitraryLengthDistinct')) {
-      return context ?
-        context.set('isPathArbitraryLengthDistinct', true) :
-        Map.of('isPathArbitraryLengthDistinct', true);
+  public async isPathArbitraryLengthDistinct(context: ActionContext, path: Algebra.Path):
+  Promise<{context: ActionContext; operation: IActorQueryOperationOutputBindings | undefined}> {
+    if (!context || !context.get(ActorAbstractPath.isPathArbitraryLengthDistinctKey)) {
+      context = context ?
+        context.set(ActorAbstractPath.isPathArbitraryLengthDistinctKey, true) :
+        ActionContext({ [ActorAbstractPath.isPathArbitraryLengthDistinctKey]: true });
+      return { context,
+        operation: ActorQueryOperation.getSafeBindings(await this.mediatorQueryOperation.mediate({
+          operation: ActorAbstractPath.FACTORY.createDistinct(path),
+          context,
+        })) };
     }
 
-    return context.set('isPathArbitraryLengthDistinct', false);
+    context = context.set(ActorAbstractPath.isPathArbitraryLengthDistinctKey, false);
+    return { context, operation: undefined };
   }
 
   // Based on definition in spec https://www.w3.org/TR/sparql11-query/
--- a/packages/actor-abstract-path/lib/ActorAbstractPath.ts
+++ b/packages/actor-abstract-path/lib/ActorAbstractPath.ts
@@ -4,9 +4,9 @@ import {
   IActorQueryOperationTypedMediatedArgs,
 } from '@comunica/bus-query-operation';
 import { ActionContext, IActorTest } from '@comunica/core';
-import { blankNode } from '@rdfjs/data-model';
+import { blankNode, variable } from '@rdfjs/data-model';
 import { AsyncIterator, BufferedIterator } from 'asynciterator';
-import { BlankNode, Term } from 'rdf-js';
+import { BlankNode, Term, Variable } from 'rdf-js';
 import { termToString } from 'rdf-string';
 import { Algebra, Factory } from 'sparqlalgebrajs';
 
@@ -45,6 +45,20 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
     }
 
     return blankNode(name);
+  }
+
+  // Generates a blank node that does not yet occur in the path
+  public generateVariable(path?: Algebra.Path, name?: string): Variable {
+    if (!name) {
+      return this.generateVariable(path, 'b');
+    }
+
+    // Path predicates can't contain variables/blank nodes
+    if (path && (path.subject.value === name || path.object.value === name)) {
+      return this.generateVariable(path, `${name}b`);
+    }
+
+    return variable(name);
   }
 
   // Based on definition in spec https://www.w3.org/TR/sparql11-query/
--- a/packages/actor-abstract-path/lib/ActorAbstractPath.ts
+++ b/packages/actor-abstract-path/lib/ActorAbstractPath.ts
@@ -136,6 +136,11 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
           (<any> it)._push(Bindings({ [subjectString]: subjectVal, [objectString]: object }));
         }
       }
+      if (--counter.count === 0) {
+        it.close();
+      }
+
+      return;
     }
 
     // Construct promise to calculate all reachable nodes from this object
--- a/packages/actor-abstract-path/lib/ActorAbstractPath.ts
+++ b/packages/actor-abstract-path/lib/ActorAbstractPath.ts
@@ -4,9 +4,10 @@ import {
   IActorQueryOperationTypedMediatedArgs,
 } from '@comunica/bus-query-operation';
 import { ActionContext, IActorTest } from '@comunica/core';
-import { blankNode, variable } from '@rdfjs/data-model';
+import { variable } from '@rdfjs/data-model';
 import { AsyncIterator, BufferedIterator } from 'asynciterator';
-import { BlankNode, Term, Variable } from 'rdf-js';
+import { Map } from 'immutable';
+import { Term, Variable } from 'rdf-js';
 import { termToString } from 'rdf-string';
 import { Algebra, Factory } from 'sparqlalgebrajs';
 
@@ -34,20 +35,6 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
   }
 
   // Generates a blank node that does not yet occur in the path
-  public generateBlankNode(path?: Algebra.Path, name?: string): BlankNode {
-    if (!name) {
-      return this.generateBlankNode(path, 'b');
-    }
-
-    // Path predicates can't contain variables/blank nodes
-    if (path && (path.subject.value === name || path.object.value === name)) {
-      return this.generateBlankNode(path, `${name}b`);
-    }
-
-    return blankNode(name);
-  }
-
-  // Generates a blank node that does not yet occur in the path
   public generateVariable(path?: Algebra.Path, name?: string): Variable {
     if (!name) {
       return this.generateVariable(path, 'b');
@@ -59,6 +46,19 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
     }
 
     return variable(name);
+  }
+
+  // Such connectivity matching does not introduce duplicates (it does not incorporate any count of the number
+  // of ways the connection can be made) even if the repeated path itself would otherwise result in duplicates.
+  // https://www.w3.org/TR/sparql11-query/#propertypaths
+  public isPathArbitraryLengthDistinct(context: ActionContext): ActionContext {
+    if (!context || !context.get('isPathArbitraryLengthDistinct')) {
+      return context ?
+        context.set('isPathArbitraryLengthDistinct', true) :
+        Map.of('isPathArbitraryLengthDistinct', true);
+    }
+
+    return context.set('isPathArbitraryLengthDistinct', false);
   }
 
   // Based on definition in spec https://www.w3.org/TR/sparql11-query/
--- a/packages/actor-abstract-path/lib/ActorAbstractPath.ts
+++ b/packages/actor-abstract-path/lib/ActorAbstractPath.ts
@@ -80,14 +80,14 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
 
     (<any> it)._push(x);
     termHashes[termString] = x;
+    counter.count++;
 
-    const thisBlankNode = this.generateBlankNode();
+    const thisBlankNode = this.generateVariable();
     const bString = termToString(thisBlankNode);
     const path = ActorAbstractPath.FACTORY.createPath(x, predicate, thisBlankNode);
     const results = ActorQueryOperation.getSafeBindings(
       await this.mediatorQueryOperation.mediate({ operation: path, context }),
     );
-    counter.count++;
     results.bindingsStream.on('data', async bindings => {
       const result = bindings.get(bString);
       await this.ALP(result, predicate, context, termHashes, it, counter);
--- a/packages/actor-abstract-path/lib/ActorAbstractPath.ts
+++ b/packages/actor-abstract-path/lib/ActorAbstractPath.ts
@@ -3,6 +3,7 @@ import {
   ActorQueryOperationTypedMediated,
   IActorQueryOperationTypedMediatedArgs,
   IActorQueryOperationOutputBindings,
+  Bindings,
 } from '@comunica/bus-query-operation';
 import { ActionContext, IActorTest } from '@comunica/core';
 import { variable } from '@rdfjs/data-model';
@@ -36,13 +37,13 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
     return true;
   }
 
-  // Generates a blank node that does not yet occur in the path
+  // Generates a variable that does not yet occur in the path
   public generateVariable(path?: Algebra.Path, name?: string): Variable {
     if (!name) {
       return this.generateVariable(path, 'b');
     }
 
-    // Path predicates can't contain variables/blank nodes
+    // Path predicates can't contain variables
     if (path && (path.subject.value === name || path.object.value === name)) {
       return this.generateVariable(path, `${name}b`);
     }
@@ -91,15 +92,54 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
     termHashes[termString] = x;
     counter.count++;
 
-    const thisBlankNode = this.generateVariable();
-    const bString = termToString(thisBlankNode);
-    const path = ActorAbstractPath.FACTORY.createPath(x, predicate, thisBlankNode);
+    const thisVariable = this.generateVariable();
+    const vString = termToString(thisVariable);
+    const path = ActorAbstractPath.FACTORY.createPath(x, predicate, thisVariable);
     const results = ActorQueryOperation.getSafeBindings(
       await this.mediatorQueryOperation.mediate({ operation: path, context }),
     );
     results.bindingsStream.on('data', async bindings => {
-      const result = bindings.get(bString);
+      const result = bindings.get(vString);
       await this.ALP(result, predicate, context, termHashes, it, counter);
+    });
+    results.bindingsStream.on('end', () => {
+      if (--counter.count === 0) {
+        it.close();
+      }
+    });
+  }
+
+  public async ALPTwoVariables(subjectString: string, objectString: string, subjectVal: Term,
+    x: Term, predicate: Algebra.PropertyPathSymbol, context: ActionContext,
+    termHashes: {[id: string]: Term}, it: BufferedIterator<Bindings>, counter: any): Promise<void> {
+    const termString = termToString(subjectVal) + termToString(x);
+    if (termHashes[termString]) {
+      return;
+    }
+
+    (<any> it)._push(Bindings({ [subjectString]: subjectVal, [objectString]: x }));
+    termHashes[termString] = x;
+    counter.count++;
+
+    const thisVariable = this.generateVariable();
+    const vString = termToString(thisVariable);
+    const path = ActorAbstractPath.FACTORY.createPath(x, predicate, thisVariable);
+    const results = ActorQueryOperation.getSafeBindings(
+      await this.mediatorQueryOperation.mediate({ operation: path, context }),
+    );
+    results.bindingsStream.on('data', async bindings => {
+      const result = bindings.get(vString);
+      await this.ALPTwoVariables(
+        subjectString,
+        objectString,
+        subjectVal,
+        result,
+        predicate,
+        context,
+        termHashes,
+        it,
+        counter,
+      );
     });
     results.bindingsStream.on('end', () => {
       if (--counter.count === 0) {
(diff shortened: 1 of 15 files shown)


rubensworks:
I assume this also requires a version bump of asyncjoin in our package.json's?

FlorianFV:
That's correct.

packages/actor-abstract-path/lib/ActorAbstractPath.ts
```diff
@@ -47,6 +47,20 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
     return blankNode(name);
   }
 
+  // Generates a blank node that does not yet occur in the path
+  public generateVariable(path?: Algebra.Path, name?: string): Variable {
```
rubensworks, line 51: Can the `generateBlankNode` method be removed?

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
```diff
@@ -19,17 +20,35 @@ export class ActorQueryOperationPathOneOrMore extends ActorAbstractPath {
   }
 
   public async runOperation(path: Algebra.Path, context: ActionContext): Promise<IActorQueryOperationOutputBindings> {
+    // Such connectivity matching does not introduce duplicates (it does not incorporate any count of the number
+    // of ways the connection can be made) even if the repeated path itself would otherwise result in duplicates.
+    // https://www.w3.org/TR/sparql11-query/#propertypaths
+    if (!context || !context.get('isPathArbitraryLengthDistinct')) {
+      context = context ?
+        context.set('isPathArbitraryLengthDistinct', true) :
+        Map.of('isPathArbitraryLengthDistinct', true);
```
rubensworks, line 29: There should be an `ActionContext` function somewhere you can use for this instead.

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
```diff
@@ -19,17 +20,35 @@ export class ActorQueryOperationPathOneOrMore extends ActorAbstractPath {
...
+
     const predicate = <Algebra.OneOrMorePath> path.predicate;
 
     const sVar = path.subject.termType === 'Variable';
     const oVar = path.object.termType === 'Variable';
 
     if (!sVar && oVar) {
       // Get all the results of applying this once, then do zeroOrMore for those
-      const single = ActorAbstractPath.FACTORY.createPath(path.subject, predicate.path, path.object, path.graph);
+      const single = ActorAbstractPath.FACTORY.createDistinct(
```
rubensworks, line 45: Is this distinct wrapping needed again here?

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
```diff
@@ -59,8 +78,41 @@ export class ActorQueryOperationPathOneOrMore extends ActorAbstractPath {
...
+      const subjectString = termToString(path.subject);
+      const objectString = termToString(path.object);
+
+      const subjects: Set<string> = new Set();
+
+      const bindingsStream: MultiTransformIterator<Bindings, Bindings> = new MultiTransformIterator(
+        results.bindingsStream,
+        {
+          multiTransform: (bindings: Bindings) => {
+            if (subjects.has(bindings.get(subjectString).value)) {
```
rubensworks, line 95: Use `termToString` instead of `.value`.

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
```diff
@@ -59,8 +78,41 @@ export class ActorQueryOperationPathOneOrMore extends ActorAbstractPath {
...
+              return new EmptyIterator();
+            }
+            subjects.add(bindings.get(subjectString).value);
```
rubensworks, line 98: Same as above.

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
```diff
@@ -72,22 +124,21 @@ export class ActorQueryOperationPathOneOrMore extends ActorAbstractPath {
...
-        filter: item => item.get(bString).equals(path.object),
-        transform(item, next, push) {
-          push(Bindings({ }));
-          next();
-        },
-      });
-      return { type: 'bindings', bindingsStream, variables: []};
     }
+    // If (!sVar && !oVar)
+    const blankNode = this.generateBlankNode();
```
rubensworks, line 129: Should this be a variable instead?

packages/actor-query-operation-path-seq/lib/ActorQueryOperationPathSeq.ts
```diff
@@ -44,7 +44,10 @@ export class ActorQueryOperationPathSeq extends ActorAbstractPath {
       },
     });
 
-    return { type: 'bindings', bindingsStream, variables: join.variables };
+    const variables = join.variables;
```
rubensworks, line 47: It's unclear to me what is happening here. Could you add a small comment?

packages/actor-query-operation-path-zero-or-more/test/ActorQueryOperationPathZeroOrMore-test.ts
```diff
@@ -2,12 +2,15 @@ import { ActorQueryOperation, Bindings } from '@comunica/bus-query-operation';
...
 import { namedNode, variable } from '@rdfjs/data-model';
 import { ArrayIterator } from 'asynciterator';
+import { Map } from 'immutable';
 import { termToString } from 'rdf-string';
 import { QUAD_TERM_NAMES } from 'rdf-terms';
 import { Algebra, Factory } from 'sparqlalgebrajs';
 import { ActorQueryOperationPathZeroOrMore } from '../lib/ActorQueryOperationPathZeroOrMore';
 const arrayifyStream = require('arrayify-stream');
 
+jest.setTimeout(30000);
```
rubensworks, line 12: Unneeded?

packages/actor-query-operation-path-zero-or-one/lib/ActorQueryOperationPathZeroOrOne.ts
```diff
@@ -36,6 +37,21 @@ export class ActorQueryOperationPathZeroOrOne extends ActorAbstractPath {
       throw new Error('ZeroOrOne path expressions with 2 variables not supported yet');
     }
 
+    // Such connectivity matching does not introduce duplicates (it does not incorporate any count of the number
+    // of ways the connection can be made) even if the repeated path itself would otherwise result in duplicates.
+    // https://www.w3.org/TR/sparql11-query/#propertypaths
+    if (!context || !context.get('isPathArbitraryLengthDistinct')) {
```
rubensworks, line 43: Could you abstract this block into the abstract path actor? Because it seems so occur more than once. (Copy-pasting is bad for maintainability)

rubensworks:
@joachimvh Could you also have a look at this one? You probably know this codebase a lot better than me.

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
FlorianFV: It's not necessary but improves performance, I think. If it is not there and the path without the + has multiple times the same solution as endpoint, there will be started a search with ALP for every time the same solution is encountered.

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
```diff
@@ -59,8 +71,41 @@ export class ActorQueryOperationPathOneOrMore extends ActorAbstractPath {
...
+        {
+          multiTransform: (bindings: Bindings) => {
+            if (subjects.has(termToString(bindings.get(subjectString)))) {
+              return new EmptyIterator();
+            }
+            subjects.add(termToString(bindings.get(subjectString)));
+            const val = bindings.get(subjectString);
+            return new TransformIterator<Bindings>(
+              async() => ActorQueryOperation.getSafeBindings(await this.mediatorQueryOperation.mediate({
+                context, operation: ActorAbstractPath.FACTORY.createPath(val, predicate, path.object, path.graph),
```
joachimvh, line 95: Seems correct, but I'm wondering if you could use a shared subjects map for all the separate streams as is done [above](https://github.com/comunica/comunica/blob/master/packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts#L35-L58). That should cut out all duplicate work that would currently be done.

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
rubensworks: I may be missing something, but due to `isPathArbitraryLengthDistinct` in the context, isn't it already guaranteed that this operation will be wrapped in dinstinct? So AFAICS , this will cause two dinstincts on top of each other.

packages/actor-abstract-path/lib/ActorAbstractPath.ts
```diff
@@ -34,17 +35,30 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
...
+  }
+
+  // Such connectivity matching does not introduce duplicates (it does not incorporate any count of the number
+  // of ways the connection can be made) even if the repeated path itself would otherwise result in duplicates.
+  // https://www.w3.org/TR/sparql11-query/#propertypaths
+  public isPathArbitraryLengthDistinct(context: ActionContext): ActionContext {
+    if (!context || !context.get('isPathArbitraryLengthDistinct')) {
+      return context ?
+        context.set('isPathArbitraryLengthDistinct', true) :
+        Map.of('isPathArbitraryLengthDistinct', true);
```
rubensworks, line 58: Could you replace `Map.of` with `ActionContext`? We don't want the immutable dependency to leak too much.

packages/actor-abstract-path/lib/ActorAbstractPath.ts
```diff
@@ -34,17 +35,30 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
...
 
-    return blankNode(name);
+    return variable(name);
+  }
+
+  // Such connectivity matching does not introduce duplicates (it does not incorporate any count of the number
+  // of ways the connection can be made) even if the repeated path itself would otherwise result in duplicates.
+  // https://www.w3.org/TR/sparql11-query/#propertypaths
+  public isPathArbitraryLengthDistinct(context: ActionContext): ActionContext {
+    if (!context || !context.get('isPathArbitraryLengthDistinct')) {
```
rubensworks, line 55: Could you put `'isPathArbitraryLengthDistinct'` in a static variable on this class, and always refer to that one? This will make it impossible to have typo's.

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
```diff
@@ -19,17 +20,28 @@ export class ActorQueryOperationPathOneOrMore extends ActorAbstractPath {
   }
 
   public async runOperation(path: Algebra.Path, context: ActionContext): Promise<IActorQueryOperationOutputBindings> {
+    context = this.isPathArbitraryLengthDistinct(context);
+    if (context.get('isPathArbitraryLengthDistinct')) {
```
rubensworks, line 24: Also refer to that static field here.

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
```diff
@@ -72,22 +117,21 @@ export class ActorQueryOperationPathOneOrMore extends ActorAbstractPath {
...
+      operation: ActorAbstractPath.FACTORY.createPath(path.subject, predicate, blankNode, path.graph),
+    }));
+    const bindingsStream = results.bindingsStream.transform<Bindings>({
+      filter: item => item.get(bString).equals(path.object),
+      transform(item, next, push) {
+        push(Bindings({ }));
+        next();
+      },
+    });
+    return { type: 'bindings', bindingsStream, variables: []};
```
rubensworks, line 135: `variables` is empty, I suspect this is wrong. But it looks like this used to be like this before as well. Was this intentional @joachimvh?

packages/actor-query-operation-path-one-or-more/test/ActorQueryOperationPathOneOrMore-test.ts
```diff
@@ -2,6 +2,7 @@ import { ActorQueryOperation, Bindings } from '@comunica/bus-query-operation';
 import { Bus } from '@comunica/core';
 import { namedNode, variable } from '@rdfjs/data-model';
 import { ArrayIterator } from 'asynciterator';
+import { Map } from 'immutable';
```
rubensworks, line 5: You can use `ActionContext` instead.

packages/actor-query-operation-path-zero-or-more/lib/ActorQueryOperationPathZeroOrMore.ts
```diff
@@ -16,14 +19,59 @@ export class ActorQueryOperationPathZeroOrMore extends ActorAbstractPath {
   }
 
   public async runOperation(path: Algebra.Path, context: ActionContext): Promise<IActorQueryOperationOutputBindings> {
+    context = this.isPathArbitraryLengthDistinct(context);
+    if (context.get('isPathArbitraryLengthDistinct')) {
```
rubensworks, line 23: Also refer to the static field here.

But now that I've seen this pattern multiple times already, I would suggest abstracting this a bit more.
Instead of having `isPathArbitraryLengthDistinct`, I'd create something like `getDistinctOperation`, which will return a wrapped distinct operation if needed, or undefined if not needed.
Every time you call `getDistinctOperation`, you'd simply have to do an if-check, and return directly if the result is not undefined.

packages/actor-query-operation-path-zero-or-more/test/ActorQueryOperationPathZeroOrMore-test.ts
```diff
@@ -2,6 +2,7 @@ import { ActorQueryOperation, Bindings } from '@comunica/bus-query-operation';
 import { Bus } from '@comunica/core';
 import { namedNode, variable } from '@rdfjs/data-model';
 import { ArrayIterator } from 'asynciterator';
+import { Map } from 'immutable';
```
rubensworks, line 5: ActionContext

packages/actor-query-operation-path-zero-or-one/lib/ActorQueryOperationPathZeroOrOne.ts
```diff
@@ -36,6 +37,16 @@ export class ActorQueryOperationPathZeroOrOne extends ActorAbstractPath {
       throw new Error('ZeroOrOne path expressions with 2 variables not supported yet');
     }
 
+    context = this.isPathArbitraryLengthDistinct(context);
```
rubensworks, line 40: Abstract

packages/actor-query-operation-path-zero-or-more/lib/ActorQueryOperationPathZeroOrMore.ts
```diff
@@ -16,14 +19,59 @@ export class ActorQueryOperationPathZeroOrMore extends ActorAbstractPath {
...
 
     if (sVar && oVar) {
-      throw new Error('ZeroOrMore path expressions with 2 variables not supported yet');
-    } else if (!sVar && !oVar) {
+      // Get all the results of subjects, then fill in first variable for those
+      const predVar = this.generateVariable(path);
+      const single = ActorAbstractPath.FACTORY.createUnion(
+        ActorAbstractPath.FACTORY.createPattern(path.subject, predVar, path.object, path.graph),
+        ActorAbstractPath.FACTORY.createPattern(path.object, predVar, path.subject, path.graph),
+      );
```
joachimvh, line 41: Why is this union here? Am I wrong in thinking that this entire block should be identical to the one in the OneOrMore operator?

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
FlorianFV: It's a path where both object and subject are not variables (see line 121) something like :a :p+ :b, so no variables there. Not sure what this means semantically

joachimvh: This is actually correct since this corresponds to the cases were both ends are not variables, so something like `:s :p+ :o.`. This will result in a potential empty binding.

rubensworks: Ah, I see, thanks for clarifying!

joachimvh: So semantically it is pretty much a check to see if this pattern exists at the endpoint.

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
FlorianFV: So, what we do is: we apply the path once first, and then we start a search with all the found endpoints with ALP. From all those ALP's together, we have first wrapped this in a distinct, so no solution will be returned twice by the oneOrMore operator without the distinct you are talking about, that's true. But, we also want to start the ALP search with only different values, and that's where this distinct is for.

So this distinct makes sure we don't start an ALP search with twice the same path, the distinct from the AbstractPath (with the context etc) makes sure we dont return the same solution after our whole computation (it filters for example 2 solutions from ALP, obtained with a different startnode).

rubensworks: Ah ok, that's clear to me. So for performance-reasons, you do something like `distinct( stuff( distinct(pp) ) )`.

packages/actor-query-operation-path-zero-or-more/lib/ActorQueryOperationPathZeroOrMore.ts
FlorianFV: In the zeroOrMore, every namedNode in the dataset exists in a binding to itself, so this is how I fetch all entities in the dataset (all subjects + all objects) because they all apply to the query using the path 0 times, this was not necessary in the oneOrMore.

joachimvh: But this block corresponds to the case where both subject and object are a variable, so both queries in the union would generate identical results. Those specific cases would be caught when you call this function recursively below with an instantiated variable.

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
FlorianFV: What work is duplicate in this code then? I also think it's not that easy, because above it can be passed to ALP to emit right there directly, but here it would require a change in the mediate function?

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
FlorianFV: That's it!

packages/actor-query-operation-path-zero-or-more/lib/ActorQueryOperationPathZeroOrMore.ts
FlorianFV: I'm not sure that will work. A namedNode that only occurs as object in the dataset, not even with the requested predicate in our zeroOrMore should also be included. A possibility is without a union, but then extracting the returned bindings in two seperate results (the subject and the object).

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
joachimvh: Currently, both `a` and `b` would be in the subjects Set. This would generate 2 TransformIterators, one starting for each subject. In the first case this would call the ALP function with `a`, which would then recursively call itself again with `b`, since the ALP function starts with an empty map by default. The second stream would call the ALP function once with `b`. If they shared the same term `termHashes` either the second call to the ALP in the first stream, or the first call in the second stream wouldn't happen (depending on timing).

packages/actor-query-operation-path-zero-or-more/lib/ActorQueryOperationPathZeroOrMore.ts
FlorianFV: The false set on the context will then again have to be out of this abstracted function.

packages/actor-query-operation-path-zero-or-more/lib/ActorQueryOperationPathZeroOrMore.ts
joachimvh: Ah true, both subject and object would need to be checked for this specific instance (potentially, would have to think a bit more about the details to be sure). But this is not the way to do it from what I can see. To convert to SPARQL what is happening here, the input pattern is `?s :p+ ?o.`. In the lines above you convert this to `{ ?s :p+ ?o. } UNION { ?o :p+ ?s. }`. Since variable names don't matter, this is a union of 2 identical streams.

packages/actor-query-operation-path-zero-or-more/lib/ActorQueryOperationPathZeroOrMore.ts
rubensworks: Ah right, the context. Maybe you can include it in the return value of the abstracted function instead?

So basically you'd return `{ context, operation }`, where `operation` could be undefined.

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
FlorianFV: Aha, allright, I see. Is it possible to call ALP with a variable as predicate?

joachimvh: I'm assuming you mean subject instead of predicate? But no, you need a constant term. So your first call to generate the subjects is still necessary here. But in the streams generated in the multitransformiterator you can call ALP instead of the mediator and use the same map for al those calls.

packages/actor-query-operation-path-zero-or-more/lib/ActorQueryOperationPathZeroOrMore.ts
FlorianFV: Actually I convert it to `{ ?s ?p ?o} UNION { ?o ?p ?s }`. The question is actually: "How to get put a query that returns a binding of size 1 with every namedNode in the dataset?". This was where I came up with, but it's indeed not very good.

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
FlorianFV: Yes, I figured. That's why I deleted the post :)

packages/actor-query-operation-path-zero-or-more/lib/ActorQueryOperationPathZeroOrMore.ts
FlorianFV: Do you have an alternative to get al those bindings of size 1?

joachimvh: Ah yes, I remember now why this part wasn't implemented, because requesting a stream of all triples to an endpoint is not really ideal. You could do it with a single `?s ?p ?o` pattern and then apply a transform iterator to it that checks both subject and object, which would "only" have to create 1 stream of all triples. Another optimization could be to first return all the results for if the query was `?s :p+ ?o.` (instead of `*`) and only append those single bindings afterwards since those are going to be way more inefficient to calculate.

packages/actor-query-operation-path-one-or-more/lib/ActorQueryOperationPathOneOrMore.ts
FlorianFV: I'm struggling implementing it. Directly calling ALP with the subjects is not possible, because this would generate to many solutions. So I called it with the objects, and then we transform `it` with as subject the subject of the single operation and as object the result of the ALP. But that also does not work with a centralized it, because then it's not clear which subject to add of which single operation.

joachimvh: Could be that the ALP function is not the correct one to call for this and perhaps some other function would be needed (it's been a while so I don't recall exactly how ALP works). It just felt to me like this is something that should be able to be optimized since you're looking for a bunch of triples with the same predicate every time. But I'm not the one writing it right now so I could be wrong.

packages/actor-query-operation-path-zero-or-more/lib/ActorQueryOperationPathZeroOrMore.ts
```diff
@@ -34,36 +34,64 @@ export class ActorQueryOperationPathZeroOrMore extends ActorAbstractPath {
...
-      );
+      const single = ActorAbstractPath.FACTORY.createPattern(path.subject, predVar, path.object, path.graph);
       const results = ActorQueryOperation.getSafeBindings(
         await this.mediatorQueryOperation.mediate({ context, operation: single }),
       );
       const subjectString = termToString(path.subject);
       const objectString = termToString(path.object);
 
-      const subjects: Set<string> = new Set();
+      const entities: Set<string|undefined> = new Set();
```
rubensworks, line 44: Why do we want undefined in our set?

packages/actor-query-operation-path-zero-or-more/lib/ActorQueryOperationPathZeroOrMore.ts
```diff
@@ -34,36 +34,64 @@ export class ActorQueryOperationPathZeroOrMore extends ActorAbstractPath {
     if (sVar && oVar) {
       // Get all the results of subjects, then fill in first variable for those
       const predVar = this.generateVariable(path);
-      const single = ActorAbstractPath.FACTORY.createUnion(
-        ActorAbstractPath.FACTORY.createPattern(path.subject, predVar, path.object, path.graph),
-        ActorAbstractPath.FACTORY.createPattern(path.object, predVar, path.subject, path.graph),
-      );
+      const single = ActorAbstractPath.FACTORY.createPattern(path.subject, predVar, path.object, path.graph);
```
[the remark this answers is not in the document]
rubensworks: Not sure what you mean exactly...
I suspect that this implementation here will work, but is not optimal, so this solution should then be good for now.
If so, could you open an issue that explains the problem, and what we could do to solve it?

packages/actor-abstract-path/lib/ActorAbstractPath.ts
```diff
@@ -107,4 +108,43 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
...
+    const path = ActorAbstractPath.FACTORY.createPath(x, predicate, thisVariable);
+    const results = ActorQueryOperation.getSafeBindings(
+      await this.mediatorQueryOperation.mediate({ operation: path, context }),
+    );
+    results.bindingsStream.on('data', async bindings => {
+      const result = bindings.get(vString);
+      await this.ALPTwoVariables(
+        subjectString,
+        objectString,
+        subjectVal,
```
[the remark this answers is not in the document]
FlorianFV: No, it's still subjectVal. Let's say the function started with as subjectVal :a. The query is ?s :p+ ?o, and you found after one step for example :b, you return a binding with ?s :a and ?o :b. Then you start looking one step further (starting with result (as 4th parameter)). If you get then :c, you want to return ?s :a and ?o :c, so we still need the :a, which is the subjectVal we got in the first place.

packages/actor-query-operation-path-zero-or-more/lib/ActorQueryOperationPathZeroOrMore.ts
FlorianFV: It's because that's is the returntype of termToString. I can cast it before I put it in the set, if that's better?

packages/actor-query-operation-path-zero-or-more/lib/ActorQueryOperationPathZeroOrMore.ts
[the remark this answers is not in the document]
joachimvh: More in-depth description in #709

packages/actor-abstract-path/lib/ActorAbstractPath.ts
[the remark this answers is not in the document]
FlorianFV: Simpler said: we want a reference to the ?s this all started with, so this does not need to shift along the way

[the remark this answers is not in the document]
joachimvh: Ah yes, I misread and thought that one was also used for creating the followup path. But is the termString generated correct then? If your dataset is `:a :p :b. :b :p :c.` and you start with the `:a` your termHashes are going to be `:a:b` and `:a:c`. If the call to this function is then done with `:b` it is still going to run since it doesn't see the termhash `:b:c`, right?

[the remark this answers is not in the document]
joachimvh: Note that I also realize the binding with subjectVal is necessary so I'm not immediately sure what the solution would be if I'm right.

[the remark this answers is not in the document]
FlorianFV: It is gonna run, that's true. But indeed for another subject value.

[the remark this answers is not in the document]
joachimvh: So at this point, does this function actually work differently from the original ALP function besides adding the bindings of the subject inside of the functions instead of outside of the function as was done previously?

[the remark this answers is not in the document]
FlorianFV: Could this be a solution:
The termstring in ALPTwoVariables would be the object. So in your case first the b. We set it the termHashes on []. And then we call normal ALP from b and we listen to dataevents and then add it to that []. When the second b comes in ALPTwoVariables, we check if termstring exists and if so push our new subject with all the objects in the list.

packages/actor-query-operation-path-zero-or-more/lib/ActorQueryOperationPathZeroOrMore.ts
rubensworks: The return type should always be `string` for any given defined term: https://github.com/rubensworks/rdf-string.js/blob/master/lib/TermUtil.ts#L22

packages/actor-abstract-path/lib/ActorAbstractPath.ts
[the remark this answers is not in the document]
FlorianFV: There is only one problem with this I think. If the end event has not yet occured on the ALP run on b and a second b comes in as object in ALPTwoVariables, the list may not be complete...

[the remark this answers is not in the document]
FlorianFV: > So at this point, does this function actually work differently from the original ALP function besides adding the bindings of the subject inside of the functions instead of outside of the function as was done previously?

I does. When :a :b en :z :b are both solutions, this is now possible because of the changed termhashes, the :z :b would not be a solution before, because the :b would be in the termhashes and it wouldn't be emitted for the z. But this can be fixed by not sharing the termshashes anymore I see now.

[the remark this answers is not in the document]
joachimvh: Yes that's what I meant, this solution is the same as the previous solution (where the termHashes weren't shared). But I do think there is potential in this solution somewhere to change that.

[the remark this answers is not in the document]
joachimvh: Not again btw that I could totally be wrong and there actually isn't a feasible way to do this more efficient.

rubensworks:
@FlorianFV Oh, there's a merge conflict in yarn.lock. Rebasing on master, and re-installing should fix it.

packages/actor-abstract-path/lib/ActorAbstractPath.ts
```diff
@@ -109,42 +109,71 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
...
+      const objects = await termHashesGlobal[termString];
+      for (const object of objects) {
+        if (!termHashesCurrentSubject[termToString(object)]) {
+          termHashesCurrentSubject[termToString(object)] = true;
+          (<any> it)._push(Bindings({ [subjectString]: subjectVal, [objectString]: object }));
+        }
       }
+    }
+
+    // Construct promise to calculate all reachable nodes from this object
```
joachimvh, line 141: If I understand correctly, I think everything below the `if` statement should be in an `else` block? (Or the `if` statement should return out of the function). Otherwise I think you're going to keep overwriting the promises in `termHashesGlobal`. But besides that this seems to look good.

FlorianFV: You're right, there's missing a return;

packages/actor-abstract-path/lib/ActorAbstractPath.ts
```diff
@@ -84,4 +108,77 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
...
+      results.bindingsStream.on('end', () => {
+        if (--counter.count === 0) {
+          it.close();
+        }
+        resolve(objectsArray);
+      });
+    });
+
+    // Set it in the termHashesGlobal when this object occurs again they can wait for this promise
+    termHashesGlobal[termString] = promise;
```
joachimvh, line 182: Just realized this, but should there not be a `return promise;` below this line? Because currently the promise gets generated, but if the `if` statement above is not entered (which can happen based on the data structure), the promise above could potentially never be explicitly executed.

FlorianFV: This is what I found on https://medium.com/@adityaprakashjoshi1/promises-in-javascript-cf81f8e56f80 : The executorfunction is executed immediately by the Promise implementation, passing resolve and reject functions (the executor is called before the Promise constructor even returns the created object).
Return is actually not possible because the return type of the function is Promise<void>.

rubensworks: Indeed, promises are started immediately upon creation.

Issue #613, opened 2020-03-04
support auth HTTP requests in default Comunica


mathib:
#### Issue type:

- :heavy_plus_sign: Feature request

____
#### Description:
It would be nice if the default Comunica query engine also allowed to query Linked Data Fragments that need authentication, e.g. a secured triplestore needing basic auth.

According to question https://github.com/comunica/comunica/issues/612, it should be doable to add a HTTP actor that can handle authenticated HTTP requests, e.g. by starting from [`actor-http-solid-auth-fetch`](https://github.com/comunica/actor-http-solid-auth-fetch).

rubensworks:
For reference, this can be used until we add a concrete package for this: https://github.com/comunica/examples/tree/master/packages/configure-sparql-http-solid

jaw111:
In case this only concerns requests initiated from a client browser, it may be sufficient to modify `Requester-browser.ts` in `actor-http-native` to set `withCredentials` to `true`. That would allow the browser to send the authentication and cookies for the target authority on those cross-origin requests.

If there are concerns that `withCredentials` cannot always be set to `true`, then it could be configurable such that, for example, a user of the jQuery widget is able to choose if credentials are sent by checking a box in the settings.

rubensworks:
Will be added in the next release.
Documentation: https://comunica.dev/docs/query/advanced/basic_auth/

PR #708, opened 2020-08-12
send http requests with credentials by default

jaw111:
modify actor-http-native to send requests initiated from a browser with credentials by default

see also: https://github.com/comunica/comunica/issues/613


rubensworks:
@FlorianFV Did you also have a solution for #613 in the works? If so, is this compatible with this PR?

rubensworks:
Thanks for the PR @jaw111!
I've added some tests, made it work for our fetch-based actor as well, and allowed it to be configured via the context: 2492fe93b8afa30a866b023958f5e0847c4fbc94

Will publish a new release in a bit. (1.15.0)

See here how it can be enabled: https://comunica.dev/docs/query/advanced/context/#11--include-credentials-in-http-requests

PR #710, opened 2020-08-13
Remove indexOf with ...new Set

FlorianFV:


FlorianFV:
Ah right, I searched for uniq, but I also used it for union..

PR #716, opened 2020-08-18
Fix named graph tests pp

FlorianFV:


rubensworks:
As far as I understand this code, this looks like a good way forward to make graphs work.

@joachimvh do you think this is the right approach? Or do you think some things should be done differently?

packages/actor-abstract-path/lib/ActorAbstractPath.ts
```diff
@@ -73,15 +73,52 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
...
-    await this.ALP(x, predicate, context, {}, it, { count: 0 });
+    await this.ALP(objectVal, predicate, graph, context, {}, it, { count: 0 });
 
-    return it;
+    return it.transform<Bindings>({
+      transform(item, next, push) {
+        push(Bindings({ [termToString(object)]: item }));
+        next();
+      },
+    });
```
joachimvh, line 118: Would not add this (and the extra function parameters for this) to the ALP function and create an external function to do this part since this is changing what the ALP function does compared to the [spec](https://www.w3.org/TR/sparql11-query/#PropertyPathPatterns) (or rename the function I guess).

packages/actor-abstract-path/lib/ActorAbstractPath.ts
```diff
@@ -131,10 +168,19 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
     if (termString in termHashesGlobal) {
       const objects = await termHashesGlobal[termString];
       for (const object of objects) {
-        if (!termHashesCurrentSubject[termToString(object)]) {
-          termHashesCurrentSubject[termToString(object)] = true;
-          (<any> it)._push(Bindings({ [subjectString]: subjectVal, [objectString]: object }));
-        }
+        await this.ALPTwoVariables(
```
joachimvh, line 171: This seems like a drastic change compared to what was here originally. Was there something wrong previously?

joachimvh (changes requested):

packages/actor-abstract-path/lib/ActorAbstractPath.ts
FlorianFV: Yes, there was. We only added the results of the mediate from one step to the array, and then we just pushed those. But we have to look further and add their neighbours and so on. But still will there be no work done twice, because the neighbour of the neighbour will already have been calculated as well, so they will have to check that array as well, but not mediate for the same subject more then once.

FlorianFV: I found this mistake: 
:a : p : c
:c : p : d
:d : p : e
: b : p : c
with a query as ?x :p* ?y : 
:b :c en :b :d where found (as c is common so one step further d is added as waiting on the array in the termhashes)
:b :e was not found, because it is 2 steps from the common c.

packages/actor-abstract-path/lib/ActorAbstractPath.ts
FlorianFV: Suggestion for the new name of the function?

FlorianFV: And then maybe rename ALPeval to thatNamEval.

joachimvh: No opinion, it's just that if a function is called ALP/ALPeval it should (sort of) do what the corresponding function in the spec does. Giving it a readable name that sort of explains what it does might be better anyway.

joachimvh (approved):

joachimvh (approved):
At this point I'm probably too far removed from the code to still see if there are any inherent problems 😅 , but this seems fine.

rubensworks:
Full property path compliance thanks to @FlorianFV and @joachimvh 🎉

Issue #552, opened 2019-10-18
Implement different join algorithms


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request

____
#### Description:

While we already have nested loop join, hash join and symmetric hash join, other are still possible:

* [Leapfrog Triejoin (multiway)](http://aidanhogan.com/docs/SPARQL_worst_case_optimal.pdf)
* ...

Related to #35 and #548.

PR #720, opened 2020-08-22
Feature/join undefs

### LINKED ISSUE #711: Join actors do not consider undefined values in bindings
#### Issue type:

- :bug: Bug

____
#### Description:

In some situations, we can have bindings with undefined values in some cases. This can occur with things like optionals and values (see query below).
In these cases, our join actors (symmetric hash is our default) may not be able to find matches, since the undefined binding values should be seen as wildcards (matching everything).

Therefore, we probably have to drastically rethink the way we perform joins, so that we can handle these undefined value cases.
One possible solution would be to do something similar like our minus actor, and make use of an undefined-binding-value-aware [`BindingsIndex`](https://github.com/comunica/comunica/blob/master/packages/actor-query-operation-minus/lib/BindingsIndex.ts).

Relevant for SPARQL 1.1 spec tests values4, values5, values7, values8: #287

Example query where this occurs:
```
PREFIX : <http://example.org/> 

SELECT ?s ?o1 ?o2
{
  ?s ?p1 ?o1 .
  ?s ?p2 ?o2 .
} VALUES (?o1 ?o2) {
 ("Alan" UNDEF)
}
```

Dataset:
```
@prefix : <http://example.org/> .
@prefix foaf: <http://xmlns.com/foaf/0.1/> .

:a foaf:name "Alan" .
:a foaf:mbox "alan@example.org" .
:b foaf:name "Bob" .
:b foaf:mbox "bob@example.org" .
:a foaf:knows :b .
```

Expected output (comunica finds no results):
```
[
  {
    "?s": {
      "value": "http://example.org/a"
    },
    "?o1": {
      "value": "Alan",
      "datatype": {
        "value": "http://www.w3.org/2001/XMLSchema#string"
      },
      "language": ""
    },
    "?o2": {
      "value": "Alan",
      "datatype": {
        "value": "http://www.w3.org/2001/XMLSchema#string"
      },
      "language": ""
    }
  },
  {
    "?s": {
      "value": "http://example.org/a"
    },
    "?o1": {
      "value": "Alan",
      "datatype": {
        "value": "http://www.w3.org/2001/XMLSchema#string"
      },
      "language": ""
    },
    "?o2": {
      "value": "alan@example.org",
      "datatype": {
        "value": "http://www.w3.org/2001/XMLSchema#string"
      },
      "language": ""
    }
  },
  {
    "?s": {
      "value": "http://example.org/a"
    },
    "?o1": {
      "value": "Alan",
      "datatype": {
        "value": "http://www.w3.org/2001/XMLSchema#string"
      },
      "language": ""
    },
    "?o2": {
      "value": "http://example.org/b"
    }
  }
]
```

____
#### Environment:

Master 1.15.0

rubensworks:
This PR does two things:

1. adds `canContainUndefs` flag to bindings results. (this commit is huge, no need to check this one in detail)
2. enables nested loop join actor to handle bindingsStreams with undefs.

Closes #711.

This also fixes the last remaining failing spec tests (#287).


joachimvh (approved):
Looks fine. I don't think it completely solves #711 though (or maybe a new issue is needed), since a nested loop join is very inefficient and this can quite probably be optimized.

rubensworks:
It fixes #711 in the sense that it at least works now :-)
Performance can be an issue indeed, let's track that in #552.

PR #737, opened 2020-10-01
resolve eager accept header boundary bug, fixes #534

### LINKED ISSUE #534: Wildcard accept header appending sometimes to eager
#### Issue type:

- :bug: Bug

____
#### Description:

The code in https://github.com/comunica/comunica/commit/53c7e7519f6125054ac5e80c79ca4b038f067a18 would cut off the last part of an accept header that perfectly fits into 127 characters, just to make room for the `*/*;q=0.1` it does not need.

Minor thing related to this:

might want to first convert toFixed before comparing to '1.00' because of rounding; and the regex might have `+` instead of `*`.

tbaccaer:
This pull request is supposed to fix the bug mentioned in issue 534.

### DIFF
--- a/packages/actor-rdf-dereference-http-parse/lib/ActorRdfDereferenceHttpParseBase.ts
+++ b/packages/actor-rdf-dereference-http-parse/lib/ActorRdfDereferenceHttpParseBase.ts
@@ -135,20 +135,25 @@ export abstract class ActorRdfDereferenceHttpParseBase extends ActorRdfDereferen
   }
 
   public mediaTypesToAcceptString(mediaTypes: { [id: string]: number }, maxLength: number): string {
-    // Ensure a ',*/*;q=0.1' suffix
-    maxLength -= 10;
-
+    const wildcard = '*/*;q=0.1';
     const parts: string[] = [];
     const sortedMediaTypes = Object.keys(mediaTypes)
       .map(mediaType => ({ mediaType, priority: mediaTypes[mediaType] }))
       .sort((left, right) => right.priority - left.priority);
-    let partsLength = 0;
+    // Take into account the ',' characters joining each type
+    const separatorLength = sortedMediaTypes.length - 1;
+    let partsLength = separatorLength;
     for (const entry of sortedMediaTypes) {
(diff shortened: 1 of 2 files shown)


rubensworks (approved):
Looks good to me, thanks!

@RubenVerborgh since you first noticed this issue, maybe you want to have a look at it as well?

RubenVerborgh (approved):
Looks good to me!

packages/actor-rdf-dereference-http-parse/lib/ActorRdfDereferenceHttpParseBase.ts
```diff
@@ -135,20 +135,25 @@ export abstract class ActorRdfDereferenceHttpParseBase extends ActorRdfDereferen
...
+    // Take into account the ',' characters joining each type
+    const separatorLength = sortedMediaTypes.length - 1;
+    let partsLength = separatorLength;
     for (const entry of sortedMediaTypes) {
       const part = entry.mediaType + (entry.priority !== 1 ?
         `;q=${entry.priority.toFixed(3).replace(/0*$/u, '')}` :
         '');
       if (partsLength + part.length > maxLength) {
-        parts.push('*/*;q=0.1');
+        while (partsLength + wildcard.length > maxLength) {
```
RubenVerborgh, line 151: You could change this into a `do … while` since we know the condition will be true the first time.

tbaccaer: The condition is not necessarily true the first time. The while-condition covers the situation where there is enough space left to concatenate the wildcard, but not enough space left to concatenate a media type. Doing so anyway would result in an eager removal of a type, which would cause the original bug again.

 This scenario is covered by the first test, by the way:
Screenshot 2020-10-02 at 13 34 10

RubenVerborgh: My bad, thought the `if` and `while` had the same conditions. Good testing!

PR #738, opened 2020-10-02
Add Microdata parser

rubensworks:
I was recently made aware of the fact that [Microdata is the most used semantic format on Web pages](http://webdatacommons.org/structureddata/2019-12/stats/stats.html), so I've implemented [a parser](https://github.com/rubensworks/microdata-rdf-streaming-parser.js) for it, and plugged it in here.

### DIFF
--- a/packages/actor-rdf-parse-html-microdata/lib/ActorRdfParseHtmlMicrodata.ts
+++ b/packages/actor-rdf-parse-html-microdata/lib/ActorRdfParseHtmlMicrodata.ts
@@ -18,8 +18,7 @@ export class ActorRdfParseHtmlMicrodata extends ActorRdfParseHtml {
 
   public async run(action: IActionRdfParseHtml): Promise<IActorRdfParseHtmlOutput> {
     const mediaType = action.headers ? action.headers.get('content-type') : null;
-    // eslint-disable-next-line no-implicit-coercion
-    const xmlMode = !!mediaType && mediaType.includes('xml');
+    const xmlMode = mediaType?.includes('xml');
 
     const htmlParseListener = new MicrodataRdfParser({ baseIRI: action.baseIRI, xmlMode });
     htmlParseListener.on('error', action.error);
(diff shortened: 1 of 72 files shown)


packages/actor-rdf-parse-html-microdata/lib/ActorRdfParseHtmlMicrodata.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  public async test(action: IActionRdfParseHtml): Promise<IActorTest> {
+    return true;
+  }
+
+  public async run(action: IActionRdfParseHtml): Promise<IActorRdfParseHtmlOutput> {
+    const mediaType = action.headers ? action.headers.get('content-type') : null;
+    // eslint-disable-next-line no-implicit-coercion
+    const xmlMode = !!mediaType && mediaType.includes('xml');
```
joachimvh, line 22: If you don't want to disable the rule you can do `Boolean(mediaType && mediaType.includes('xml'))`, or even `Boolean(mediaType?.includes('xml'))`

packages/actor-rdf-parse-html-microdata/lib/ActorRdfParseHtmlMicrodata.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    const htmlParseListener = new MicrodataRdfParser({ baseIRI: action.baseIRI, xmlMode });
+    htmlParseListener.on('error', action.error);
+    htmlParseListener.on('data', action.emit);
+    const onTagEndOld = htmlParseListener.onEnd;
+    htmlParseListener.onEnd = () => {
+      onTagEndOld.call(htmlParseListener);
+      action.end();
+    };
```
joachimvh, line 31: Would `htmlParseListener.on('end', () => action.end());` not achieve the same since listeners get executed in the same order they were added?

joachimvh (approved):

packages/actor-rdf-parse-html-microdata/lib/ActorRdfParseHtmlMicrodata.ts
rubensworks: Thanks, good suggestions!

packages/actor-rdf-parse-html-microdata/lib/ActorRdfParseHtmlMicrodata.ts
rubensworks: The super call of `MicrodataRdfParser` must be called, since internally it has other end-listeners that must be invoked.

PR #752, opened 2020-10-25
pure-js hash operators

jacoscaz:
This PR, which is still a WIP, introduces pure-js implementations of 

- `actor-query-operation-reduced-hash` (`actor-query-operation-reduced-purejs-hash`)
- `actor-query-operation-distinct-hash` (`actor-query-operation-distinct-purejs-hash`)
- `actor-abstract-bindings-hash` (`and actor-abstract-bindings-purejs-hash`)

The motivation behind this PR is that the `actor-abstract-bindings-hash` causes Webpack to shim the entire `crypto` module, which is not available in browsers, as it allows to use any hash that is available in Node.js. This leads to browser-side having a ton of crypto-related code for no good reason, given that everything seems to be using the default hash algorithm `sha1`. See the attached screenshot of Webpack's bundle analyzer.

I've worked on this after talking about it with @rubensworks on Gitter. I sort of rushed into a solution while it could have done with a few more exchanges but I'm going to have limited time next week and I have wanted to get my hands on Comunica for a while. Hopefully this stands a chance of being merged but even if it doesn't, it was well worth the time. ~~All tests are passing~~ but this is my first PR to Comunica and also my first time working with dependency injection in JS-land and with a RDF-based framework: there's no way I can get it right on the first try. I look forward to your feedback!

EDIT: actually no, not all tests are passing. All `yarn run test` tests are passing, yes, but [Travis points out that other tests are not passing](https://travis-ci.org/github/comunica/comunica/jobs/738779252). I'll try to figure out what's going over the next few days... 

As for the choice of which library to use for the pre-js operators, [rusha](https://www.npmjs.com/package/rusha) seemed to be a good compromise between size (no dependencies), speed (it is said to be one of the faster implementations of `sha1` in pure Javascript) and the fact that it implements the algorithm that is used by default in the crypto-based, already-present operators. We could also add `md5` as an alternative via [spark-md5](https://www.npmjs.com/package/spark-md5). 

I've defaulted to `hex` as the digest algorithm instead of `base64` because most pure-js libraries seem to only support `hex` and `ArrayBuffer` (I have not investigated why).

Screen Shot 2020-10-23 at 22 56 08

### DIFF
--- a/packages/actor-abstract-bindings-hash/lib/AbstractFilterHash.ts
+++ b/packages/actor-abstract-bindings-hash/lib/AbstractFilterHash.ts
@@ -1,5 +1,3 @@
-import type { Hash } from 'crypto';
-import { createHash, getHashes } from 'crypto';
 import type { Bindings, IActorQueryOperationOutputBindings,
   IActorQueryOperationTypedMediatedArgs } from '@comunica/bus-query-operation';
 import {
@@ -7,6 +5,7 @@ import {
 } from '@comunica/bus-query-operation';
 import type { ActionContext } from '@comunica/core';
 import { termToString } from 'rdf-string';
+import { createHash } from 'rusha';
 import type { Algebra } from 'sparqlalgebrajs';
 
 /**
@@ -33,7 +32,7 @@ export abstract class AbstractFilterHash<T extends Algebra.Operation> extends Ac
      * @return {boolean} If it exists.
      */
   public static doesHashAlgorithmExist(hashAlgorithm: string): boolean {
-    return getHashes().includes(hashAlgorithm);
+    return [ 'sha1' ].includes(hashAlgorithm);
   }
 
   /**
@@ -42,7 +41,7 @@ export abstract class AbstractFilterHash<T extends Algebra.Operation> extends Ac
      * @return {boolean} If it exists.
      */
   public static doesDigestAlgorithmExist(digestAlgorithm: string): boolean {
-    return [ 'latin1', 'hex', 'base64' ].includes(digestAlgorithm);
+    return [ 'hex' ].includes(digestAlgorithm);
   }
 
   /**
@@ -53,7 +52,13 @@ export abstract class AbstractFilterHash<T extends Algebra.Operation> extends Ac
      * @return {string} The object's hash.
      */
   public static hash(hashAlgorithm: string, digestAlgorithm: string, bindings: Bindings): string {
-    const hash: Hash = createHash(hashAlgorithm);
+    if (hashAlgorithm !== 'sha1') {
+      throw new Error(`Unsupported hashAlgorithm "${hashAlgorithm}"`);
+    }
+    if (digestAlgorithm !== 'hex') {
+      throw new Error(`Unsupported digestAlgorithm "${digestAlgorithm}"`);
+    }
+    const hash = createHash();
     hash.update(require('canonicalize')(bindings.map(x => termToString(x))));
     return hash.digest(<any> digestAlgorithm);
   }
--- a/packages/actor-abstract-bindings-hash/lib/AbstractFilterHash.ts
+++ b/packages/actor-abstract-bindings-hash/lib/AbstractFilterHash.ts
@@ -13,60 +13,24 @@ import type { Algebra } from 'sparqlalgebrajs';
  */
 export abstract class AbstractFilterHash<T extends Algebra.Operation> extends ActorQueryOperationTypedMediated<T>
   implements IActorInitRdfDereferencePagedArgs {
-  public readonly hashAlgorithm: string;
-  public readonly digestAlgorithm: string;
-
   public constructor(args: IActorInitRdfDereferencePagedArgs, operator: string) {
     super(args, operator);
-    if (!AbstractFilterHash.doesHashAlgorithmExist(this.hashAlgorithm)) {
-      throw new Error(`The given hash algorithm is not present in this version of Node: ${this.hashAlgorithm}`);
-    }
-    if (!AbstractFilterHash.doesDigestAlgorithmExist(this.digestAlgorithm)) {
-      throw new Error(`The given digest algorithm is not present in this version of Node: ${this.digestAlgorithm}`);
-    }
   }
 
   /**
-     * Check if the given hash algorithm (such as sha1) exists.
-     * @param {string} hashAlgorithm A hash algorithm.
-     * @return {boolean} If it exists.
-     */
-  public static doesHashAlgorithmExist(hashAlgorithm: string): boolean {
-    return [ 'sha1' ].includes(hashAlgorithm);
-  }
-
-  /**
-     * Check if the given digest (such as base64) algorithm exists.
-     * @param {string} digestAlgorithm A digest algorithm.
-     * @return {boolean} If it exists.
-     */
-  public static doesDigestAlgorithmExist(digestAlgorithm: string): boolean {
-    return [ 'hex' ].includes(digestAlgorithm);
-  }
-
-  /**
-     * Create a string-based hash of the given object.
-     * @param {string} hashAlgorithm A hash algorithm.
-     * @param {string} digestAlgorithm A digest algorithm.
-     * @param bindings The bindings to hash.
-     * @return {string} The object's hash.
-     */
-  public static hash(hashAlgorithm: string, digestAlgorithm: string, bindings: Bindings): string {
-    if (hashAlgorithm !== 'sha1') {
-      throw new Error(`Unsupported hashAlgorithm "${hashAlgorithm}"`);
-    }
-    if (digestAlgorithm !== 'hex') {
-      throw new Error(`Unsupported digestAlgorithm "${digestAlgorithm}"`);
-    }
-    const hash = createHash();
-    hash.update(require('canonicalize')(bindings.map(x => termToString(x))));
-    return hash.digest(<any> digestAlgorithm);
+   * Create a string-based hash of the given object.
+   * @param bindings The bindings to hash.
+   * @return {string} The object's hash.
+   */
+  public static hash(bindings: Bindings): string {
+    return createHash()
+      .update(require('canonicalize')(bindings.map(x => termToString(x))))
+      .digest('hex');
   }
 
   public abstract async runOperation(pattern: T, context: ActionContext): Promise<IActorQueryOperationOutputBindings>;
 }
 
 export interface IActorInitRdfDereferencePagedArgs extends IActorQueryOperationTypedMediatedArgs {
-  hashAlgorithm: string;
-  digestAlgorithm: string;
+
 }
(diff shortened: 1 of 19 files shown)


rubensworks:
Thanks for the PR!

I did some more thinking after our discussion on gitter, and I thought of some other ways forward:

# 1. A new hash provider bus

Right now, we have several actors extending from `AbstractBindingsHash`, which I never really liked.
We're wrongly using inheritance to achieve common logic, while composition would be better here.

It would be better if we'd have some kind of hash bus, that provides hash function implementations.
The actors that currently depend on `AbstractBindingsHash` would simply have a (mediator over the) hash bus, and call it when hashing needs to be done.
This way, we can easily plug in different hashing implementations via the bus.

The only major downside is that this would be a breaking change.
As such, I would **postpone** this solution until the time is right for a Comunica 2.x.x.

# 2. Hardcoding hash function in `AbstractBindingsHash`

Since no one is modifying the default `hashAlgorithm` and `digestAlgorithm`, I would hardcode it to a sane default (which is fast and small in the browser).
This would in practise be a non-breaking change, and a **good temporary solution** until we can go forward with option 1.

jacoscaz:
I agree with both 1. and 2. and I was actually wondering about that specific use of inheritance which felt like the odd one out.

As for 2., it would be trivial for me to change this PR so as to change `AbstractBindingsHash` directly. Before I do so, are you ok with the following?

1. using `sha1` as the hash algorithm
2. using the [rusha](https://www.npmjs.com/package/rusha) implementation of `sha1`
3. using `hex` as the digest algorithm

There's no easy alternative for n. 3 but we could use other algorithms like `md5` via [spark-md5](https://www.npmjs.com/package/spark-md5) for the hashing itself. As we're not dealing with security the usual considerations about `md5` being _broken_ do not really apply here and `md5` is generally faster than `sha1`.  Plus, there are typings available for `spark-md5` while I had to manually add a `.d.ts` file for `rusha`.

rubensworks:
@jacoscaz Your suggestions sound good to me!

The only concern I have is with the rusha typings. Given that the package seems to have quite a bit of users, it might make sense to PR the typings you've created directly into DefinitelyTyped?

jacoscaz:
Yes, that's a good point. I would suggest keeping the two separate, thus adding the `.d.ts` file to Comunica and addressing the upstreaming of its declaration in a separate PR. 

I think a better alternative would be to have the `.d.ts` file in `rusha` itself, keeping DefinitelyTyped as an alternative solution, but I don't know how responsive the maintainer is.

rubensworks:
PRs to DefinitelyTyped should go very fast though, so there's probably not too much overhead there.
This will make our repo structure here a bit nicer, so that we can just depend on `@types/rusha` instead of having to include a custom `.d.ts` file :-)

jacoscaz:
PR to DefinitelyTyped: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/49123 - @rubensworks would you have to review it? It's my first one on DT and the resident bot tells me I should find someone to do so...

packages/actor-abstract-bindings-hash/lib/AbstractFilterHash.ts
```diff
@@ -33,7 +32,7 @@ export abstract class AbstractFilterHash<T extends Algebra.Operation> extends Ac
      * @return {boolean} If it exists.
      */
   public static doesHashAlgorithmExist(hashAlgorithm: string): boolean {
-    return getHashes().includes(hashAlgorithm);
+    return [ 'sha1' ].includes(hashAlgorithm);
```
jacoscaz, line 35: Should I keep using this form or should I move to something like `return 'sha1' === hashAlgorithm`? I guess it depends on how long you'd like to keep this module around and how much you'd like for others to feel like they can extend it by adding other hashing algorithms. Same applies for line 55, which could to with using the same style as this one regardless of which style you prefer.

rubensworks: I'm ok with removing those functions, along with the parameters: https://github.com/comunica/comunica/blob/2d0818c64e5bfbbb334ecbccb7b5a98a69263d1c/packages/actor-abstract-bindings-hash/lib/AbstractFilterHash.ts#L17-L18

jacoscaz: Done!

rubensworks (changes requested):
Thanks @jacoscaz, this looks great!

I have no comments on the changes.
The are just a few more places that reference the removed parameters:
* https://github.com/comunica/comunica/blob/3a9160199493a0aaf2f229809386868107bb3d88/packages/actor-abstract-bindings-hash/components/context.jsonld#L10-L11
* https://github.com/comunica/comunica/blob/cf916839f99b7141e0f1d4e2e70f90aeb597789a/packages/actor-query-operation-distinct-hash/README.md#config-parameters
* https://github.com/comunica/comunica/blob/cf916839f99b7141e0f1d4e2e70f90aeb597789a/packages/actor-query-operation-reduced-hash/README.md#config-parameters

Could you remove them here as well?

jacoscaz:
@rubensworks done! I had totally forgot to check `README.md` files, sorry about that!

rubensworks:
No worries, thanks a lot! :-)

I plan to make a new release early next week.
Just want to add one minor enhancement first.

jacoscaz:
That would be awesome as it more or less coincides with the timing I had in mind for quadstore. Releasing `quadstore@7.0.1` with this change would cut down the size of the browser-side bundle by almost a third (or more), which would really help us.

rubensworks:
@jacoscaz I probably should've checked this sooner, but sparqlee makes use of this library: https://www.npmjs.com/package/create-hash
This looks like a drop-in replacement for crypto, which makes use of the browser's hashing functionality outside of Node.
Do you think it would be feasible to use that one instead of rusha?

jacoscaz:
I would have to evaluate the use cases inside `sparqlee` but the issue with using `crypto` was not so much about `crypto` itself but the fact that only a _very_ small part of it was actually being used (the `sha1` implementation) with no tree-shaking of its unused parts by Webpack (such tree-shaking being basically impossible due to CommonJS + using a single entry point `createHash` function for all hashing algorithms).

Based on this, I would look at your question in reverse: does `sparqlee` need the entire functionality of `crypto` or could it also be modified to use only one or two hashing algorithms?

jacoscaz:
That said, `create-hash` is vastly smaller than `crypto-browserify`, so using it in place of `rusha` would be a good compromise I think.

rubensworks:
> create-hash is vastly smaller than crypto-browserify, so using it in place of rusha would be a good compromise I think.

I would be interested to know what the difference in (browser) size is between rusha and create-hash.
If create-hash is of comparable size, then create-hash might be a better option, since we're already needing it in sparqlee anyways due to the SPARQL functions spec.
If rusha is on the other hand significantly smaller, then we may want to keep using that.

jacoscaz:
I suspect that, given that `create-hash` should be a lot smaller than `crypto-browserify` anyway, consolidating around one fewer dependency would be of more value than aiming for the lowest possible bundle size. That said, if you'd like me to I can look at comparing the two libraries during the weekend.

rubensworks:
Sure, that would be great!

jacoscaz:
@rubensworks I've done a bit of testing with `rusha`, `create-hash` and `crypto` using Webpack v4.44.2. 

My webpack configuration:

```js
const path = require('path');
module.exports = {
  mode: 'production',
  entry: path.join(__dirname, 'main.js'),
  output: {
    path: __dirname,
    filename: 'main.bundle.js',
    library: 'test',
    libraryTarget: 'umd',
  },
  target: 'web',
};
```

My main file:

```js
// const rusha = require('rusha');
// const createHash = require('create-hash');
// const md5 = require('spark-md5');
// const crypto = require('crypto');
// const sha256 = require('hash.js/lib/hash/sha/512');
const hashjs = require('hash.js');
```

Bundle sizes:

- `rusha`: 13k
- `create-hash`: 117k
- `crypto`: 632k
- `spark-md5`: 9.1k
- `hash.js/sha256`: 8.0k
- `hash.js`: 17k

Given that `create-hash` is used in `sparqlee` with more than one hashing algorithm and thus cannot be swapped out for a smaller library without breaking [support for some SPARQL functions][1], given these findings I think it would be best to replace `rusha` with `create-hash` as the latter will always end up in any configuration of comunica anyway. 

Alternatively, we could evaluate whether to replace `create-hash` with `hash.js`, which seems to take up vastly less bundle space. I haven't looked into how much of `crypto#createHash` is covered by `hash.js` and/or any difference in performance between the two.

[1]: https://github.com/comunica/sparqlee/blob/4b873834a38c35329495d142eaf1c59f56fc0038/lib/functions/RegularFunctions.ts#L631-L669

Based on these findings, I would either get rid of `rusha` and use `create-hash` in case

jacoscaz:
I just noticed - the combination of `spark-md5` + `hash.js` sits at 25k while getting us support for all of those `sparqlee` functions...

jacoscaz:
Actually, I just tried replacing `create-hash` with `hash.js`+`spark-md5` in `sparqlee` and everything seems to work just fine: https://github.com/jacoscaz/sparqlee/blob/feature/lighter-hash-dependencies/lib/functions/RegularFunctions.ts#L632-L670

jacoscaz:
As a further addendum, I've used https://github.com/dominictarr/crypto-bench to compare the performance of `rusha` with that of `hash.js` and the latter seems to be faster for small payloads, which fits the use cases of both `sparqlee` and the abstract actor in `comunica`. 

At this point I think it'd be worth replacing both `rusha` and `create-hash` with `spark-md5` + `hash.js` in both projects.

rubensworks:
Those are some interesting findings!

>At this point I think it'd be worth replacing both rusha and create-hash with spark-md5 + hash.js in both projects.

Sounds good to me.

The sparqlee change should be trivial I guess.

How do you think this change would fit best in Comunica?
Should we revert the last commit to master (== this PR), since hash.js supports different functions?
Or do you think it would be better to build upon the version applied by this PR?

jacoscaz:
Mh... Good point. I'm not sure. I think I would build on this PR rather than revert back to having the custom parameters. Nobody is using them and less code is usually better than more code in the long run. That said, you're in a much better place than I am to choose which way to go. I'm happy with both.

rubensworks:
No strong opinion myself either. If we can keep the current parameters, then this may be nice to have.
But I think the most important consideration here is probably just ease of implementation.
What do you think would be easier to implement spark-md5 + hash.js over? The current or reverted state?

jacoscaz:
> But I think the most important consideration here is probably just ease of implementation.

Then, IMHO, it'd be better to keep things as they are now, without those parameters. The `digestAlgorithm` parameter would have to be removed anyway as there's very little support for anything outside of `hex`.

rubensworks:
Alright, no reverting then, sounds good!

PR #757, opened 2020-10-31
Replaces rusha with hash.js

jacoscaz:
Follows up to the ideas discussed in https://github.com/comunica/comunica/pull/752 by replacing `rusha` with `hash.js`. This brings `comunica` to use the same pure-js hashing libraries as `sparqlee` (https://github.com/comunica/sparqlee/pull/78), which should lead to a dramatic decrease in browser-side bundle size.


rubensworks:
Great, thanks!

rubensworks:
For reference, all of these changes reduces bundle size from 1.47 MiB to 1.15 MiB, which is quite a significant reduction.

jacoscaz:
That’s not as dramatic a difference as I had expected but, still, a really nice one considering that it comes at no loss in functionality.  

I wonder why it seems to be more towards 350kb than the 700kb I had envisioned -  maybe some dependencies of crypto-browserify are still being pulled in by other modules?

I’ll have a look with the analyzer as soon as I update quadstore to the new versions of these packages.

Issue #780, opened 2021-02-10
Use weaker query engine interface in bus-rdf-metadata-extract


brechtvdv:
#### Issue type:

- :bug: Bug 

____
#### Description:

I added bus-rdf-metadata-extract to my custom package (yarn add @comunica/bus-rdf-metadata-extract), but when I run "yarn install" it doesn't install @comunica/actor-init-sparql, because this is a peerDependency I guess?. This returns following error when running typescript:
____
#### Crash log:

`node_modules/@comunica/bus-rdf-metadata-extract/lib/ActorRdfMetadataExtractQuery.d.ts:1:38 - error TS2307: Cannot find module '@comunica/actor-init-sparql' or its corresponding type declarations. `

I guess @comunica/actor-init-sparql should be added to devDependencies or dependencies in ´package.json´ of @comunica/bus-rdf-metadata-extract ?

rubensworks:
Making `'@comunica/actor-init-sparql'` an actual or dev dependency might cause some problems, so I would not do that.

What we could do is making the [query engine contract](https://github.com/comunica/comunica/blob/master/packages/bus-rdf-metadata-extract/lib/ActorRdfMetadataExtractQuery.ts#L24) a bit weaker.
This should be possible by moving the query engine interface so a separate types-only interface.

(\cc @jacoscaz, this could also be solved via this `@comunica/types` package)

jacoscaz:
Hi all! Yes, that is precisely what I'm aiming for with `@comunica/types`, having the query engine contract defined as an interface implemented by `ActorInitSparql`.

PR #784, opened 2021-02-21
Move actor-init-sparql's query engine contract types into the new package @comunica/types

jacoscaz:
This PR moves actor-init-sparql's contract types, i.e. the types that constitute the query engine contract of Comunica, into a separate types-only package `@comunica/types`. 

The goal of this new package is to play a similar role to `@types/rdf-js`, that is to provide shared typings without an explicit dependency on `@comunica/actor-init-sparql` and all of its dependencies.

The `@comunica/types` package should help with the following issues:

- https://github.com/beautifulinteractions/node-quadstore/issues/129
- https://github.com/comunica/comunica/issues/780

~~Something that could be improved is doing away with the re-export of the `IBindings` interface as the `Bindings` type in https://github.com/jacoscaz/comunica/blob/package/types/packages/bus-query-operation/lib/Bindings.ts#L10 and referencing the `IBindings` interface directly throughout the project. @rubensworks what do you think?~~ Actually, this turned out to be easier than I expected, so I just went ahead and did it.


packages/bus-query-operation/lib/ActorQueryOperation.ts
```diff
@@ -177,55 +183,6 @@ export abstract class ActorQueryOperation extends Actor<IActionQueryOperation, I
...
 }
 
-export interface IExpressionContext {
-  now?: Date;
-  baseIRI?: string;
-  // Exists?: (expr: Algebra.ExistenceExpression, bindings: Bindings) => Promise<boolean>;
-  // bnode?: (input?: string) => Promise<RDF.BlankNode>;
-}
-
-export interface IActionQueryOperation extends IAction {
```
rubensworks, line 187: As to not require making this a breaking change, I would re-export the imported types here.

packages/bus-query-operation/lib/Bindings.ts
```diff
@@ -1,38 +1,18 @@
...
 
-/**
- * A stream of bindings.
- *
- * Next to the list of available variables,
- * an optional metadata hash can be present.
- *
- * @see Bindings
- */
-export type BindingsStream = AsyncIterator<Bindings>;
```
rubensworks, line 16: Also re-export this and following.

packages/bus-query-operation/package.json
```diff
@@ -41,7 +41,8 @@
     "@comunica/core": "^1.0.0"
   },
   "devDependencies": {
-    "@comunica/core": "^1.19.2"
+    "@comunica/core": "^1.19.2",
+    "@comunica/types": "^1.19.2"
```
rubensworks, line 45: This would actually have to be an actual dependency. Also for all other packages (that use these types) for that matter. Strange that the linter didn't trip over this, as I thought I defined strict linter rules for this...

packages/core/lib/Actor.ts
```diff
@@ -1,8 +1,12 @@
+import type { IAction, IActionContext } from '@comunica/types';
 import { Map } from 'immutable';
 import type { Bus } from './Bus';
 import type { Logger } from './Logger';
 import { KEY_CONTEXT_LOG } from './Logger';
 
+export type ActionContext = IActionContext;
```
rubensworks, line 7: Ah, here you re-export, great :-)

packages/types/index.ts
```diff
@@ -0,0 +1,161 @@
+import type { AsyncIterator } from 'asynciterator';
+import type { Map } from 'immutable';
+import type * as RDF from 'rdf-js';
+import type { Algebra } from 'sparqlalgebrajs';
+
+/**
```
rubensworks, line 6: Just wondering if we want to split this up into separate files, not sure yet...
In any case, we can still do this in a non-breaking future change if needed.

packages/types/index.ts
```diff
@@ -0,0 +1,161 @@
...
+ * An immutable solution mapping object.
+ * This maps variables to a terms.
+ *
+ * Variables are represented as strings containing the variable name prefixed with '?'.
+ * Blank nodes are represented as strings containing the blank node name prefixed with '_:'.
+ * Terms are named nodes, literals or the default graph.
+ */
+export type IBindings = Map<string, RDF.Term>;
```
rubensworks, line 14: Any reason why this is prefixed with `I`, while the old type wasn't?
(Also applicable for other types in this file)

If no particular reason, I'd keep the type names exactly the same as before, as to not confuse users.

packages/types/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "@comunica/types",
+  "version": "1.19.2",
+  "description": "Typings module for Comunica",
+  "lsd:module": false,
```
rubensworks, line 5: Or just removing this line would also be ok :-)

packages/types/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "files": [
+    "components",
+    "lib/**/*.d.ts",
+    "lib/**/*.js",
+    "bin/**/*.d.ts",
+    "bin/**/*.js",
+    "index.d.ts",
+    "index.js"
+  ],
+  "devDependencies": {
```
rubensworks, line 34: All of these would have to become actual dependencies based on how the types are defined now.

This makes me wonder if we really want this, or if we somehow want to define some more minimal interfaces of these packages internally, without having our types package blow up in number of interfaces.

rubensworks (changes requested):
Great! See my comments inline.

packages/bus-query-operation/lib/ActorQueryOperation.ts
jacoscaz: Ah yes, I will put those back in. I had dropped them to get the TS compiler to help me refactor stuff internally but not putting them back would likely break other packages depending on Comunica.

packages/bus-query-operation/lib/Bindings.ts
jacoscaz: Will do.

packages/bus-query-operation/package.json
jacoscaz: The general issue of errors not being 100% deterministic across environments is an annoying aspect of TS and TS linting, at least insofar as my limited experience goes. Will do.

packages/types/index.ts
jacoscaz: No strong opinion on this one.

packages/types/index.ts
jacoscaz: I think `IBindings` is a clearer indicator of what that is - a TS typing - and I think the `@comunica/types` could benefit from the additional clarity (given that most other types and interfaces do start with `I`). Same applies to `IBindingsStream`.

packages/types/package.json
jacoscaz: Will do!

packages/types/package.json
jacoscaz: I share this concern, although I have no idea of which percentage of these modules' typings we're actually using. I'll try to have a go at this and report back here.

packages/types/index.ts
rubensworks: We currently follow the convention that `interface`'s start with `I`, and types don't start with `I`.
I'm open to changing this, but then we'd have to make this consistent across the whole codebase, and make this a breaking change.

jacoscaz: Ah, you're right! My bad. Will drop the `I` from types. Apologies for the latency, I'm currently on a business trip and have very little time for anything else but the tests we're carrying out. At least the facility is nice!

rubensworks: No worries, there's no hurry on my end. Have fun there!

packages/types/package.json
jacoscaz: Done!

packages/bus-query-operation/lib/ActorQueryOperation.ts
jacoscaz: This should be resolved now. I've opted to rename types using the `T` prefix instead of the `I` prefix in the new `@comunica/types` package (feels more coherent) but re-exported them without any prefix to maintain backward compatibility.

packages/bus-query-operation/lib/Bindings.ts
jacoscaz: Further to my comments on Gitter, I gave up and simply re-added and re-exported the old `Bindings` and `BindingsStream` types while using the new `TBindings` and `TBindingsStream` ones throughout the module. They are effectively identical and the TS compiler is able to figure out that they are interchangeable. This likely comes at a cost in terms of compilation performance but I haven't seen tangible differences in compilation times, so I went with this workaround hoping that the night will bring some new idea as to how to approach this.

packages/types/index.ts
jacoscaz: Actually, I would rather keep them in the same file until the latter becomes exceedingly long and unmaintainable.

packages/types/index.ts
jacoscaz: See my comments above. I've replaced the `I` with `T` in the `@comunica/types` package as that felt like the most coherent solution but kept the original names when re-exporting types in packages like `@comunica/bus-query-operation` to maintain backward compatibility.

packages/bus-query-operation/package.json
jacoscaz: Done!

packages/types/package.json
jacoscaz: Gave a go at this one but found that Comunica is using quite a lot of both `AsyncIterator` and Immutable's `Map`, which would require quite a bit of typings duplication. Any ideas?

packages/types/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "@types/rdf-js": "4.0.1",
```
rubensworks, line 34:
```suggestion
    "@types/rdf-js": "*",
```

packages/types/package.json
rubensworks: Yeah, I was afraid of that. Let's just keep it as-is then.

packages/types/index.ts
rubensworks: Sounds good to me.

packages/bus-query-operation/lib/ActorQueryOperation.ts
rubensworks: > I've opted to rename types using the T prefix instead of the I prefix in the new @comunica/types package

While I'm not opposed to the convention of prefixing types with `T`, such a change will have far-stretching consequences. If we want to be consistent, we'd have to prefix _all_ types in this repo with `T`.
Also, this may confuse users of the current types (without `T`).

For these reasons, I would suggest omitting the `T` prefix.

packages/bus-query-operation/lib/Bindings.ts
rubensworks: Did you experience any problems with just re-exporting those from the `types` package?
If not, I would propose to do that. Otherwise we'd have identical type and interface definitions in two different places, which will certainly cause maintainability problems in the future. (I would also expect jest's coverage requirement to trip over this)

packages/bus-query-operation/lib/ActorQueryOperation.ts
jacoscaz: No objections on my end. I thought that, given we're adding a new package, it might be a good idea to adopt a less confusing (IMHO) naming convention right from the start. But, I do understand the argument for consistency. I'll rename without the `T`.

packages/bus-query-operation/lib/Bindings.ts
jacoscaz: Yes, once I drop the `T` and try to re-export the `Bindings` type as imported from `@comunica/types`, the TS compiler starts complaining about duplicate exports. I don't really know how to get out of this. Will test some more later on.

packages/types/index.ts
jacoscaz: I guess you can resolve this as we're also discussing the renaming issue in [the thread above](https://github.com/comunica/comunica/pull/784#discussion_r587715184).

packages/bus-query-operation/lib/ActorQueryOperation.ts
jacoscaz: Done!

packages/bus-query-operation/lib/Bindings.ts
jacoscaz: I might have found something that works. It's a bit ugly but it should prevent the maintainability issues you mentioned. The gist of it is as follows:

```js
import type { Bindings as _Bindings } from '@comunica/types';
export type Bindings = _Bindings;
function Bindings(hash) { /* ... */ }
```

The compiler seems happy and I've thus switched to this approach in my last commit. Done?

.eslintrc.js
```diff
@@ -108,6 +108,7 @@ module.exports = {
     '@typescript-eslint/no-use-before-define': 'off',
     '@typescript-eslint/prefer-nullish-coalescing': 'off',
     '@typescript-eslint/consistent-type-imports': ['error', { prefer: 'type-imports' }],
+    '@typescript-eslint/indent': 'off',
```
rubensworks, line 111: Was this disabled intentionally?

packages/actor-abstract-path/lib/ActorAbstractPath.ts
```diff
@@ -205,10 +205,11 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
      */
   // Let the iterator `it` emit all bindings of size 2, with subjectStringVariable as value subjectVal
   // and objectStringVariable as value all nodes reachable through predicate* beginning at objectVal
+  /* eslint max-len: off */
```
rubensworks, line 208: This can probably go if the next diff is undone.

packages/types/index.ts
```diff
@@ -0,0 +1,161 @@
...
+  context?: ActionContext;
+}
+
+/**
+ * An immutable key-value mapped context that can be passed to any (@link IAction}.
+ * All actors that receive a context must forward this context to any actor, mediator or bus that it calls.
+ * This context may be transformed before forwarding.
+ *
+ * Each bus should describe in its action interface which context entries are possible (non-restrictive)
+ * and expose a `KEY_CONTEXT_${ENTRY_NAME}` constant for easy reuse.
```
rubensworks, line 145: I suspect some recent changes were lost. E.g. https://github.com/comunica/comunica/commit/12b9ee3e8e5bc2d0fadd662a3d6aeef838b87619#diff-cfc7a9b5296e9c5c4950a2fde534bebbe7053cc63ae511026bfa251f9e56a1b6

packages/types/index.ts
```diff
@@ -0,0 +1,161 @@
...
+ *   Each key must be prefixed with the package name followed by a `:`.
+ *   For example, the `rdf-resolve-quad-pattern` bus declares the `sources` entry,
+ *   which should be named as `@comunica/bus-rdf-resolve-quad-pattern:sources`.
+ *
+ * This context can contain any information that might be relevant for certain actors.
+ * For instance, this context can contain a list of datasources over which operators should query.
+ */
+export type ActionContext = Map<string, any>;
+
+export interface IActorInitSparql {
```
rubensworks, line 159: Since this is a brand new interface, we may consider giving this a more generic name. Perhaps something like `IQueryEngine`?
What do you think?

packages/types/index.ts
```diff
@@ -0,0 +1,161 @@
...
+   * The query operation to handle.
+   */
+  operation: Algebra.Operation;
+}
+
+/**
+ * Query operation output.
+ * @see IActorQueryOperationOutputBindings, IActorQueryOperationOutputQuads, IActorQueryOperationOutputBoolean
+ */
+export type ActorQueryOperationOutput =
```
rubensworks, line 97:
```suggestion
export type IActorQueryOperationOutput =
```

I'm aware this is not an interface, but I would do this to be better in line with the currently existing interface/type (so that developers only have to change the import target, not the rest of their code).

We may want to add a `TODO` comment above this to properly rename upon our next major update. Perhaps we could then even use a name such as `QueryOutput`.

rubensworks (changes requested):
Just some minor nits. Should be good to go after these suggestions.

.eslintrc.js
jacoscaz: I just removed this line. Something in my workflow had altered param indentation in a few places but I got tricked into thinking I had inherited such indentation from `master`.

packages/types/index.ts
jacoscaz: Agreed! And done.

packages/types/index.ts
jacoscaz: Done!

jacoscaz:
Something apparently unrelated to this PR but that comes up at least twice a day: running `yarn run test` or `yarn run test-ci` often fails with the following error:

```
 FAIL  packages/actor-rdf-resolve-quad-pattern-hypermedia/test/LinkedRdfSourcesAsyncRdfIterator-test.ts
  ● LinkedRdfSourcesAsyncRdfIterator › A LinkedRdfSourcesAsyncRdfIterator instance › handles metadata event emitted before the end event

    Unhandled error. (TypeError: require(...) is not a function



      at Function.toNodeReadable (packages/bus-http/lib/ActorHttp.ts:376:32)
      at ActorRdfDereferenceHttpParse.run (packages/actor-rdf-dereference-http-parse/lib/ActorRdfDereferenceHttpParseBase.ts:1480:53)
      at MediatedLinkedRdfSourcesAsyncRdfIterator.getSource (packages/actor-rdf-resolve-quad-pattern-hypermedia/lib/MediatedLinkedRdfSourcesAsyncRdfIterator.ts:905:32))Error [ERR_UNHANDLED_ERROR]: Unhandled error. (TypeError: require(...) is not a function
      at Function.toNodeReadable (packages/bus-http/lib/ActorHttp.ts:376:32)
      at ActorRdfDereferenceHttpParse.run (packages/actor-rdf-dereference-http-parse/lib/ActorRdfDereferenceHttpParseBase.ts:1480:53)
      at MediatedLinkedRdfSourcesAsyncRdfIterator.getSource (packages/actor-rdf-resolve-quad-pattern-hypermedia/lib/MediatedLinkedRdfSourcesAsyncRdfIterator.ts:905:32))
      at SimpleTransformIterator.<anonymous> (packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts:3039:19)
      at MediatedLinkedRdfSourcesAsyncRdfIterator.destinationEmitError (node_modules/asynciterator/dist/asynciterator.cjs:1067:23)
      at node_modules/asynciterator/dist/asynciterator.cjs:177:26
      at MediatedLinkedRdfSourcesAsyncRdfIterator._destroy (node_modules/asynciterator/dist/asynciterator.cjs:868:9)
      at MediatedLinkedRdfSourcesAsyncRdfIterator.destroy (node_modules/asynciterator/dist/asynciterator.cjs:174:18)
      at packages/actor-rdf-resolve-quad-pattern-hypermedia/lib/LinkedRdfSourcesAsyncRdfIterator.ts:200:30
```

Whenever this happens, running the test suite again always succeeds without errors. I don't think it's related to something I've done but I thought I'd mention it anyway, if only to keep track of it.

packages/actor-abstract-path/lib/ActorAbstractPath.ts
jacoscaz: Done!

rubensworks:
> running yarn run test or yarn run test-ci often fails with the following error:

Hmm, I never seen that error before (locally or CI).
Any chance you're using a very recent (experimental?) Node version?
If not, feel free to open an issue for this, just in case others would also run into this problem.

jacoscaz:
I generally tend to stick with the LTS branch, I'm currently using `v14.15.4`. Will open an issue.

packages/types/index.ts
jacoscaz: I completely missed the updated version of this jsdoc comment in moving the `ActionContext` type to `@comunica/types` and then rebasing my branch onto `master`. I've now updated the comment in `@comunica/types`  to match the version in `master` and modified the latter to add the `@deprecated` flag. Should be done.

jacoscaz:
Perhaps I should squash all of these commits into a single one after your review? It'd make the commit history a bit tidier.

rubensworks:
Oh, that's ok, I can squash here via the GitHub UI before merging :-)

packages/actor-query-operation-leftjoin-left-deep/lib/ActorQueryOperationLeftJoinLeftDeep.ts
```diff
@@ -34,6 +32,7 @@ export class ActorQueryOperationLeftJoinLeftDeep extends ActorQueryOperationType
    *                                                                       of an operation.
    * @return {BindingsStream}
    */
+  /* eslint max-len: off */
```
rubensworks, line 35: I assume this line can also go?

jacoscaz: Done!

rubensworks:
Perfect, thanks!

Any urgency on a release from your end?

jacoscaz:
It'd be nice to be able to reference the new `@comunica/types` with the next release of `quadstore` and `quadstore-comunica` so as to remove the duplicated typings but there's plenty to do WRT to the query spec, so no rush at all.

jacoscaz:
Thank you for merging this!

Issue #745, opened 2020-10-16
Timeout on SPARQL queries


ddeboer:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

In our user-facing application, we would like to stop waiting for SPARQL responses if they take longer than a couple of seconds to arrive. While this can be solved in the calling code, it would be nice if Comunica supported client timeouts out of the box.

rubensworks:
For reference, we've just implemented timeout support when running Comunica as a SPARQL endpoint (#797).

For direct usages of Comunica via the JS API, this might be more difficult, as it's not really possible to interupt a function call in a purely single-threaded JS environment.
So if we want to add this, we may want to create a webworker-based wrapper around Comunica, which can create one or more workers for executing a query, and killing such a worker one a timeout is exceeded.
Since inter-worker communication may be expensive, we'll have to investigate if it's feasible to pipe all query results across this worker communication API at a sufficient level of efficiency.

rubensworks:
Considering this resolved (#70)

PR #797, opened 2021-03-19
Enable worker threads in SPARQL endpoints

rubensworks:
My original goal was to fix the problem where timeouts did not really kill long-running query processes (as this is not really possible in single-treaded Node).
So I solved this using the `'cluster'` API, which also allowed me to easily add support for more than one worker.

Related to #745

### DIFF
--- a/packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
+++ b/packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
@@ -1,3 +1,4 @@
+import * as cluster from 'cluster';
 import type { EventEmitter } from 'events';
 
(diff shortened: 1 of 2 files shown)


packages/actor-init-sparql/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -56,6 +59,7 @@ Options:
     this.context = args.context || {};
     this.timeout = args.timeout ?? 60_000;
```
joachimvh, line 60: Huh, TIL you can write numbers like that.

joachimvh (approved):
Looks fine but I have never used cluster so not sure how valuable that review is :D

PR #803, opened 2021-03-26
Fix blank node correlation (closes #795)

### LINKED ISSUE #795: @comunica/actor-init-sparql-rdfjs does not correlate blank nodes
#### Issue type:

- :bug: Bug

____
#### Description:

The actor implemented in `@comunica/actor-init-sparql-rdfjs` seems unable to correlate blank nodes when querying upon a RDF/JS source. 

The problem can be reproduced by running the test against the following manifest:

```sh
cd packages/actor-init-sparql-rdfjs
 ../../node_modules/.bin/rdf-test-suite-ldf spec/sparql-engine.js http://jacoscaz.com/manifest-ldf-tests/sparql-rdfjs/rdfjs-manifest.ttl -d 200000 -c ../../.rdf-test-suite-ldf-cache/
```

The manifest run a simple `SELECT * FROM { ?s ?p ?o}` query on the following quads:

```turtle
@prefix : <http://example.org/> .

:s1 :p1 _:b1 .
_:b1 :p2 :o2 .
```

The actor returns the following bindings:

```js
 Bindings:  [
  {
    "?s": {  "termType": "NamedNode",  "value": "http://example.org/s1" },
    "?p": { "termType": "NamedNode", "value": "http://example.org/p1" },
    "?o": {
      "termType": "BlankNode", "value": "bc_0_b4_b11",
      "skolemized": { "termType": "NamedNode", "value": "urn:comunica_skolem:source_0:b4_b1" }
    }
  },
  {
    "?s": {
      "termType": "BlankNode", "value": "bc_0_b4_b12",
      "skolemized": { "termType": "NamedNode", "value": "urn:comunica_skolem:source_0:b4_b1" }
    },
    "?p": { "termType": "NamedNode", "value": "http://example.org/p2" },
    "?o": { "termType": "NamedNode", "value": "http://example.org/o2" }
  }
]
```

While the skolemized versions of the two blank nodes are correlated, the labels themselves are not (but they should be!).

I stumbled into this issue while troubleshooting https://github.com/beautifulinteractions/node-quadstore/issues/134 and @rubensworks noted it might be related to https://github.com/comunica/comunica/issues/773 .
____
#### Environment:

| Comunica Init Actor | 1.19.2 (dev)
| node                | v14.15.4
| npm                 | 7.4.1
| yarn                | 1.22.10
| Operating System    | darwin (Darwin 20.3.0)

jacoscaz:
This PR (~~which depends on https://github.com/comunica/sparqlee/pull/81~~ MERGED) fixes the blank node correlation issue raised in issue https://github.com/comunica/comunica/issues/795 (and may fix other similar issues). 

#### TODO

- [x] modify unit tests to match the new behavior of `packages/actor-query-operation-project/lib/ActorQueryOperationProject.ts`
- [x] modify `ActorQueryOperation` tests to account for the `bnode` functions in expression contexts
- [x] add a test to verify blank node correlation
- [x] update to new version of `sparqlee` once https://github.com/comunica/sparqlee/pull/81 is merged and published to NPM
- [x] general cleanup and linting

### DIFF
--- a/packages/actor-query-operation-construct/lib/BindingsToQuadsIterator.ts
+++ b/packages/actor-query-operation-construct/lib/BindingsToQuadsIterator.ts
@@ -75,9 +75,6 @@ export class BindingsToQuadsIterator extends MultiTransformIterator<Bindings, RD
   public static localizeBlankNode(blankNodeCounter: number,
     term: RDF.Term): RDF.Term {
     if (term.termType === 'BlankNode') {
-      if (term instanceof BlankNodeBindingsScoped) {
-        return new BlankNodeBindingsScoped(`${term.value}${blankNodeCounter}`);
-      }
       return DF.blankNode(`${term.value}${blankNodeCounter}`);
     }
     return term;
--- a/packages/actor-query-operation-construct/lib/BindingsToQuadsIterator.ts
+++ b/packages/actor-query-operation-construct/lib/BindingsToQuadsIterator.ts
@@ -1,4 +1,3 @@
-import { BlankNodeBindingsScoped } from '@comunica/data-factory';
 import type { Bindings, BindingsStream } from '@comunica/types';
 import type { AsyncIterator } from 'asynciterator';
 import { ArrayIterator, MultiTransformIterator } from 'asynciterator';
--- a/packages/actor-query-operation-construct/lib/BindingsToQuadsIterator.ts
+++ b/packages/actor-query-operation-construct/lib/BindingsToQuadsIterator.ts
@@ -1,10 +1,10 @@
+import { BlankNodeBindingsScoped } from '@comunica/data-factory';
 import type { Bindings, BindingsStream } from '@comunica/types';
 import type { AsyncIterator } from 'asynciterator';
 import { ArrayIterator, MultiTransformIterator } from 'asynciterator';
 import { DataFactory } from 'rdf-data-factory';
 import type * as RDF from 'rdf-js';
 import { mapTerms } from 'rdf-terms';
-import {BlankNodeBindingsScoped} from '@comunica/data-factory';
 const DF = new DataFactory();
 
 /**
--- a/packages/actor-query-operation-construct/lib/BindingsToQuadsIterator.ts
+++ b/packages/actor-query-operation-construct/lib/BindingsToQuadsIterator.ts
@@ -4,6 +4,7 @@ import { ArrayIterator, MultiTransformIterator } from 'asynciterator';
 import { DataFactory } from 'rdf-data-factory';
 import type * as RDF from 'rdf-js';
 import { mapTerms } from 'rdf-terms';
+import {BlankNodeBindingsScoped} from '@comunica/data-factory';
 const DF = new DataFactory();
 
 /**
@@ -74,6 +75,9 @@ export class BindingsToQuadsIterator extends MultiTransformIterator<Bindings, RD
   public static localizeBlankNode(blankNodeCounter: number,
     term: RDF.Term): RDF.Term {
     if (term.termType === 'BlankNode') {
+      if ((<RDF.BlankNode & { singleBindingsScope?: boolean }>term).singleBindingsScope) {
+        return new BlankNodeBindingsScoped(`${term.value}${blankNodeCounter}`);
+      }
       return DF.blankNode(`${term.value}${blankNodeCounter}`);
     }
     return term;
--- a/packages/actor-query-operation-construct/lib/BindingsToQuadsIterator.ts
+++ b/packages/actor-query-operation-construct/lib/BindingsToQuadsIterator.ts
@@ -75,7 +75,7 @@ export class BindingsToQuadsIterator extends MultiTransformIterator<Bindings, RD
   public static localizeBlankNode(blankNodeCounter: number,
     term: RDF.Term): RDF.Term {
     if (term.termType === 'BlankNode') {
-      if ((<RDF.BlankNode & { singleBindingsScope?: boolean }>term).singleBindingsScope) {
+      if (term instanceof BlankNodeBindingsScoped) {
         return new BlankNodeBindingsScoped(`${term.value}${blankNodeCounter}`);
       }
       return DF.blankNode(`${term.value}${blankNodeCounter}`);
--- a/packages/actor-query-operation-extend/lib/ActorQueryOperationExtend.ts
+++ b/packages/actor-query-operation-extend/lib/ActorQueryOperationExtend.ts
@@ -33,7 +33,7 @@ export class ActorQueryOperationExtend extends ActorQueryOperationTypedMediated<
     );
 
     const extendKey = termToString(variable);
-    const config = { ...ActorQueryOperation.getExpressionContext(context, this.mediatorQueryOperation) };
+    const config = { ...ActorQueryOperation.getAsyncExpressionContext(context, this.mediatorQueryOperation) };
     const evaluator = new AsyncEvaluator(expression, config);
 
     // Transform the stream by extending each Bindings with the expression result
--- a/packages/actor-query-operation-project/lib/ActorQueryOperationProject.ts
+++ b/packages/actor-query-operation-project/lib/ActorQueryOperationProject.ts
@@ -1,9 +1,10 @@
 import type { IActorQueryOperationTypedMediatedArgs } from '@comunica/bus-query-operation';
 import { ActorQueryOperation, ActorQueryOperationTypedMediated } from '@comunica/bus-query-operation';
 import type { ActionContext, IActorTest } from '@comunica/core';
-import { BlankNodeScoped } from '@comunica/data-factory';
+import { BlankNodeBindingsScoped, BlankNodeScoped } from '@comunica/data-factory';
 import type { Bindings, BindingsStream, IActorQueryOperationOutputBindings } from '@comunica/types';
-import { DataFactory } from 'rdf-data-factory';
+import {BlankNode, DataFactory} from 'rdf-data-factory';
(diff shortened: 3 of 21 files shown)


packages/actor-query-operation-construct/lib/BindingsToQuadsIterator.ts
```diff
@@ -74,6 +75,9 @@ export class BindingsToQuadsIterator extends MultiTransformIterator<Bindings, RD
   public static localizeBlankNode(blankNodeCounter: number,
     term: RDF.Term): RDF.Term {
     if (term.termType === 'BlankNode') {
+      if (term instanceof BlankNodeBindingsScoped) {
```
rubensworks, line 78: Do we really need to preserve `BlankNodeBindingsScoped` in the context of construct queries?
AFAICS `BlankNodeBindingsScoped` is only really useful for bindings results, while we will always return quads here.

packages/actor-query-operation-extend/lib/ActorQueryOperationExtend.ts
```diff
@@ -33,7 +33,7 @@ export class ActorQueryOperationExtend extends ActorQueryOperationTypedMediated<
     );
 
     const extendKey = termToString(variable);
-    const config = { ...ActorQueryOperation.getExpressionContext(context, this.mediatorQueryOperation) };
+    const config = { ...ActorQueryOperation.getAsyncExpressionContext(context, this.mediatorQueryOperation) };
```
rubensworks, line 36: Is there a certain reason why this was changed?

We probably want the async context indeed, but I'm just wondering why it wasn't async in the first place.

packages/actor-query-operation-project/lib/ActorQueryOperationProject.ts
```diff
@@ -57,12 +58,15 @@ export class ActorQueryOperationProject extends ActorQueryOperationTypedMediated
     bindingsStream = bindingsStream.transform({
       map(bindings: Bindings) {
         blankNodeCounter++;
+        const scopedBlankNodesCache = new Map<string, RDF.BlankNode>();
         return <Bindings> bindings.map(term => {
-          if (term && term.termType === 'BlankNode') {
-            if (term instanceof BlankNodeScoped) {
-              return new BlankNodeScoped(`${term.value}${blankNodeCounter}`, term.skolemized);
```
rubensworks, line 63: Unless I'm missing something, we also need to keep `BlankNodeScoped`, right?

packages/bus-query-operation/lib/ActorQueryOperation.ts
```diff
@@ -11,7 +12,9 @@ import type {
   Bindings,
   PatternBindings,
 } from '@comunica/types';
+import type * as RDF from 'rdf-js';
 import type { Algebra } from 'sparqlalgebrajs';
+import * as uuid from 'uuid';
```
rubensworks, line 17: Can we use a simple counter to keep things more lightweight?

packages/data-factory/README.md
```diff
@@ -18,3 +18,4 @@ $ yarn add @comunica/data-factory
 ## Exposed classes
 
 * [`BlankNodeScoped`](https://comunica.github.io/comunica/classes/data_factory.blanknodescoped-1.html): A blank node that is scoped to a certain source, and exposes a skolemized named node.
+* [`BlankNodeScoped`](https://comunica.github.io/comunica/classes/data_factory.blanknodescoped-1.html): A blank node that is scoped to a certain set of bindings.
```
rubensworks, line 21: `BlankNodeBindingsScoped`?

packages/actor-query-operation-extend/lib/ActorQueryOperationExtend.ts
jacoscaz: sparqlee's typings specify two different types of contexts, one the sync evaluator and another one for the async evaluator. The former one expects synchronous `bnode()` and `exists()` functions while the latter one expects asynchronous versions of those functions. The `exists()` function provided by comunica is always asynchronous and thus can only be included in the asynchronous context, hence I tried to use the asynchronous context as much as possible so as not to introduce breaking changes.

packages/bus-query-operation/lib/ActorQueryOperation.ts
jacoscaz: `uuid` is already a dependency of `sparqlee` so we're not introducing additional dependencies here. That said, this might change one day so +1 on this, will do.

packages/data-factory/README.md
jacoscaz: Will do!

packages/actor-query-operation-construct/lib/BindingsToQuadsIterator.ts
jacoscaz: Mh... Right, I hadn't thought of that. Will do.

packages/actor-query-operation-project/lib/ActorQueryOperationProject.ts
jacoscaz: Why? A blank node scoped to a source is still a blank node that is not scoped to a single set of bindings. Modifying its label with a per-binding counter would break correlations, would it not?

packages/actor-query-operation-extend/lib/ActorQueryOperationExtend.ts
rubensworks: I think Comunica only makes use of the async context, or am I wrong there? If not, perhaps we can just have `getExpressionContext` always return an async context?

packages/actor-query-operation-project/lib/ActorQueryOperationProject.ts
rubensworks: We don't need to modify the label, we just need to ensure that we keep the skolemized term. Because we don't want to break cases like this that depend on these skolemized terms: https://github.com/solid/query-ldflex/issues/34

jacoscaz: Oooh my bad, I had misinterpreted the question. Will do!

packages/actor-query-operation-extend/lib/ActorQueryOperationExtend.ts
jacoscaz: That's what I thought, too, but then I found https://github.com/comunica/comunica/blob/7e77c0134b8354eb735ab2b805f8f6689f7c6b34/packages/actor-query-operation-group/lib/GroupsState.ts#L32  which uses the `SyncEvaluatorConfig` type.

rubensworks: Ah yes, aggregation happens sync. Then I guess there's no other way around this :-)

packages/actor-query-operation-project/lib/ActorQueryOperationProject.ts
jacoscaz: Doesn't the current version of the code already preserve `BlankNodeScoped`, though? All terms are returned as-is with the exception of `BlankNodeBindingsScoped` instances.

rubensworks: Ah yes, you're right, I didn't interpret the code correctly.
Then this change makes things even a bit cleaner :-)

packages/data-factory/README.md
jacoscaz: Done!

packages/bus-query-operation/lib/ActorQueryOperation.ts
jacoscaz: Done!

packages/actor-query-operation-construct/lib/BindingsToQuadsIterator.ts
jacoscaz: Done!

packages/actor-query-operation-construct/lib/BindingsToQuadsIterator.ts
```diff
@@ -1,3 +1,4 @@
+import { BlankNodeBindingsScoped } from '@comunica/data-factory';
```
rubensworks, line 1: This can be removed now?

packages/data-factory/lib/BlankNodeBindingsScoped.ts
```diff
@@ -0,0 +1,20 @@
...
+  public readonly termType: 'BlankNode' = 'BlankNode';
+  public readonly singleBindingsScope: true = true;
+  public readonly value: string;
+
+  public constructor(value: string) {
+    this.value = value;
+  }
+
+  public equals(other: RDF.Term | null | undefined): boolean {
+    // TODO: should we always return false given the nature of this class?
```
rubensworks, line 16: I would say no. I think the current implementation is fine.
Even if we always return false here, invoking `equals` from another bnode would still return true, which could cause more problems than we would fix.

packages/actor-query-operation-construct/lib/BindingsToQuadsIterator.ts
jacoscaz: Done!

packages/data-factory/lib/BlankNodeBindingsScoped.ts
jacoscaz: Done!

rubensworks:
Super, thanks for all your work on this!

jacoscaz:
My pleasure! Do you think it'd be possible to publish a new version of Comunica soon-ish? Even a beta one, doesn't have to be an official release.

PR #806, opened 2021-04-09
Add hypermedia-based update bus and Solid server actor

rubensworks:
This adds a new hypermedia-based update bus, similar to the existing hypermedia-based quad pattern resolving bus.

This also adds one actor to this bus that handles servers that accept (simple) SPARQL Update queries as PATCH requests (such as Solid servers). This actor requires another metadata extraction actor that checks whether or not the server exposes a valid `Accept-Patch` header.

@jeswr Since you've expressed interest in implementing additional update actors, feel free to let me know what you think of this new bus.
The SPARQL endpoint actor will probably become a new actor on this new hypermedia-based update bus, and will probably require a minor addition to `actor-rdf-metadata-extract-sparql-service` check for the update-capability in the service description.
I think the file-based actor will be best suited on the regular update quads bus (so not using hypermedia).


packages/bus-rdf-update-quads/lib/IQuadDestination.ts
```diff
@@ -0,0 +1,42 @@
...
+    dropGraphs: boolean,
+  ) => Promise<void>;
+  /**
+   * Create the given (empty) graph.
+   * @param graph The graph name to create.
+   * @param requireNonExistence If true, an error MUST be thrown when the graph already exists.
+   *                            For destinations that do not record empty graphs,
+   *                            this should only throw if at least one quad with the given quad already exists.
+   */
+  createGraph: (graph: RDF.NamedNode, requireNonExistence: boolean) => Promise<void>;
```
jeswr, line 41: It may be useful to enable the creation and deletion multiple graphs at once in the `IQuadDestination` interface with the type becoming something like:

```ts
deleteGraphs: (
    graphs: (RDF.DefaultGraph | 'NAMED' | 'ALL' | RDF.NamedNode)[] | RDF.DefaultGraph | 'NAMED' | 'ALL' | RDF.NamedNode,
    requireExistence: boolean,
    dropGraphs: boolean,
  ) => Promise<void>;
  /**
   * Create the given (empty) graph.
   * @param graph The graph name to create.
   * @param requireNonExistence If true, an error MUST be thrown when the graph already exists.
   *                            For destinations that do not record empty graphs,
   *                            this should only throw if at least one quad with the given quad already exists.
   */
  createGraph: (graph: RDF.NamedNode[] | RDF.NamedNode, requireNonExistence: boolean) => Promise<void>;
```

I realise that creation/deletion of multiple graphs is currently not enabled by SPARQL `clear`/`drop` operations - but it seems like something that would be reasonably straightforward to implement for most sources (the case of SPARQL endpoints may just require multiple queries to be issued).

Reasons to do this would be:
 - It would make it easier to update comunica if the SPARQL standard ever changed to allow such querying behavior.
 - Gives dev's more flexibility if they want to use the internal API's of comunica.

jeswr:
I've left my 2 cents worth on the `bus-rdf-update-quads`. I don't have anything to add for the hypermedia based update bus.

P.S. I'm in the last semester of my degree & overloading. I'm on mid-semester break at the moment, hence the offer to work on the file and sparql-endpoint actors, but since that soon ends I'm unsure if I'll have time to make too much of a contribution.

packages/bus-rdf-update-quads/lib/IQuadDestination.ts
rubensworks: Good idea, will look into that.

rubensworks:
@jeswr Thanks for the feedback!

> P.S. I'm in the last semester of my degree & overloading. I'm on mid-semester break at the moment, hence the offer to work on the file and sparql-endpoint actors, but since that soon ends I'm unsure if I'll have time to make too much of a contribution.

No worries, there's no obligation at all :-)
There's no urgency for file and endpoint support on my end, but I may eventually implement support for it should you not be able to get to it.
I'll probably look into other update-related features first, such as exposing the graph store protocol.

packages/bus-rdf-update-quads/lib/IQuadDestination.ts
rubensworks: Done in https://github.com/comunica/comunica/commit/e695c5e468c402e1cd6ade35254c14a92963ba43

PR #808, opened 2021-04-11
Fixes missing context for actors called by MediatorCombinePipeline

jacoscaz:
While working on https://github.com/beautifulinteractions/node-quadstore/issues/115 I noticed that my custom optimizer actor had no `context` entry in its `action` argument. I tracked the issue to the fact that, while `MediatorCombinePipeline` does pass the output of one actor to the next one, the output of each actor is not guaranteed to have the `context` property in it and therefore only the first actor in a chain of mediated actors is guaranteed to receive the context.

This fix is backward-compatible and ensures that the properties of the original action are always preserved unless they are overwritten by one of the actors mediated over.  

Furthermore, this fix also modifies how the return value of the optimize mediator is handled. If the mediator returns a context, then the original context is replaced with the mediator's; if the mediator doesn't return a context, then the original context is used. This allows optimize actors to use the context to propagate information to other actors in Comunica.

### DIFF
--- a/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
+++ b/packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
@@ -199,7 +199,9 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs, 
     }
 
     // Optimize the query operation
-    operation = (await this.mediatorOptimizeQueryOperation.mediate({ context, operation })).operation;
+    const mediatorResult = await this.mediatorOptimizeQueryOperation.mediate({ context, operation });
+    operation = mediatorResult.operation;
+    context = mediatorResult.context || context;
 
     // Pre-processing the context, this time with the operation
     context = (await this.mediatorContextPreprocess.mediate({ context, operation })).context;
--- a/packages/actor-init-sparql/test/ActorInitSparql-test.ts
+++ b/packages/actor-init-sparql/test/ActorInitSparql-test.ts
@@ -623,21 +623,15 @@ describe('ActorInitSparql', () => {
       });
 
       it('should allow optimize actors to modify the context', async() => {
-        const optimizeMediate = mediatorOptimizeQueryOperation.mediate;
         mediatorOptimizeQueryOperation.mediate = (action: any) => {
-          action = {
+          return Promise.resolve({
             ...action,
             context: action.context.set('the-answer', 42),
-          };
-          return optimizeMediate(action);
+          });
         };
-        const queryMediate = mediatorQueryOperation.mediate;
-        mediatorQueryOperation.mediate = (action: any) => {
-          expect(action.context.get('the-answer')).toBe(42);
-          return queryMediate(action);
-        };
-        await expect(actor.query('SELECT * WHERE { ?s ?p ?o }'))
-          .resolves.toBeTruthy();
+        const result = await actor.query('SELECT * WHERE { ?s ?p ?o }');
+        expect(result).toHaveProperty('context');
+        expect(result.context!.get('the-answer')).toEqual(42);
       });
     });
 
--- a/packages/actor-init-sparql/test/ActorInitSparql-test.ts
+++ b/packages/actor-init-sparql/test/ActorInitSparql-test.ts
@@ -622,6 +622,15 @@ describe('ActorInitSparql', () => {
           });
       });
 
+      it('should pass down the provided context if optimize actors do not return one', async() => {
+        mediatorOptimizeQueryOperation.mediate = (action: any) => {
+          return Promise.resolve({ ...action, context: undefined });
+        };
+        const result = await actor.query('SELECT * WHERE { ?s ?p ?o }', { 'the-answer': 42 });
+        expect(result).toHaveProperty('context');
+        expect(result.context!.get('the-answer')).toEqual(42);
+      });
+
       it('should allow optimize actors to modify the context', async() => {
         mediatorOptimizeQueryOperation.mediate = (action: any) => {
           return Promise.resolve({
--- a/packages/actor-init-sparql/test/ActorInitSparql-test.ts
+++ b/packages/actor-init-sparql/test/ActorInitSparql-test.ts
@@ -621,6 +621,24 @@ describe('ActorInitSparql', () => {
             '@comunica/actor-init-sparql:baseIRI': 'myBaseIRI',
           });
       });
+
+      it('should allow optimize actors to modify the context', async() => {
+        const optimizeMediate = mediatorOptimizeQueryOperation.mediate;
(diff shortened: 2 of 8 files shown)


packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
```diff
@@ -199,7 +199,9 @@ export class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs,
     }
 
     // Optimize the query operation
-    operation = (await this.mediatorOptimizeQueryOperation.mediate({ context, operation })).operation;
+    const mediatorResult = await this.mediatorOptimizeQueryOperation.mediate({ context, operation });
+    operation = mediatorResult.operation;
+    context = mediatorResult.context || context;
```
jacoscaz, line 204: I don't understand why coverage doesn't flag this line as tested. The test I've added would not work without this.

packages/core/lib/Actor.ts
```diff
@@ -222,5 +222,5 @@ export interface IActorTest {
  * Data interface for the type of an actor run result.
  */
 export interface IActorOutput {
-
+  context?: ActionContext;
```
rubensworks, line 225: Instead of adding this here, could we add it to the output interface of the optimization bus?
As long as it's an optional field, this should still be backwards-compatible.

packages/mediator-combine-pipeline/lib/MediatorCombinePipeline.ts
```diff
@@ -27,7 +27,7 @@ export class MediatorCombinePipeline<A extends Actor<H, T, H>, H extends IAction
     // and each actor output as input to the following actor.
     let handle: H = action;
     for (const actor of testResults.map(result => result.actor)) {
-      handle = await actor.runObservable(handle);
+      handle = { ...handle, ...await actor.runObservable(handle) };
```
rubensworks, line 30: Ah yes, this will be needed now.
Could we add a unit tests in `mediator-combine-pipeline` that tests this new behaviour?

packages/actor-init-sparql/test/ActorInitSparql-test.ts
```diff
@@ -621,6 +621,24 @@ describe('ActorInitSparql', () => {
             '@comunica/actor-init-sparql:baseIRI': 'myBaseIRI',
           });
       });
+
+      it('should allow optimize actors to modify the context', async() => {
+        const optimizeMediate = mediatorOptimizeQueryOperation.mediate;
```
rubensworks, line 626: I would just copy-paste this method impl instead of extending it like this, to make the test more self-contained (not clear what this method does when just looking at this test).

packages/actor-init-sparql/test/ActorInitSparql-test.ts
```diff
@@ -621,6 +621,24 @@ describe('ActorInitSparql', () => {
...
+        mediatorOptimizeQueryOperation.mediate = (action: any) => {
+          action = {
+            ...action,
+            context: action.context.set('the-answer', 42),
+          };
+          return optimizeMediate(action);
+        };
+        const queryMediate = mediatorQueryOperation.mediate;
```
rubensworks, line 634: Same as above.

packages/actor-init-sparql/test/ActorInitSparql-test.ts
```diff
@@ -621,6 +621,24 @@ describe('ActorInitSparql', () => {
...
+        mediatorQueryOperation.mediate = (action: any) => {
+          expect(action.context.get('the-answer')).toBe(42);
```
rubensworks, line 636: For test safety, I would move this `expect` to after `actor.query`. Because if this method is not called, then this assertion will never be tested, and the test may not show up as a failure.

packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
rubensworks: Not sure myself, but perhaps my suggested changes to the test could help finding the cause.

rubensworks (changes requested):

rubensworks:
I like this change!
It's actually a much cleaner solution to a similar problem I had before, and for which I introduced [this](https://github.com/comunica/comunica/commit/5685224235fdda8cd106c70ae984ebfd6748791f). I'll probably just revert my change once this is merged (since it was not part of a release yet).

jacoscaz:
Ah, good! Happy this will benefit use cases other than mine alone. I did wonder about that repeated use of the pre-processing mediator... Thanks for reviewing, will work on this later today and tomorrow.

packages/mediator-combine-pipeline/lib/MediatorCombinePipeline.ts
jacoscaz: Done!

packages/core/lib/Actor.ts
jacoscaz: Done!

packages/actor-init-sparql/test/ActorInitSparql-test.ts
jacoscaz: Done!

packages/actor-init-sparql/test/ActorInitSparql-test.ts
jacoscaz: Done!

packages/actor-init-sparql/test/ActorInitSparql-test.ts
jacoscaz: Done! And while doing this, I realized I could rewrite the test in a much simpler way.

packages/actor-init-sparql/lib/ActorInitSparql-browser.ts
jacoscaz: Done! Added a dedicated unit tests for the left side of  `||`.

packages/bus-optimize-query-operation/lib/ActorOptimizeQueryOperation.ts
```diff
@@ -1,5 +1,6 @@
 import type { IAction, IActorArgs, IActorOutput, IActorTest } from '@comunica/core';
 import { Actor } from '@comunica/core';
+import type { ActionContext } from '@comunica/types';
```
rubensworks, line 3: Could you also add `@comunica/types` as dependency to this package?

jacoscaz: Done!

rubensworks:
Super, thanks!

rubensworks:
Released prerelease 1.20.1-alpha.2.0 (still publishing at the time of writing)

jacoscaz:
Oh nice! Thank you!

Issue #732, opened 2020-09-15
Emit provenance: explaining query results


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request

____

This issue describes a high-level plan to implement provenance generation in Comunica to explain SPARQL query results.

## Different levels of provenance

* High-level: On the query result object
	* SPARQL query
	* Optimized query plan
	* Used ActionContext (containing sources)
	* The used Comunica config (IRI)
* Low-level: Annotation on each separate query result in the bindings stream.
	* Tree of SPARQL query operators

## Example

The `Bindings` in the resulting bindings stream can be annotated with a `'__prov'` entry,
which contains provenance information from the actor that produced this bindings object.

For example:
```
{
  ?name:"Ruben",
  __prov:{
    actor:"actor-query-operation-project",
    variables:[
      "?name"
    ],
    bindings:{
      ?name:"Ruben",
      ?country:"Belgium",
      __prov:{
        actor:"actor-query-operation-bgp",
        join:[
          bindings:{
            ?name:"Ruben",
            __prov:{
              actor:"actor-query-operation-quad-pattern",
              quad:"<#me> foaf:name \"Ruben\"",
              resolvedFrom:{
                actor:"actor-rdf-resolve-quad-pattern-hypermedia",
                metadata:{},
                url:"https://www.rubensworks.net/",
                contentType:"text/html",
                responseHeaders:{
                  
                }
              }
            }
          },
          bindings:{
            ?name:"Belgium",
            __prov:{
              actor:"actor-query-operation-quad-pattern",
              quad:"<#me> foaf:country \"Ruben\"",
              resolvedFrom:{
                actor:"actor-rdf-resolve-quad-pattern-hypermedia",
                metadata:{ next:"..." },
                url:"http://fragments.dbpedia.org/2016-04/en",
                contentType:"text/turtle",
                responseHeaders:{
                  
                }
              }
            }
          }
        ]
      }
    }
  }
}
```

_The `BindingsStream` will in practise emit many of these `Bindings` objects, so this will definitely lead to some overhead._

## Implementation notes

* Only emit provenance if `GENERATE_PROVENANCE` flag is enabled in context
* New bus: bus-query-operation-provenance
    * Actors on this bus can listen to certain types of query operation, and append provenance information.
    * For example actor-query-operation-quad-pattern
	* Not all query operations necessarily require a corresponding provenance actor.
* New actor: actor-query-operation-annotate-provenance
[...]

PR #821, opened 2021-05-12
suggestion: include provenance in FederatedQuadSource

gertjandemulder:
This PR serves as a starting point towards a more generic solution for including provenance data from federated sources.

> Note: tests are not updated and fail due to the suggested changes

Related to #732

### DIFF
--- a/packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
+++ b/packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
@@ -172,7 +172,7 @@ export class FederatedQuadSource implements IQuadSource {
 
   public match(subject: RDF.Term, predicate: RDF.Term, object: RDF.Term, graph: RDF.Term): AsyncIterator<RDF.Quad> {
     // Counters for our metadata
-    const metadata: Record<string, any> = { totalItems: 0 };
+    const metadata: Record<string, any> = { totalItems: 0, prov: new Array<Record<string, any>>() };
     let remainingSources = this.sources.length;
 
     // Anonymous function to handle totalItems from metadata
@@ -223,6 +223,12 @@ export class FederatedQuadSource implements IQuadSource {
 
       // Handle the metadata from this source
       output.data.getProperty('metadata', (subMetadata: Record<string, any>) => {
+        if (subMetadata.prov) {
+          const prov = subMetadata.totalItems ?
+              { ...subMetadata.prov, totalItems: subMetadata.totalItems } : subMetadata.prov;
+          metadata.prov.push(prov);
+        }
+
         if ((!subMetadata.totalItems && subMetadata.totalItems !== 0) || !Number.isFinite(subMetadata.totalItems)) {
           // We're already at infinite, so ignore any later metadata
           metadata.totalItems = Number.POSITIVE_INFINITY;


rubensworks:
Could we think of a more generic approach to this?
Because now this is hardcoded to `prov`, which makes our actors too tightly coupled.

Perhaps some kind of `bus-rdf-metadata-aggregate` bus might be valuable here, with an `actor-rdf-metadata-aggregate-prov` (and perhaps also a `actor-rdf-metadata-aggregate-totalitems`)?
This bus would then be responsible for combining the metadata of one source, into the aggregated metadata.

gertjandemulder:
> Could we think of a more generic approach to this?
> Because now this is hardcoded to `prov`, which makes our actors too tightly coupled.
> 
> Perhaps some kind of `bus-rdf-metadata-aggregate` bus might be valuable here, with an `actor-rdf-metadata-aggregate-prov` (and perhaps also a `actor-rdf-metadata-aggregate-totalitems`)?
> This bus would then be responsible for combining the metadata of one source, into the aggregated metadata.

@rubensworks , something like the following diagram ? 
mermaid-diagram-20210514133044

Actors on the RDF Metadata Aggregate bus (`bus-rdf-metadata-aggregate`) should be executed prior to the Federated actor, right?

rubensworks:
@gertjandemulder Almost; I would place the RDF Metadata Aggregate bus underneath the Federated actor. Because the Federated actor will have to invoke it (multiple times).

Concretely, the stuff on line 226 in your PR will then have to be replaced by a call to a mediator on this RDF Metadata Aggregate bus.

gertjandemulder:
## New components & configurations
- created & configured the RDF Metadata Aggregate bus 
- created & configured the RDF Metadata Aggregate Total Items actor
- added a new configuration set (`rdf-metadata-aggregate.json`) to `actor-init-sparql/config/sets/`

## Implementation
The Federated Actor calls the RDF Metadata Aggregate bus for every source, and it passes on the corresponding quad pattern output (because this output contains the data iterator, which contains the (sub)metadata ).
Each time the RDF Metadata Aggregate Total Items actor is run, it will collect the submetadata from the data iterator for that source,  and returns a promise that resolves to the aggregated total items as well as the original records on which was aggregated.

Given the following query, for example: 
```SPARQL
PREFIX foaf: <http://xmlns.com/foaf/0.1/>

SELECT ?s ?interest
WHERE {
  ?s foaf:topic_interest ?interest.
}
LIMIT 10
```

The metadata of the query result would look like: 
```
{
"totalItems": 10,
"https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-init-sparql/^1.0.0/config/sets/rdf-metadata-aggregate.json#myActorRdfMetadataAggregateTotalItems": {
  "totalItems": 12,
  "records": [
    {
      "totalItems": 6,
      "resolvedFrom": "https://ruben.verborgh.org/profile/"
    },
    {
      "totalItems": 6,
      "resolvedFrom": "https://www.rubensworks.net/"
    }
  ]
}
```

@rubensworks , what are your thoughts on this?

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -216,24 +212,21 @@ export class FederatedQuadSource implements IQuadSource {
         this.isSourceEmpty(source, pattern = this.algebraFactory
           .createPattern(patternS, patternP, patternO, patternG))) {
         output = { data: new ArrayIterator([], { autoStart: false }) };
-        output.data.setProperty('metadata', { totalItems: 0 });
```
rubensworks, line 219: This will break things, we should somehow still invoke the bus, in order to obtain `totalItems: 0`.

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -216,24 +212,21 @@ export class FederatedQuadSource implements IQuadSource {
...
-          remainingSources = 0;
-          checkEmitMetadata(Number.POSITIVE_INFINITY, source, pattern, subMetadata);
-        } else {
-          metadata.totalItems += subMetadata.totalItems;
-          remainingSources--;
-          checkEmitMetadata(subMetadata.totalItems, source, pattern, subMetadata);
-        }
-      });
+      // Call RdfMetadataAggregate bus
+      const metadataPromise = this.mediatorRdfMetadataAggregate
```
rubensworks, line 221: This bus should be fast, so I would just `await` it immediately here.

packages/actor-rdf-metadata-aggregate-total-items/lib/ActorRdfMetadataAggregateTotalItems.ts
```diff
@@ -0,0 +1,56 @@
+import { ActorRdfMetadataAggregate, IActionRdfMetadataAggregate, IActorRdfMetadataAggregateOutput } from '@comunica/bus-rdf-metadata-aggregate';
+import { IActorArgs, IActorTest } from '@comunica/core';
+
+/**
+ * A comunica Total Items RDF Metadata Aggregate Actor.
+ */
+export class ActorRdfMetadataAggregateTotalItems extends ActorRdfMetadataAggregate {
+  public records: Record<string,any>[] = [];
```
rubensworks, line 8: Actors should be stateless, so this shouldn't be necessary.

rubensworks (changes requested):
The interlinking of new actors and buses looks good!

The metadata of the query result does however look a bit strange to me, can we just keep it as `{ "totalItems": 10 }`?
It's possible that you want to accumulate more in the provenance actor, but that should require changes in the totalItems actor.

gertjandemulder:
@rubensworks , I updated the way that the metadata is being aggregated in the Federated Quad Source, and removed the state being collected in the RDF Metadata Aggregate Total Items actor. Checking for infinity is moved to the latter actor. 
Feedback is welcome :) !

rubensworks (changes requested):
Looks good!
In any case, before we merge this, it would be good if you could first finalize and test the prov metadata aggregate actor, so we're certain that this approach is good for it.

Just some minor comments.

Also, the exec flags for `.github/deploy_browser.sh ` seem to be changed, which should not be necessary.

I also noticed that CI is not running for (only) your PR. Are you pushing in a special way by any chance?

gertjandemulder:
@rubensworks , I made the following updates:

## bus-rdf-metadata-aggregate/lib/ActorRdfMetadataAggregate.ts

For the aggregation of metadata, the `IActionRdfMetadataAggregate` and `IActorRdfMetadataAggregateOutput` are defined as follows

```typescript
export interface IActionRdfMetadataAggregate extends IAction {
  metadata: Record<string, any>;
  subMetadata?: Record<string, any>;
}

export interface IActorRdfMetadataAggregateOutput extends IActorOutput {
  aggregatedMetadata: Record<string, any>;
}
```

The `IActorRdfMetadataAggregateOutput` returned contains `aggregatedMetadata`. How and which parts of the `metadata` is aggregated, is up to the Rdf Metadata Aggregate actor.

## packages/actor-rdf-metadata-aggregate-total-items

This `ActorRdfMetadataAggregateTotalItems` actor subscribes to the `ActorRdfMetadataAggregate`-bus.
The actor looks for `totalItems` in the `metadata` and `subMetadata` , and aggregates by summing them up.
Also the cases for when to return infinite total items are handled.

### Tests

I have added some tests to verify whether the `totalItems` are correctly handled.

The test results are:

```
----------------------------------------|---------|----------|---------|---------|-------------------
File                                    | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
----------------------------------------|---------|----------|---------|---------|-------------------
All files                               |   81.48 |       75 |      80 |      84 |                   
 ActorRdfMetadataAggregateTotalItems.ts |   81.48 |       75 |      80 |      84 | 29,47-50          
----------------------------------------|---------|----------|---------|---------|-------------------
Test Suites: 1 passed, 1 total
Tests:       6 passed, 6 total
Snapshots:   0 total
Time:        0.838 s, estimated 1 s
Ran all test suites.
Done in 1.29s.
```

## packages/actor-rdf-resolve-quad-pattern-federated

The following changes were made to FederatedQuadSource

The `match` function 

1. Maps the sources to so-called `metadataCallbacks: Promise<Record<string, any> | undefined>[]` . These `metadataCallbacks`  hook into every source-output and resolve to its metadata (if any).
2. Reduces the `metadataCallbacks` by calling the `mediatorAggregate` to aggregating the metadata from each source.

### Tests

I created a `MediatorRdfMetadataAggregate`-mock that will is used by both 

- `ActorRdfResolveQuadPatternFederated-test.ts`
- `FederatedQuadSource-test.ts`

However, for some tests, I changed from `.toEqual` to `.toMatchObject` because the test would fail if the metadata contained additional fields (other than `totalItems`). I assume these additional fields (fields that could come from other aggregation actors, for example) should actually be allowed? 

```
----------------------------------------|---------|----------|---------|---------|-------------------
File                                    | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
----------------------------------------|---------|----------|---------|---------|-------------------
All files                               |   97.71 |    98.61 |   96.67 |   97.62 |                   
 ActorRdfResolveQuadPatternFederated.ts |     100 |      100 |     100 |     100 |                   
 FederatedQuadSource.ts                 |   97.52 |    98.57 |    96.3 |   97.41 | 263,268,305       
----------------------------------------|---------|----------|---------|---------|-------------------

Test Suites: 2 passed, 2 total
Tests:       103 passed, 103 total
Snapshots:   0 total
Time:        1.324 s, estimated 5 s
Ran all test suites.
```

packages/actor-init-sparql/config/config-default.json
```diff
@@ -15,6 +15,8 @@
     "files-cais:config/sets/rdf-update-hypermedia.json",
     "files-cais:config/sets/rdf-update-rdfjs.json",
     "files-cais:config/sets/resolve-federated.json",
+    "files-cais:config/sets/rdf-metadata-aggregate.json",
+
```
rubensworks, line 19: Unnecessary newline.

packages/actor-rdf-metadata-aggregate-total-items/README.md
```diff
@@ -0,0 +1,39 @@
...
+```text
+{
+  "@context": [
+    ...
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-rdf-metadata-aggregate-total-items/^1.0.0/components/context.jsonld"  
+  ],
+  "actors": [
+    ...
+    {
+      "@id": TODO,
```
rubensworks, line 30: Some TODO's to fill in here.

packages/actor-rdf-metadata-aggregate-total-items/lib/ActorRdfMetadataAggregateTotalItems.ts
rubensworks: Ping

packages/actor-rdf-metadata-aggregate-total-items/lib/ActorRdfMetadataAggregateTotalItems.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  public async test(action: IActionRdfMetadataAggregate): Promise<IActorTest> {
+    return true;
+  }
+
+  public async run(action: IActionRdfMetadataAggregate): Promise<IActorRdfMetadataAggregateOutput> {
+    const {metadata, subMetadata} = action
+
+    let newTotalItems: number = 0
+    const isEmptyRecord = (r: Record<string, any>) => Object.keys(r).length === 0
```
rubensworks, line 23: Due to the simplicity of these functions, I would just inline them.

packages/actor-rdf-metadata-aggregate-total-items/lib/ActorRdfMetadataAggregateTotalItems.ts
```diff
@@ -0,0 +1,69 @@
...
+      const subMetadataHasTotalItems = hasTotalItems(subMetadata)
+
+
+      if (metadataHasTotalItems && subMetadataHasTotalItems) {
+        // metadata has total items & submetadata has total items
+        newTotalItems = metadata.totalItems + subMetadata.totalItems
+      } else if (!metadataHasTotalItems && !subMetadataHasTotalItems) {
+        // neither metadata or submetadata has totalItems
+        newTotalItems = Number.POSITIVE_INFINITY
+      } else if (!metadataHasTotalItems && subMetadata) {
```
rubensworks, line 42:
```suggestion
      } else if (!metadataHasTotalItems && subMetadataHasTotalItems) {
```

packages/actor-rdf-metadata-aggregate-total-items/lib/ActorRdfMetadataAggregateTotalItems.ts
```diff
@@ -0,0 +1,69 @@
...
+  public constructor(args: IActorArgs<IActionRdfMetadataAggregate, IActorTest, IActorRdfMetadataAggregateOutput>) {
+    super(args);
+  }
+
+  public async test(action: IActionRdfMetadataAggregate): Promise<IActorTest> {
+    return true;
+  }
+
+  public async run(action: IActionRdfMetadataAggregate): Promise<IActorRdfMetadataAggregateOutput> {
+    const {metadata, subMetadata} = action
```
rubensworks, line 20: Did you somehow disable the linter? Because I would expect the linter to complain about this line (and others).

packages/actor-rdf-metadata-aggregate-total-items/lib/ActorRdfMetadataAggregateTotalItems.ts
```diff
@@ -0,0 +1,69 @@
...
+      // submetadata not defined
+      resultRecord = metadata
+    } else {
+      // submetadata IS defined
+      const metadataHasTotalItems = hasTotalItems(metadata)
+      const subMetadataHasTotalItems = hasTotalItems(subMetadata)
+
+
+      if (metadataHasTotalItems && subMetadataHasTotalItems) {
+        // metadata has total items & submetadata has total items
```
rubensworks, line 37: These comments can go, the code is sufficiently self-explanatory.

packages/actor-rdf-metadata-aggregate-total-items/lib/ActorRdfMetadataAggregateTotalItems.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  public async run(action: IActionRdfMetadataAggregate): Promise<IActorRdfMetadataAggregateOutput> {
+    const {metadata, subMetadata} = action
+
+    let newTotalItems: number = 0
+    const isEmptyRecord = (r: Record<string, any>) => Object.keys(r).length === 0
+    const hasTotalItems = (r: Record<string, any>) => r.totalItems !== undefined
+    let resultRecord = {}
+
+    if (!subMetadata) {
```
rubensworks, line 27: Can you double-check if the semantics of this actor are still the same as https://github.com/comunica/comunica/blob/master/packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts#L226-L235 ?

Because I suspect there will be some difference. E.g. because `Number.isFinite` is never checked in here.

packages/actor-rdf-metadata-aggregate-total-items/lib/ActorRdfMetadataAggregateTotalItems.ts
```diff
@@ -0,0 +1,69 @@
...
+        newTotalItems = Number.POSITIVE_INFINITY
+      } else if (!metadataHasTotalItems && subMetadata) {
+        // metadata does not have total items
+        newTotalItems = subMetadata.totalItems
+
+      } else {
+        if (isEmptyRecord(subMetadata)) {
+          newTotalItems = Number.POSITIVE_INFINITY;
+        } else {
+          throw  Error('CASE NOT COVERED YET')
```
rubensworks, line 50: Can this ever happen? If so, we may want to handle this, or produce a more descriptive error.

packages/actor-rdf-metadata-aggregate-total-items/lib/ActorRdfMetadataAggregateTotalItems.ts
```diff
@@ -0,0 +1,69 @@
...
+    resultRecord = {
+      ...subMetadata,
+      ...metadata,
+      totalItems: newTotalItems
+    }
+
+    return Promise.resolve(
+        {
+          aggregatedMetadata: {
+            ...resultRecord
```
rubensworks, line 64: No need to copy again.

packages/actor-rdf-resolve-quad-pattern-federated/components/Actor/RdfResolveQuadPattern/Federated.jsonld
```diff
@@ -24,7 +24,14 @@
           "range": "xsd:boolean",
           "default": false,
           "unique": true
+        },
+        {
+          "@id": "carrqpf:Actor/RdfResolveQuadPattern/Federated/mediatorRdfMetadataAggregate",
+          "comment": "todo",
```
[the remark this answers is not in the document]
rubensworks: Ping

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -12,6 +12,12 @@ import { mapTerms } from 'rdf-terms';
 import type { Algebra } from 'sparqlalgebrajs';
 import { Factory } from 'sparqlalgebrajs';
 
+import type {
+  IActionRdfMetadataAggregate,
+  IActorRdfMetadataAggregateOutput
+} from '@comunica/bus-rdf-metadata-aggregate';
+import {Record} from "immutable";
```
rubensworks, line 19: This seems incorrect. We should be using `Record` from Node, so just removing this line is fine.

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -129,6 +144,24 @@ export class FederatedQuadSource implements IQuadSource {
     return term;
   }
 
+  static deskolemizeQuad(s: RDF.Term,
```
rubensworks, line 147: Why is this new function needed?

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -170,30 +203,111 @@ export class FederatedQuadSource implements IQuadSource {
     return sourceId;
   }
 
+
+
+  checkPushEmptyPattern = (md: Record<string,any> | undefined, source: IDataSource,
```
rubensworks, line 208:
```suggestion
  checkPushEmptyPattern = (metadata: Record<string,any> | undefined, source: IDataSource,
```

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -170,30 +203,111 @@ export class FederatedQuadSource implements IQuadSource {
...
-      }
-      if (!remainingSources) {
-        if (lastMetadata && this.sources.length === 1) {
-          // If we only had one source, emit the metadata as-is.
-          it.setProperty('metadata', lastMetadata);
-        } else {
-          it.setProperty('metadata', metadata);
-        }
+
+    const metadataCallbacks =
```
rubensworks, line 217: Not sure I understand what is going on here, but I suspect you may be overcomplicating things.

If I understand your code correctly, your iterating over the sources a second time, and invoking `mediatorResolveQuadPattern` manually again to obtain the metadata.
This is not something we want to do, because we're otherwise calling expensive querying code twice, thereby making queries twice as slow, which is not what we want.

I would suggest analyzing the original code again wrt metadata handling, and modifying the parts where `totalItems` are hardcoded to invoke the new metadata-aggregate bus instead.

packages/actor-rdf-resolve-quad-pattern-federated/package.json
```diff
@@ -62,7 +62,10 @@
       "ts",
       "js"
     ],
-    "collectCoverage": true
+    "collectCoverage": true,
+    "testPathIgnorePatterns": [
```
rubensworks, line 66: This should not be necessary.

packages/actor-rdf-resolve-quad-pattern-federated/test/__mocks__/MediatorRdfMetadataAggregateMock.ts
```diff
@@ -0,0 +1,55 @@
+import {Record} from "immutable";
+
+export const mockedMediatorAgg = {
```
rubensworks, line 3: Not need to make this a separate file, just add it in the test file.

Also, we can probably make a much more simplified version of this mediator instead of copy-pasting the existing totalItems actor.

packages/bus-rdf-metadata-aggregate/lib/ActorRdfMetadataAggregate.ts
```diff
@@ -0,0 +1,33 @@
+import { Actor, IAction, IActorArgs, IActorOutput, IActorTest } from '@comunica/core';
+import {IDataSource} from "@comunica/bus-rdf-resolve-quad-pattern";
+
+/**
+ * A comunica actor for rdf-metadata-aggregate events.
+ *
+ * Actor types:
+ * * Input:  IActionRdfMetadataAggregate:      TODO: fill in.
```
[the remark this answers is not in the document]
rubensworks: Ping

rubensworks (changes requested):
I have the feeling you're overcomplicating some things, which will lead to a massive impact on performance.
See my comments inline.

Could you also try rebasing on master? That may be the reason why CI is not running on this PR.
Also make sure to run the pre-commit checks.

packages/actor-rdf-resolve-quad-pattern-federated/test/__mocks__/MediatorRdfMetadataAggregateMock.ts
gertjandemulder: The mocked RDF Metadata Aggregate mediator (`mockedMediatorAgg`) is used by both `FederatedQuadSource-test.ts` and `ActorRdfResolveQuadPatternFederated-test.ts`, so that would imply duplicating that code twice?

packages/actor-rdf-resolve-quad-pattern-federated/package.json
gertjandemulder: If we keep the mocked RDF Metadata Aggregate mediator in a separate file (cfr. [this comment](https://github.com/comunica/comunica/pull/821#discussion_r692218718)), the mock-folder is likely to remain in the ignored testpaths?

gertjandemulder:
@rubensworks , I made the following changes:

- Worked away occurrences of `totalItems` in `FederatedQuadSource`, under the assumption that setting the `totalItems` to zero should now be the responsibility of `ActorRdfMetadataAggregateTotalItems`.
- Fixed coverage.
- Added code documentation.
- Updated readme’s.
- Updated configuration-files.

For now, the RDF Metadata Aggregate mock is still in a separate file to avoid duplicating the logic in both test-files `FederatedQuadSource-test.ts` and `ActorRdfResolveQuadPatternFederated-test.ts`, as mentioned in [this comment](https://github.com/comunica/comunica/pull/821#discussion_r692218718).

Feedback is welcome!

PR #837, opened 2021-07-06
Do not reset blank-node counter (closes #836)

### LINKED ISSUE #836: Blank-node counter gets reset
#### Issue type:

- :bug: Bug 

____
#### Description:

When repeating a construct query, all the constructed blank-node IDs are the same across queries, because `ActorSparqlParseAlgebra` resets a global blank-node counter every time Comunica runs a query:

- `ActorSparqlParseAlgebra` resets the blank node counter here: https://github.com/comunica/comunica/blob/master/packages/actor-sparql-parse-algebra/lib/ActorSparqlParseAlgebra.ts#L29
- by calling this function: https://github.com/RubenVerborgh/SPARQL.js/blob/main/lib/sparql.jison#L175,
- which modifies a global blank node counter.

I.e., each time we execute a query with Comunica, we reset this counter for sparqljs, affecting all other places (possibly unrelated to Comunica) where sparqljs is used.

Would it be possible to remove the call to `_resetBlanks()`, as the comment seems to indicate that it is merely for presentation purposes.

____
#### Environment:

| Comunica Init Actor | 1.21.3 
| node                | v12.22.1
| npm                 | 6.14.12
| yarn                | 1.22.5
| Operating System    | linux (Linux 5.8.0-55-generic)

MPvHarmelen:


rubensworks:
Super, thanks!

PR #841, opened 2021-07-20
Report FILTER errors as warnings in the logger (closes #767)

### LINKED ISSUE #767: Report FILTER errors as warnings in the logger
#### Issue type:

____
#### Description:

Queries such as the following (e.g. on DBpedia) will skip results for `?o` with IRIs, as these cause an internal casting error, which must be ignored according to the SPARQL 1.1 spec:

```sparql
    SELECT ?o WHERE {
      <http://dbpedia.org/resource/Landscaping> <http://www.w3.org/2002/07/owl#sameAs> ?o .
      FILTER regex(?o, "wikidata")
    }
```

However, in many cases, this is a user error, where the user should manually cast the variable to a string.
In order to help users debug this, we should report these errors via the logger as warnings instead of voiding the errors.

Concretely, we need an else-clause here that emits a warning to the logger: https://github.com/comunica/comunica/blob/master/packages/actor-query-operation-filter-sparqlee/lib/ActorQueryOperationFilterSparqlee.ts#L44-L46

jitsedesmet:


packages/actor-query-operation-filter-sparqlee/test/ActorQueryOperationFilterSparqlee-test.ts
```diff
@@ -148,6 +149,22 @@ describe('ActorQueryOperationFilterSparqlee', () => {
       await new Promise<void>(resolve => output.bindingsStream.on('error', () => resolve()));
     });
 
+    it('Should log warning for an expressionError', async() => {
+      // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
```
rubensworks, line 153: `eslint-disable-next-line` should not be used unless absolutely necessary. I think they can be resolved here, so there is no need for them I think.

packages/actor-query-operation-filter-sparqlee/test/ActorQueryOperationFilterSparqlee-test.ts
```diff
@@ -148,6 +149,22 @@ describe('ActorQueryOperationFilterSparqlee', () => {
...
+      const logWarnSpy = jest.spyOn(<any> actor, 'logWarn');
+      // eslint-disable-next-line no-import-assign
+      Object.defineProperty(sparqlee, 'isExpressionError', { writable: true });
+      (<any> sparqlee).isExpressionError = jest.fn(() => true);
```
rubensworks, line 157: Instead of overwriting this, can we create a new type of `expression` that results in an expression error?
(you may have to look into the sparqlee code to see how this can be done)

packages/actor-query-operation-filter-sparqlee/test/ActorQueryOperationFilterSparqlee-test.ts
```diff
@@ -148,6 +149,22 @@ describe('ActorQueryOperationFilterSparqlee', () => {
...
+      const op = { operation: { type: 'filter', input: {}, expression: erroringExpression }};
+      const output: IActorQueryOperationOutputBindings = <any> await actor.run(op);
+      await new Promise<void>(resolve => output.bindingsStream.on('end', () => resolve()));
```
rubensworks, line 160: Not sure, but this may also work.
```suggestion
      await new Promise<void>(resolve => output.bindingsStream.on('end', resolve));
```

packages/actor-query-operation-filter-sparqlee/test/ActorQueryOperationFilterSparqlee-test.ts
```diff
@@ -148,6 +149,22 @@ describe('ActorQueryOperationFilterSparqlee', () => {
...
+      expect(logWarnSpy).toHaveBeenCalled();
```
rubensworks, line 161: If you use `toHaveBeenCalledTimes(1)`, you can remove the following loop to just check `logWarnSpy.mock.calls[0][2]`.

rubensworks (changes requested):
Code looks good! Just some minor suggestions for the test.

packages/actor-query-operation-filter-sparqlee/test/ActorQueryOperationFilterSparqlee-test.ts
jitsedesmet: So what you mean is just like we pass 
```JS
const erroringExpression = parse('?a + ?a');
```
right now, we should pass an expression for which we are sure runOperation would result in an ExpressionError?

rubensworks: Yes, exactly!

packages/actor-query-operation-filter-sparqlee/test/ActorQueryOperationFilterSparqlee-test.ts
jitsedesmet: This doesn't work because the error gets called 3 times?

rubensworks: Ah yes, because `bindingsStream` contains 3 bindings. Nvm then :-)

But do check that `toHaveBeenCalledTimes(3)` then.

packages/actor-query-operation-filter-sparqlee/test/ActorQueryOperationFilterSparqlee-test.ts
jitsedesmet: I'm in luck because the quoted expression is one that does this. However, the previous tactic is also used in another test. If this is what you mean I'll find a way to avoid using the tactic in the other test to.

jitsedesmet: Well I probably won't find an expression that creates an error which is not an ExpressionError. That was stupid of me.

rubensworks (approved):
Nice, looks perfect to me!

Before we merge, could you show me (e.g. via a screenshot) how the warning is shown in the logger output when running a query over any datasource that results in this error on the CLI?
If needed, you could create a dummy local RDF file and host is via https://www.npmjs.com/package/http-server.

jitsedesmet:
Sure. I will just use the query mentioned in the original issue.
image
Hope this sufficient.

rubensworks:
@jitsedesmet The error display looks good, but perhaps the bindings display could be improved.
Could you try invoking `toJS()` (or something like that) on the bindings object to see what output this gives?
Ideally, we should be able to see the RDF terms within the bindings, because the user can not see them at the moment.

jitsedesmet:
That's true. I hadn't tried this. The output is indeed readable right now. 
```
bindings: { '?o': NamedNode { id: 'http://www.wikidata.org/entity/Q8070591' } }
```
If this is better for you I will edit the tests. I fear I will now not be able to test whether "bindings" isBidings().

jitsedesmet:
> If this is better for you I will edit the tests. I fear I will now not be able to test whether "bindings" isBidings().

Should I just drop this check?

rubensworks:
That looks perfect!

> If this is better for you I will edit the tests. I fear I will now not be able to test whether "bindings" isBidings().

Instead, you can expect the bindings toEqual a custom object. (You may have to call the RDFJS datafactory to create NamedNodes)

packages/actor-query-operation-filter-sparqlee/test/ActorQueryOperationFilterSparqlee-test.ts
```diff
@@ -147,12 +147,15 @@ describe('ActorQueryOperationFilterSparqlee', () => {
...
       await new Promise<void>(resolve => output.bindingsStream.on('end', resolve));
       expect(logWarnSpy).toHaveBeenCalledTimes(3);
-      for (const call of logWarnSpy.mock.calls) {
+      logWarnSpy.mock.calls.forEach((call, index) => {
         const dataCB = <() => { error: any; bindings: Bindings }> call[2];
         const { error, bindings } = dataCB();
         expect(isExpressionError(error)).toBeTruthy();
-        expect(isBindings(bindings)).toBeTruthy();
-      }
+        const dataFact = new DataFactory();
```
rubensworks, line 154: There's already a factory you can use here: https://github.com/comunica/comunica/blob/3889390dd0b21928e4683723b310000462f1c256/packages/actor-query-operation-filter-sparqlee/test/ActorQueryOperationFilterSparqlee-test.ts#L14

jitsedesmet: sorry

rubensworks:
Great work!

PR #845, opened 2021-08-02
Feature/use sparqlee async aggregator

jitsedesmet:
Updating the ActorQueryOperationGroup to use the new AsyncAggregateEvaluator implemented into sparqlee in [this issue](https://github.com/comunica/sparqlee/issues/86). This will allow the use of the recently added extension functions to be used within groups. After this pull request comunica should have full support for extension functions.

### DIFF
--- a/packages/actor-init-sparql/test/system/Engine-test.ts
+++ b/packages/actor-init-sparql/test/system/Engine-test.ts
@@ -188,24 +188,45 @@ describe('System test: ActorInitSparql', () => {
           );
         });
 
-        it('handles complex queries with groupBy', async() => {
-          const context = <any> { sources: [ store ]};
-          const complexQuery = `PREFIX func: <http://example.org/functions#>
+        describe('handles complex queries with groupBy', () => {
+          let context: any;
+          let complexQuery: string;
+          let extensionBuilder: (timout: boolean) => (args: RDF.Term[]) => Promise<RDF.Term>;
+
+          beforeEach(() => {
+            context = <any> { sources: [ store ]};
+            complexQuery = `PREFIX func: <http://example.org/functions#>
         SELECT (SUM(func:count-chars(?o)) AS ?sum) WHERE {
               ?s ?p ?o.
         }
           `;
-          context[KeysInitSparql.extensionFunctions] = {
-            async 'http://example.org/functions#count-chars'(args: RDF.Term[]) {
+            extensionBuilder = (timout: boolean) => async(args: RDF.Term[]) => {
               const arg = args[0];
               if (arg.termType === 'Literal' && arg.datatype.equals(DF.literal('', stringType).datatype)) {
+                if (timout) {
+                  await new Promise(setImmediate);
+                }
                 return DF.literal(String(arg.value.length), integerType);
               }
               return arg;
-            },
-          };
-          const result = <IQueryResultBindings> await engine.query(complexQuery, context);
-          expect((await result.bindings()).map(res => res.get('?sum').value)).toEqual([ '20' ]);
+            };
+          });
+
+          it('can be evaluated', async() => {
+            context[KeysInitSparql.extensionFunctions] = {
+              'http://example.org/functions#count-chars': extensionBuilder(false),
+            };
+            const result = <IQueryResultBindings> await engine.query(complexQuery, context);
+            expect((await result.bindings()).map(res => res.get('?sum').value)).toEqual([ '20' ]);
+          });
+
+          it('can be truly async', async() => {
+            context[KeysInitSparql.extensionFunctions] = {
+              'http://example.org/functions#count-chars': extensionBuilder(true),
+            };
+            const result = <IQueryResultBindings> await engine.query(complexQuery, context);
+            expect((await result.bindings()).map(res => res.get('?sum').value)).toEqual([ '20' ]);
+          });
         });
       });
     });
--- a/packages/actor-init-sparql/test/system/Engine-test.ts
+++ b/packages/actor-init-sparql/test/system/Engine-test.ts
@@ -97,9 +97,11 @@ describe('System test: ActorInitSparql', () => {
         let quads: RDF.Quad[];
         let stringType: RDF.NamedNode;
         let booleanType: RDF.NamedNode;
+        let integerType: RDF.NamedNode;
         beforeEach(() => {
           stringType = DF.namedNode('http://www.w3.org/2001/XMLSchema#string');
           booleanType = DF.namedNode('http://www.w3.org/2001/XMLSchema#boolean');
+          integerType = DF.namedNode('http://www.w3.org/2001/XMLSchema#integer');
           funcAllow = 'allowAll';
           baseFunctions = {
             'http://example.org/functions#allowAll': async(args: RDF.Term[]) => DF.literal('true', booleanType),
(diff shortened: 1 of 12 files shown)


packages/actor-init-sparql/test/system/Engine-test.ts
```diff
@@ -185,6 +187,26 @@ describe('System test: ActorInitSparql', () => {
...
+        it('handles complex queries with groupBy', async() => {
+          const context = <any> { sources: [ store ]};
+          const complexQuery = `PREFIX func: <http://example.org/functions#>
+        SELECT (SUM(func:count-chars(?o)) AS ?sum) WHERE {
+              ?s ?p ?o.
+        }
+          `;
+          context[KeysInitSparql.extensionFunctions] = {
+            async 'http://example.org/functions#count-chars'(args: RDF.Term[]) {
+              const arg = args[0];
```
rubensworks, line 200: Could you also add a variant of this test that is slower (to check for race-conditions)?

We could do this by adding something like this on the current line: `await new Promise((resolve) => setImmediate(resolve, 1));`

packages/actor-query-operation-group/lib/ActorQueryOperationGroup.ts
```diff
@@ -23,7 +23,9 @@ export class ActorQueryOperationGroup extends ActorQueryOperationTypedMediated<A
   public async testOperation(pattern: Algebra.Group, context: ActionContext): Promise<IActorTest> {
     for (const aggregate of pattern.aggregates) {
       // Will throw for unsupported expressions
-      const _ = new SyncEvaluator(aggregate.expression);
+      const _ = new AsyncEvaluator(aggregate.expression, {
```
rubensworks, line 26: Unneeded object spread (just pass `ActorQueryOperation.getAsyncExpressionContext(context)` directly)

packages/actor-query-operation-group/lib/ActorQueryOperationGroup.ts
```diff
@@ -43,9 +45,7 @@ export class ActorQueryOperationGroup extends ActorQueryOperationTypedMediated<A
       ...aggregates.map(agg => termToString(agg.variable)),
     ];
 
-    // TODO: we should not override extensionFunctionCreator. Instead we should use an AsyncAggregateEvaluator.
-    const sparqleeConfig = { ...ActorQueryOperation.getExpressionContext(context),
-      extensionFunctionCreator: undefined };
+    const sparqleeConfig = { ...ActorQueryOperation.getAsyncExpressionContext(context) };
```
rubensworks, line 48: Unneeded object spread

packages/actor-query-operation-group/lib/GroupsState.ts
```diff
@@ -44,56 +49,69 @@ export class GroupsState {
...
-    let group: IGroup | undefined = this.groups.get(groupHash);
-    if (!group) {
-      // Initialize state for all aggregators for new group
-      const aggregators: Record<string, AggregateEvaluator> = {};
-      for (const aggregate of this.pattern.aggregates) {
-        const key = termToString(aggregate.variable);
-        aggregators[key] = new AggregateEvaluator(aggregate, this.sparqleeConfig);
-        aggregators[key].put(bindings);
-      }
+    const groupTest: Promise<IGroup> | undefined = this.groupsInitializer.get(groupHash);
```
rubensworks, line 61:
```suggestion
    const groupInitializer: Promise<IGroup> | undefined = this.groupsInitializer.get(groupHash);
```

packages/actor-query-operation-group/lib/GroupsState.ts
```diff
@@ -44,56 +49,69 @@ export class GroupsState {
...
 
-      group = { aggregators, bindings: grouper };
-      this.groups.set(groupHash, group);
+    let res: Promise<any>;
+    if (!groupTest) {
+      // Initialize state for all aggregators for new group
+      const groupPromise = (async() => {
```
rubensworks, line 66:
```suggestion
      groupInitializer = (async() => {
```

packages/actor-query-operation-group/lib/GroupsState.ts
```diff
@@ -44,56 +49,69 @@ export class GroupsState {
...
-      }
+        if (this.distinctHashes) {
+          const bindingsHash = this.hashBindings(bindings);
+          this.distinctHashes.set(groupHash, new Set([ bindingsHash ]));
+        }
+        return { aggregators, bindings: grouper };
+      })();
+      this.groupsInitializer.set(groupHash, groupPromise);
+      res = groupPromise.then(group => {
+        this.groups.set(groupHash, group);
```
rubensworks, line 82: Instead of doing this via a separate `.then` call, could we just move this into the actual initializer promise a few lines above?

packages/actor-query-operation-group/lib/GroupsState.ts
```diff
@@ -44,56 +49,69 @@ export class GroupsState {
...
-      // Group already exists
-      // Update all the aggregators with the input binding
-      for (const aggregate of this.pattern.aggregates) {
-        // If distinct, check first wether we have inserted these values already
-        if (aggregate.distinct) {
-          const hash = this.hashBindings(bindings);
-          if (this.distinctHashes!.get(groupHash)!.has(hash)) {
-            continue;
-          } else {
+      const defGroupPromise = groupTest;
```
rubensworks, line 85:
```suggestion
      const groupInitializerDefined = groupInitializer;
```

packages/actor-query-operation-group/lib/GroupsState.ts
```diff
@@ -44,56 +49,69 @@ export class GroupsState {
...
-        const variable = termToString(aggregate.variable);
-        group.aggregators[variable].put(bindings);
-      }
+          const variable = termToString(aggregate.variable);
+          await group.aggregators[variable].put(bindings);
+        }));
+      })();
+    }
+    return res.then(() => {
+      this.subtractCounter();
```
rubensworks, line 104: Instead of doing this via a separate `.then` call, could we just move this into the res promise a few lines above?

packages/actor-query-operation-group/lib/GroupsState.ts
```diff
@@ -44,56 +49,69 @@ export class GroupsState {
...
+    });
+  }
+
+  private subtractCounter(): void {
```
rubensworks, line 108:
```suggestion
  private subtractWaitCounterAndCollect(): void {
```

rubensworks (changes requested):
Looks great!
Just some minor comments here and there.

Could you also add some additional comments? Mainly the parts around `groupsInitializer`, `waitCounter` and `waitResolver` may not be trivial for others to understand in the future.

packages/actor-query-operation-group/lib/GroupsState.ts
jitsedesmet: I suppose we could yes.

packages/actor-query-operation-group/lib/GroupsState.ts
jitsedesmet: I don't think this is possible because the promise above this is a list of promises. The subtract needs to happen after all those promises.

rubensworks: You have an `await Promise.all` there. So if you move it after that promise resolves, it should be fine, right?

jitsedesmet: oh you just mean before I put the promise into `res`? That would be possible. I would the just need to add it in both cases of the if.

packages/actor-init-sparql/test/system/Engine-test.ts
rubensworks: Ah, I actually meant `setTimeout`, not `setImmediate`. (my bad...)

packages/actor-query-operation-group/lib/GroupsState.ts
```diff
@@ -120,15 +145,31 @@ export class GroupsState {
...
       rows = [ Bindings(single) ];
     }
+    this.waitResolver(rows);
+  }
 
-    return rows;
+  /**
+   * Collect the result of the final state. This returns a Bindings per group,
+   * and a (possibly empty) Bindings in case no Bindings have been consumed yet.
+   * you can only call this method once. Once the promise resolves calling @{consumeBindings} will not alter this value.
```
rubensworks, line 161: > Once the promise resolves calling @{consumeBindings} will not alter this value.

Makes me wonder if we also want to throw in `consumeBindings` if `this.resultHasBeenCalled` is true.

packages/actor-query-operation-group/lib/GroupsState.ts
```diff
@@ -120,15 +145,31 @@ export class GroupsState {
...
+   */
+  public collectResults(): Promise<Bindings[]> {
+    if (this.resultHasBeenCalled) {
+      return new Promise((resolve, reject) => reject(new Error('collectResult should only be called once.')));
```
rubensworks, line 165: Can be simplified to `Promise.reject(...)`.

rubensworks (changes requested):
Some minor nits, should be good to go after this.

packages/actor-query-operation-group/lib/GroupsState.ts
jitsedesmet: I think that would be a valid thing to do.

rubensworks:
Great work @jitsedesmet!

Issue #844, opened 2021-07-28
Send original SPARQL query string to endpoints


sdevalk:
_This issue has been modified by @rubensworks to describe the intended solution_

The problem is that we always parse the incoming query, and serialize it again when we send it to a SPARQL endpoint. Since this parsing-serializing process sometimes results in slightly different queries, this may cause issues for certain use cases. To solve this, we need to keep track of the original query string, and use that one in the case of querying a SPARQL endpoint.

In actor-init-sparql, we need to add the original query string into the context.
In `actor-query-operation-sparql-endpoint`, we need to check this context entry, and use that one if it exists.

The only difficulty here is that we have to make sure that this context entry does not propagate to places where it shouldn't. For example, if this actor is only used because of a `SERVICE` keyword, we need to make sure that the actor still uses the `SERVICE` operation instead of the full original query. We could do this by just making sure that the service operator unsets or overrides this entry.

-----

Original issue:

#### Issue type:

- :question: Question 

____
#### Question:

I want to query SPARQL endpoint `https://data.muziekschatten.nl/sparql`, backed by Virtuoso, using the query underneath:

```sparql
SELECT * WHERE {
  ?uri <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Person>;
    <http://schema.org/name> ?label.
  FILTER(<bif:contains>(?label, "emile"))
  BIND("no name" AS ?name)
}
LIMIT 100
```

It's a silly query - but the crucial parts are the `FILTER` and the `BIND`. If I execute this using Comunica...

`npx comunica-sparql sparql@https://data.muziekschatten.nl/sparql -f query.rq`

...where file `query.rq` contains the query above...then the result is:

`[Invalid SPARQL endpoint (https://data.muziekschatten.nl/sparql) response: undefined`

My guess is that Comunica is rewriting the query, e.g. to:

```sparql
SELECT * WHERE {
  ?uri <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://schema.org/Person>;
    <http://schema.org/name> ?label.
  BIND("no name" AS ?name)
  FILTER(<bif:contains>(?label, "emile"))
}
LIMIT 100
```

`FILTER` is placed after `BIND`. If I execute this query via the endpoint's query editor on `https://data.muziekschatten.nl/sparql`, then Virtuoso yields the error:

[...]

PR #896, opened 2021-11-25
Send original SPARQL query string to endpoints

jasmineleonard:
This is in an initial attempt to solve https://github.com/comunica/comunica/issues/844.

As suggested in the issue description, in `actor-init-sparql` we save the original query string in the context, and in `actor-query-operation-sparql-endpoint`, we check this context entry and use it if it exists.

### Questions

1. I've added `@comunica/context-entries` to the dependencies in `actor-query-operation-sparql-endpoint/package.json`, but running `yarn run lint` produces the following error:

> workspace/comunica/packages/actor-query-operation-sparql-endpoint/lib/ActorQueryOperationSparqlEndpoint.ts
  9:1  error  '@comunica/context-entries' should be listed in the project's dependencies. Run 'npm i -S @comunica/context-entries' to add it  import/no-extraneous-dependencies

Is there something else I need to do?  Also, do I need to update the version of `@comunica/context-entries` to reflect the addition of the new key?

2. How should I go about testing`ActorQueryOperationSparqlEndpoint.ts`?  I've added one additional test to attempt to test the new functionality added in line 89, but the test output says that both branches of this line are not covered.

3.  The issue description states that "we have to make sure that this context entry does not propagate to places where it shouldn't" and suggests "making sure that the service operator unsets or overrides this entry".  Are there any other places that we need to do this besides the service operator?

### DIFF
--- a/packages/actor-query-operation-service/lib/ActorQueryOperationService.ts
+++ b/packages/actor-query-operation-service/lib/ActorQueryOperationService.ts
@@ -34,10 +34,10 @@ export class ActorQueryOperationService extends ActorQueryOperationTypedMediated
     context = context || ActionContext({});
     let subContext: ActionContext = context
       .delete(KeysRdfResolveQuadPattern.source)
-      .delete(KeysRdfResolveQuadPattern.sources);
+      .delete(KeysRdfResolveQuadPattern.sources)
+      .delete(KeysInitSparql.queryString);
     const sourceType = this.forceSparqlEndpoint ? 'sparql' : undefined;
     subContext = subContext.set(KeysRdfResolveQuadPattern.sources, [{ type: sourceType, value: endpoint }]);
-    subContext = subContext.delete(KeysInitSparql.queryString);
     // Query the source
     let output: IActorQueryOperationOutputBindings;
     try {
--- a/packages/actor-query-operation-service/lib/ActorQueryOperationService.ts
+++ b/packages/actor-query-operation-service/lib/ActorQueryOperationService.ts
@@ -1,7 +1,7 @@
 import type { IActorQueryOperationTypedMediatedArgs } from '@comunica/bus-query-operation';
 import { ActorQueryOperation, ActorQueryOperationTypedMediated,
   Bindings } from '@comunica/bus-query-operation';
-import { KeysRdfResolveQuadPattern } from '@comunica/context-entries';
+import { KeysInitSparql, KeysRdfResolveQuadPattern } from '@comunica/context-entries';
 import type { IActorTest } from '@comunica/core';
 import { ActionContext } from '@comunica/core';
 import type { IActorQueryOperationOutputBindings } from '@comunica/types';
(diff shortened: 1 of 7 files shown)


packages/actor-query-operation-service/lib/ActorQueryOperationService.ts
```diff
@@ -37,7 +37,7 @@ export class ActorQueryOperationService extends ActorQueryOperationTypedMediated
       .delete(KeysRdfResolveQuadPattern.sources);
     const sourceType = this.forceSparqlEndpoint ? 'sparql' : undefined;
     subContext = subContext.set(KeysRdfResolveQuadPattern.sources, [{ type: sourceType, value: endpoint }]);
-
+    subContext = subContext.delete(KeysInitSparql.queryString);
```
rubensworks, line 40: This operation can be chained onto the previous line, so we don't need the second override of `subContext`.

rubensworks (changes requested):
Great work! This is very aligned with what I had in mind :-)
Just one tiny. comment

> I've added @comunica/context-entries to the dependencies in actor-query-operation-sparql-endpoint/package.json, but running yarn run lint produces the following error:

Ah, probably just an eslint caching issue (can be fixed by removing `.eslintcache`). The CI passes the lint job.

> How should I go about testingActorQueryOperationSparqlEndpoint.ts? I've added one additional test to attempt to test the new functionality added in line 89, but the test output says that both branches of this line are not covered.

I suspect the branch where the context is undefined may not be covered, so perhaps we can try adding a second test where the context is undefined (and the algebra-based query will be used).

> The issue description states that "we have to make sure that this context entry does not propagate to places where it shouldn't" and suggests "making sure that the service operator unsets or overrides this entry". Are there any other places that we need to do this besides the service operator?

AFAIK, the service operator is the only one, so your PR should have everything covered :-)

jasmineleonard:
Thanks @rubensworks!

> Ah, probably just an eslint caching issue (can be fixed by removing .eslintcache)

Yes, that was it! 🤦‍♀️ 

> I suspect the branch where the context is undefined may not be covered, so perhaps we can try adding a second test where the context is undefined (and the algebra-based query will be used).

I rewrote the code slightly, as setting the context as undefined threw an error due to there not being a sparql endpoint source defined (line 77 in `ActorQueryOperationSparqlEndpoint.ts`).  This meant the code never actually reached line 89 if the context was undefined, so I've just made that explicit.  Is that an acceptable workaround?

>  AFAIK, the service operator is the only one, so your PR should have everything covered :-)

Wonderful!

rubensworks (approved):
Looks perfect!

Will merge once the CI finishes.

rubensworks:
One of the integration test fails, but this is actually expected to fail.
I will look into fixing this one myself, since our integration tests setup is a bit underdocumented atm...

jasmineleonard:
> I will look into fixing this one myself, since our integration tests setup is a bit underdocumented atm...

Thanks so much @rubensworks! 🙌

Issue #902, opened 2021-12-28
Utility type for Medator<Actor<...>...> pattern


jeswr:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

The pattern `Medator<Actor<...>...>` is quite common and a little cumbersome when writing a bunch of new packages. I'm thinking it would be useful to have the following utility type in `@comunica/types` that captures this pattern (though I'm not sure if I have chosen the best name for it).

```ts
type Mediated<I extends IAction, O extends IActorOutput, T extends IActorTest = IActorTest> = Mediator<Actor<I, T, O>, I, T, O>;
```

The result being that

```ts
public readonly mediatorRdfParse: Mediator<Actor<IActionRdfParse, IActorTest, IActorRdfParseOutput>,
  IActionRdfParse, IActorTest, IActorRdfParseOutput>;
```
becomes
```ts
public readonly mediatorRdfParse: Mediated<IActionRdfParse, IActorRdfParseOutput>;
```

If you think it is appropriate I'm happy to create a PR that adds this type and upgrade the rest of the codebase to use it.

rubensworks:
I've added something similar in the `next/major` branch quite recently: https://github.com/comunica/comunica/commit/d160c8f074e98bf0e857a9a530322a7f54dcd807
So this will land in the next major release.
Would this meet your needs?

jeswr:
Ah nice! I'll follow that pattern of exporting a custom mediator type in the [reasoning packages](https://github.com/comunica/comunica-feature-reasoning/tree/experimental-development/packages) I am working on.

PR #908, opened 2022-01-18
chore: mediated utility type

jeswr:
Adds a `Mediate` utility type as described in https://github.com/comunica/comunica/issues/902

### DIFF
--- a/packages/core/lib/Mediator.ts
+++ b/packages/core/lib/Mediator.ts
@@ -112,3 +112,6 @@ export interface IMediatorArgs<A extends Actor<I, T, O>,
    */
   bus: Bus<A, I, T, O>;
 }
+
+export type Mediated<I extends IAction, O extends IActorOutput, T extends IActorTest = IActorTest> =
+Mediator<Actor<I, T, O>, I, T, O>;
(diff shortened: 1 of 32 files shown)


packages/core/lib/Mediator.ts
```diff
@@ -112,3 +112,6 @@ export interface IMediatorArgs<A extends Actor<I, T, O>,
    */
   bus: Bus<A, I, T, O>;
 }
+
+export type Mediate<I extends IAction, O extends IActorOutput, T extends IActorTest = IActorTest> =
```
rubensworks, line 116: Any particular reason why the order is I-O-T instead of the usual I-T-O?
If not, I would favor using the latter for consistency.

rubensworks (changes requested):
Looks great! Just one comment about generic type order.

packages/core/lib/Mediator.ts
jeswr: `T` can often be defaulted to `IActorTest` e.g. `export type MediatorInit = Mediate<IActionInit, IActorOutputInit>;` - and default parameters must come after required parameters.

rubensworks: Ah, I missed the default param definition.
Ok, good enough a reason for me :-)

Issue #909, opened 2022-01-18
breaking feature: generic dereference bus


jeswr:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

Extending discussion in #907 .

Now that there only appear to be 3 dereference actors (fallback, file and http) this should be more straightforward (I'm assuming actor-rdf-dereference-paged-next is deprecated rather than moved to another package location).

A rough implementation would be:

The `bus-dereference` package would be similar to `bus-rdf-dereference` with the exception of the output being a Readable stream of `string`s rather than `quad`s.

`actor-rdf-dereference-fallback` would become `actor-dereference-fallback` with almost no other changes

`actor-rdf-dereference-file` would become `actor-dereference-file`
Rather than having `mediaMappings` provided in the constructor, this would instead be provided as an input parameter on the `run` and `test` calls (This change would actually happen in `ActorRdfDereferenceMediaMappings`).

`actor-rdf-dereference-http` would become `actor-dereference-http`
Same change with `mediaMappings` as input and a 'hook' to a `ParseMediatypes` call.
Though I feel like there has to be a better way of doing this than providing a hook to a function call.

Then any usages of `mediatorRdfDereference` can be replaced with a composition of `mediatorDereference` and `mediatorRdfParse`.

rubensworks:
> The bus-dereference package would be similar to bus-rdf-dereference with the exception of the output being a Readable stream of strings rather than quads.

Perhaps this should be a generic type so that the user can still fill in that it will return a stream of quads?

> Same change with mediaMappings as input and a 'hook' to a ParseMediatypes call.
> Though I feel like there has to be a better way of doing this than providing a hook to a function call.

An async callback may be a good solution indeed (probably 2 will be needed).

I initially had something in mind where the new dereference actors would become abstract, and accept an abstract parsing mediator in its constructor, so that the dereference actors could be instantiated multiple times for different parsing mediators. But I actually think your solution is simpler, and more flexible.

jeswr:
> Perhaps this should be a generic type so that the user can still fill in that it will return a stream of quads?
Sure I'll do this for the draft PR - it shouldn't affect the rest of the implementation much anyway; but what would be the use case for this?

> An async callback may be a good solution indeed (probably 2 will be needed).
Ok - I'll draft a PR with this approach, and can move to your alternative if any problems arise.

rubensworks:
[...]

PR #912, opened 2022-01-19
DRAFT: Dereference Refactor

jeswr:
I think I have found a middle ground between callbacks and abstract bus from the discussion in #909 .

The structure is as follows:
The dereference bus simply dereferences a file to a stream of `string`s.
We then have abstract dereference-parse actors that are instantiated with chosen parse mediators via the constructor. A rough instantiation for rdf parsing should look somewhat like [this](https://github.com/comunica/comunica/compare/next/major...jeswr:feat/dereference-refactor?expand=1#diff-b39cfb92367e3b5a84ece5967ce0acda86b9555796e68819ff81a4a8cdf5ce87) - I'm in the stage of figuring out how to make this abstract as I haven't really done this before.
The rest of the actors then use the instantiated `dereference-parse` actors in a similar manner to the current use of `rdf-dereference`. 

Ignore changes to anything that is not `bus-dereference` or `bus-rdf-dereference`. I made these when I was trying a different approach and will likely revert them if I folow this direction

Question:
`Why are the `headers` a `Record<string, string>` rather than `Headers` in `IActorRdfDereferenceOutput`? It feels like it would be easier to pass around a `Headers` object especially given this is what the parser uses.


jeswr:
@rubensworks - Do you think this is appropriate? Especially the current `bus-dereference` and associated actors which I don't anticipate making many/any more changes to

rubensworks:
> I think I have found a middle ground between callbacks and abstract bus from the discussion in #909 .

Seems like an excellent solution!
So basically `dereference` is a _new_ bus, but it doesn't replace `rdf-dereference`, but merely abstracts it.

I haven't gone through the PR in detail, but I agree with going forward with this approach.

> Why are the headersaRecord<string, string>rather thanHeadersinIActorRdfDereferenceOutput? It feels like it would be easier to pass around a Headers` object especially given this is what the parser uses.

Ah, that's probably an artifact from the past...
Feel free to change it to `Headers` if that's more convenient. (but perhaps we want to do this in a separate PR)

PR #913, opened 2022-01-20
Feat: use headers type

jeswr:
Replaces use of `Record<string, string>` type with `Header` type and removes the return of headers from `bus-rdf-dereference-file`.

I've also snuck in a couple of replacements of `any` for `IActorRdfMetadataExtractOutput['metadata']` in a couple of the files I had to change.


rubensworks (approved):
Perfect! Thanks for this.

PR #917, opened 2022-01-29
DRAFT: Dereference Refactor

jeswr:
Supercedes #912

@rubensworks - this is almost complete barring adding some documentation.

I am running into a couple of errors when I run the `build:components` command which look like:
```
Could not load class or interface or other type EventEmitter from /home/jeswr/Documents/Github18/comunica/node_modules/events/events:
ENOENT: no such file or directory, open '/home/jeswr/Documents/Github18/comunica/node_modules/events/events.d.ts'
```

So far I've diagnosed that the issue is occurring when a type is imported as part of a type parameter that is used - for instance the `EventEmitter` referenced above is used in the `RDF.Stream` parameter in the following code.

```ts
export interface IActorRdfDereferenceArgs extends IAbstractDereferenceParseArgs<RDF.Stream, IActionRdfParseMetadata, IActorRdfParseOutputMetadata> {
 ...
}
```

Is this an error in Components.js that can be fixed, or do these interfaces need to be reworked?


jeswr:
Note to self - I also need to fix up some prefixes in the configuration files

rubensworks:
> I am running into a couple of errors when I run the build:components command which look like:

Certain component types are not supported yet by the Components.js generator, so those will have to be ignored until they are supported.

> So far I've diagnosed that the issue is occurring when a type is imported as part of a type parameter that is used - for instance the EventEmitter referenced above is used in the RDF.Stream parameter in the following code.

You can ignore unsupported components by adding them to the list here: https://github.com/comunica/comunica/blob/next/major/.componentsjs-generator-config.json

PR #921, opened 2022-02-09
feat: Filter combine pipeline

jeswr:
This PR adds an optional parameter to `MediatorCombinePipeline` which filters out actors with erroring tests.

This contains *no breaking changes*.

I'm not tied to this design - just proposing this now as it is useful to have something like this in the reasoning components I am working on

### DIFF
--- a/packages/mediator-combine-pipeline/lib/MediatorCombinePipeline.ts
+++ b/packages/mediator-combine-pipeline/lib/MediatorCombinePipeline.ts
@@ -28,7 +28,7 @@ export class MediatorCombinePipeline
     // Pass action to first actor,
     // and each actor output as input to the following actor.
     let handle: H = action;
-    for (const actor of testResults.map(result => result.actor)) {
+    for (const { actor } of testResults) {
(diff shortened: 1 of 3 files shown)


packages/mediator-combine-pipeline/lib/MediatorCombinePipeline.ts
```diff
@@ -23,12 +25,24 @@ export class MediatorCombinePipeline
     }
 
     // Delegate test errors.
+    if (this.filterErrors) {
+      const _testResults: IActorReply<A, H, T, H>[] = [];
+      for (const result of testResults) {
+        try {
+          await result.reply;
+          _testResults.push(result);
+          // eslint-disable-next-line no-empty
```
rubensworks, line 34: Instead of disabling this rule, we usually do something like this instead:
```
} catch {
  // Ignore errors
}
```

rubensworks (changes requested):
Looks great! Just one minor nit.

jeswr:
@rubensworks this is ready to merge - I won't do it myself in case you're currently working directly on the `next/major` branch

PR #926, opened 2022-02-11
chore: create tests for overlapping BGP patterns

jeswr:
Created these when I was trying to debug an error with another error I was running into [here](https://github.com/comunica/comunica/tree/next/major-experimental-document-extender).

Won't be offended if you don't think they are needed


rubensworks:
Normally I'd say that more unit tests can't hurt, but since the tests here are are actually integration tests that are imprecise (they don't check the exact results, and the data may change in the future), I'm not sure this is the correct place for these kinds of tests.

If we want to include such tests, perhaps a minimal custom Store with precise result checking might be better?
But in any case, I don't think they are required per se, since this functionality should also be tested within the SPARQL spec tests.

PR #927, opened 2022-02-12
chore: fix bus-rdf-dereference deps

jeswr:


rubensworks:
> This package is missing the .npmignore file as well

Oh, that may be a problem indeed.

> though I'm wondering if any of the packages in the monrepo need .npmignore files since there are no .gitignore files in those folders?

To be honest, I don't fully remember why we have the empty `.npmignore` files (probably should have documented this somewhere...). But it had something to do with the `files` entry in `package.json`.
But it _may_ not be needed indeed if we don't have a `.gitignore` in that package.
We definitely need some testing if we want to change this.

rubensworks (approved):

PR #928, opened 2022-02-12
Feat: set safe

jeswr:
This PR introduces `setSafe` to the action context which only sets if the key is not already available. This is a common pattern when applying defaults - both in the main repo and in other packages (ie for reasoning).

I've also gone done a rabbit hole and refactored a bit more to remove code that is not necessary now that the `context` required in `IAction`.


jeswr:
Converting to draft until I fix a couple of problems with test coverage @rubensworks

packages/core/lib/ActionContext.ts
```diff
@@ -11,6 +11,13 @@ export class ActionContext implements IActionContext {
     this.map = Map<string, any>(data);
   }
 
+  /**
+   * Will only set the value if the key is not already set.
+   */
+  public setSafe<V>(key: IActionContextKey<V>, value: V): IActionContext {
```
rubensworks, line 17: Perhaps `setDefault` might be a more descriptive name?

rubensworks:
Oh, this is neat. Will check in more detail soon.

PR #944, opened 2022-03-08
Add USER to Dockerfile

MPParsley:
By default, containers run as root. A container running as root has full control of the host system. As container technology matures, more secure default options may become available. For now, requiring root is dangerous for others and may not be available in all environments. Your image should use the USER instruction to specify a non-root user for containers to run as. If your software does not create its own user, you can create a user and group in the Dockerfile.

Source: https://redhatgov.io/workshops/security_containers/exercise1.2/


rubensworks:
Thanks for the PR @MPParsley!
It looks like the Docker build is failing though. Probably because `/var/www/` is owned by the root?

rubensworks:
Sure, the home folder is ok!

But it looks like the build still fails with that change. (I haven't used the `USER` functionality of Docker before, so not sure what the right way of solving that would be)

engines/query-sparql/Dockerfile
```diff
@@ -1,7 +1,13 @@
 FROM node:16
 
+# Create user
+RUN addgroup comunica && adduser -S -G comunica comunica
```
MPParsley, line 4:
```suggestion
RUN addgroup comunica && adduser --system --group comunica comunica
```

MPParsley:
Not sure why this occurs:
`sh: 1: cannot create engine-default.js: Permission denied`

User comunica should have access to the home folder.

Adding debug.

MPParsley:
Tests are failing randomly on `Error: Could not retrieve https://fragments.dbpedia.org/2016-04/en?page=5 (HTTP status 500):`

rubensworks:
Ouch, created https://github.com/comunica/comunica/issues/947 for this. Restarted the job.

engines/query-sparql/Dockerfile
```diff
@@ -1,7 +1,16 @@
 FROM node:16
 
 # Install location
-ENV dir /var/www/@comunica/query-sparql/
+ENV dir /home/comunica/
+
+# Create user
+RUN useradd --create-home --shell /bin/bash comunica
```
MPParsley, line 7:
```suggestion
RUN useradd --create-home --shell /bin/sh comunica
```
```suggestion
RUN useradd --create-home --shell /bin/bash comunica
```

MPParsley:
Instead of creating our own user we can/should reuse the default node user.

rubensworks:
> Instead of creating our own user we can/should reuse the default node user.

Sure, that sounds reasonable to me.

MPParsley:
@rubensworks, could you restart the job once more?

rubensworks:
Done.

(locally building the Dockerfile might be easier though)

engines/query-sparql/Dockerfile
```diff
@@ -1,23 +1,24 @@
...
+COPY --chown=node:node bin/*.js bin/
+COPY --chown=node:node engine-default.js package.json ./
 
 # Set the npm registry
 ARG NPM_REGISTRY=https://registry.npmjs.org/
 RUN npm config set @comunica:registry $NPM_REGISTRY
 
 # Install the node module
-RUN cd ${dir} && npm install --only=production
+RUN npm install --only=production
```
MPParsley, line 19: npm WARN config only Use `--omit=dev` to omit dev dependencies from the install.

```suggestion
RUN npm install --omit=dev
```

MPParsley:
Docker CI pipeline is still complaining about an access error. Locally on Docker build/run works.

I'm not familiar with lerna docker but it would seem that a specific user is being configured in the ci workflow or node_modules is cached under a different user:

```
  docker:
    needs:
      - test
      - lint
    runs-on: ubuntu-latest
    steps:
      - name: Use Node.js
        uses: actions/setup-node@v3
        with:
          node-version: '14.x'
      - name: Check out repository
        uses: actions/checkout@v3
      - name: Load cache
        uses: actions/cache@v2
        with:
          path: '**/node_modules'
          key: ${{ runner.os }}-docker-modules-${{ secrets.CACHE_VERSION }}-${{ hashFiles('**/yarn.lock') }}
      - name: Install dependencies
        run: yarn install --pure-lockfile
      - name: Install Lerna Docker
        run: sh -c "`curl -fsSl https://raw.githubusercontent.com/rubensworks/lerna-docker/master/install.sh`"
      - name: Build Docker images
        run: ~/.lerna-docker/bin/lerna-docker comunica build
      - name: Deploy Docker images
        if: startsWith(github.ref, 'refs/heads/master') || startsWith(github.ref, 'refs/tags/')
        run: ~/.lerna-docker/bin/lerna-docker comunica push
        env:
          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
          DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
```

engines/query-sparql/Dockerfile
```diff
@@ -1,23 +1,24 @@
...
+COPY --chown=node:node bin/*.js bin/
+COPY --chown=node:node engine-default.js package.json ./
 
 # Set the npm registry
 ARG NPM_REGISTRY=https://registry.npmjs.org/
 RUN npm config set @comunica:registry $NPM_REGISTRY
 
 # Install the node module
-RUN cd ${dir} && npm install --only=production
+RUN npm install --omit=dev
```
MPParsley, line 19: Add debug;
```suggestion
RUN echo $(whoami) && ls -al ./ && chown -R node:node /var/www/@comunica/query-sparql/
RUN echo $(whoami) && npm install --omit=dev
```

MPParsley:
@rubensworks, the Docker image is fixed.

Issue #942, opened 2022-03-07
Utility types


jeswr:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

Similar to the introduction of `Mediate` I have found myself wanting the following utility types in some test suites I am writing

```ts

type IReply<I extends IAction, O extends IActorOutput, T extends IActorTest = IActorTest> = IActorReply<Actor<I, T, O>, I, T, O>;
type IBus<I extends IAction, O extends IActorOutput, T extends IActorTest = IActorTest> = Bus<Actor<I, T, O>, I, T, O>

```

PR #946, opened 2022-03-10
Developer experience improvements

### LINKED ISSUE #945: `getSafe` method for context
#### Issue type:

____
#### Description:

Implemented as

```ts
getSafe<V>(key: IActionContextKey<V>): V {
   if (!this.has(key)) {
      throw new Error(`Context entry ${key.name} is required but not available`);
   }
   return this.get(key);
}
```

jeswr:
Resolves #945 #942

### DIFF
--- a/packages/core/test/ActionContext-test.ts
+++ b/packages/core/test/ActionContext-test.ts
@@ -48,6 +48,30 @@ describe('ActionContext', () => {
       });
     });
 
+    describe('getSafe', () => {
+      beforeEach(() => {
+        context = context
+          .set(key1, 'abc')
+          .set(key3, [ true, false ]);
+      });
+
+      it('should get entries', () => {
+        expect(context.getSafe(key1)).toEqual('abc');
+        expect(() => context.getSafe(key2)).toThrow(`Context entry ${key2.name} is required but not available`);
+        expect(context.getSafe(key3)).toEqual([ true, false ]);
+      });
+
+      it('should fail during compilation for an incorrect key value', () => {
+        // @ts-expect-error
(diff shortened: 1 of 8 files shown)


packages/core/test/ActionContext-test.ts
```diff
@@ -48,6 +48,30 @@ describe('ActionContext', () => {
...
+      });
+
+      it('should get entries', () => {
+        expect(context.getSafe(key1)).toEqual('abc');
+        expect(() => context.getSafe(key2)).toThrow(`Context entry ${key2.name} is required but not available`);
+        expect(context.getSafe(key3)).toEqual([ true, false ]);
+      });
+
+      it('should fail during compilation for an incorrect key value', () => {
+        // @ts-expect-error
```
rubensworks, line 65: Instead of `@ts-expect-error`, could we explicitly check for the error?
So something like `expect(() => context.getSafe(key1)).toThrow('expected error message...')`

This allows us to test for exact error messages as well

rubensworks (changes requested):
Great! Just a minor nit about the tests.

packages/core/test/ActionContext-test.ts
jeswr: We are already doing this here https://github.com/comunica/comunica/blob/37074b7ddf407210352c8754b8821cc74c5c42d9/packages/core/test/ActionContext-test.ts#L60

rubensworks: Oh, my bad, glanced over it too quickly :-)

PR #948, opened 2022-03-11
fix: require @rdfjs/types version

woutermont:
@rubensworks, as you suggested, this should fix `Namespace '"[...]/node_modules/@rdfjs/types/index"' has no exported member [...]` errors when running `tsc` on a package depending on `@comunica/types@^2.0.1`.

### DIFF
--- a/packages/types/package.json
+++ b/packages/types/package.json
@@ -31,7 +31,7 @@
     "bin/**/*.js"
   ],
   "dependencies": {
-    "@rdfjs/types": "*",
+    "@rdfjs/types": "^1.1.0",
     "@types/yargs": "^17.0.2",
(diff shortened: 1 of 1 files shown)


packages/types/package.json
```diff
@@ -31,7 +31,7 @@
     "bin/**/*.js"
   ],
   "dependencies": {
-    "@rdfjs/types": "*",
+    "@rdfjs/types": "^1.1.0",
```
rubensworks, line 34:
```suggestion
    "@rdfjs/types": ">=1.1.0",
```
I'll just broaden the range a bit since this is a types package.

rubensworks:
Thanks for this!

PR #967, opened 2022-03-28
Chore/depcheck

jeswr:
To avoid issues like that in #966 I suggest we add `depcheck`.

As you can see the tests are failing here because there are several remaining dependency problems throughout the monrepo; ~I don't have time to go through and address them all right now so if someone else wants to take over this PR and fix those problems feel free!~


jeswr:
Acutally don't do this manually - I've almost completed a script to do this for the reasoning components

jeswr:
Ok, I've added a depcheck script which will fix all of the package.json's - however; it runs into some issues caused by
1. Circular dependencies (e.g. between `@comunica/core` and `@comunica/types`)
2. Self references (e.g. `@comunica/types` importing from `@comunica/types`) - once these are resolved the script *should* work.

jeswr:
All working - the only manual change I made was moving the logger into `@comunica/types` and then re-exporting it from `@comunica/core` so as to remove a circular dependency there.

rubensworks (changes requested):
This is really great!
We used to do this using eslint, but it didn't really handle type-only packages that well, so this new approach here is a big improvement.

packages/actor-query-operation-group/package.json
```diff
@@ -36,7 +36,7 @@
     "@comunica/bus-query-operation": "^2.0.1",
     "@comunica/core": "^2.0.1",
     "@comunica/types": "^2.0.1",
-    "@types/json-stable-stringify": "^1.0.32",
+    "@rdfjs/types": "^1.1.0",
```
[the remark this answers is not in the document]
jeswr: Rather than detecting if it is a type-only package, why don't we detect if there is a resolution defined in the global package.json? https://github.com/comunica/comunica/blob/448bfbadd805dcad95841789cd12b0d33a672585/package.json#L91-L93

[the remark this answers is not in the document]
rubensworks: Sure, that might be fine for now.
In the long-term, we probably want this for _all_ type packages (not just the ones we're overriding via resolutions), but that's not very urgent atm.

packages/actor-init-query/package.json
```diff
@@ -47,13 +46,18 @@
     "@comunica/types": "^2.0.1",
     "@rdfjs/types": "*",
     "@types/yargs": "^17.0.2",
+    "arrayify-stream": "^2.0.0",
```
rubensworks, line 49: The added dependencies here seem incorrect. I think these should be dev dependencies? (or even no deps at all?)

jeswr: Its being triggered by this https://github.com/comunica/comunica/blob/23af30008e7696b81b9f52c267c1f0f6a1c1df9b/packages/actor-init-query/spec/sparql-engine-base.js#L22

The script only excludes subfolders matching `test`, should it exclude `spec` and `__mocks__` as well?

rubensworks: Yes, definitely.

But instead of blacklisting, maybe we can use a whitelist? Perhaps we can hook into the `files` entry of the `package.json` file? That should be pretty foolproof.

PR #972, opened 2022-04-07
chore: Remove unecessary TransformIterator mapping

jeswr:


rubensworks:
I think we _do_ need this wrapping, since `match` can not be made async, and `proxyIt` is only determined async.
Unless I'm missing something @jeswr?

jeswr:
Ah, my bad - we should just wrap it when https://github.com/RubenVerborgh/AsyncIterator/pull/63 is merged since this will remove unnecessary buffering behaviour.

rubensworks:
Sounds good!

rubensworks:
Any updates on this?

Issue #969, opened 2022-03-31
Error inserting data into Graph with SPARQL


UniMa007:
#### Issue type:

- :bug: Bug 

____
#### Description:

I've been playing around in the context of blockchain and decentralized knowledge graphs lately and stumbled over your tool.
I really like your 'decentralized RDF' approach and think the web3 world can definitely benefit from your toolset. (Disclaimer: I'm new at nodeJS)

While trying to use the @comunica/query-sparql I wanted to use inserts some data into my sparql endpoint (using blazegraph)

My query is: 

```
                                 PREFIX schema: <http://schema.org/> 
                                  INSERT DATA
                                  { GRAPH <did:dkg:1e62550721611b96321c7459e7790498240431025e46fce9cd99f2ea9763ffb0> 
                                  { <did:dkg:1e62550721611b96321c7459e7790498240431025e46fce9cd99f2ea9763ffb0> schema:hasBlockchain "polygon::testnet"   
                                  } 
                                  }

```

Running this directly inside the blazegraph sparql editor works, now I realized the query using comunica

```
   const askQuery = `ASK WHERE { GRAPH <${rootHash}> { ?s ?p ?o } }`;
        const exists = await this.queryEngine.queryBoolean(askQuery, this.context);

        const insertion = `
                                  PREFIX schema: <http://schema.org/> 
                                  INSERT DATA
                                  { GRAPH <${rootHash}> 
                                  { ${triples}  
                                  } 
                                  }`;
        if (!exists) {
            await this.queryEngine.queryVoid(insertion, this.context);
```

Triggering the queryVpid method will result in the following error:

```
TypeError: Cannot use 'in' operator to search for 'getReader' in 
                                  PREFIX schema: <http://schema.org/> 
                                  INSERT DATA
                                  { GRAPH <did:dkg:90e62550721611b96321c7459e7790498240431025e46fce9cd99f2ea9763ffb0> 
                                  { <did:dkg:90e62550721611b96321c7459e7790498240431025e46fce9cd99f2ea9763ffb0> schema:hasBlockchain "polygon::testnet"   
                                  } 
                                  }
```

Changing the code to use INSERT instead of INSERT DATA makes the code run forever and timeout.
Am I doing something wrong? Seems like the SPARQL Code is working, yet I might call the lib wrong or it is a bug.

Thanks in advance!

____
#### Environment:

Using client SDK : @comunica/query-sparql  

#### Crash log:

See above

UniMa007:
[...]

PR #976, opened 2022-04-11
fix: body of request can be a string

Maximvdw:
Fixes: https://github.com/comunica/comunica/issues/969


rubensworks:
This is great, thanks @Maximvdw!

Maximvdw:
Thanks for the fast merge ;)

PR #981, opened 2022-04-16
Draft: Allow sources to be a promise

jeswr:
This adds an actor that handles sources wrapped in a promise.

The use case for this functionality is to wrap a source that is having reasoning applied to it wrapped in a Promise until reasoning is completed - this allows for query results to be returned from the raw sources _while_ reasoning is taking place, rather than the current behavior of blocking results until reasoning is completed.

### DIFF
--- a/.componentsjs-generator-config.json
+++ b/.componentsjs-generator-config.json
@@ -59,7 +59,9 @@
     "Factory",
 
     "UriTemplate",
-    "LRUCache"
+    "LRUCache",
+
+    "Promise"
   ],
   "modulePrefix": {
     "@comunica/actor-query-operation-service": "caqoser",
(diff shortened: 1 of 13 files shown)


rubensworks (changes requested):
Looks great! No major comments here.

I'm actually (positively) surprised no further changes were needed to make this work.
So the federated actor will start processing non-promise sources even if some promise sources haven't resolved yet?

jeswr:
> So the federated actor will start processing non-promise sources even if some promise sources haven't resolved yet?

I had assumed that was the case - upon adding some integration tests I discovered this was not actually happening.

The key problem is that the `getMetadata` call in each of the joins is a blocking action.

An example of such a point at which it 'blocks' actually appears to be within the `test` method for `ActorRdfJoinMultiEmpty`

```ts
public async test(action: IActionRdfJoin): Promise<IMediatorTypeJoinCoefficients> {
    if ((await ActorRdfJoin.getMetadatas(action.entries))
      .every(metadata => ActorRdfJoin.getCardinality(metadata).value > 0)) {
      throw new Error(`Actor ${this.name} can only join entries where at least one is empty`);
    }
    return super.test(action);
  }
```

rubensworks:
Any updates on this?

Issue #70, opened 2018-01-15
Allow HTTP timeout configuration


rubensworks:
We should allow users to pass an HTTP timeout value via the context (`httpTimeout`).

This could be implemented using our own `setTimeout` and the fetch `AbortController`: https://github.com/bitinn/node-fetch/issues/95

We should keep in mind here that we should clear our own timeout once the request completes (response object is available).

Additionally, we need an extra context option (boolean:  `httpTimeoutOnBody`) to make it so that the timeout not only applies to the time until response starts coming in, but also to the time until the response body is fully available. The latter could take longer, or potentially be infinite for e.g. continuous data streams. This should also take into account that response bodies can be cancelled from within Comunica.

-----

## Bounty

A [bounty](https://comunica.dev/association/bounties/) has been placed on this issue by:

| |
|-|
| [<img alt="Netwerk Digitaal Erfgoed" src="https://user-images.githubusercontent.com/440384/134636829-f730230c-bfec-4d40-86d1-67f28479cdcc.png" height="25">](https://netwerkdigitaalerfgoed.nl/) |
| €1088 |

[Click here to learn more](https://comunica.dev/association/bounty_process/) if you're interested in claiming this bounty by resolving this issue.

rubensworks:
node-fetch now supports cancellation: https://github.com/bitinn/node-fetch#request-cancellation-with-abortsignal

rubensworks:
This is important for cases where the application hangs, such as

> comunica-sparql https://dbpedia.org/sparql "construct where { ?s ?p ?o } limit 5"

rubensworks:
@Tpt Once everything has been implemented, could you document this functionality briefly as a new section on https://comunica.dev/docs/query/advanced/context/ ?

PR #1009, opened 2022-06-15
Adds the "httpTimeout" context option and implement it inside actor-http-fetch

Tpt:
First step to solve #70

### DIFF
--- a/packages/actor-http-fetch/lib/ActorHttpFetch.ts
+++ b/packages/actor-http-fetch/lib/ActorHttpFetch.ts
@@ -58,22 +58,43 @@ export class ActorHttpFetch extends ActorHttp {
       action.init.headers = ActorHttp.headersToHash(action.init.headers);
     }
 
-    // Perform request
-    const customFetch: ((input: RequestInfo, init?: RequestInit) => Promise<Response>) | undefined = action
-      .context?.get(KeysHttp.fetch);
-    return await (customFetch || fetch)(action.input, await this.fetchInitPreprocessor.handle({
-      ...action.init,
-      ...action.context.get(KeysHttp.includeCredentials) ? { credentials: 'include' } : {},
-    })).then(response => {
+    let requestInit = { ...action.init };
+
+    if (action.context.get(KeysHttp.includeCredentials)) {
+      requestInit.credentials = 'include';
+    }
+
+    const httpTimeout: number | undefined = action.context?.get(KeysHttp.httpTimeout);
+    let requestTimeout;
+    if (httpTimeout !== undefined) {
+      const controller = await this.fetchInitPreprocessor.createAbortController();
+      requestInit.signal = controller.signal;
+      requestTimeout = setTimeout(() => controller.abort(), httpTimeout);
+    }
+
+    try {
+      requestInit = await this.fetchInitPreprocessor.handle(requestInit);
+
+      // Perform request
+      const customFetch: ((input: RequestInfo, init?: RequestInit) => Promise<Response>) | undefined = action
+        .context?.get(KeysHttp.fetch);
+
+      const response = await (customFetch || fetch)(action.input, requestInit);
+
       // Node-fetch does not support body.cancel, while it is mandatory according to the fetch and readablestream api.
       // If it doesn't exist, we monkey-patch it.
       if (response.body && !response.body.cancel) {
         response.body.cancel = async(error?: Error) => {
-          (<Readable> <any> response.body).destroy(error);
+          (<Readable><any>response.body).destroy(error);
         };
       }
       return response;
-    });
+    } finally {
+      // We abort timeouts
+      if (requestTimeout !== undefined) {
+        clearTimeout(requestTimeout);
+      }
+    }
   }
 }
 
--- a/packages/actor-http-fetch/lib/FetchInitPreprocessor-browser.ts
+++ b/packages/actor-http-fetch/lib/FetchInitPreprocessor-browser.ts
@@ -38,4 +38,8 @@ export class FetchInitPreprocessor implements IFetchInitPreprocessor {
     // Only enable keepalive functionality if we are not sending a body (some browsers seem to trip over this)
     return { keepalive: !init.body, ...init };
   }
+
+  public async createAbortController(): Promise<AbortController> {
+    return new AbortController();
+  }
 }
--- a/packages/actor-http-fetch/test/ActorHttpFetch-test.ts
+++ b/packages/actor-http-fetch/test/ActorHttpFetch-test.ts
@@ -338,5 +338,28 @@ describe('ActorHttpFetch', () => {
       expect(fetch).not.toHaveBeenCalled();
       expect(customFetch).toHaveBeenCalled();
     });
+
+    it('should work with a large timeout', async() => {
+      await expect(actor.run({
+        input: <Request> { url: 'https://www.google.com/' },
+        context: new ActionContext({ [KeysHttp.httpTimeout.name]: 100_000 }),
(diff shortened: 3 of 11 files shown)


packages/actor-http-fetch/test/ActorHttpFetch-test.ts
```diff
@@ -338,5 +338,26 @@ describe('ActorHttpFetch', () => {
...
     });
+
+    it('should work with a timeout', async() => {
+      await expect(actor.run({
+        input: <Request> { url: 'https://www.google.com/' },
+        context: new ActionContext({ [KeysHttp.httpTimeout.name]: 100 }),
+      })).resolves.toMatchObject({ status: 200 });
+    });
+
+    it('should abort properly with a timeout', async() => {
```
Tpt, line 349: This test is a bit fragile and very slow because it relies on timeout behavior.
I am not sure how to write it differently without a lot of dependency injection.

packages/actor-http-fetch/lib/ActorHttpFetch.ts
```diff
@@ -58,22 +58,44 @@ export class ActorHttpFetch extends ActorHttp {
...
+    }
+
+    const httpTimeout: number | undefined = action.context?.get(KeysHttp.httpTimeout);
+    let setupHttpTimeout: (() => void) = () => {
+      // No op by default
+    };
+    if (httpTimeout !== undefined) {
+      // Fallback to abort-controller for Node 14 backward compatibility
+      /* istanbul ignore next */
+      const AbortController = global.AbortController || await import('abort-controller');
```
rubensworks, line 74: I would recommend moving this logic into [FetchInitPreprocessor.ts](https://github.com/comunica/comunica/blob/master/packages/actor-http-fetch/lib/FetchInitPreprocessor.ts) and [FetchInitPreprocessor-browser.ts](https://github.com/comunica/comunica/blob/master/packages/actor-http-fetch/lib/FetchInitPreprocessor-browser.ts), since `AbortController` is guaranteed to exist in the browser, while for Node we may indeed need a polyfill.

That reminds me, we definitely need to test this in a browser environment as well (manually, since we don't have proper unit tests for browsers yet).

packages/actor-http-fetch/lib/ActorHttpFetch.ts
```diff
@@ -58,22 +58,44 @@ export class ActorHttpFetch extends ActorHttp {
...
+      const controller = new AbortController();
+      requestInit.signal = controller.signal;
+      setupHttpTimeout = () => {
+        setTimeout(() => controller.abort(), httpTimeout);
```
rubensworks, line 78: We need to save the timeout handler, and clear it if the response comes in before the timeout.
Otherwise, the process will hang if queries end faster than this timeout.

packages/actor-http-fetch/lib/ActorHttpFetch.ts
```diff
@@ -58,22 +58,44 @@ export class ActorHttpFetch extends ActorHttp {
...
-      // If it doesn't exist, we monkey-patch it.
-      if (response.body && !response.body.cancel) {
-        response.body.cancel = async(error?: Error) => {
-          (<Readable> <any> response.body).destroy(error);
-        };
-      }
-      return response;
-    });
+    const preprocessedRequestInit = await this.fetchInitPreprocessor.handle(requestInit);
+    // We setup the timout here to avoid waiting while we await the preprocessor
```
rubensworks, line 86: I wouldn't worry too much about the preprocessor delay, we can assume it's constant. So I would suggest applying the timeout immediately (to reduce complexity).

packages/context-entries/lib/Keys.ts
```diff
@@ -39,6 +39,10 @@ export const KeysHttp = {
    * Fetch function implementation.
    */
   fetch: new ActionContextKey<typeof fetch>('@comunica/bus-http:fetch'),
+  /**
+   * HTTP request timeout in milliseconds.
+   */
+  httpTimeout: new ActionContextKey<number>('@comunica/bus-http:http-timeout'),
```
rubensworks, line 45: We'll also want to expose this via the CLI.
This can be done here:
- https://github.com/comunica/comunica/blob/master/packages/actor-init-query/lib/ActorInitQueryBase.ts#L101-L118
- https://github.com/comunica/comunica/blob/master/packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts#L78

packages/actor-http-fetch/test/ActorHttpFetch-test.ts
rubensworks: We can make it faster using Jest's timer mocks: https://jestjs.io/docs/timer-mocks

rubensworks (changes requested):

packages/context-entries/lib/Keys.ts
Tpt: Sure!

packages/actor-http-fetch/lib/ActorHttpFetch.ts
Tpt: Genuine question: Are you sure? The "should work with a timeout" test seems to pass even if the response comes in before the timeout. I also made some real world hand tests that did not exposed this problem either. But I'm maybe missing something.

rubensworks: I'm pretty sure, yes. I remember I had to do similar things in the past when working with timeouts.
Unless something changed in Node recently, but I can't find any news on that AFAICS.

You could try by setting the timeout to a very high value (e.g. 1 minute). If you then issue a query, then the query results will arrive quickly, but the process will only end after one minute, while it is allowed to exit earlier.

Tpt: Indeed, my bad.  Thank you!

packages/actor-http-fetch/lib/ActorHttpFetch.ts
Tpt: I left the logic in `ActorHttpFetch` to make the cancelation easier but added a provider for "AbortController" to `FetchInitPreprocessor`. I hope it's fine for you.

packages/actor-http-fetch/lib/FetchInitPreprocessor-browser.ts
```diff
@@ -38,4 +38,8 @@ export class FetchInitPreprocessor implements IFetchInitPreprocessor {
     // Only enable keepalive functionality if we are not sending a body (some browsers seem to trip over this)
     return { keepalive: !init.body, ...init };
   }
+
+  public async abortController(): Promise<new () => AbortController> {
```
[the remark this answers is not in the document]
Tpt: Done.

packages/actor-http-fetch/lib/ActorHttpFetch.ts
```diff
@@ -58,22 +58,43 @@ export class ActorHttpFetch extends ActorHttp {
...
       // If it doesn't exist, we monkey-patch it.
       if (response.body && !response.body.cancel) {
         response.body.cancel = async(error?: Error) => {
-          (<Readable> <any> response.body).destroy(error);
+          (<Readable><any>response.body).destroy(error);
         };
       }
       return response;
-    });
+    } finally {
```
Tpt, line 92: Not very nice but needed to ensure that the `requestTimeout` is cleared even if the request fails with an exception.

rubensworks: Ah yes, good catch. I don't see a cleaner solution either, so this is fine :-)

rubensworks:
Perfect, thanks!

Issue #166, opened 2018-07-03
HTTP caching


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request

____
#### Description:

We should add support for proper client-side caching of HTTP resources.
We should check both the `Expires`, `Cache-Control`, `ETag` and `Last-Modified` headers, and appropriately send requests with `If-None-Match`/`If-Modified-Since` if needed.

We should make the cache actor work on the http bus.

pietercolpaert:
I believe we should rely on https://www.npmjs.com/package/http-cache-semantics

The cache probably should not be mediated, but it should be statically configured: e.g., if you have a redis cache, you will need a config file to define how to connect to this, or maybe you’d rather want to use something like [LevelDB](https://github.com/Level/levelup) or [cacache](https://www.npmjs.com/package/cacache). As long as somethin adheres to https://developer.mozilla.org/en-US/docs/Web/API/Cache, I would allow it through a kind of strategy pattern that reads a config. What do you think?

rubensworks:
That package looks like it might do the trick.

> The cache probably should not be mediated, but it should be statically configured

If with "not mediated", you mean that it should not be on a separate "caching" bus, then I agree :-) But it should still listen on the HTTP bus IMO.

We could have this caching actor observe all HTTP calls passively (which will cache all performed HTTP requests), and also actively be registered on the HTTP bus (for replying with results from cache).

> a kind of strategy pattern that reads a config

Yes, that sounds good. The caching storage strategy (redis, LevelDB, ...) could be defined in the config file via a parameter. We'd then need separate _caching storage_ components (not necessarily actors (I think)) as well.

rubensworks:
To consider: https://github.com/solid/web-access-control-spec/pull/67

jaxoncreed:
As discussed with @rubensworks, I will work on this issue via the Comunica Association.

jaxoncreed:
I've encountered a few problems with this task:

> A simple default LRUCache-based Cache implementation should be set here if the user provided none in the context: https://github.com/comunica/comunica/blob/master/packages/actor-init-query/lib/QueryEngineBase.ts#L110
It can be implemented using https://www.npmjs.com/package/http-cache-semantics

The `http-cache-semantics` library is primarily concerned with the "cache-control" header and the web-cache interface doesn't care about that (for example see https://w3c.github.io/ServiceWorker/#cache-addAll the only thing it cares about is the status of the response and whether or not its http)

[...]

PR #1010, opened 2022-06-16
Feat/http cache

jaxoncreed:
Implementation for https://github.com/comunica/comunica/issues/166


packages/http-cache-storage-lru/lib/HttpCacheStorageLru.ts
```diff
@@ -0,0 +1,34 @@
+import type { IHttpCacheStorage, IHttpCacheStorageValue } from '@comunica/http-cache-storage';
+import * as LRU from 'lru-cache';
```
rubensworks, line 2: @jeswr Any idea why the depcheck task might be reporting `lru-cache` as an extra dependency that should be removed?

https://github.com/comunica/comunica/runs/6961732697?check_suite_focus=true

jeswr: It seems that the [associated package.json](https://github.com/comunica/comunica/blob/95d7313f7454c7de25fad7c501394c80b4684617/packages/http-cache-storage-lru/package.json) is missing the "files" parameter which depcheck is using to determine which files to include dependencies from.

rubensworks: Aha, thanks @jeswr!

@jaxoncreed It looks like some other package.json entries are missing as well. This might be a good reference: https://github.com/comunica/comunica/blob/95d7313f7454c7de25fad7c501394c80b4684617/packages/types/package.json

jaxoncreed: Got it

jaxoncreed:
@rubensworks  I've tried it again, and it still fails for 14.x and 16.x. You said before that this can be ignored. Is it time to check this in?

rubensworks:
Thanks @jaxoncreed!
I will look into this soon!

packages/actor-http-cache/lib/ActorHttpCache.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  /**
+   * Fetches the document and splits it into parts
+   * @param action Fetch action
+   * @returns
+   */
+  private async fetchDocument(action: IActionHttp): Promise<[bodyInit?: BodyInit, responseInit?: ResponseInit]> {
+    const response = await this.actorHttpFetch.run(action);
+    return [
+      await response.text(),
```
rubensworks, line 123: We will need a more elaborate method for copying response bodies.

Streamability is an important element within Comunica's design.
This means that responses should be parseable in a streaming manner.
One the one hand, this is so that Comunica can already start processing parts of responses from slow servers.
On the other hand, this also means that Comunica can handle infinitely long documents (such as server-sent events).

Therefore, we should copy the body while preserving its streaming nature.
I suspect this one may help here: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/tee

packages/actor-http-cache/lib/headerConversionHelpers.ts
```diff
@@ -0,0 +1,66 @@
...
+  response: Response,
+): CacheResponse {
+  return { ...response, headers: headersToHash(response.headers) };
+}
+
+/**
+ * Convert the given headers object into a raw hash.
+ * @param headers A headers object.
+ */
+export function headersToHash(headers: Headers): HeaderHash {
```
rubensworks, line 35: I guess this one can be removed in favor of https://github.com/comunica/comunica/blob/master/packages/bus-http/lib/ActorHttp.ts#L56

packages/actor-http-cache/lib/headerConversionHelpers.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Converts a header hash into a standard header
+ * @param hash the hash to be converted into a header
+ */
+export function addHashHeadersToObject(
+  hash: HeaderHash,
+  reqRes: Request | Response,
+): void {
+  Object.entries(hash).forEach(([ headerKey, headerValue ]) => {
```
rubensworks, line 51: It might be possible to simplify this by first converting the hash to a `Headers` object using `new Headers(hash)`, and then iterating+appending.

packages/http-cache-storage-lru/.npmignore
```diff
@@ -0,0 +1,2 @@
+!dist
```
rubensworks, line 1: This file can remain empty.

packages/http-cache-storage-lru/Readme.md
```diff
@@ -0,0 +1,18 @@
+# Comunica Http Cache Storage
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Fhttp-cache-storage-lru.svg)](https://www.npmjs.com/package/@comunica/http-cache-storage-lru)
+
+A comunica implementation of cache-storage using a Line Replacement Unit implementation.
```
rubensworks, line 5: Least Recently Used

packages/actor-http-cache/package.json
```diff
@@ -0,0 +1,49 @@
...
+  },
+  "homepage": "https://comunica.dev/",
+  "files": [
+    "components",
+    "lib/**/*.d.ts",
+    "lib/**/*.js"
+  ],
+  "devDependencies": {
+    "@comunica/http-cache-storage": "^0.0.1",
+    "@types/http-cache-semantics": "^4.0.1"
```
rubensworks, line 35: Can be an actual dependency

packages/actor-http-cache/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "url": "https://github.com/comunica/comunica/issues"
+  },
+  "homepage": "https://comunica.dev/",
+  "files": [
+    "components",
+    "lib/**/*.d.ts",
+    "lib/**/*.js"
+  ],
+  "devDependencies": {
+    "@comunica/http-cache-storage": "^0.0.1",
```
rubensworks, line 34: Must be an actual dependency

packages/http-cache-storage/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "url": "https://github.com/comunica/comunica/issues"
+  },
+  "homepage": "https://github.com/comunica/comunica#readme",
+  "files": [
+    "components",
+    "lib/**/*.d.ts",
+    "lib/**/*.js"
+  ],
+  "devDependencies": {
+    "http-cache-semantics": "^4.1.0"
```
rubensworks, line 29: Can be an actual dependency.

packages/http-cache-storage/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@comunica/http-cache-storage",
```
rubensworks, line 2: I guess we don't need a separate package for this one.
We can just move the interface to `@comunica/types`.

packages/actor-http-cache/lib/ActorHttpCache.ts
```diff
@@ -0,0 +1,205 @@
...
+ * A comunica Cache Http Actor.
+ */
+export class ActorHttpCache extends ActorHttp {
+  private readonly cacheStorage: IHttpCacheStorage;
+  private readonly mediatorHttpInvalidate: MediatorHttpInvalidate;
+  private readonly actorHttpFetch: ActorHttp;
+
+  public constructor(args: IActorHttpCacheArgs) {
+    super(args);
+    this.cacheStorage = args.cacheStorage;
```
rubensworks, line 34: As described in #166, we should also cover the case where a custom `cacheStorage` is supplied by the user inside the context object (in the `run` method).

(feel free to ask me for pointers if this is not fully clear)

rubensworks (changes requested):
Great work!
I think all necessary parts are in here already, some things just need moving around a bit.

@pietercolpaert could you double-check on your end that this does what you had in mind for #166?

pietercolpaert:
@brechtvdv Can you validate this and check whether this would be usable within the LDES client?

@jaxoncreed In the browser I believe you cannot read the cache-control/etag headers when doing HTTP requests: this is only a NodeJS contribution, right?

brechtvdv:
> 

Yes, actor-http-cache would allow us to rewrite the getPage function (https://github.com/TREEcg/event-stream-client/blob/main/packages/actor-init-ldes-client/lib/EventStream.ts#L512).
 @jaxoncreed: does it matter whether its an HTTP or HTTPS request? 
One thing that we need is follow-redirects. This is probably not handled currently?

packages/actor-http-cache/lib/ActorHttpCache.ts
jaxoncreed: Did something change from the conversation here https://github.com/comunica/comunica/issues/166#issuecomment-1151448769? I'm happy to implement a custom cache storage, but at the moment, I don't see how it would work.

packages/actor-http-cache/lib/ActorHttpCache.ts
```diff
@@ -0,0 +1,205 @@
...
+  addHashHeadersToObject,
+} from './headerConversionHelpers';
+
+/**
+ * A comunica Cache Http Actor.
+ */
+export class ActorHttpCache extends ActorHttp {
+  private readonly cacheStorage: IHttpCacheStorage;
+  private readonly mediatorHttpInvalidate: MediatorHttpInvalidate;
+  private readonly actorHttpFetch: ActorHttp;
```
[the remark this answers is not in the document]
jaxoncreed: Ah yes, okay I've fixed that.

packages/actor-http-cache/README.md
```diff
@@ -0,0 +1,41 @@
...
+      "@id": "urn:comunica:default:http/actors#cache",
+      "@type": "ActorHttpCache"
+    }
+  ]
+}
+```
+
+### Config Parameters
+
+TODO: fill in parameters (this section can be removed if there are none)
```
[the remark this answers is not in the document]
jaxoncreed: Fixed

jaxoncreed:
@pietercolpaert Yes, it would seem that this would only work in a NodeJS environment, so I guess you should have a separate configuration for an in-browser environment.

@brechtvdv No follow-redirects are not handled

packages/actor-http-cache/lib/ActorHttpCache.ts
jaxoncreed: I tried redoing this using `tee`. Unfortunately, NodeJS bodies don't have a `tee` method and causes the error: `TypeError: response.body?.tee is not a function`

rubensworks: Oh, that's annoying, but not surprising actually, given [the workaround we have to add `body.cancel`](https://github.com/comunica/comunica/blob/master/packages/actor-http-fetch/lib/ActorHttpFetch.ts#L100-L110).

We really need to keep streamability though. One low-effort solution would be to monkey-patch body (like `body.cancel`) to add `tee` support, which can be implemented in the usual way of non-blocking cloning Node streams. What do you think?

packages/actor-http-cache/lib/ActorHttpCache.ts
rubensworks: > Did something change from the conversation here https://github.com/comunica/comunica/issues/166#issuecomment-1151448769?

No, that still holds.

I did not mean that we should implement a custom cache, but that it should be possible for users to pass a custom cache (that implements the new `IHttpCacheStorage` interface) via the context. In the same way users can provide options such as baseIri and so on.

rubensworks:
> @pietercolpaert Yes, it would seem that this would only work in a NodeJS environment, so I guess you should have a separate configuration for an in-browser environment.

Could you look into creating such a separate configuration? So that we don't break browser builds when we merge this.
I guess we can follow a similar approach as we do for `@comunica/query-sparql-file`, where we just have a browser config, and a node config that extends from it and includes the cache actor.
See https://github.com/comunica/comunica/tree/master/engines/config-query-sparql/config, https://github.com/comunica/comunica/blob/master/engines/query-sparql-file/package.json#L174, https://github.com/comunica/comunica/blob/master/engines/query-sparql-file/package.json#L178-L181, and 

> @brechtvdv No follow-redirects are not handled

If the http-fetch actor is configured properly, then redirects will be followed correctly.

rubensworks:
Note to self: before merging, we should do an in-depth manual check to see if everything still works correctly with this. For example, we should test and inspect TPF queries.

packages/actor-http-cache/lib/ActorHttpCache.ts
jaxoncreed: Okay, I think that I've fixed it using NodeJS's streams

packages/actor-http-cache/lib/headerConversionHelpers.ts
jaxoncreed: Yep, done

packages/actor-http-cache/lib/ActorHttpCache.ts
jaxoncreed: Got it. Updated

jaxoncreed:
@rubensworks Okay, I think I've covered all the comments.

engines/config-query-sparql/config/http/cache-storage.json
```diff
@@ -0,0 +1,13 @@
+{
+  "@context": [
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/mediator-number/^2.0.0/components/context.jsonld",
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/http-cache-storage-lru/^0.0.0/components/context.jsonld"
+  ],
+  "@graph": [
+    {
+      "@id": "urn:comunica:default:http/cache-storage#main",
+      "@type": "HttpCacheStorageLru",
+      "args_max": 1000
```
rubensworks, line 10:
```suggestion
      "args_max": 1024
```

packages/actor-http-cache/README.md
```diff
@@ -0,0 +1,38 @@
...
+    {
+      "@id": "urn:comunica:default:http/actors#cache",
+      "@type": "ActorHttpCache",
+      "cacheStorage": { "@id": "urn:comunica:default:http/cache-storage#main" },
+      "mediatorHttpInvalidate": { "@id": "urn:comunica:default:http-invalidate/mediators#main" },
+      "mediatorHttp": { "@id": "urn:comunica:default:http/mediators#main" }
+    },
+  ]
+}
+```
```
rubensworks, line 38: The convention is to explicitly list config parameters in the readme, like here: https://github.com/comunica/comunica/tree/master/packages/actor-http-proxy
Could we add them here as well?

packages/context-entries/lib/Keys.ts
```diff
@@ -64,6 +65,17 @@ export const KeysHttpProxy = {
...
 
+export const KeysHttpCache = {
+  /**
+   * Will skip the HTTP cache if this is set to true
+   */
+  doNotCheckHttpCache: new ActionContextKey<boolean>('@communica/actor-http-cache:do-not-check-http-cache'),
+  /**
+   * A storage object that can overwrites the default http cache
+   */
+  httpCacheStorage: new ActionContextKey<IHttpCacheStorage>('@communica/actor-http-cache:http-cache-storage'),
```
rubensworks, line 76: This one can be moved to `KeysHttp`, and 
```suggestion
  httpCacheStorage: new ActionContextKey<IHttpCacheStorage>('@communica/bus-http:cache-storage'),
```

packages/http-cache-storage-lru/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "main": "lib/index.js",
+  "typings": "lib/index",
+  "scripts": {
+    "build": "npm run build:ts && npm run build:components",
+    "build:ts": "node \"../../node_modules/typescript/bin/tsc\"",
+    "build:components": "componentsjs-generator"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/comunica/comunica.git"
```
rubensworks, line 15: `"directory"` should be included in here, like the other packages.

packages/http-cache-storage-lru/lib/HttpCacheStorageLru.ts
```diff
@@ -0,0 +1,34 @@
+import type { IHttpCacheStorage, IHttpCacheStorageValue } from '@comunica/types';
+import * as LRU from 'lru-cache';
+
+export class HttpCacheStorageLru implements IHttpCacheStorage {
+  private readonly cache: LRU<string, IHttpCacheStorageValue>;
+
+  public constructor(args: IHttpCacheLruArgs) {
+    this.cache = new LRU({ max: args.max });
```
rubensworks, line 8: Since this LRU cache can also independently evict cache entries once the cache becomes full, we probably also want to link the `dispose` method of this cache to the HTTP invalidation bus.

packages/actor-http-cache/lib/ActorHttpCache.ts
```diff
@@ -0,0 +1,233 @@
...
+
+  private async teeAndCacheResponse(
+    request: Request,
+    response: Response,
+    cachePolicy: CachePolicy,
+    cacheStorage: IHttpCacheStorage,
+  ): Promise<Response> {
+    // Node-fetch does not support body.tee, while it is mandatory according to the fetch and readablestream api.
+    // If it doesn't exist, we monkey-patch it.
+    if (response.body && !response.body.tee) {
```
rubensworks, line 172: Can we move this monkey-patch to after this line [here](https://github.com/comunica/comunica/blob/e89bc9f943c7358159eb720d1f638d092df2bef5/packages/actor-http-fetch/lib/ActorHttpFetch.ts#L110)?
Just so that we have all the fetch-related monkey-patching in a single place.

packages/actor-http-cache/lib/ActorHttpCache.ts
```diff
@@ -0,0 +1,233 @@
...
+      response.body.tee = (): [ReadableStream, ReadableStream] => {
+        // eslint-disable-next-line import/no-nodejs-modules
+        const Readable = require('stream').Readable;
+        return [
+          Readable.from(response.body),
```
rubensworks, line 177: Not sure this way of cloning is fully safe.
Just learned about this package though, which looks like it meets all of our requirements: https://www.npmjs.com/package/cloneable-readable

packages/actor-http-cache/lib/ActorHttpCache.ts
```diff
@@ -0,0 +1,233 @@
...
+          Readable.from(response.body),
+        ];
+      };
+    }
+
+    // There almost will always be a body, so ignore next
```
rubensworks, line 183: I'm not a big fan of ignoring coverage :P
Shouldn't be too much effort to test this case as well?

rubensworks (changes requested):
Thanks for the changes @jaxoncreed.
Just a few minor issues here and there.

After these are resolved, I'll do some manual tests with this locally on my machine, and then things should be ok to merge :-)

packages/actor-http-cache/lib/ActorHttpCache.ts
jaxoncreed: Unfortunately, the value of Body from the NodeJS fetch doesn't seem to be a stream. It's a `Buffer`. So, this library will not work. I'll look for other solutions. What are the requirements that need to be met?

jaxoncreed: clonedStream

jaxoncreed: I see. In order to make a `Response` object and maintain the stream that was passed in as the body, the `body instanceof Stream` must evaluate to `true`. This library is of type `Cloneable` which does not evaluate to stream. As discussed here (https://stackoverflow.com/questions/9898875/good-way-to-change-an-objects-prototype-to-change-results-of-instanceof), there is no way to modify the prototype of an object while maintaining its classes.

rubensworks: > In order to make a Response object and maintain the stream that was passed in as the body, the body instanceof Stream must evaluate to true

Where does this cause an error exactly? Is it a typing error? Or a runtime error (in the Response constructor)?

jaxoncreed: When you construct a `Request` object and pass in a stream, it will go here (https://github.com/node-fetch/node-fetch/blob/main/src/body.js#L53). However, if the object that you provide is not an `instanceof` stream, the code will hop here (https://github.com/node-fetch/node-fetch/blob/main/src/body.js#L59), and automatically turn the stream into a buffer. Because this is behavior built into the request object, it can't be changed. Perhaps we could build an entirely new request object that follows the same API. I'll try that and see if it works.

jaxoncreed: Interestingly, the following works in a clean project, but doing the same thing in this project doesn't work. 

```
const stream = require("stream");
import "cross-fetch/polyfill";

const store = [];

async function main() {
  const response = await fetch("https://example.com");
  // @ts-ignore
  const stream1 = response.body.pipe(new stream.PassThrough());
  // @ts-ignore
  const stream2 = response.body.pipe(new stream.PassThrough());

  // @ts-ignore
  stream1.on("data", (chunk: any) => {
    console.log("chunk 1");
    console.log(chunk.toString());
  });
  // @ts-ignore
  stream1.on("end", () => {
    console.log("Stream 1 Ended");
  });

  setTimeout(() => {
    stream2.on("data", (chunk: any) => {
      console.log("chunk 2");
      console.log(chunk.toString());
    });
    stream2.on("end", () => {
      console.log("Stream 2 Ended");
    });
  }, 5000);
}

main();
```

jaxoncreed: @rubensworks I've done some experimenting with stream cloning techniques. In isolation, this one seems to work (https://github.com/jaxoncreed/clone-stream-demo). You can clone this repo and check it yourself.

But, implementing this solution inside communica causes problems. I've created a demo of that here (https://github.com/jaxoncreed/comunica/pull/1/files). The `end` event is never called if two streams are created. (You can see that it is successful by removing the second clone line here https://github.com/jaxoncreed/comunica/pull/1/files#diff-032bb79de64c61798f786bec348f8bbabe74a11da61cafbde03f44aeb797305eR92)

I can't figure out why this solution works in the first codebase, but not the second. Do you have any ideas?

rubensworks: Most tests use a minimal body mock: https://github.com/jaxoncreed/comunica/blob/9eddbeae4488a5e0e349d4ecb7161cc7935a0535/packages/actor-http-fetch/test/ActorHttpFetch-test.ts#L17

For tests that don't override the mocked body, I guess this could explain hanging streams?

jaxoncreed: @rubensworks I'll need to update those tests in the future, so I agree that the unit tests for the project are a problem, but the problem persists with integrations tests as well. For example, I'm running the following to test the app: `jest engines/query-sparql/test/QuerySparql-test.ts -t "for the single source context entry" --coverage false` and it still hangs.

rubensworks: I vaguely remember hanging Node problems when the body was left unconsumed.
Perhaps the problem is that the cloned and cached body is never consumed, which makes Node hang?
You can probably easily test this by removing the data and end listeners on the second stream of your clone-stream-demo.

If this is the problem, then we can't really force stream consumption after end I think, as this would defeat the purpose of the cache. But perhaps there is a way to make that stream not hang.

jaxoncreed: Ah, indeed, that is the case. Removing the stream listeners for the second stream will cause the demo to hang for requests that have more than one chunk.

But, there's some more bad news. The problem goes beyond the fact that `stream2` keeps Node running. It also keeps the the subsequent chunks from triggering events on `stream1`.

For example given the following code and a stream with 3 chunks...

```typescript
  const stream1 = response.body.pipe(new stream.PassThrough());
  const stream2 = response.body.pipe(new stream.PassThrough());

  let stream1ChunkCount = 0;
  stream1.on("data", (chunk: any) => console.log(`stream1 chunk ${++stream1ChunkCount}`));
  stream1.on("end", () => console.log("Stream 1 Ended"));

  setTimeout(() => {
    let stream2ChunkCount = 0;
    stream2.on("data", (chunk: any) => console.log(`stream2 chunk ${++stream2ChunkCount}`));
    stream2.on("end", () => console.log("Stream 2 Ended"));
  }, 5000);
```

This is logged
```
stream1 chunk 1
<-- Waits for 5 seconds -->
stream1 chunk 2
stream2 chunk 1
stream2 chunk 2
stream1 chunk 3
stream2 chunk 3
Stream 1 Ended
Stream 2 Ended
```

So, the streams, while cloned, are still intertwined.

rubensworks: > But, there's some more bad news. The problem goes beyond the fact that stream2 keeps Node running. It also keeps the the subsequent chunks from triggering events on stream1.

I see that you're snippet is using the `PassThrough`-approach for cloning, so this behaviour makes sense.

Wasn't it possible to make it work using `cloneable-readable`? Because I suspect that one should solve all of the problems we have. We just may have to create a custom `Stream` wrapper for it ourselves.

jaxoncreed: No the `cloneable-readable` causes the same result. The stream will not trigger events until both the clone and the original have registered their listeners:

For example, the following code

```typescript
const cloneableBody = cloneable(response.body);
  const stream1 = cloneableBody.clone();
  const stream2 = cloneableBody;

  let stream1ChunkCount = 0;

  stream1.on("data", (chunk: any) => {
    console.log(`stream1 chunk ${++stream1ChunkCount}`);
  });
  stream1.on("end", () => {
    console.log("Stream 1 Ended");
  });

  setTimeout(() => {
    console.log('Timeout')
    let stream2ChunkCount = 0;
    stream2.on("data", (chunk: any) => {
      console.log(`stream2 chunk ${++stream2ChunkCount}`);
    });
    stream2.on("end", () => {
      console.log("Stream 2 Ended");
    });
  }, 5000);
```
prints
```
Timeout Complete
stream1 chunk 1
stream2 chunk 1
stream1 chunk 2
stream2 chunk 2
stream1 chunk 3
stream2 chunk 3
Stream 2 Ended
Stream 1 Ended
```

Stream 1 doesn't start sending chunks until after the timeout is complete

jaxoncreed: @rubensworks what do you think we should do? It doesn't look possible to keep a stream suspended while using another stream.

rubensworks: Hmm, that's too bad. I thought `cloneable-readable` would enable this.

I would suggest to solve the problem by fully consuming the stream in `HttpCacheStorageLru`, and storing it directly (e.g. via a `Buffer`). So I would keep the `cloneable-readable` logic, as we'll need to duplicate the stream in any case. But instead of storing the stream, we store the data inside the stream. And when the data is requested from the cache, we replay the stream.

Some things to keep in mind for this:
- We should make sure that the replayed streams are lazy, and don't start flowing before data listeners are attached. This is to avoid expensive streams to start if only things like headers or so are read from the cache.
- We should add a (configurable?) limit on the number of bytes we store from streams. And when the body stream exceeds that threshold, we should just not cache this request. This is to avoid memory issues with very large or infinite files.
- We should be able to handle cases where a cached body is requested while the cached body is still in progress of being read. We can do this by storing the body in 2 phases. 1) While the body that we are caching is being read to store in the `HttpCacheStorageLru`, we temporarily store a reference to this stream, so that we can make another clone of it when it is being read from cache. 2) Once the body stream that we are caching has ended, we stop storing this stream's reference, and instead store the bytes directly, so that we can replay it as a stream once retrieved from cache.

Does this make sense?

jaxoncreed: Okay. I'll figure it out.

jaxoncreed: All right, I whipped up a demo that I think satisfies all the requirements. @rubensworks  would you mind looking it over to confirm that this is what you want me to do? If that looks good, I'll get to work on refactoring and writing tests:

https://github.com/jaxoncreed/clone-stream-demo/blob/buffer-solution/lib/index.ts

This code logs the following:
```
Request made
stream1 chunk 1
Short Term Cached value is needed
shortTermCachedStream chunk 1
stream1 chunk 2
shortTermCachedStream chunk 2
stream1 chunk 3
shortTermCachedStream chunk 3
Stream 1 Ended
shortTermCachedStream Ended
Request is too big to store
Timeout complete. Long Term Cached value is needed
longTermCachedStream chunk 1
longTermCachedStream Ended
All stream values are the same: true
```

rubensworks: @jaxoncreed The approach in your demo looks very good to me!

Some thoughts:
- I guess we can use `cloneable-readable` instead of passthrough-based cloning.
- The short-term caching approach is probably very sensitive to bugs, so we'll need some good unit tests to check all edge-cases.

jaxoncreed: Sounds good. A few quick questions:
* What is wrong with using the "stream" pass-through solution? Is it because of concerns over NodeJS libraries being available in browser, in which case I suggest using the [readable stream library](https://www.npmjs.com/package/readable-stream).
* Here's a list of the edge cases that I think need to be tested for the short term cache. Can you think of anymore?:
   * Cache needed when the request has been sent, but even the headers haven't been returned yet (synchronous - the request for the cache was made in the same closure as the first request)
   * Cache needed when the request has been sent, but even the headers haven't been returned yet (asynchronous - the request for the cache was made in a different closure separated by some kind of 0second await)
   * Cache needed when the request has come in, and there is no body (synchronous)
   * Cache needed when the request has come in, and there is no body (asynchronous)
   * Cache is needed when the first part of a single chunk body has come in (synchronous)
   * Cache is needed when the first part of a single chunk body has come in (asynchronous)
   * Cache is needed when the first part of a multi-chunk body has come in (synchronous)
   * Cache is needed when the first part of a multi-chunk body has come in (asynchronous)
   * Cache is needed when the middle part of a multi-chunk body has come in (synchronous)
   * Cache is needed when the middle part of a multi-chunk body has come in (asynchronous)
   * Cache is needed when the final part of a multi-chunk body has come in (synchronous)
   * Cache is needed when the middle part of a multi-chunk body has come in (asynchronous)

jaxoncreed:
Hey @rubensworks sorry for the delay. I've now refactored the code. There's a few things to point out:

Firstly, remember that `request.body` is, in actuality, either a `ReadableStream` or a `NodeJS.Readable` when it comes in, depending on the environment. [Later in the code](https://github.com/jaxoncreed/comunica/blob/feat/http-cache/packages/actor-dereference-http/lib/ActorDereferenceHttpBase.ts#L92), it is normalized to a `NodeJS.Readable`, but down in the HttpLayer it remains in the superposition of being one or the other. For this code to work, `response.body` requires methods that are exclusive to both `ReadableStream` (`body.cancel`, `body.tee`) and `NodeJS.Readable` (`body.on`). Would it make sense to normalize the body type to `NodeJS.ReadableStream` (ie in `ActorHttpFetch`)? Keeping it in this superposition requires a lot of forced casting and ts-ignores.

Secondly, Components.js is not able to recognize built-in web types like `Buffer` or `ReadableStream`. I've left these blank for now, but left a comment where they can't be used (https://github.com/jaxoncreed/comunica/blob/feat/http-cache/packages/actor-http-cache/lib/HttpCacheStorageStream.ts#L149)

rubensworks:
> Hey @rubensworks sorry for the delay. I've now refactored the code.

Super!
Is this ready for review?

> Would it make sense to normalize the body type to NodeJS.ReadableStream (ie in ActorHttpFetch)? Keeping it in this superposition requires a lot of forced casting and ts-ignores.

We should only make assumptions about the (non-Node) `ReadableStream` interface, so I would not do this.
Once we migrate away from `node-fetch` (probably only after Node < 18 LTS support is dropped), we'll be able to migrate fully to fetch interfaces, and the weird superposition of `ReadableStream` and `NodeJS.Readable` will be a thing of the past.

> Secondly, Components.js is not able to recognize built-in web types like Buffer or ReadableStream

You can add these to https://github.com/comunica/comunica/blob/master/.componentsjs-generator-config.json

jaxoncreed:
Fantastic, I've updated the comunica issue.

And yep, we're ready for a review.

packages/actor-http-cache/lib/HttpCacheStorageStream.ts
```diff
@@ -0,0 +1,146 @@
...
+      const incompleteStreamInfo = this.incompleteStreams[key.url];
+      let passThrough = new PassThrough();
+      const currentBufferStream = Readable.from(Buffer.concat(incompleteStreamInfo.buffers));
+      passThrough = currentBufferStream.pipe(passThrough, { end: false });
+      // @ts-expect-error In actuality, this is a NodeJS.Readable
+      passThrough = incompleteStreamInfo.stream.pipe(passThrough, { end: false });
+      // @ts-expect-error In actuality, this is a NodeJS.Readable
+      incompleteStreamInfo.stream.on('end', () => passThrough.emit('end'));
+      // @ts-expect-error In actuality, this is a NodeJS.Readable
+      ActorHttp.normalizeResponseBody(passThrough);
```
rubensworks, line 83: Instead of calling this, does `ActorHttp.toWebReadableStream` already do what we want?

If it does, then we probably don't need to create this `normalizeResponseBody` function anymore.

packages/bus-http/lib/ActorHttp.ts
```diff
@@ -60,6 +61,37 @@ export abstract class ActorHttp extends Actor<IActionHttp, IActorTest, IActorHtt
...
+        }
+      };
+    }
+
+    // Node-fetch does not support body.tee, while it is mandatory according to the fetch and readablestream api.
+    // If it doesn't exist, we monkey-patch it.
+    if (body && !body.tee) {
+      body.tee = (): [ReadableStream, ReadableStream] => {
+        // eslint-disable-next-line import/no-nodejs-modules
+        const stream = require('stream');
```
rubensworks, line 88: I guess this should be `readable-stream`.
Also, can we move this to a top-level import?

rubensworks (changes requested):
Really nice work! Just mainly some comments about Node/Web stream handling.

After these are resolved, I'll do some hands-on testing myself with this.

packages/actor-http-cache/lib/ActorHttpCache.ts
```diff
@@ -0,0 +1,270 @@
...
+    }
+    const cacheStorage = this.getCacheStorage(action.context);
+    if (await cacheStorage.has(new Request(action.input, action.init))) {
+      return { time: 1 };
+    }
+    const newAction = {
+      ...action,
+      context: action.context.set(KeysHttpCache.doNotCheckHttpCache, true),
+    };
+    return await (
```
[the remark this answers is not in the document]
jaxoncreed: ActorHttpCache should be called both when the cache has a value and when it doesn't have a value. Think of it like a Proxy for ActorHttpFetch. If you were to throw an error when there isn't an item in the cache the response would never be cached as the program would skip directly to ActorHttpFetch which doesn't touch caching at all.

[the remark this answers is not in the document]
rubensworks: > ActorHttpCache should be called both when the cache has a value and when it doesn't have a value.

Right, good point.

In that case, we can remove the check if the cache contains the entry and the mediator call, and just always return `{ time: 1 };`, just like the proxy.
We only need to keep the `doNotCheckHttpCache` check.

packages/bus-http/lib/ActorHttp.ts
jaxoncreed: I'll add a comment explaining this, but we need to use 'stream' to simulate tee. We can't use 'readable-stream'. Later down the road, we want to reconsutrct a `Response` using `new Response(teedStream)`. When calling the constructor of Response in `node-fetch`, it checks to see what type of body was provided to it. If the body is `instanceof Stream` it will leave the body alone (https://github.com/node-fetch/node-fetch/blob/main/src/body.js#L53), if not it tried to convert it to a buffer (https://github.com/node-fetch/node-fetch/blob/main/src/body.js#L62).

So, when you pass in a `readable-stream` stream, `node-fetch` gets confused because a `readable-stream` stream is not an instanceof the base NodeJS stream. So, it tries to convert it into a buffer which causes the result of `await response.text()` to be `[object Object]` because, of course, `Buffer.from()` expects a string, and `[object Object]` is the stratification of a stream object.

This is an inline import because we don't want to require `stream` at the top for browser use cases as the `stream` library won't be available in browser. Though the browser should never execute line 88 because its response does have a `tee` method.

packages/actor-http-cache/lib/HttpCacheStorageStream.ts
jaxoncreed: `ActorHttp.toWebReadableStream` cannot be used because of the problems listed below.

jaxoncreed:
@rubensworks I did a little experiment and I'm pulling my hair out. This will not work in the web browser. Ugh. I'll need to refactor everything and get back to you :( :( :( :( :( :( :(

jaxoncreed:
Then again, does this need to work in the web browser? The web browser has its own cache, so would there be any instance were a browser-focused comunica would use this cache?

rubensworks:
> Then again, does this need to work in the web browser? The web browser has its own cache, so would there be any instance were a browser-focused comunica would use this cache?

Yes, this would be important in cases where we want other actors to be notified of cache invalidations via the HTTP invalidation actions.

> For example, I could just update the test function to throw an error if it's used in a web browser. Then, I could always just assume that the stream is a Node Stream. Does that work for you?

Given all the work that went into this PR already, I'm ok with scoping this as a Node-only actor.

However, just checking for browser presence in the test function will not work, as Webpack will already error at build-time.
Instead, we should create a dedicated Node and browser config file of the `query-sparql` engine, similar as how it's done in `query-sparql-file`, so that this actor is only included in Node builds.



I remember that you said at some point that you wanted to use this functionality in one of your libraries, which I assume is browser-side? Would this then still work for you?

jaxoncreed:
Sorry about the frustration I expressed before. I had a goal to have this complete by the end of the week, and that wasn't hit.

Anyway, I looked back into getting things to work in the web browser, and I figured out how to do it, but many of the techniques used to clone streams and create new streams are completely different. (Here's the demo I made. I've tested to be sure it works in the browser https://gist.github.com/jaxoncreed/d6c8c4ec59e08b6bb8710ed34a12f41b)

So, from an architecture perspective, would you rather have separate classes that work in each environment then have those classes wired up in componentsJS, or keep the same classes but provide a bunch of conditionals depending on the environment?

jaxoncreed:
@rubensworks which architecture would you like?

rubensworks:
I'll get back to you in January!

rubensworks:
@jaxoncreed If we can make it work in both Node and the browser, that would be best option!
But as I've said before, given all the effort that went into this PR already, it would be sufficient to have only Node support for now.

> So, from an architecture perspective, would you rather have separate classes that work in each environment then have those classes wired up in componentsJS, or keep the same classes but provide a bunch of conditionals depending on the environment?

If you think supporting both Node and the browser will be feasible, then I would recommend implementing everything in one package, and using the [`"browser"` package.json field](https://gist.github.com/defunctzombie/4339901/49493836fb873ddaa4b8a7aa0ef2352119f69211) to override Node-specific classes with browser-specific classes. Similar to how the [`FetchInitPreprocessor`](https://github.com/comunica/comunica/blob/master/packages/actor-http-fetch/package.json#L44-L46) is overridden for the browser in our fetch actor.
The trick here will be to make sure that logic that is different between Node and the browser is placed into one or more distinct file(s), and common code is shared.

To make sure that we test both variants of the code, we'll have to test them via system tests (which are run both on Node and on browsers in our CI).

jaxoncreed:
Okay excellent. I'll get to work.

rubensworks:
> @rubensworks Do you know if there's an easy way to run tests in the web browser? Just to be sure everything's working properly.

@jaxoncreed We recently started also running our system tests (that we already ran in Node) in browsers using Karma: https://github.com/comunica/comunica/blob/master/package.json#L88
For example, tests like [these](https://github.com/comunica/comunica/blob/master/engines/query-sparql/test/QuerySparql-test.ts).
So you could try writing such system tests as well (next to the specific actor unit tests, which I would run only in Node).
The only caveat is that no jest-specific things can be used in this case, but I think this is not required here.

packages/actor-http-cache/lib/HttpCacheStorageStream.ts
```diff
@@ -0,0 +1,146 @@
...
+    this.incompleteStreams[key.url] = {
+      policy: value.policy,
+      init: value.init,
+      buffers: [],
+      streamId,
+      stream: value.body,
+    };
+    // @ts-expect-error In actuality, this is a NodeJS.Readable
+    value.body.on('data', (chunk: any) => {
+      if (this.incompleteStreams[key.url] && this.incompleteStreams[key.url].streamId === streamId) {
```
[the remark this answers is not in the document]
jaxoncreed: As it exists right now, old streams are dumped when a new stream comes in for the same url. Do you think that there's a reason to keep track of the old streams? I can't think of one right now.

packages/actor-http-cache/lib/ActorHttpCache.ts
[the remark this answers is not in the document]
jaxoncreed: Do you mean return `return { time: Number.POSITIVE_INFINITY };` like the proxy?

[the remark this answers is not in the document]
rubensworks: Ah, yes, infinity will do :-)

packages/actor-http-cache/lib/HttpCacheStorageStream.ts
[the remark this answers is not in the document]
rubensworks: I think we need to keep track of them for ensuring proper cleanup.
I suspect that if we don't explicitly consume or destroy the incomplete streams, that Node will hang in certain cases. (if this is not the case, then we can just ignore them indeed)

In cases where a previous incomplete stream is still being consumed downstream, we can not clean up the stream just yet until it is fully consumed downstream.

jaxoncreed:
Hey @rubensworks  I'm having a bit of trouble getting the system tests to work. Even on a clean install, they time out.

I'm cloning the repo, running `yarn install`, then running `yarn test`. Is there anything else I need to do?

rubensworks:
> I'm cloning the repo, running yarn install, then running yarn test. Is there anything else I need to do?

Nope, that's all you need to do.

But they did work before for you, or not?

rubensworks:
> System tests did work before. But after updating to the most recent version on master, they fail by timing out

That's weird. Nothing really changed though. Since the CI still works, and they don't time-out on my machine and on the machines from other people I know, I suspect it will be an issue on your end somehow.

jaxoncreed:
Okay! It looks all the tests are passing except for any tests that involve solid-authn-js, and I have not yet found a solution for that.

The tests will have no problem with running the setup (https://github.com/jaxoncreed/comunica/blob/feat/http-cache/engines/query-sparql/test/QuerySparql-solid-test.ts#L133-L136), but when it runs the actual test (https://github.com/jaxoncreed/comunica/blob/feat/http-cache/engines/query-sparql/test/QuerySparql-solid-test.ts#L141-L143), jest freezes. Tests won't even time out.

More specifically, they will freeze because of a low-level crypto function that signs DPoP tokens (https://github.com/panva/jose/blob/main/src/runtime/node/sign.ts#L30). There may be some kind of problem with promisifying the crypto function, but the confounding part is that this has nothing to do with the cache.

Even more strange, when you force the cache to always return undefined, the stream will be consumed prematurely. The stream is still open inside `ActorHttpCache` but is closed by the time it reaches `ActorRdfHypermediaNone`. For some reason, the stream is being consumed somewhere after the cache code.

@rubensworks  if any of this sounds like something you've dealt with before, let me know, otherwise I'll keep at it.

rubensworks:
@jaxoncreed Ok, interesting. Since you also get these freezing tests on the master branch, this probably isn't a caching problem indeed. So I would suggest not losing too much time on this. As long as the tests pass in the CI, it's fine by me.

@jeswr have you run into this problem that @jaxoncreed is describing before? If so, how can it be resolved?

jaxoncreed:
Unfortunately, I fixed the master branch and the freezing only happens with the cache now. It does not pass CI.

jeswr:
> @jeswr have you run into this problem that @jaxoncreed is describing before? If so, how can it be resolved?

Yes - tests freeze a lot on me when I run the whole test suite; though I had pinned it on my machine since the Linux Kernel wasn't fully supporting all of the hardware on my work machine until recently.

Locally, I've just been testing only the suites for the files I have changed and then skipping the pre-commit check and then relying on CI to run all the tests.

jaxoncreed:
Thanks @jeswr . It looks like problem persists in the CI when you include the cache I wrote. Also, it looks like the integration tests freeze as well. I'm going to see if they're freezing for the same reason.

rubensworks:
> Unfortunately, I fixed the master branch and the freezing only happens with the cache now. It does not pass CI.

Ok, then it's probably a race condition in the cache logic somewhere.

jaxoncreed:
Yep, I'll continue to look into it. My laptop's in the shop right now, so I'll get back to it when it's out.

rubensworks:
Hi @jaxoncreed, any updates on this PR?

jaxoncreed:
Unfortunately, no. I've spent a lot of time trying to figure out the stream problem, and we still haven't found an answer. I've done a workaround in my current work that doesn't require a cache. Eventually we'll need to return to this, but I am unsure how to at this point.

rubensworks:
@jaxoncreed FYI, we disabled the Solid integration issues, as they were causing other issues.

jaxoncreed:
@rubensworks That's good to hear. Perhaps I'll come back to this issue, but at the moment, it's low in the priority list. Thanks for the heads up.

rubensworks:
I'm going to close this PR due to staleness.

PR #1011, opened 2022-06-17
Adds httpTimeoutOnBody option

Tpt:


packages/actor-http-fetch/test/ActorHttpFetch-test.ts
```diff
@@ -361,5 +372,60 @@ describe('ActorHttpFetch', () => {
...
+    // FIXME: does not work
+    // it('should work with a large timeout including body if the body is not consumed', async() => {
+    // await expect(actor.run({
+    //     input: <Request> { url: 'https://www.google.com/' },
+    //     context: new ActionContext({
+    //       [KeysHttp.httpTimeout.name]: 100_000,
+    //       [KeysHttp.httpTimeoutOnBody.name]: true
+    //     }),
+    // })).resolves.toMatchObject({ status: 200 });
+    // });
```
Tpt, line 413: This test does not work yet because there is now way to act if a steam is not read but just "dropped" by the code and GCed.

Tpt (commented):
There are still to problems with this change:

1. The timer does not get cleared if the body stream is not read.
2. I attempted to add a "httpTimeoutOnBody" CLI option but the current CLI parser does not support some argument name (here `httpTimeout`) being the prefix of an other one `httpTimeoutOnBody`. A new name might be welcomed.

Tpt:
> I think we can assume that the body will either be read, or it will be destroyed. (We had a bug a while ago where the process would hang if an unconsumed body was not explicitly destroyed: [45fe4a0](https://github.com/comunica/comunica/commit/45fe4a0316efd6d6b3837a5a0a927fe7d1e70a1c)) So listening for the end-event and whatever event `.cancel()` produces should be sufficient I think.

Done. The `"end"` event is called in both cases.

> What about `httpTimeout` and `httpBodyTimeout`?

Indeed. It's much better. Done.

packages/actor-http-fetch/lib/ActorHttpFetch.ts
```diff
@@ -88,12 +90,26 @@ export class ActorHttpFetch extends ActorHttp {
           (<Readable><any>response.body).destroy(error);
         };
       }
+
+      // We remove or update the timeout
+      if (requestTimeout !== undefined) {
+        const httpBodyTimeout = action.context?.get(KeysHttp.httpBodyTimeout) || false;
+        if (httpBodyTimeout && response.body) {
+          onTimeout = () => response.body?.cancel(new Error(`HTTP timeout of ${httpTimeout}ms enabled using the "httpTimeout" and "httpBodyTimeout" options`));
```
rubensworks, line 98: I would suggest the following as error message:

```
HTTP timeout when reading the body of ${response.url}.
This error can be disabled by modifying the 'httpBodyTimeout' and/or 'httpTimeout' options.
```

packages/actor-http-fetch/lib/ActorHttpFetch.ts
```diff
@@ -88,12 +90,26 @@ export class ActorHttpFetch extends ActorHttp {
...
+          (<Readable><any>response.body).on('end', () => {
+            clearTimeout(requestTimeout!);
+          });
+        } else {
+          clearTimeout(requestTimeout);
+        }
+      }
+
       return response;
-    } finally {
```
rubensworks, line 92: Is the finally block not required anymore?

packages/actor-http-fetch/test/ActorHttpFetch-test.ts
```diff
@@ -361,5 +372,67 @@ describe('ActorHttpFetch', () => {
...
+        context: new ActionContext({
+          [KeysHttp.fetch.name]: customFetch,
+          [KeysHttp.httpTimeout.name]: 100_000,
+          [KeysHttp.httpBodyTimeout.name]: true,
+        }),
+      });
+      const body = <Readable><any> response.body;
+      for await (const chunk of body) {
+        // We just want to consume everything
+      }
```
rubensworks, line 401: Could we somehow check here if `clearTimeout` is correctly called?

packages/actor-http-fetch/test/ActorHttpFetch-test.ts
```diff
@@ -361,5 +372,67 @@ describe('ActorHttpFetch', () => {
...
+      });
+      const response = await actor.run({
+        input: <Request> { url: 'https://www.google.com/' },
+        context: new ActionContext({
+          [KeysHttp.fetch.name]: customFetch,
+          [KeysHttp.httpTimeout.name]: 100_000,
+          [KeysHttp.httpBodyTimeout.name]: true,
+        }),
+      });
+      await response.body?.cancel();
```
rubensworks, line 419: Same as above.

packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts
```diff
@@ -198,6 +202,14 @@ export class CliArgsHandlerBase implements ICliArgsHandler {
       context[KeysHttp.httpTimeout.name] = args.httpTimeout;
     }
 
+    // Define HTTP body timeout
+    if (args.httpBodyTimeout) {
+      if (!args.httpTimeout) {
+        throw new Error('The --httpBodyTimeout option requires the --httpTimeout option to be set');
```
rubensworks, line 208: Ah yes indeed, good catch :-)

rubensworks (changes requested):
Great work! I only have a few minor comments.

packages/actor-http-fetch/lib/ActorHttpFetch.ts
Tpt: Done

packages/actor-http-fetch/test/ActorHttpFetch-test.ts
Tpt: Done. It allowed me to uncover that the node-fetch response stream does not always emit the `"close"` event... I have added a work-around for that.

packages/actor-http-fetch/test/ActorHttpFetch-test.ts
Tpt: Done using `jest.spyOn`.

Tpt (commented):
Thank you for your review!

rubensworks (approved):
Excellent!

Tpt:
Thank you!

PR #1020, opened 2022-06-29
WebPack: Uses explicit overrides instead of NodePolyfillPlugin

Tpt:
Allows to track explicitly which pollyfills are useful and reduce the number of transitive dependencies.


rubensworks:
@Tpt Before we merge this, could you confirm that the webpacked version still works as intended (i.e., does a simple query executed within the browser still work)?

Ideally, we'd have automated tests for this as well, but we're not there yet :-)

Tpt:
> @Tpt Before we merge this, could you confirm that the webpacked version still works as intended (i.e., does a simple query executed within the browser still work)?

Yes, I have checked it and it works. Doing so allowed me to discover that I initialy forgot the `process: 'process/browser'` piece.

rubensworks:
Alright, perfect!

PR #1021, opened 2022-07-05
Removes dependency on 'web-streams-node'

Tpt:
The only useful line in this module was exporting globally the ReadableStream polyfill provided by 'web-streams-ponyfill'

### DIFF
--- a/packages/actor-rdf-resolve-hypermedia-sparql/test/RdfSourceSparql-test.ts
+++ b/packages/actor-rdf-resolve-hypermedia-sparql/test/RdfSourceSparql-test.ts
@@ -10,6 +10,9 @@ const quad = require('rdf-quad');
 const streamifyString = require('streamify-string');
 const DF = new DataFactory();
 
+// TODO: Remove when targeting NodeJS 18+
+global.ReadableStream = global.ReadableStream || require('web-streams-ponyfill').ReadableStream;
+
 describe('RdfSourceSparql', () => {
   const context = new ActionContext({});
   const mediatorHttp: any = {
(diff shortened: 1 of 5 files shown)


packages/actor-rdf-resolve-hypermedia-sparql/test/RdfSourceSparql-test.ts
```diff
@@ -136,7 +136,7 @@ describe('RdfSourceSparql', () => {
           const query = action.init.body.toString();
           return {
             headers: new Headers({ 'Content-Type': 'application/sparql-results+json' }),
-            body: require('web-streams-node').toWebReadableStream(query.indexOf('COUNT') > 0 ?
```
rubensworks, line 139: Same as above?

packages/bus-http/lib/ActorHttp.ts
```diff
@@ -2,8 +2,9 @@ import type { IAction, IActorArgs, IActorOutput, IActorTest, Mediate } from '@co
 import { Actor } from '@comunica/core';
 import { ReadableWebToNodeStream } from 'readable-web-to-node-stream';
 
+global.ReadableStream = global.ReadableStream || require('web-streams-ponyfill').ReadableStream;
```
rubensworks, line 5: Setting globals in JS is not considered good practise, as it can pollute the global space when working with multiple libraries together.

Can we remove this line?

rubensworks (changes requested):

packages/bus-http/lib/ActorHttp.ts
Tpt: Sadly we cant because `web-streams-node` expects `ReadableStream` to be availlable globally. This global expose was done by `web-streams-node` (it's the only line of  `web-streams-node` that was useful for Comunica).

rubensworks: Ah, you're right. It's not nice, but we should indeed keep the existing behaviour.

Can you put a `// TODO:` above it, and say that this can be removed once we make Node.js 18 the minimal version?

packages/actor-rdf-resolve-hypermedia-sparql/test/RdfSourceSparql-test.ts
Tpt: Yes!

packages/bus-http/lib/ActorHttp.ts
Tpt: Done.

rubensworks:
Thanks!

PR #1022, opened 2022-07-12
Removes xml dependencies

Tpt:
- Fixes "xmlns" attribute
- The "header" node is mandatory in the schema

### DIFF
--- a/packages/actor-query-result-serialize-sparql-xml/lib/ActorQueryResultSerializeSparqlXml.ts
+++ b/packages/actor-query-result-serialize-sparql-xml/lib/ActorQueryResultSerializeSparqlXml.ts
@@ -8,7 +8,7 @@ import type {
   IQueryOperationResultBoolean,
 } from '@comunica/types';
 import type * as RDF from '@rdfjs/types';
-import * as xml from 'xml';
+import { XmlSerializer, type IXmlNode } from './XmlSerializer';
 
 /**
  * A comunica sparql-results+xml Serialize Actor.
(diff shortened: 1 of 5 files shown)


packages/actor-query-result-serialize-sparql-xml/lib/ActorQueryResultSerializeSparqlXml.ts
```diff
@@ -89,20 +88,22 @@ export class ActorQueryResultSerializeSparqlXml extends ActorQueryResultSerializ
...
+          children: [ ...bindings ]
+            .map(([ key, value ]) => ActorQueryResultSerializeSparqlXml.bindingToXmlBindings(value, key)) });
       });
 
       // Close streams
       resultStream.on('end', () => {
-        results.close();
-        root.close();
+        serializer.close();
+        serializer.close();
```
rubensworks, line 99: Should only be called once?

rubensworks (changes requested):
This is a very clean implementation! :-)

Just one question.

packages/actor-query-result-serialize-sparql-xml/lib/ActorQueryResultSerializeSparqlXml.ts
Tpt: No because we want to close both the `<results>` node and the `<sparql>` node.

rubensworks: Ok, makes sense, thanks for clarifying!

PR #1023, opened 2022-07-13
Move away from Node.js built-ins

Tpt:
This makes it possible to easily build this project using tools such as Webpack 5.

- Generalizes readable-stream instead of node:stream
- Upgrades dependencies to versions without node:stream dependency
- polyfills node:process in the ActorInitQuery entry point
- uses object-inspect instead of utils.inspect
- Uses the import/no-nodejs-modules eslint rule to enforce compatibility


rubensworks (approved):
Amazing work!

@surilindur, can you check on your end if this change works fine on your end as well?
I.e., if you can use this version of Comunica in the React app without making any Webpack config changes, and if queries still work fine.

surilindur:
Hi and apologies for the massive delay! I finally got around to testing this, but instead of the to-do app I used the [jQuery Widget](https://github.com/comunica/jQuery-Widget.js) as you suggested.

There appears to be some use of the `util` module from Node, just the `inspect` function, at `packages/logger-pretty/lib/LoggerPretty.ts`. I swapped it for `JSON.stringify` for testing purposes, but I am not sure what the best replacement would actually be. There does not seem to be any other use of `util`, either, so just that one function.

Other than that, everything worked fine (at least with the default config). This is awesome work! :slightly_smiling_face:

Tpt:
Thank you so much @surilindur! I have updated the PR to use `JSON.stringify`.
@rubensworks Is it fine for you?

surilindur:
Oh, I meant that I replaced it with `JSON.stringify` to have something that sort of does similar things. But there are some cases where the stringify function can (and will) fail, I guess, if not all input can be serialised as JSON strings. I also do not know how it handles all the potential inputs, which is probably why the Node built-in was used in the first place. :grimacing:

I hope @rubensworks has an idea for whether that replacement makes sense or not, or what type of alternatives exist for it. There appear to be some browser versions of the entire `util` module, but for a single function those feel slightly overkill?

rubensworks:
Thanks @surilindur and @Tpt!

> I have updated the PR to use JSON.stringify.

>  But there are some cases where the stringify function can (and will) fail, I guess, if not all input can be serialised as JSON strings.

Indeed, @surilindur is right, there are some differences between `JSON.stringify` and Node's `inspect`. The latter is generally safer to use. For example, it can safely handle recursive objects and very deep objects, while the former may crash on those.

I'm sure we can find some good browser-safe alternatives for it. Just found out about `object-inspect` for example, but perhaps there are better ones.

rubensworks:
@Tpt Just found out about this: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-nodejs-modules.md
Could you have a look to see if this rule can help in our case?

Tpt:
> I'm sure we can find some good browser-safe alternatives for it. Just found out about `object-inspect` for example, but perhaps there are better ones.

I was also considering `object-inspect`. It seems the most used and maintained one. It supports recursive structs and objects not supported by `JSON.stringify` like `BigInt` and we already have it in the dependency tree. I have updated the PR with it (and added some tests).

rubensworks:
Perfect! Will do a final pass on this PR soon.

Tpt:
> @Tpt Just found out about this: https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-nodejs-modules.md Could you have a look to see if this rule can help in our case?

I have just added this rule. I have disabled it for tests, "actor-dereference-file", "actor-http-native" and "logger-bunyan" that are not compatible with web browsers (or are deprecated and very cumbersome to port).
The rule also had a false-positive in `ActorInitQuery-browser.ts` (there is a polyfill the rule is not aware of).

packages/logger-pretty/lib/LoggerPretty.ts
```diff
@@ -1,5 +1,5 @@
-import { inspect } from 'util';
 import { Logger } from '@comunica/types';
+const inspect = require('object-inspect');
```
rubensworks, line 2: This one should have type definitions: https://www.npmjs.com/package/@types/object-inspect
Could you check if those work, so that this can become an import statement?

rubensworks (changes requested):
Just one question on `object-inspect`. Other than that, this looks good to be merged :-)

packages/logger-pretty/lib/LoggerPretty.ts
Tpt: Done!

packages/logger-pretty/lib/LoggerPretty.ts
```diff
@@ -1,5 +1,5 @@
-import { inspect } from 'util';
 import { Logger } from '@comunica/types';
+import objectInspect = require('object-inspect');
```
rubensworks, line 2: Is the `require` really needed here?
It might be possible to replace it with `import * as objectInspect from 'object-inspect'`

Tpt: Indeed. That's much better. I have updated the PR.

rubensworks:
Thanks @Tpt, really great work as always!

Feel free to follow up on finalizing the bounty via open collective.

New release with this change will be coming soon.

PR #1033, opened 2022-08-03
WebPack: Avoids applying @babel/polyfill twice on the output code

### LINKED ISSUE #1032: Webpack: @babel/polyfill is applied twice on the output code
#### Issue type:

- :bug: Bug 

____
#### Description:

The webpack builds of Communica contains the warning `@babel/polyfill is loaded more than once on this page. This is probably not desirable/intended and may have consequences if different versions of the polyfills are applied sequentially.`.

This is caused because both `@babel/preset-env` is loaded in `.babelrc` and that `@babel/polyfill` is loaded manually in the webpack configuration. One of them should be removed (probably `@babel/polyfill`).

Tpt:
Closes #1032

Tested using the latest Firefox, Chrome and Gnome Web.


rubensworks:
Oh, good catch!

PR #1038, opened 2022-08-08
Replace setImmediate by setTimeout

Tpt:
setTimeout(X) runs the callback on the next event cycle
except if there is more than 5 nested setTimeout.
In this case it waits at least 4ms.
This should be fine because we are not overusing setTimeout in Comunica.


rubensworks (approved):
For reference, `setTimeout` with delay 0 is slower than `setImmediate`.
However, since all of these usages are in non-performance-critical parts of the code, this is not a problem.
Just something to take into account should this change in the future.

Issue #1034, opened 2022-08-04
resultToString throws stream._fillBuffer is not a function when querying GraphDB solely


vterzic:
#### Issue type:

- :bug: Bug 

____
#### Description:

`queryEngine.resultToString()` throws error when querying GraphDB solely. Once you add additional source, everything works as expected.

Reproducible code example: https://gist.github.com/vterzic/66d2718fec73f4e842731380533b1ce1
____
#### Environment:

| software         | version
| ---------------- | -------
| Comunica Engine  | 2.3.0 
| node             | v16.13.2
| npm              | 8.1.2
| yarn             | _Yarn is unavailable_
| Operating System | darwin (Darwin 20.5.0)
| GraphDB | 9.10.1 |
#### Crash log:
https://gist.github.com/vterzic/66d2718fec73f4e842731380533b1ce1?permalink_comment_id=4255805#gistcomment-4255805

rubensworks:
@jeswr @jacoscaz This code makes use of the `wrap` function of AsyncIterator.
Do you know if this code does not expose the `_fillBuffer` method anymore?

jeswr:
`wrap` can now bypass the `WrappingIterator` in certain cases (e.g. if it was already an AsyncIterator, it will just return the input). In such cases `_fillBuffer` will not exist.

jeswr:
Not sure why this would be a problem on some sparql endpoints and not others though - it shouldn't change the nature of the iterator

jacoscaz:
Why are we using a private method of `AsyncIterator`, though?

rubensworks:
I'm sure it will also break in many other places :-)

Is there another method we can use to "start" the iterator?

rubensworks:
For reference: https://github.com/comunica/comunica/blob/master/packages/actor-query-operation-sparql-endpoint/lib/ActorQueryOperationSparqlEndpoint.ts#L168-L171

jacoscaz:
Oh I think I understand what is going on. Metadata can only get to the end of the chain once the entire chain begins to read, so we're invoking `_fillBuffer()` to force that even when we haven't actually begun to read items.

Mh... We should probably temporarily fix this by wrapping into a `BufferingIterator` within Comunica (or something like that) and then figure out whether we can decouple metadata propagation.

I'll have a look, I think we can do something relatively simple to address this.

rubensworks:
@jacoscaz Jesse suggested a similar fix (he's sitting next to me), he's going to look into it.

[...]

PR #1043, opened 2022-08-11
Fix/stop  fill buffer usage

jeswr:
Draft fix to #1034 - still some bugs to iron out


lerna.js
```diff
@@ -104,3 +104,33 @@ async function depcheckTask(log) {
 
 module.exports.depfixTask = depfixTask
 module.exports.depcheckTask = depcheckTask
+
+const ncu = require('npm-check-updates');
```
rubensworks, line 108: This script seems really valuable!
Perhaps we want to run it in a separate PR, so it doesn't pollute this one too much?

jeswr: Sure - only problem is that it bumps AsyncIterator to 3.6.1 which breaks some unit tests because of the `_fillBuffer` usage so the fixes here will need to be merged first.

jeswr:
Closing now that most changes are in main and the check-updates script is in #1078

PR #1045, opened 2022-08-14
fix: remove usage of _fillBuffer internal from AsyncIterator

jeswr:


packages/actor-query-operation-sparql-endpoint/lib/LazyCardinalityIterator.ts
```diff
@@ -0,0 +1,109 @@
...
+    if (this._source.done) {
+      this.close();
+    }
+
+    if (this.done) {
+      this._cardinality = Promise.resolve(this._count);
+    } else {
+      this._buffer = new LinkedList();
+      this._cardinality = new Promise(resolve => {
+        // TODO: See if this section should reject on error
```
jeswr, line 66: @rubensworks Should we reject this promise on the `error` event. Similarly to what is done in `.toArray` https://github.com/RubenVerborgh/AsyncIterator/blob/a94fbf50dddfe0ee09b6dc608bc68d83340d5c76/asynciterator.ts#L337

rubensworks: Yes, good catch, we should definitely do this!

rubensworks (approved):
Looks great!
Definitely a much cleaner solution than the one we had before.

The only concern I have is the complexity of the new iterator (and the LinkedList). I would guess there's a lot of logic we could be reusing from AsyncIterator?

jeswr:
> I would guess there's a lot of logic we could be reusing from AsyncIterator?

Indeed - this is just awaiting the release of https://github.com/RubenVerborgh/AsyncIterator/pull/82/files.

packages/actor-query-operation-sparql-endpoint/lib/LazyCardinalityIterator.ts
jeswr: One more case I forgot to think about - should we also reject if an error event occured before `getCardinality` was called?

rubensworks: Good point, that's indeed something that could occur in cases where the stream is not being used, but only the metadata is of relevance.

jeswr: Why only when the stream is not being used. What if `getCardinality` is called before / after / during the consumption of the stream.

rubensworks: Those as well indeed, this case was just the first that popped in my mind :-)

jeswr:
This is ready for review @rubensworks

packages/actor-query-operation-sparql-endpoint/lib/LazyCardinalityIterator.ts
```diff
@@ -0,0 +1,125 @@
...
+    this._source.removeListener('readable', destinationSetReadable);
+    this._source.removeListener('end', destinationSetReadable);
+    this._source.removeListener('error', destinationEmitError);
+    delete (<any> this._source)[DESTINATION];
+    this._source.destroy();
+    super.close();
+  }
+}
+
+const DESTINATION = Symbol('destination');
```
rubensworks, line 115: Can we reuse the symbol that's exported by the AsyncIterator package?

packages/actor-query-operation-sparql-endpoint/package.json
```diff
@@ -41,7 +41,7 @@
     "@comunica/mediatortype-httprequests": "^2.3.0",
     "@comunica/types": "^2.3.0",
     "@rdfjs/types": "*",
-    "asynciterator": "^3.4.2",
+    "asynciterator": "^3.7.0",
```
rubensworks, line 44: Could you also update this dep in all our other packages? Just so we avoid multiple different versions of asynciterator. (there are two of them now in the `yarn.lock` file)

rubensworks: Related to this, we may also want to depend on the latest asyncjoin version, to avoid dep issues downstream.

rubensworks (changes requested):
Thanks @jeswr!
Just a couple minor comments. Other than that, this looks good to be merged.

PR #1047, opened 2022-08-18
perf: remove unecessary TransformIterator

jeswr:


rubensworks:
I just pushed some changes to that file this morning, so it has some conflicts now...

But I think you're right in that this TransformIterator wrapper is not required anymore (it definitely used to be at some point).

However, are you sure that there's actually a performance benefit? Because the wrapper still avoids `getDuplicateElementLinks` being called in some cases (but its impact should be minor in any case).

jeswr:
Would need to double check but I think the setup meant that all elements were being buffered through the transform iterator which can have non-negligible impacts depending on the source.

rubensworks:
Any updates on this?

rubensworks:
Closing this due to staleness. Feel free to re-open again later.

PR #1055, opened 2022-09-05
fix: polyfill Buffer in webpack

### LINKED ISSUE #1053: Buffer is not defined when using latest rdfjs browser build of Comunica
#### Issue type:

- :bug: Bug 

____
#### Description:

I think this is because readable-stream still requires a `buffer` polyfill https://www.npmjs.com/package/readable-stream but that polyfill was removed in https://github.com/comunica/comunica/pull/1023/files.

Pinging @Tpt 
____

jeswr:
Resolves #1053 

My main question is - does the `ProvidePlugin` removed in https://github.com/comunica/comunica/pull/1023/files need to be introduced

### DIFF
--- a/packages/actor-init-query/webpack.config.js
+++ b/packages/actor-init-query/webpack.config.js
@@ -20,6 +20,11 @@ module.exports = {
       },
     ]
   },
+  resolve: {
+    fallback: {
+      buffer: require.resolve("buffer/"),
+    }
+  },
   plugins: [
     new webpack.ProgressPlugin()
   ]


Tpt:
> My main question is - does the ProvidePlugin removed in https://github.com/comunica/comunica/pull/1023/files need to be introduced

I don't think so. I belive it's enough if the `buffer` module is globally loaded. But I might be wrong.

rubensworks:
Thanks @jeswr!

In theory, this workaround can be removed once sparqljson-parse.js migrates to https://github.com/comunica/json-event-parser.js

PR #1059, opened 2022-09-08
Add Buffer global polyfill for Webpack

surilindur:
This should fix the issue with the browser builds producing an error about `Buffer` not being defined. The `Buffer` global needs to be filled in separately from the `buffer` import (which has already been added). This was done for the jQuery-Widget.js but apparently Comunica does browser builds directly from here, as well. Apologies for that!

~~When it will be fixed here, [the corresponding fix](https://github.com/comunica/jQuery-Widget.js/commit/049b4f3828338bfb0b11a998218c7e229513755a) can likely be removed from the jQuery-Widget.js.~~ Actually it probably cannot... need to try.

The workaround here needs to be removed when the parsing library will be updated to no longer need it, I think: [sparqljson-parse.js#42](https://github.com/rubensworks/sparqljson-parse.js/issues/42)

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "lerna-script": "^1.4.0",
     "manual-git-changelog": "^1.0.0",
     "memory-streams": "^0.1.3",
+    "node-polyfill-webpack-plugin": "^2.0.1",
     "nodemon": "^2.0.15",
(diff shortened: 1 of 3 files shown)


rubensworks:
Thanks @surilindur! :-)

PR #1060, opened 2022-09-12
ignore also .vscode file

constraintAutomaton:
For the convenience of vs code users, since web storm config file are already ignored.


rubensworks:
Thanks!

PR #1063, opened 2022-09-14
Fix/describe blank node not working as expected

### LINKED ISSUE #870: Describe query for blank node not working as expected
#### Issue type:

- :bug: Bug 

____
#### Description:

While using `DESCRIBE` query to retrieve the triples linked to a `blank node`, the returned triples seems to be erroneous. In the returned results, triples that should have similar subjects (i.e the same blank node uri), they're becoming different uri.

Example:

consider the following triples:
```ttl
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix ex: <http://example.com/> .

ex:a a ex:b;
    ex:c ex:d;
    ex:e [
        a ex:f;
        ex:v ex:m ;
    ]
.
```
and the following `DESCRIBE` query:
```sparql
DESCRIBE ?o {
    ?s ?p ?o .
        FILTER(isBlank(?o)) 
}
```

____
#### Environment:
The current results with current version `1.20.0` looks like this:
```
subject                      predicate                                            object                                                                                         
--------------------------------------------------------------------------------------------------------------
bc_0_n3-00            http://www.w3.org/1999/02/22-rdf-syntax-ns#type            http://example.com/f                                                                                 
bc_0_n3-01              http://example.com/v                                     http://example.com/m  
```

#### Expected output:
What to expect, might looks like:
```
subject                      predicate                                            object                                                                                         
--------------------------------------------------------------------------------------------------------------
bc_0_n3            http://www.w3.org/1999/02/22-rdf-syntax-ns#type            http://example.com/f                                                                                 
bc_0_n3             http://example.com/v                                      http://example.com/m  
```

constraintAutomaton:
Fixing the blank node inconsistency with `DESCRIBE` queries by adding a flag to the [`ActorQueryOperationConstruct`](https://github.com/comunica/comunica/blob/master/packages/actor-query-operation-construct/lib/ActorQueryOperationConstruct.ts) to ignore the [blank node localization](https://github.com/comunica/comunica/blob/4c66bfaa44f73786bc7403504e5fe571181430c7/packages/actor-query-operation-construct/lib/BindingsToQuadsIterator.ts#L20). For context the `ActorQueryOperationDescribeSubject`  [performs under the hood a `CONSTRUCT` operation](https://github.com/comunica/comunica/blob/0a368bc779c855a5674aeed5e4cd020c1b395f6f/packages/actor-query-operation-describe-subject/lib/ActorQueryOperationDescribeSubject.ts#L43).

Since the behavior was not caught by the current tests some integration test is also added.

close #870


packages/actor-query-operation-construct/lib/ActorQueryOperationConstruct.ts
```diff
@@ -45,10 +46,19 @@ export class ActorQueryOperationConstruct extends ActorQueryOperationTypedMediat
       await this.mediatorQueryOperation.mediate({ operation, context }),
     );
 
+    // Check if the query if it's a DESCRIBE query
+    let isDescribeOperator = false;
+    const describeQuery = context.get(new ActionContextKey('@comunica/action-query-operation:root'));
```
rubensworks, line 51: Instead of defining a new context key here, can we add a constant one in https://github.com/comunica/comunica/blob/master/packages/context-entries/lib/Keys.ts ?

As name, I would suggest `@comunica/actor-query-operation-construct:localizeBlankNodes`, which just contains a boolean (that is set to `false` by the describe actor).

That way, we make the construct actor more loosely coupled to the describe actor.

rubensworks (changes requested):
Thanks @constraintAutomaton, in general, this is what we need indeed!

@jeswr @jacoscaz Since both of you have done things related to skolemization, could you confirm that this change makes sense?

packages/actor-query-operation-construct/lib/ActorQueryOperationConstruct.ts
constraintAutomaton: > Instead of defining a new context key here, can we add a constant one in https://github.com/comunica/comunica/blob/master/packages/context-entries/lib/Keys.ts ?
> 
> As name, I would suggest `@comunica/actor-query-operation-construct:localizeBlankNodes`, which just contains a boolean (that is set to `false` by the describe actor).
> 
> That way, we make the construct actor more loosely coupled to the describe actor.

if we want to add the key as a constant we might want to name it `@comunica/actor-query-operation:localizeBlankNodes` to make the name more general?

rubensworks: Yes, good suggestion, that's a bit more future-proof indeed!

jeswr:
I can't fault this - but since I haven't worked with the blank node internals in a while I also can't say its correct with complete confidence.

What would be useful to maintain long term consistency is an equivalent set of tests added that show the blank node skolemisation behavior is the same on a SELECT query as it is on a describe query.

constraintAutomaton:
> What would be useful to maintain long term consistency across queries is an equivalent set of tests added that show the blank node skolemisation behavior is the same on a SELECT query as it is on a describe query.

I think it's a good idea, but I'm sure if I should do it in this pull request. I will create an issue for that.

packages/actor-query-operation-construct/lib/ActorQueryOperationConstruct.ts
```diff
@@ -10,6 +10,7 @@ import type * as RDF from '@rdfjs/types';
 import type { AsyncIterator } from 'asynciterator';
 import { getTerms, getVariables, uniqTerms } from 'rdf-terms';
 import { Algebra } from 'sparqlalgebrajs';
+import { KeysQueryOperation } from './../../context-entries';
```
rubensworks, line 13: This should refer to the actual package name, instead of being a relative path.

packages/context-entries/lib/Keys.ts
```diff
@@ -181,6 +181,10 @@ export const KeysQueryOperation = {
    * If the default graph should also contain the union of all named graphs.
    */
   unionDefaultGraph: new ActionContextKey<boolean>('@comunica/bus-query-operation:unionDefaultGraph'),
+  /**
+   * An indicator that the operator should apply blank node localization
+   */
+  blankNodeLocalization: new ActionContextKey<boolean>('@comunica/actor-query-operation:localizeBlankNodes'),
```
rubensworks, line 187: Can you rename this to `localizeBlankNodes`, for consistency?

rubensworks (changes requested):

packages/actor-query-operation-construct/lib/ActorQueryOperationConstruct.ts
constraintAutomaton: For some reason the linter ask me to 

```
 error  '@comunica/context-entries' should be listed in the project's dependencies. Run 'npm i -S @comunica/context-entries' to add it  import/no-extraneous-dependencies
```

when I do `import { KeysQueryOperation } from '@comunica/context-entries';` and installing the dependency does't work
```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE could not resolve
```

rubensworks: >> when I do import { KeysQueryOperation } from '@comunica/context-entries';

This is correct.

You'll have to add the dependency manually to the package.json of this package, and then run a yarn install on the comunica root.

rubensworks:
Really nice work @constraintAutomaton!

As a minimal-effort follow-up PR, could you have a look at also exposing this `localizeBlankNodes` option when using comunica on the command-line, and to provide a shortcut for context entries? This can be done in https://github.com/comunica/comunica/tree/master/packages/actor-init-query/lib/cli and https://github.com/comunica/comunica/blob/master/packages/actor-init-query/lib/ActorInitQueryBase.ts#L98-L122

constraintAutomaton:
> Really nice work @constraintAutomaton!
> 
> As a minimal-effort follow-up PR, could you have a look at also exposing this `localizeBlankNodes` option when using comunica on the command-line, and to provide a shortcut for context entries? This can be done in https://github.com/comunica/comunica/tree/master/packages/actor-init-query/lib/cli and https://github.com/comunica/comunica/blob/master/packages/actor-init-query/lib/ActorInitQueryBase.ts#L98-L122

yes I'm gonna do it

PR #1068, opened 2022-09-16
Fix non-working links

constraintAutomaton:
fix the non-working links of `bus-rdf-resolve-quad-pattern`


rubensworks:
Thank you!

Issue #1069, opened 2022-09-18
Handling broken URLs / ontology documents


jeswr:
#### Issue type:

- :heavy_plus_sign: Feature request 

It would be good to have a way of defining custom behavior for when a link is broken - for instance trying to retrieve the link via the wayback machine instead (it seems like this package could be used for that https://www.npmjs.com/package/memento-client).

Actually in the case of the wayback machine it may be as simple as appending `http://wayback.archive-it.org/` to the URL.

One way of doing this would be to add custom `dereference` actors that enable this behavior.
____

rubensworks:
Sounds related to the [fallback-based datasources approach discussed recently](https://github.com/comunica/comunica/issues/1061#issuecomment-1247626851).

Note that we already have a memento actor here: https://github.com/comunica/comunica/tree/master/packages/actor-http-memento

PR #1070, opened 2022-09-18
Feat/wayback machine

jeswr:
Started this thinking there was a 30min solution to #1069 .

Turns out this is not the case as pages with 404s will still pass the test methods for `http-fetch`; and hence won't use the wayback actor.

To solve this and finish this PR I will need to select one of the following approaches:
 - Introduce a `http-fallback` bus; make that an actor on it and manually call that fallback bus where appropriate for instance in the `dereference-http` bus.
 - Introduce a new mediator that *sorts* actors by the test method; and then goes through the run method for each until it finds one that meets a certain condition (e.g. status != 404)
  - Something else
 
 @rubensworks thoughts?


rubensworks:
Oh, nice work :-)

Also not sure what the best mediation approach here is.

Another alternative might be to follow a wrapper-based approach. So that we always run the wayback actor first, but let it delegate back again to the same bus, and provide a wayback link when the delegation returned 404, but do a no-op on everything else.

This approach is probably the least invasive one. The new mediator is probably also ok in that respect.

jeswr:
Alrighty - this was a good distraction to come back to after banging my head against asynciterator internals.

This is working now - the only remaining task is to work out how to mock broken urls to do some proper testing - @rubensworks ; are there examples you can point to where this is already being done?

rubensworks:
> This is working now - the only remaining task is to work out how to mock broken urls to do some proper testing - @rubensworks ; are there examples you can point to where this is already being done?

I think the HTTP actor and/or dereference actors have some in their unit tests.

jeswr:
This should be ready to merge @rubensworks :)

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -446,6 +446,38 @@ describe('System test: QuerySparql', () => {
     });
   });
 
+  // TODO: Re-enable this once https://github.com/RubenVerborgh/AsyncIterator/pull/45 is closed
```
rubensworks, line 449: I'm not a big fan of commented code :P
Could we move this to an issue (if it still is a problem)?

packages/actor-http-intercept-wayback/README.md
```diff
@@ -0,0 +1,40 @@
+# Comunica Wayback Http Intercept Actor
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-http-intercept-wayback.svg)](https://www.npmjs.com/package/@comunica/actor-http-intercept-wayback)
+
+A Comunica actor to intercept HTTP requests to recover broken links using the WayBack  Machine
```
rubensworks, line 5:
```suggestion
A Comunica actor to intercept HTTP requests to recover broken links using the WayBack Machine
```

Can we also add a link?

packages/actor-http-intercept-wayback/lib/ActorHttpInterceptWayback.ts
```diff
@@ -0,0 +1,74 @@
...
+      });
+
+      // If the wayback machine returns a 200 status then use that result
+      if (fallbackResult.status === 200) {
+        [ result, fallbackResult ] = [ fallbackResult, result ];
+      }
+
+      // Consume stream to avoid process
+      if (fallbackResult.body) {
+        await stringifyStream(ActorHttp.toNodeReadable(fallbackResult.body));
```
rubensworks, line 64: Calling `fallbackResult.body.destroy()` would be more efficient in some cases.

engines/config-query-sparql/config/dereference/actors.json
```diff
@@ -12,7 +12,7 @@
       "@id": "urn:comunica:default:dereference/actors#http",
       "@type": "ActorDereferenceHttp",
       "beforeActors": { "@id": "urn:comunica:default:dereference/actors#fallback" },
-      "mediatorHttp": { "@id": "urn:comunica:default:http/mediators#main" }
+      "mediatorHttp": { "@id": "urn:comunica:default:http-intercept/mediators#main" }
```
rubensworks, line 15: I'm surprised this works. Wondering if it shouldn't?

Because the dereference-http actor is expecting a mediator over the HTTP bus, but it is actually receiving a mediator over the HTTP-intercept bus, which is a different thing.
So perhaps we need to change this param to `mediatorHttpIntercept`, and make it explicit typing-wise as well?

Related to this, we have several other actors that are using the HTTP mediator as well. Should we change this to this interceptor as well?

If we make these changes, this will be a breaking change, which we may not want to do (again) at this stage.

rubensworks: I think the wrapper-based approach discussed earlier might lead to less breakage.

rubensworks (changes requested):
Thanks @jeswr, the actor itself looks fine to me. Just not convinced yet about the way things are wired (see comments below).

engines/config-query-sparql/config/dereference/actors.json
jeswr: Not quite sure I understand what you mean by wrapper approach - and how it differs to what I've done here. Regardless I've cleaned all this up in #1077 by using two `http` bus' in the build (which I only just worked out how to do) rather than introducing an intercept `-bus`

Issue #1061, opened 2022-09-14
Support SPARQL endpoints with multiple SPARQL service description


marcelomachado:
#### Issue type:

- :bug: Bug 

____
#### Description:

I created a small web app that queries multiple RDF sources, including a Blazegraph SPARQL endpoint. If my only Comunica source is this Blazegraph SPARQL endpoint, I get the expected result. When I add another source and execute the query, it returns a 404 ERROR:
```
Invalid SPARQL endpoint response from http://{host}/bigdata/LBS/namespace/{namespace}/sparql (HTTP status 404).
```

However, I realized that the endpoint that appears in the error message is different than the one I was trying to use: 
```
http://{host}/blazegraph/namespace/{namespace}/sparql
```

That's when I decided to look at the service description and realized that there are multiple endpoints and in fact the one in the error message is not found:

```
<endpoint xmlns="http://www.w3.org/ns/sparql-service-description#" rdf:resource="http://{host}/blazegraph/namespace/{namespace}/sparql"/>
<endpoint xmlns="http://www.w3.org/ns/sparql-service-description#" rdf:resource="http://{host}/bigdata/LBS/namespace/{namespace}/sparql"/>
```

I don't have access to the Blazegraph service to remove such a description, actually I believe this shouldn't be the solution. I can't see why the query couldn't be completed with the endpoint that works (the endpoint I'm using in `sources`). I also don't understand why this is a problem that only happens when more than one source is selected.

FYI: This issue https://github.com/comunica/comunica/issues/535 seems to be similar, however the problem occurs when no endpoint is provided in the service description.

rubensworks:
I have no experience with Blazegraph myself, but I would guess this is a Blazegraph configuration issue.
If `http://{host}/bigdata/LBS/namespace/{namespace}/sparql` is exposed in the SD, then this would mean that it is enabled in Blazegraph, but is somehow broken.

As a workaround for this Blazegraph bug, it _may_ be possible to [force the `sparql` source type](https://comunica.dev/docs/query/advanced/source_types/).

rubensworks:
Hmm, too bad the workaround doesn't work.

The proper long-term solution would be to consider the two endpoints in the SD as alternatives to the same dataset. In this case, Comunica should first attempt to query the first one, and fallback to the second one if the first one fails (or some other load-balancing solution). However, implementing this would require quite some effort.

[...]

PR #1075, opened 2022-09-21
Add parameter to ignore sparql endpoints coming from the service description.

marcelomachado:
Added a parameter to disables/ignores the endpoints detected in the SD. This implementation is related to the issue #1061 .

In a JS application this parameter (`ignoreSparqlEndpointFromServiceDescription`) should be passed in the source list such as: 
```js
const result = await this.queryEngine.query(query, {
    sources: [
        { value: 'https://dbpedia.org/sparql', ignoreSparqlEndpointFromServiceDescription: true},
        { value: 'https://query.wikidata.org/sparql, type: 'sparql'}
    ]
});
```

Using the command line I chose to use the `+` character preceding the url that should be used without accessing the SD. Then:
```
> comunica-sparql +https://fragments.dbpedia.org/2016-04/en \
    sparql@https://query.wikidata.org/sparql \
    "SELECT * WHERE { ?s ?p ?o } LIMIT 100"
```

### DIFF
--- a/.github/workflows/new-pr-assign.yml
+++ b/.github/workflows/new-pr-assign.yml
@@ -1,0 +1,13 @@
+name: "Assign new PR"
+on:
+  issues:
+    types: [opened, edited]
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: delivery-much/actions-assigner@v1
+      with:
+        token: "${{ secrets.PAT }}"
+        team-reviewers: core


packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts
```diff
@@ -56,6 +56,12 @@ export class CliArgsHandlerBase implements ICliArgsHandler {
       sourceString = sourceString.slice(0, authMatches.index + 2) +
         sourceString.slice(authMatches.index + credentials.length + 3);
     }
+
+    if (sourceString.startsWith('+')) {
```
rubensworks, line 60: This microsyntax is not that self-explanatory, and may cause confusion in the future. So I would recommend adding a dedicated CLI argument for this.

packages/types/lib/IDataSource.ts
```diff
@@ -3,6 +3,7 @@ import type { IActionContext } from './IActionContext';
 
 export type IDataSource = string | RDF.Source | {
   type?: string;
+  ignoreSparqlEndpointFromServiceDescription?: boolean;
```
rubensworks, line 6: Instead of storing this in a new field, I would suggest storing this in the context, by introducing a new context key in the `context-entries` package. This would also align nicely with the new CLI argument by handling it in `CliArgsHandlerQuery`.

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -221,7 +226,11 @@ export class FederatedQuadSource implements IQuadSource {
       // Prepare the context for this specific source
       let context: IActionContext = getDataSourceContext(source, this.contextDefault);
       context = context.set(KeysRdfResolveQuadPattern.source,
-        { type: getDataSourceType(source), value: getDataSourceValue(source) });
+        {
```
rubensworks, line 229: The changes here won't be needed anymore when using the context to store this information.

packages/bus-rdf-resolve-quad-pattern/lib/utils.ts
```diff
@@ -98,6 +98,24 @@ export function getContextSourceUrl(source?: IDataSource): string | undefined {
   }
 }
 
+/**
+ * Check if the given source should be used without visit the service description.
+ * @param {IDataSource} source A source.
+ * @return {boolean} If the given source indicates the sparql endpoint from the service description should be ignored.
+ */
+export function shouldIgnoreSparqlEndpointFromServiceDescription(source: IDataSource): boolean {
```
rubensworks, line 106: Can be moved to the SPARQL actor

packages/actor-rdf-resolve-quad-pattern-hypermedia/lib/MediatedQuadSource.ts
```diff
@@ -15,19 +15,21 @@ export class MediatedQuadSource implements IQuadSource {
...
 
   public sourcesState: ISourcesState;
 
   private readonly cacheSize: number;
   private readonly maxIterators: number;
 
-  public constructor(cacheSize: number, context: IActionContext, firstUrl: string,
-    forceSourceType: string | undefined, maxIterators: number, mediators: IMediatorArgs) {
+  public constructor(cacheSize: number, context: IActionContext, firstUrl: string, forceSourceType: string | undefined,
+    ignoreSparqlEndpointFromServiceDescription: boolean | undefined, maxIterators: number, mediators: IMediatorArgs) {
```
rubensworks, line 27: The changes here won't be needed anymore when using the context to store this information.

rubensworks (changes requested):
Thanks for the PR!

In general, I would recommend using the context object (which can be defined on specific sources as well) to store this parameter, to align with how other source-based parameters are passed.

packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts
marcelomachado: Indeed, however this parameter should be used for specific uris. Any idea on how to map this?

rubensworks: We probably need a more expressive CLI approach to define source-level context options.
Is this something that you need for your use case?
If not, I would suggest leaving this open for future work (and making an issue(, and just enabling this option for _all_ sources for now.

PR #1081, opened 2022-09-28
Feat/wayback 2

jeswr:
Supercedes #1070 - uses bus with different name but still of type `ActorHttp`.

### DIFF
--- a/engines/config-query-sparql/config/http/actors-wayback.json
+++ b/engines/config-query-sparql/config/http/actors-wayback.json
@@ -1,0 +1,19 @@
+{
+  "@context": [
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/runner/^2.0.0/components/context.jsonld",
+
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-http-wayback/^2.0.0/components/context.jsonld",
+
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/bus-http/^2.0.0/components/context.jsonld"
+  ],
+  "@id": "urn:comunica:default:Runner",
+  "@type": "Runner",
+  "actors": [
+    {
+      "@id": "urn:comunica:default:http/actors#main",
+      "@type": "ActorHttpWayback",
+      "bus": { "@id": "ActorHttp:_fallback_bus" },
+      "mediatorHttp": { "@id": "urn:comunica:default:http/mediators#no-fallback" }
+    }
+  ]
+}
--- a/yarn.lock
+++ b/yarn.lock
@@ -2454,7 +2454,7 @@
     qs "^6.10.1"
     url-parse "^1.5.3"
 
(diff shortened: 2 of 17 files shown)


yarn.lock
```diff
@@ -4802,7 +4802,7 @@ create-hmac@^1.1.0, create-hmac@^1.1.4, create-hmac@^1.1.7:
     safe-buffer "^5.0.1"
     sha.js "^2.4.8"
 
-cross-fetch@^3.0.5, cross-fetch@^3.0.6:
+cross-fetch@^3.0.5, cross-fetch@^3.0.6, cross-fetch@^3.1.5:
```
jeswr, line 4805: Please run the `ncu` script once this is merged to resolve this problem with different packages being on different versions of things like `cross-fetch`.

rubensworks: You're welcome to submit a PR :-)

jeswr: will do after this is merged to avoid conflicts

engines/config-query-sparql/config/http/actors-wayback.json
```diff
@@ -0,0 +1,19 @@
...
+
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-http-wayback/^2.0.0/components/context.jsonld",
+
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/bus-http/^2.0.0/components/context.jsonld"
+  ],
+  "@id": "urn:comunica:default:Runner",
+  "@type": "Runner",
+  "actors": [
+    {
+      "@id": "urn:comunica:default:http/actors#main",
```
rubensworks, line 13:
```suggestion
      "@id": "urn:comunica:default:http/actors#wayback",
```

rubensworks (approved):
Having different instantiations of the http bus is a great idea, this looks like a really nice solution!

Before we merge, could you do a quick manual check of update queries (e.g. over a solid pod) (create and update) to see if this fallback approach properly handles this case as well? (since we don't have any integration tests for this yet).

engines/config-query-sparql/config/http/actors-wayback.json
jeswr: This change is by design. By default actor the `WayBack` actor is the only one subscribed to `#main` and the rest are subscribed to `#no-fallback` so that we can always do the wayback attempt on 404ing http requests in the default configuration.

rubensworks: I suppose you're talking about the mediator or bus id?
But I was talking about the actor id here, for which we follow this naming strategy: https://github.com/comunica/comunica/tree/master/engines/config-query-sparql/config#iri-strategy

rubensworks: I see you've already merged the PR, could you apply this fix in a separate PR?

PR #1085, opened 2022-09-29
chore: add solid integration tests

jeswr:
This PR adds some integration tests against a Solid Pod.

There are two TODOs
 - Fix failing tests by resolving https://github.com/comunica/comunica-feature-solid/issues/43
 - Improve test coverage with more complex operations
 - Use jest mock functionality properly rather than overriding global fetch
 - Test against different CSS versions / configs?

Currently CI/CD is failing because CSS does not work on Node 18. I'm not sure if there is an easy way to disable these tests specifically on that version of node for now @rubensworks ?

### DIFF
--- a/engines/config-query-sparql/config/http/actors-wayback.json
+++ b/engines/config-query-sparql/config/http/actors-wayback.json
@@ -1,0 +1,19 @@
+{
+  "@context": [
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/runner/^2.0.0/components/context.jsonld",
+
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-http-wayback/^2.0.0/components/context.jsonld",
+
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/bus-http/^2.0.0/components/context.jsonld"
+  ],
+  "@id": "urn:comunica:default:Runner",
+  "@type": "Runner",
+  "actors": [
+    {
+      "@id": "urn:comunica:default:http/actors#main",
+      "@type": "ActorHttpWayback",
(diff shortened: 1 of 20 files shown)


rubensworks:
This is really useful stuff @jeswr!

> Currently CI/CD is failing because CSS does not work on Node 18. I'm not sure if there is an easy way to disable these tests specifically on that version of node for now @rubensworks ?

What we could do, is running those tests in a separate GH actions job (we should be able to create another jest config file for this).
Then we can just disable (or still run, but not error) Node 18 for this GH actions job, until the Node 18 issues are resolved.

joachimvh:
> CSS does not work on Node 18

This is the error I see in the logs:

> error @inrupt/solid-client-authn-core@1.12.2: The engine "node" is incompatible with this module. Expected version "^14.0.0 || ^16.0.0". Got "18.9.1"

The issue seems to be that yarn errors if a node version is incompatible with a package whereas `npm` just warns:

> npm WARN EBADENGINE Unsupported engine {
> npm WARN EBADENGINE   package: '@inrupt/solid-client-authn-core@1.12.2',
> npm WARN EBADENGINE   required: { node: '^14.0.0 || ^16.0.0' },
> npm WARN EBADENGINE   current: { node: 'v18.9.1', npm: '8.19.1' }
> npm WARN EBADENGINE }

The above is from a CSS 18.x CI run.

Might be worth looking into if you can change this for yarn because otherwise you depend on those dependencies their configuration.

joachimvh:
I was going to suggest installing CSS without its dev dependencies as I think the issue only occurs there, but having had a look at what happens here it is actually not the CSS dependency but the `@inrupt/solid-client-authn-*` dependencies that are the cause of the issue here (as those are also the devDependencies we have that cause the same issue).

PR #1088, opened 2022-09-29
Add support for http retries

### LINKED ISSUE #1012: Feature: Support HTTPS Retries
#### Issue type:

____
#### Description:

In order to better support unstable internet connections, Comunica should have a context entry option to enable retries on failed network calls. I would say that the number and frequency of retries should be configurable.

Laurin-W:
Allow setting httpRetryCount and httpRetryDelay to ActorHttpFetch context.

Closes #1012

### DIFF
--- a/packages/actor-http-fetch/lib/ActorHttpFetch.ts
+++ b/packages/actor-http-fetch/lib/ActorHttpFetch.ts
@@ -32,6 +32,64 @@ export class ActorHttpFetch extends ActorHttp {
     return { time: Number.POSITIVE_INFINITY };
   }
 
+  /**
+   * Perform a fetch request, taking care of retries
+   * @param fetchFn
+   * @param requestInput Url or RequestInfo to pass to fetchFn
+   * @param requestInit RequestInit to pass to fetch function
+   * @param retryCount Maximum retries after which to abort
+   * @param retryDelay Time in milliseconds to wait between retries
+   * @returns a fetch `Response` object
+   */
+  private static async getResponse(
+    fetchFn: (input: RequestInfo | URL, init?: RequestInit | undefined) => Promise<Response>,
+    requestInput: RequestInfo | URL,
+    requestInit: RequestInit,
+    retryCount: number,
+    retryDelay: number,
+    throwOn500Response: boolean,
+  ): Promise<Response> {
+    let lastError: unknown;
+    // The retryCount is 0-based. Therefore, add 1 to triesLeft.
+    let triesLeft = retryCount + 1;
+
+    // When retry count is greater than 0, repeat fetch.
+    while (triesLeft-- > 0) {
+      try {
+        const response = await fetchFn(requestInput, requestInit);
+        // Check, if server sent a 5XX response.
+        if (throwOn500Response && response.status >= 500 && response.status < 600) {
+          throw new Error(`Server replied with response code ${response.status}: ${response.statusText}`);
+        }
+        return response;
+      } catch (error: unknown) {
+        lastError = error;
+        // If the fetch was aborted by timeout, we won't retry.
+        if (requestInit.signal?.aborted) {
+          throw error;
+        }
+
+        if (triesLeft > 0) {
+          // Wait for specified delay, before retrying.
+          await new Promise((resolve, reject) => {
+            setTimeout(resolve, retryDelay);
+            // Cancel waiting, if timeout is reached.
+            requestInit.signal?.addEventListener('abort', () => {
+              reject(new Error('Fetch aborted by timeout.'));
+            });
+          });
+        }
+      }
+    }
+    // The fetch was not successful. We throw.
+    if (retryCount > 0) {
+      // Feedback the last error, if there were retry attempts.
+      throw new Error(`Number of fetch retries (${retryCount}) exceeded. Last error: ${String(lastError)}`);
+    } else {
+      throw lastError;
+    }
+  }
+
   public async run(action: IActionHttp): Promise<IActorHttpOutput> {
     // Prepare headers
     const initHeaders = action.init?.headers ?? {};
@@ -76,50 +134,17 @@ export class ActorHttpFetch extends ActorHttp {
 
     try {
       requestInit = await this.fetchInitPreprocessor.handle(requestInit);
-      let retryCount: number = action.context?.get(KeysHttp.httpRetryCount) ?? 0;
-
+      // Number of retries to perform after a failed fetch.
+      const retryCount: number = action.context?.get(KeysHttp.httpRetryCount) ?? 0;
+      const retryDelay: number = action.context?.get(KeysHttp.httpRetryDelay) ?? 0;
+      const retryOn5xxResponse: boolean = action.context?.get(KeysHttp.httpRetryOn5xx) ?? false;
       const customFetch: ((input: RequestInfo, init?: RequestInit) => Promise<Response>) | undefined = action
         .context?.get(KeysHttp.fetch);
 
-      // Actual fetch respecting timeout and retry count.
-      const getResponse = async(): Promise<Response> => {
-        let lastError: unknown;
-        const retriesAllowed = retryCount > 0;
-
-        // When retry count is greater than 0, repeat fetch.
-        while (retryCount-- >= 0) {
-          try {
-            return await (customFetch || fetch)(action.input, requestInit);
-          } catch (error: unknown) {
-            lastError = error;
-            // If the fetch was aborted by timeout, we won't retry.
-            if (requestInit.signal?.aborted) {
-              throw error;
-            }
-
-            if (retryCount >= 0) {
-              // Wait for specified delay, before retrying.
-              await new Promise((resolve, reject) => {
-                setTimeout(resolve, action.context?.get(KeysHttp.httpRetryDelay) ?? 0);
-                // Cancel waiting, if timeout is reached.
-                requestInit.signal?.addEventListener('abort', () => {
-                  reject(new Error('Fetch aborted by timeout.'));
-                });
-              });
-            }
-          }
-        }
-        // The fetch was not successful. We through.
-        if (retriesAllowed) {
-          // Feedback the last error, if there were retry attempts.
-          throw new Error(`Number of fetch retries exceeded. Last error: ${String(lastError)}`);
-        } else {
-          throw lastError;
-        }
-      };
-
       // Execute the fetch (with retries and timeouts, if applicable).
-      const response = await getResponse();
+      const response = await ActorHttpFetch.getResponse(
+        customFetch || fetch, action.input, requestInit, retryCount, retryDelay, retryOn5xxResponse,
+      );
 
       // We remove or update the timeout
       if (requestTimeout !== undefined) {
--- a/packages/actor-http-fetch/lib/ActorHttpFetch.ts
+++ b/packages/actor-http-fetch/lib/ActorHttpFetch.ts
@@ -76,12 +76,50 @@ export class ActorHttpFetch extends ActorHttp {
 
     try {
       requestInit = await this.fetchInitPreprocessor.handle(requestInit);
+      let retryCount: number = action.context?.get(KeysHttp.httpRetryCount) ?? 0;
 
-      // Perform request
       const customFetch: ((input: RequestInfo, init?: RequestInit) => Promise<Response>) | undefined = action
         .context?.get(KeysHttp.fetch);
 
-      const response = await (customFetch || fetch)(action.input, requestInit);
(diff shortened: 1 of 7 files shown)


packages/actor-http-fetch/lib/ActorHttpFetch.ts
```diff
@@ -76,12 +76,46 @@ export class ActorHttpFetch extends ActorHttp {
...
+            }
+          }
+        }
+        // If retry count was 0, i.e. only one error was recorded, just throw it.
+        if (errors.length === 1) {
+          throw errors[0];
+        } else {
+          throw new Error(`Number of fetch retries exceeded. Encountered errors: ["${errors.join('", "')}"]`);
+        }
+      };
```
jeswr, line 115: Rather than reporting all the errors I would just report the first error encountered. I think it is fairly safe to assume that the same problem is going to be encountered each time and reporting all the errors is going to create massive error messages if you do a significant number of retries.

packages/actor-http-fetch/test/ActorHttpFetch-test.ts
```diff
@@ -443,5 +443,67 @@ This error can be disabled by modifying the 'httpBodyTimeout' and/or 'httpTimeou
...
+
+    it('should retry with a delay', async() => {
+      let numberOfRetries = 0;
+      const customFetch = jest.fn(async(_, init) => {
+        if (numberOfRetries < 2) {
+          numberOfRetries++;
+          throw new Error('Retry count not reached.');
+        }
+        return {};
+      });
```
jeswr, line 455: Rather than having this logic you could probably just test how many times the fetch function was called https://stackoverflow.com/questions/52885551/jest-check-how-many-times-a-mocked-module-function-was-called

packages/actor-http-fetch/test/ActorHttpFetch-test.ts
```diff
@@ -443,5 +443,67 @@ This error can be disabled by modifying the 'httpBodyTimeout' and/or 'httpTimeou
...
+        context: new ActionContext({
+          [KeysHttp.fetch.name]: customFetch,
+          [KeysHttp.httpRetryCount.name]: 2,
+          [KeysHttp.httpRetryDelay.name]: 100,
+        }),
+      });
+
+      const time = performance.now() - start;
+      expect(time).toBeGreaterThan(200);
+      expect(time).toBeLessThan(250);
```
jeswr, line 470: What guarantees it being lest than `250`?

packages/actor-http-fetch/test/ActorHttpFetch-test.ts
jeswr: Ah I see you have it below - why is this needed as well?

packages/actor-http-fetch/lib/ActorHttpFetch.ts
```diff
@@ -76,12 +76,46 @@ export class ActorHttpFetch extends ActorHttp {
...
+            if (requestInit.signal?.aborted) {
+              throw error;
+            }
+            errors.push(error);
+            // Wait for specified delay.
+            if (retryCount >= 0) {
+              await new Promise((resolve, reject) => {
+                // Cancel waiting, if timeout is reached.
+                requestInit.signal?.addEventListener('abort', () => {
+                  reject(new Error('Fetch aborted by timeout.'));
```
jeswr, line 102:
```suggestion
                requestInit.signal?.on('abort', reason => {
                  reject(reason ?? new Error('Request aborted before timeout.'));
```

packages/actor-http-fetch/lib/ActorHttpFetch.ts
```diff
@@ -76,12 +76,46 @@ export class ActorHttpFetch extends ActorHttp {
...
+          try {
+            return await (customFetch || fetch)(action.input, requestInit);
+          } catch (error: unknown) {
+            // If the fetch was aborted by timeout, we won't retry.
+            if (requestInit.signal?.aborted) {
+              throw error;
+            }
+            errors.push(error);
+            // Wait for specified delay.
+            if (retryCount >= 0) {
```
jeswr, line 98: Won't this always be true?

jeswr (changes requested):
Nice work! I've left some passing comments.

I haven't looked at the aborting behavior in a while - in particular the behavior of the `fetchInitPreprocessor` so I will leave someone who has looked into that more recently to check the abort logic; otherwise I will have a proper look tommorow.

packages/actor-http-fetch/test/ActorHttpFetch-test.ts
Laurin-W: The rest of the test does not take more than 10 ms at max (basically just calls the mockedFetch). So I thought, it might be reasonable to expect the whole test to be within that bounds and larger time spans indicate that something is going wrong.

jeswr: I think leave it out just because the tests suites are run concurrently and so is a little unpredictable and has the potential to become quite flakey as a test and cause workflow failures.

packages/actor-http-fetch/test/ActorHttpFetch-test.ts
Laurin-W: Not sure if I'm getting you right here. The point of counting the retries was that I want the fetch function to throw twice but resolve on the third time. The page you linked just talks about testing the number of function calls.

packages/actor-http-fetch/lib/ActorHttpFetch.ts
rubensworks: I agree with the above.
We could optionally append something like `(failed for x retries)` to the error message.

packages/actor-http-fetch/lib/ActorHttpFetch.ts
```diff
@@ -76,12 +76,46 @@ export class ActorHttpFetch extends ActorHttp {
...
       requestInit = await this.fetchInitPreprocessor.handle(requestInit);
+      let retryCount: number = action.context?.get(KeysHttp.httpRetryCount) ?? 0;
 
-      // Perform request
       const customFetch: ((input: RequestInfo, init?: RequestInit) => Promise<Response>) | undefined = action
         .context?.get(KeysHttp.fetch);
 
-      const response = await (customFetch || fetch)(action.input, requestInit);
+      // Actual fetch respecting timeout and retry count.
+      const getResponse = async(): Promise<Response> => {
```
rubensworks, line 85: The `run` method is starting to become quite long now.

Shall we move this inline function to a dedicated (possibly static) method on `ActorHttpFetch`?

packages/actor-http-fetch/lib/ActorHttpFetch.ts
```diff
@@ -76,12 +76,46 @@ export class ActorHttpFetch extends ActorHttp {
...
+        // When retry count is greater than 0, record the errors for failed requests.
+        const errors: unknown[] = [];
+        while (retryCount-- >= 0) {
+          try {
```
rubensworks, line 89: Open question: next to catching network errors, should we also handle server errors (probably only those in the 500 range then)?

rubensworks: By default, probably not, but an option for this might be valuable.

packages/context-entries/lib/Keys.ts
```diff
@@ -48,6 +48,15 @@ export const KeysHttp = {
    * but until the response is fully consumed.
    */
   httpBodyTimeout: new ActionContextKey<boolean>('@comunica/bus-http:http-body-timeout'),
+  /**
+   * Number of retries to make on failed network calls. This only takes effect
+   * on errors thrown during the initial fetch() call and not while streaming the body.
```
rubensworks, line 53: > and not while streaming the body.

I get why. But perhaps it should?
What do you think @jeswr?

Complexity of this PR would increase quite a bit with this though, so if we do this, we may want to do it in a separate PR.

packages/context-entries/lib/Keys.ts
```diff
@@ -48,6 +48,15 @@ export const KeysHttp = {
...
+   */
+  httpRetryCount: new ActionContextKey<number>('@comunica/bus-http:http-retry-count'),
+  /**
+   * Delay in milliseconds to wait between fetch retries.
+   */
+  httpRetryDelay: new ActionContextKey<number>('@comunica/bus-http:http-retry-delay'),
```
rubensworks, line 59: Might be good to also mention that this requires `httpRetryCount` to be set.

rubensworks (changes requested):
Really nice stuff @Laurin-W!

Just some minor nits and questions.

packages/context-entries/lib/Keys.ts
jeswr: I agree we should tackle it at some point but I also think it should be a separate PR.


In order to do it properly we would have to add code to find the point in the retried stream at which the original stream broke and continue from there (since Comunica will have already started processing the first stream which broke; and we don't want to introduce broken results).

It would also be completely different error handling implementation wise because rather than throwing an error in the actor as we do here; we would instead need to be doing error handling on the stream.

rubensworks: Yeah, it will not be trivial, that's for sure.
There's probably even a research paper in there :-)

packages/actor-http-fetch/lib/ActorHttpFetch.ts
Laurin-W: Regarding the `numberOfRetries` variable of the first comment. That's for passing the request on the third attempt. Or is there a jest-ier way of keeping track of this?

Regarding the second comment with the `signal?.on(...))` proposal - that does not seem to work as `signal` only exposes a lambda called `onabort` which is however undefined in that context.

packages/actor-http-fetch/lib/ActorHttpFetch.ts
Laurin-W: Not on the last allowed try in which case the loop is subsequently exited and and the encountered error is thrown.

packages/actor-http-fetch/lib/ActorHttpFetch.ts
Laurin-W: added an option httpRetryOn5xx :)

Laurin-W:
Thanks for the feedback!
I made the change suggested and changed a couple of variable names and comments to be more explicit, I hope :)

packages/actor-http-fetch/lib/ActorHttpFetch.ts
jeswr: > Regarding the numberOfRetries variable of the first commen

Oh - that makes sense! Apologies for the slightly rushed review on my end last night :)

> Regarding the second comment with the signal?.on(...)) proposal

Ok - no worries; I'd assumed this was extending the `EventEmitter` object but I guess not.

packages/actor-http-fetch/test/ActorHttpFetch-test.ts
jeswr: Yep - I was just rushed when I did the review and missed the intention. This looks good!

packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts
```diff
@@ -123,6 +123,18 @@ export class CliArgsHandlerBase implements ICliArgsHandler {
...
         },
+        httpRetryCount: {
+          type: 'number',
+          describe: 'The number of retries to perform on failed fetch requests',
+        },
+        httpRetryDelay: {
+          type: 'number',
+          describe: 'The number of milliseconds to wait between fetch retries',
+        },
+        httpRetryOn5xx: {
```
jeswr, line 134: I'm wondering if this would be better named as `httpRetryOnServerError`. Probably wouldn't bother changing it unless other changes need be made as well.

jeswr (approved):
Fantastic work! Just left a minor comment around naming.

packages/actor-http-fetch/lib/ActorHttpFetch.ts
rubensworks: Nice!

packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts
rubensworks: Had this in mind as well, so might be good to change still indeed.

Laurin-W: Got that changed :+1:

Laurin-W:
From my side, this is good to go then :)

rubensworks:
Really nice work @Laurin-W, thanks a lot!

PR #1092, opened 2022-10-05
Feat/solid integration tests

Laurin-W:
Addressing two of the issues of PR https://github.com/comunica/comunica/pull/1085

- Change jest mock of `globalThis.fetch`, resetting `globalThis.fetch` after tests ran (even though it should probably work as the DOM is reset for each test file iirc)
- Add condition in `jest.conf.ts` to ignore `Solid-test.no-node18.ts` and make node 18 deps optional
  - Alternatively, I could add another jest config file and do the condition logic in the github action. Any opinion here?


rubensworks:
Thanks for this @Laurin-W

Could you make this PR target the master branch? Otherwise the diffs are a bit difficult to view here on GH.

> Add condition in jest.conf.ts to ignore Solid-test.no-node18.ts and make node 18 deps optional
 Alternatively, I could add another jest config file and do the condition logic in the github action. Any opinion here?

Since this workaround should be temporary, I would suggest giving the test file another name, so that we don't have to rename the file once the Node 18 issues have been resolved (perhaps they will also occur on Node 19 and beyond!).
Then, in the jest config, we can just explicitly exclude this file. Will make it a bit easier to see explicitly which tests require fixing in Node 18.
Suggestion for new file name: `QuerySparql-solid-test.ts`

engines/query-sparql/package.json
```diff
@@ -168,6 +168,11 @@
     "@comunica/runner": "^2.4.0",
     "@comunica/runner-cli": "^2.4.0"
   },
+  "optionalDependencies": {
+    "@inrupt/solid-client-authn-core": "^1.12.2",
```
rubensworks, line 172: I guess these should be devDependencies? (if so, they can be added as devdep in the root package.json)

engines/query-sparql/test/configs/seed.json
```diff
@@ -0,0 +1,7 @@
+[
```
rubensworks, line 1: I suggest renaming this file to `solid-css-seed.json`

rubensworks (changes requested):

engines/query-sparql/test/QuerySparql-solid-test.ts
```diff
@@ -0,0 +1,296 @@
...
+// eslint-disable-next-line no-undef
+const globalFetch = globalThis.fetch;
+
+function createApp() {
+  return new AppRunner().create(
+    {
+      mainModulePath: resolveModulePath(''),
+      typeChecking: false,
+    },
+    resolveModulePath('config/default.json'),
```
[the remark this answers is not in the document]
jeswr: This resolves to the CSS configs

engines/query-sparql/package.json
Laurin-W: The reason, I made them optional was that it was failing the `yarn install --pure-lockfile` in the CI otherwise.
Should I add them to the `optionalDependencies` of the root `package.json` though?

rubensworks: > The reason, I made them optional was that it was failing the yarn install --pure-lockfile in the CI otherwise.

Why was it failing exactly?

Laurin-W: Sorry, that might've been a bit unclear. `solid-client-authn-core` and `solid-client-authn-node` do not support node 18 yet. That's why the `QuerySparql-solid-test.ts` needed to be excluded from testing as well when in a node 18 environment, as the test depends on `solid-client-authn-*`.

Laurin-W:
Do you have an idea why the CI for 16.x is failing?
I can sort of reproduce the error on my local machine: When I run all tests (`yarn run test-ci`), it fails with the same lines not being covered (`QueryEngineFactory.ts` ll. 9-12). However, the test that actually would cover the missing lines (`QueryEngineFactoryBase-test.ts`), fails on my local machine, whereas here in the GH action, it does not.

When I execute the test `QueryEngineFactoryBase-test.ts` in isolation, it works fine. Do you have an idea?

Failing GH action: https://github.com/comunica/comunica/actions/runs/3199811931/jobs/5226003870

engines/query-sparql/package.json
rubensworks: Ah, I see.

Wondering if there's a way (perhaps yarn-specific) to only install a dep if on a certain node version.

If not, then we have to resort to installing those deps in the CI.

The problem with optionalDeps is that end-users will also end up having these installed, while they are not needed.

rubensworks:
> Do you have an idea why the CI for 16.x is failing?
I can sort of reproduce the error on my local machine: When I run all tests (yarn run test-ci), it fails with the same lines not being covered (QueryEngineFactory.ts ll. 9-12). However, the test that actually would cover the missing lines (QueryEngineFactoryBase-test.ts), fails on my local machine, whereas here in the GH action, it does not.

That's strange. No idea tbh.

What could be causing this is an incorrect import in the test (e.g. pointing to the package index instead of the specific class file), but this doesn't appear to be the case here.

engines/query-sparql/package.json
Laurin-W: I found the following plugin: https://www.npmjs.com/package/optional-dev-dependency As I'm absolutely no expert with npm package management - do you think that's too hacky?

How would it work installing the deps in the CI? Adding the dependency with `yarn add` before running `yarn install`, removing it when on node 18, or something else?

rubensworks: > I found the following plugin: https://www.npmjs.com/package/optional-dev-dependency As I'm absolutely no expert with npm package management - do you think that's too hacky?

Not sure, doesn't look very well maintained.

> How would it work installing the deps in the CI? Adding the dependency with yarn add before running yarn install, removing it when on node 18, or something else?

Yes, adding them after install on Node 18 could work!

Laurin-W:
Oookay, finally this is all running fine :D I added another test covering the file.
Jesse suspects, the 16.x error might've come with adding the `testPathIgnorePatterns`.

.github/workflows/ci.yml
```diff
@@ -28,6 +28,9 @@ jobs:
             .rdf-test-suite-cache
             .rdf-test-suite-ldf-cache
           key: ${{ runner.os }}-test-modules-${{ secrets.CACHE_VERSION }}-${{ hashFiles('**/yarn.lock') }}
+      - name: Remove node 18 incompatible dependencies, if on node 18
+        if: ${{ matrix.node-version == '18.x'}}
+        run: yarn remove @inrupt/solid-client-authn-core @inrupt/solid-client-authn-node @solid/community-server -W
```
rubensworks, line 33: I would suggest adding these deps here via `yarn add` if _not_ on Node 18.
Because currently, devs cloning the Comunica repo on Node 18 and installing will see errors, which is probably not what we want.

rubensworks (changes requested):
I would suggest recommend reversing the way we install the inrupt libs, but other than that, everything looks perfect!

.github/workflows/ci.yml
Laurin-W: Mh, good point! For devs working on Node 16, the issue remaining would be the missing dependency errors when running tests, wouldn't it?.
How about, I move that functionality to the package.json `preinstall` script instead?

rubensworks: > Mh, good point! For devs working on Node 16, the issue remaining would be the missing dependency errors when running tests, wouldn't it?

Ah yes, you're right.

> How about, I move that functionality to the package.json preinstall script instead?

Yes, that sounds like a clean solution!

Laurin-W: Ok, I added the following:
```
    "COMMENT1": "echo ATTENTION: The command preinstall will also check, if on node 18 and remove node 18-incompatible dependencies or add them, if not on node 18. TODO: Remove this, once the dependencies support node 18.",
    "preinstall": "if [ $( node -v | sed -r \"s/v([0-9]+).*/\\1/g\" ) = 18 ]; then yarn remove -W --ignore-scripts --ignore-engines @inrupt/solid-client-authn-core @inrupt/solid-client-authn-node @solid/community-server 2> /dev/null; else yarn add -DW --ignore-scripts @inrupt/solid-client-authn-core @inrupt/solid-client-authn-node @solid/community-server 2> /dev/null ;fi; node -e \"if (!fs.existsSync('packages/runner-cli/bin/run.js')) { var os = fs.createWriteStream('packages/runner-cli/bin/run.js'); os.write('#!/usr/bin/env node\\nconsole.error(\\'Temporary runner script during lerna installation. If you see this, please run yarn install again.\\');\\nprocess.exit(1);'); os.end(); }\" && node -e \"if (!fs.existsSync('packages/runner/bin/compile-config.js')) { var os = fs.createWriteStream('packages/runner/bin/compile-config.js'); os.write('#!/usr/bin/env node\\nconsole.error(\\'Temporary compile-config script during lerna installation. If you see this, please run yarn install again.\\');\\nprocess.exit(1);'); os.end(); }\"",
```

Or in a more readable form: Add or remove the dependencies depending on node version. I added the `yarn add`, to prevent missing dependency issues when switching node versions.
```
if [ $( node -v | sed -r \"s/v([0-9]+).*/\\1/g\" ) = 18 ]; then
    yarn remove -W --ignore-scripts --ignore-engines @inrupt/solid-client-authn-core @inrupt/solid-client-authn-node @solid/community-server 2> /dev/null;
else
    yarn add -DW --ignore-scripts @inrupt/solid-client-authn-core @inrupt/solid-client-authn-node @solid/community-server 2> /dev/null ;
fi; 
```

I'm not too happy with the `COMMENT1` "script", but I guess it's better than not having a hint as to what this does, especially if running in errors.

jeswr: > I'm not too happy with the COMMENT1 "script", but I guess it's better than not having a hint as to what this does, especially if running in errors.

Perhaps this could be the following:

```
"temp:remove_broken_node_18_dependencies": "if [ $( node -v | sed -r \"s/v([0-9]+).*/\\1/g\" ) = 18 ]; then yarn remove -W --ignore-scripts --ignore-engines @inrupt/solid-client-authn-core @inrupt/solid-client-authn-node @solid/community-server 2> /dev/null; else yarn add -DW --ignore-scripts @inrupt/solid-client-authn-core @inrupt/solid-client-authn-node @solid/community-server 2> /dev/null ;fi; node -e \"if (!fs.existsSync('packages/runner-cli/bin/run.js')) { var os = fs.createWriteStream('packages/runner-cli/bin/run.js'); os.write('#!/usr/bin/env node\\nconsole.error(\\'Temporary runner script during lerna installation. If you see this, please run yarn install again.\\');\\nprocess.exit(1);'); os.end(); }\" && node -e \"if (!fs.existsSync('packages/runner/bin/compile-config.js')) { var os = fs.createWriteStream('packages/runner/bin/compile-config.js'); os.write('#!/usr/bin/env node\\nconsole.error(\\'Temporary compile-config script during lerna installation. If you see this, please run yarn install again.\\');\\nprocess.exit(1);'); os.end(); }\""
"preinstall": "npm run temp:remove_broken_node_18_dependencies"
```

rubensworks:
Really nice work @Laurin-W, thanks a lot for this!

Issue #949, opened 2022-03-11
Improve injection of context key shortcuts


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

We should make it easier to inject context key shortcuts when creating custom engines.

Concretely:

- [ ] It should be possible to easily extend/overwrite the default context shortcuts here: https://github.com/comunica/comunica/blob/master/packages/actor-init-query/lib/ActorInitQueryBase.ts#L99-L118
- [ ] Custom engines should be able to pass custom TS interfaces to `QueryEngineBase` to make the custom context shortcuts available for type-safety.

rubensworks:
The second step (passing custom context interfaces to `QueryEngineBase`) should be fairly straightforward.
Each engine extension could expose a custom `IQueryContextCommon` extension into their [types](https://github.com/comunica/comunica-feature-link-traversal/tree/master/packages/types-link-traversal) package, which can then be hooked into `QueryEngineBase` (may require a new generic param).

The first step however (extend/overwrite the default context shortcuts) will be more tricky.
The easiest option would be to override `contextKeyShortcuts` for each new engine, but this has the downside that we'll have to do a lot of copy-pasting for the common shortcuts.
A better option would be to _extend_ `contextKeyShortcuts` instead of overriding it, but this would probably require an addition to Components.js, so that we can extend default JSON-based string values.
Another option would be to add a new optional parameter `contextKeyShortcutsExtensions`, which is a `Record<string, string>[]` that can be set by engines wanting to extend the default shortcuts, which are then squashed together by `ActorInitQueryBase` after construction. (this is an array, because multiple engines may extend from each other, and may want to inherit each other's overrides)

PR #1102, opened 2022-10-17
Chore: improve context key injection

Laurin-W:
Hey @Ruben
addressing issue https://github.com/comunica/comunica/issues/949, I created this PR.

I had to add an ignoreComponent record to `.componentsjs-generator-config.json`'s as it does not seem to support Partial correctly which is used here for context type support in the IDE.

I hope this has caught the right direction and happy for feedback :)

### DIFF
--- a/packages/actor-init-query/lib/ActorInitQueryBase.ts
+++ b/packages/actor-init-query/lib/ActorInitQueryBase.ts
@@ -9,12 +9,13 @@ import type { MediatorQueryResultSerializeHandle,
   MediatorQueryResultSerializeMediaTypes,
   MediatorQueryResultSerializeMediaTypeFormats } from '@comunica/bus-query-result-serialize';
 import type { IActorTest } from '@comunica/core';
-import type { Logger } from '@comunica/types';
+import type { IQueryContextCommon, Logger } from '@comunica/types';
 
 /**
  * A browser-safe comunica Query Init Actor.
  */
-export class ActorInitQueryBase extends ActorInit implements IActorInitQueryBaseArgs {
+export class ActorInitQueryBase<QueryContext extends IQueryContextCommon = IQueryContextCommon>
+  extends ActorInit implements IActorInitQueryBaseArgs<QueryContext> {
   public readonly mediatorOptimizeQueryOperation: MediatorOptimizeQueryOperation;
   public readonly mediatorQueryOperation: MediatorQueryOperation;
   public readonly mediatorQueryParse: MediatorQueryParse;
@@ -28,9 +29,10 @@ export class ActorInitQueryBase extends ActorInit implements IActorInitQueryBase
   public readonly queryString?: string;
   public readonly defaultQueryInputFormat?: string;
   public readonly context?: string;
-  public readonly contextKeyShortcuts: Record<string, string>;
+  public readonly contextKeyShortcuts: Record<string, string> & Partial<Record<keyof QueryContext, string>>;
   /** Array of `contextKeyShortcuts` appended to `contextKeyShortcuts` during construction. */
-  public readonly contextKeyShortcutsExtensions?: Record<string, string>[];
+  public readonly contextKeyShortcutsExtensions?: (Record<string, string>
+  | Partial<Record<keyof Omit<QueryContext, keyof IQueryContextCommon>, string>>)[];
 
   /**
    * Create new ActorInitQueryBase object.
@@ -60,7 +62,8 @@ export class ActorInitQueryBase extends ActorInit implements IActorInitQueryBase
   }
 }
 
-export interface IActorInitQueryBaseArgs extends IActorInitArgs {
+export interface IActorInitQueryBaseArgs<QueryContext extends IQueryContextCommon = IQueryContextCommon>
+  extends IActorInitArgs {
   /**
    * The query operation optimize mediator
    */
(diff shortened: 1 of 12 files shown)


Laurin-W:
And there seems to be an issue with `QuerySparql-solid-test.ts` on node 14 but I can't reproduce it on my local machine..

packages/actor-init-query/lib/ActorInitQueryBase.ts
```diff
@@ -28,9 +29,27 @@ export class ActorInitQueryBase extends ActorInit implements IActorInitQueryBase
...
+   * Create new ActorInitQueryBase object.
+   * @param args.contextKeyShortcutsExtensions Array of `contextKeyShortcuts` that are merged
+   *   with the `contextKeyShortcuts` field. This allows adding shortcuts to the defaults.
+   * @throws When duplicate keys are present in `args.contextKeyShortcuts`
+   *  and `args.contextKeyShortcutsExtensions`.
+   */
   public constructor(args: IActorInitQueryBaseArgs) {
+    // Add additional contextKeyShortcuts.
+    args.contextKeyShortcutsExtensions?.forEach(extensionShortcuts => {
+      // Throw, f there are duplicate keys that are to be added to `contextKeyShortcuts`.
```
rubensworks, line 47:
```suggestion
      // Throw, if there are duplicate keys that are to be added to `contextKeyShortcuts`.
```

packages/actor-init-query/lib/ActorInitQueryBase.ts
```diff
@@ -126,5 +146,25 @@ export interface IActorInitQueryBaseArgs extends IActorInitArgs {
...
    * }}
    */
-  contextKeyShortcuts: Record<string, string>;
+  contextKeyShortcuts: Record<string, string> | Partial<Record<keyof QueryContext, string>>;
+  /**
+   * An array of `contextKeyShortcuts` that are to be appended to the (default) `contextKeyShortcuts`
+   * (which are by default injected by component.js).
+   *
+   * The appending happens in the constructor call. Conflicting keys will cause an error.
+   * If you extend `ActorInitQueryBase` and want to add custom shortcuts, do so as follows:
```
rubensworks, line 155: This is what we need indeed! But I would go a step further, and make it so that the TS code below is not needed anymore.
Using Components.js annotations magic, we should be able to define these extensions via JSDoc annotations, similar to [here](https://github.com/comunica/comunica/blob/master/packages/actor-init-query/lib/ActorInitQueryBase.ts#L97-L128).

rubensworks (changes requested):
Nice work! We're on the right track, but I think we need one more thing for better developer experience.

rubensworks:
Opened an issue for the test failure, since I saw it popping up in another unrelated PR: https://github.com/comunica/comunica/issues/1106

packages/actor-init-query/lib/ActorInitQueryBase.ts
Laurin-W: I'm not quite sure how to achieve this with the annotation magic referenced here (and couldn't find documentation).

Since the goal after all would be to extend the annotated default context (in the current case with `contextKeyShortcutsExtensions`), I am not sure how this would work for chained inherits (without overriding `contextKeyShortcutsExtensions` which would make the `Record<string, string>[]` array useless).

Am I getting that wrong or missing something?

rubensworks: You're right, this isn't possible yet using Components.js annotations.
We'd have to extend the Components.js generator to make this array append logic possible.
Instead of doing that, it would be better to just extend Components.js to extend JSON objects.

But in any case, that's beyond the scope of this PR, so let's leave that for future work. So you can ignore my comment here :-)

.github/workflows/ci.yml
```diff
@@ -9,8 +9,6 @@ jobs:
         os: [ubuntu-latest]
         node-version:
           - 14.x
-          - 16.x
```
rubensworks, line 12: Change not needed?

packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -48,6 +49,16 @@ const argsDefault = {
 };
 
 describe('HttpServiceSparqlEndpoint', () => {
+  let originalCluster: typeof cluster;
```
rubensworks, line 52: Is this a fix for #1106? If so, could we move this to a separate PR?

.github/workflows/ci.yml
Laurin-W: Hey Ruben, that was just for testing purposes since I wasn't able to reproduce the issues locally (and didn't want the succeeding tests to be re-run to save some resources).

rubensworks: Ah ok. Just ping me when this is ready for review :)

packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
Laurin-W: Yep, sorry! Was experimenting with that last night since I needed a test execution environment.

packages/actor-init-query/lib/ActorInitQueryBase.ts
Laurin-W: addressed with https://github.com/comunica/comunica/pull/1102/commits/4a7853fb9bd44530c5bcabb271a85f8086cc7f89

packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
Laurin-W: Will push reversion once the fix (https://github.com/comunica/comunica/pull/1109) is merged.
Or would you recommend to push it immediately?

rubensworks: It's ok, we can wait :)

Laurin-W:
Hey @rubensworks! This should be good to go then :)

rubensworks:
Good stuff, thanks @Laurin-W!

PR #1118, opened 2022-11-16
corrected bindingsStreamToGraphQl return type

### LINKED ISSUE #1114: return type of bindingsStreamToGraphQl is wrong
#### Issue type:

- :bug: Bug 

____
#### Description:
The return type is set to Promise<string>, but it should be Promise<any>.

____
#### Environment:
| software         | version
| ---------------- | -------
| Comunica Engine  | 2.4.3 
| node             | v16.18.1
| npm              | 8.19.2
| yarn             | 1.22.19
| Operating System | linux (Linux 5.15.0-52-generic)

#### Observation:

This line in a Typescript application:
```
const data = (await bindingsStreamToGraphQl((await result.execute()) as BindingsStream, result.context, {materializeRdfJsTerms: true})) as object[];
```
results in this Typscript error:
```
error TS2352: Conversion of type 'string' to type 'object[]' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```

#### Temporary workaround:
As a workaround, the `as unknown` in between helps:
```
const data = (await bindingsStreamToGraphQl((await result.execute()) as BindingsStream, result.context, {materializeRdfJsTerms: true})) as unknown as object[];
```

mvanbrab:
This fixes #1114.

### DIFF
--- a/packages/actor-query-result-serialize-tree/lib/ActorQueryResultSerializeTree.ts
+++ b/packages/actor-query-result-serialize-tree/lib/ActorQueryResultSerializeTree.ts
@@ -33,7 +33,7 @@ export class ActorQueryResultSerializeTree extends ActorQueryResultSerializeFixe
    */
   public static bindingsStreamToGraphQl(bindingsStream: BindingsStream,
     context: IActionContext | Record<string, any> | undefined,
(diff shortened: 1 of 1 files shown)


rubensworks:
Thanks!

PR #1120, opened 2022-11-18
fix: fix broken bindingsStreamToGraphQL export

### LINKED ISSUE #1119: bindingsStreamToGraphQl export is broken
#### Issue type:

- :bug: Bug 

____
#### Description:

Since Comunica 2.5.0, the `bindingsStreamToGraphQl`, which seems to be caused by the change in `actor-query-result-serialize-tree/lib/index.ts` that were done in https://github.com/comunica/comunica/commit/ade1186d091072c074e02d98172753d00a783e5f.

@jeswr I missed this change myself when reviewing the PR, but AFAICS, this change is in fact incorrect.
Was there a particular reason why this change was introduced?
If not, could you submit a PR to revert this change?

jeswr:
Fixes #1119


rubensworks:
Thanks for the quick fix :-)

PR #1125, opened 2022-11-26
Add relative file path support to engine command line arguments

### LINKED ISSUE #1044: Invalid baseIRI when querying a local JSON-LD file
#### Issue type:

- :bug: Bug 

____
#### Description:

When I query a local JSON-LD file with this command

```shell
comunica-sparql-file "/Users/pheyvaer/Development/test.jsonld" -q "SELECT * WHERE {?s ?p ?o}"
````

 I get the following error 

```
Found invalid baseIRI '/Users/pheyvaer/Development/test.jsonld' for value '/Users/pheyvaer/Development/test.jsonld'
```
The error doesn't occur when adding `file://` to the path:

```shell
comunica-sparql-file "file:///Users/pheyvaer/Development/test.jsonld" -q "SELECT * WHERE {?s ?p ?o}"
````
____
#### Environment:
```shell
$ comunica-sparql-file -v
```
| software         | version
| ---------------- | -------
| Comunica Engine  | 2.3.0 
| node             | v14.19.2
| npm              | 8.10.0
| yarn             | 1.22.5
| Operating System | darwin (Darwin 21.5.0)

#### Crash log:

-----

## Bounty

| |
|-|
| €136 |

andrewzhurov:
Closes #1044


packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts
```diff
@@ -145,6 +147,45 @@ export class CliArgsHandlerBase implements ICliArgsHandler {
...
+
+        // Windows drive letter must be prefixed with a slash
+        if (!pathName.startsWith('/')) {
+            pathName = `/${pathName}`;
+        }
+
+        return encodeURI(`file://${pathName}`);
+    }
+
+    return isURI(source) ? source : fileUri(source);
```
rubensworks, line 186: I think the logic above can be simplified quite a lot.
I guess we should only be checking for the presence of a scheme (`http:`, `https://`, ...), and if it is not present, fallback to the `file://` scheme.

packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts
```diff
@@ -185,7 +226,8 @@ export class CliArgsHandlerBase implements ICliArgsHandler {
     if (args.sources.length > 0) {
       context.sources = context.sources || [];
       args.sources.forEach((sourceValue: string) => {
-        const source = CliArgsHandlerBase.getSourceObjectFromString(sourceValue);
+        let sourceValueURI: string = CliArgsHandlerBase.URIfy(sourceValue);
+        const source = CliArgsHandlerBase.getSourceObjectFromString(sourceValueURI);
```
rubensworks, line 230: The problem with adding this logic to the CLI handler, is that this will not work when the engine is invoked directly via the JS API.

An alternative place to add this would be in https://github.com/comunica/comunica/blob/master/packages/actor-dereference-file/lib/ActorDereferenceFile.ts#L36

packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts
andrewzhurov: Yeah, that would be great.
I've made a wrong assumption that values such as `C:` are valid URIs.

Looks like all we need is check whether it's a (tagged) URI, and if not - fallback to deriving a file URI for it, as you suggest. Thanks for the tip.

packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts
andrewzhurov: Ah, that's a good point. My bad going head-first without getting myself familiar with architecture.
Great to see that the same actor is responsible for file dereferencing with no regard where it happens.
Thanks for the pointer.

andrewzhurov:
Ah, I have accidentely force-pushed this branch to the same commit as in `master`, and github considered it as a clue that this PR has been merged and closed it. ~_~
Created a new one #1135

PR #1127, opened 2022-12-09
Feat/shaclc support

jeswr:


packages/context-entries/lib/Keys.ts
```diff
@@ -208,6 +208,13 @@ export const KeysQueryOperation = {
   localizeBlankNodes: new ActionContextKey<boolean>('@comunica/actor-query-operation:localizeBlankNodes'),
 };
 
+export const KeysRdfParse = {
+  /**
+   * @range {boolean}
+   */
+  parseNonRecommendedFormats: new ActionContextKey<boolean>('@comunica/actor-rdf-parse:parseNonRecommendedFormats'),
```
rubensworks, line 215: Instead of a new context option, should we consider exposing this parser in an alternative config of the query engine?
Since the vast majority of users will not enable this option, it might be unnecessary for them to have this parser dependency installed.

jeswr:
Happy to put it into a separate config; though I think the context entry could still be useful (e.g. If I want to use that custom config to retrieve only recommended formats from a Solid Pod, but then also optionally accept SHACLC from a shape repo).

Do you agree or should I remove the context entry entirely?

rubensworks:
> though I think the context entry could still be useful (e.g. If I want to use that custom config to retrieve only recommended formats from a Solid Pod, but then also optionally accept SHACLC from a shape repo)

Would there really be cases where you do want parsing to succeed, even if you have a parser available for it?

jeswr:
> Would there really be cases where you do want parsing to succeed, even if you have a parser available for it?

The main reason I want to add these parsers (and possibly others down the line) at this point is so that [rdf-parse](https://github.com/rubensworks/rdf-parse.js), [rdf-dereference](https://github.com/rubensworks/rdf-dereference.js) and [rdf-serialize](https://github.com/rubensworks/rdf-serialize.js) can opt-in to support non-recommended syntax's; which then allows SHACLC support to be easily integrated into work like https://github.com/CommunitySolidServer/CommunitySolidServer/pull/1000#issuecomment-1321710099, and also allows us to easily build a Shape Repo that includes support for content negotiation with SHACLC.

In each of those instances I think it is valuable to be able to toggle between recommended only, and all syntaxes.

jeswr:
@rubensworks so you are aware - there was a timeout on this TPF test which I have re-run https://github.com/comunica/comunica/actions/runs/3673335414/jobs/6210325676. It does not seem to be related to any changes made in this PR.

rubensworks:
> In each of those instances I think it is valuable to be able to toggle between recommended only, and all syntaxes.

I think it's fine to just include the parser, without the flag.
If we really want to provide only-recommended parsers, then we could ship another config in rdf-parse (and related), possibly in a different package.

PR #1129, opened 2022-12-21
Feat/shaclc support no context entry

jeswr:
Supersedes #1127


rubensworks:
There may have been a misunderstanding.
I [meant that](https://github.com/comunica/comunica/pull/1127#issuecomment-1348542953) we can just include the parser and serializer in the default config.
This is because we're already shipping with non-recommended parsers (see the microdata parser).
Since the new parser and serializer here are quite small, it doesn't hurt to add them to the default config.
We therefore also don't need a new engine config+package.
However, we could create `query-sparql-strict` if we really wanted to, but I don't think we need it at this stage (as it would also require the microdata parser to be re-organized).

This should simplify this PR quite a bit :-)

packages/actor-rdf-parse-shaclc/README.md
```diff
@@ -0,0 +1,35 @@
+# Comunica SHACL Compact Syntax RDF Parse Actor
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-rdf-parse-shaclc.svg)](https://www.npmjs.com/package/@comunica/actor-rdf-parse-shaclc)
+
+A comunica SHACL Compact Syntax RDF Parse Actor.
```
rubensworks, line 5: Can we add a link to the spec?

packages/actor-rdf-serialize-shaclc/README.md
```diff
@@ -0,0 +1,35 @@
+# Comunica SHACL Compact Syntax RDF Serialize Actor
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-rdf-serialize-shaclc.svg)](https://www.npmjs.com/package/@comunica/actor-rdf-serialize-shaclc)
+
+A comunica SHACL Compact Syntax RDF Serialize Actor.
```
rubensworks, line 5: Can we add a link to the spec?

packages/actor-rdf-parse-shaclc/lib/ActorRdfParseShaclc.ts
```diff
@@ -0,0 +1,45 @@
...
+    super(args);
+  }
+
+  public async runHandle(action: IActionRdfParse, mediaType: string, context: IActionContext):
+  Promise<IActorRdfParseOutput> {
+    return { data: <any> wrap(toString(action.data).then(str => parse(str))), metadata: { triples: true }};
+  }
+}
+
+function toString(stream: NodeJS.ReadableStream): Promise<string> {
```
rubensworks, line 36: Can we use `stream-to-string`? Other packages already depend on it.

packages/actor-rdf-serialize-shaclc/lib/ActorRdfSerializeShaclc.ts
```diff
@@ -0,0 +1,52 @@
...
+   *       "text/shaclc": "http://www.w3.org/ns/formats/Shaclc"
+   *     }} mediaTypeFormats
+   */
+  public constructor(args: IActorRdfSerializeFixedMediaTypesArgs) {
+    super(args);
+  }
+
+  public async runHandle(action: IActionRdfSerialize, mediaType: string, context: IActionContext):
+  Promise<IActorRdfSerializeOutput> {
+    // TODO: Get prefixes (from somewhere)
```
rubensworks, line 31: Could be done using the context.

But may not be strictly needed at this stage, since we also don't do this (yet) in the JSON-LD and N3 serializers.

packages/actor-rdf-serialize-shaclc/lib/ActorRdfSerializeShaclc.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    const data = new Readable();
+    data._read = () => {
+      // Do nothing
+    };
+
+    try {
+      const { text } = await wrap(action.quadStream).toArray().then(quads => write(quads, { errorOnUnused: true }));
```
rubensworks, line 39: Instead of using asynciterator as dependency here, I would suggest using `arrayify-stream` instead.

packages/actor-rdf-serialize-shaclc/lib/ActorRdfSerializeShaclc.ts
```diff
@@ -0,0 +1,52 @@
...
+      data.push(text);
+    } catch (error: unknown) {
+      // Push the error into the stream
+      data.push(error);
```
rubensworks, line 43:
```suggestion
      data.emit('error', error);
```

packages/actor-rdf-parse-shaclc/lib/ActorRdfParseShaclc.ts
```diff
@@ -0,0 +1,45 @@
...
+   *       "text/shaclc": "http://www.w3.org/ns/formats/Shaclc"
+   *     }} mediaTypeFormats
+   */
+  public constructor(args: IActorRdfParseFixedMediaTypesArgs) {
+    super(args);
+  }
+
+  public async runHandle(action: IActionRdfParse, mediaType: string, context: IActionContext):
+  Promise<IActorRdfParseOutput> {
+    return { data: <any> wrap(toString(action.data).then(str => parse(str))), metadata: { triples: true }};
```
rubensworks, line 32: I think we don't strictly require `asynciterator` here, we can probably do this using regular Node streams.

packages/actor-rdf-serialize-shaclc/lib/ActorRdfSerializeShaclc.ts
rubensworks: We'll probably also want to wrap this inside `data._read` (once), because otherwise the error event could be emitted before listeners are attached.

rubensworks (changes requested):

jeswr:
> There may have been a misunderstanding. I [meant that](https://github.com/comunica/comunica/pull/1127#issuecomment-1348542953) we can just include the parser and serializer in the default config. This is because we're already shipping with non-recommended parsers (see the microdata parser). Since the new parser and serializer here are quite small, it doesn't hurt to add them to the default config. We therefore also don't need a new engine config+package. However, we could create `query-sparql-strict` if we really wanted to, but I don't think we need it at this stage (as it would also require the microdata parser to be re-organized).
> 
> This should simplify this PR quite a bit :-)

Ah ok great; will do that @rubensworks . Can I also add the changes in https://github.com/comunica/comunica/pull/1133 to this PR or would you like to discuss and merge those changes separately.

rubensworks:
> Can I also add the changes in https://github.com/comunica/comunica/pull/1133 to this PR or would you like to discuss and merge those changes separately.

Sure, adding those to this PR sounds good to me!

I see that a (transitive) dependency on `@jeswr/prefixcc` was added in that PR (which is fine). Could its fetch calls be delegated to the HTTP bus?

packages/actor-rdf-parse-shaclc/lib/ActorRdfParseShaclc.ts
jeswr: I'm not immediately sure how to achieve this with the current logic to emit prefixes in there as well; is having `asynciterator` here problematic in any way?

jeswr:
> I see that a (transitive) dependency on @jeswr/prefixcc was added in that PR (which is fine). Could its fetch calls be delegated to the HTTP bus?

The package never be invoked as it requires the `mintNewPrefixes` option to be enabled in the shaclc writer. This is something to be mindful of if we do choose to enable that parameter.

packages/actor-rdf-parse-shaclc/lib/ActorRdfParseShaclc.ts
rubensworks: > I'm not immediately sure how to achieve this with the current logic to emit prefixes in there as well

Not sure I understand. Is asynciterator doing anything with prefixes here?

> is having asynciterator here problematic in any way?

I just think that the Node-way would be a simpler way to solve the problem, as we don't need the additional logic complexity of asynciterator.

rubensworks: Ok, let's keep asynciterator here, will indeed be simpler :-)

rubensworks (changes requested):

rubensworks:
Thanks!

rubensworks:
I won't make a new release of this just yet, but if you need one urgently, please let me know.

PR #1135, opened 2023-01-09
Derive file url out of successfully resolved file path

### LINKED ISSUE #1044: Invalid baseIRI when querying a local JSON-LD file

andrewzhurov:
Closes #1044

### DIFF
--- a/packages/actor-dereference-file/lib/ActorDereferenceFile.ts
+++ b/packages/actor-dereference-file/lib/ActorDereferenceFile.ts
@@ -1,5 +1,5 @@
 import { accessSync, createReadStream, constants } from 'fs';
-import { fileURLToPath } from 'url';
+import { fileURLToPath, pathToFileURL } from 'url';
 import type { IActionDereference, IActorDereferenceArgs, IActorDereferenceOutput } from '@comunica/bus-dereference';
 import { ActorDereference } from '@comunica/bus-dereference';
 import type { IActorTest } from '@comunica/core';
@@ -21,6 +21,11 @@ export class ActorDereferenceFile extends ActorDereference {
     return true;
   }
 
+  private static isURI(str: string): boolean {
+    const URIRegex = /\w[\w+.-]*:.*/u;
+    return URIRegex.exec(str) !== null;
+  }
+
   public async run({ url }: IActionDereference): Promise<IActorDereferenceOutput> {
     const requestTimeStart = Date.now();
     return {
@@ -28,7 +33,7 @@ export class ActorDereferenceFile extends ActorDereference {
       // This should always be after the creation of the read stream
       requestTime: Date.now() - requestTimeStart,
       exists: true,
-      url,
+      url: ActorDereferenceFile.isURI(url) ? url : pathToFileURL(url).href,
     };
   }
 }
--- a/packages/actor-dereference-file/test/ActorDereferenceFile-test.ts
+++ b/packages/actor-dereference-file/test/ActorDereferenceFile-test.ts
@@ -1,5 +1,6 @@
 import * as fs from 'fs';
 import * as path from 'path';
+import { pathToFileURL } from 'url';
 import type { IActorDereferenceOutput } from '@comunica/bus-dereference';
 import { ActorDereference } from '@comunica/bus-dereference';
 import { ActionContext, Bus } from '@comunica/core';
@@ -69,7 +70,22 @@ describe('ActorDereferenceFile', () => {
         {
           data: expect.anything(),
           exists: true,
-          url: p,
+          url: `file://${p}`,
+        },
+      );
+    });
+
+    it('should run for relative paths', async() => {
+      // Relative paths are considered to start from the directory that `node` is ran in
+      // for tests it will be the root of the repo
+      const p = path.join(path.relative(process.cwd(), __dirname), 'dummy.ttl');
+      const result = await actor.run({ url: p, context });
+      expect(await streamToString(result.data)).toEqual(fs.readFileSync(p).toString());
+      expect(result).toMatchObject<Partial<IActorDereferenceOutput>>(
+        {
+          data: expect.anything(),
+          exists: true,
+          url: pathToFileURL(p).href,
         },
       );
     });
--- a/engines/query-sparql-file/test/QuerySparqlFile-test.ts
+++ b/engines/query-sparql-file/test/QuerySparqlFile-test.ts
@@ -1,0 +1,32 @@
+/** @jest-environment setup-polly-jest/jest-environment-node */
+
+import * as path from 'path';
+import type { QueryStringContext } from '@comunica/types';
+import 'jest-rdf';
+import arrayifyStream from 'arrayify-stream';
+import { QueryEngine } from '../lib/QueryEngine';
+
+const stringifyStream = require('stream-to-string');
+
+describe('System test: QuerySparqlFile', () => {
+  let engine: QueryEngine;
+  let spyQueryOrExplain: any;
+
+  beforeEach(() => {
+    engine = new QueryEngine();
+  });
+
+  describe('query', () => {
+    describe('simple SPO on a raw RDF document', () => {
+      const p = path.join(path.relative(process.cwd(), __dirname), 'assets/dummy.jsonld');
+      const query = `SELECT * WHERE { ?s ?p ?o. }`;
+
+      it('should return a result for a jsonld file by a relative file path', async() => {
+        const context: QueryStringContext = { sources: [{ value: p }]};
+
+        const result = await arrayifyStream(await engine.queryBindings(query, context));
+        expect(result.length).toBe(2);
+      });
+    });
+  });
+});
--- a/engines/query-sparql-file/test/assets/dummy.jsonld
+++ b/engines/query-sparql-file/test/assets/dummy.jsonld
@@ -1,0 +1,5 @@
+{
+  "@context": "https://schema.org",
+  "predicate1": "value 1",
+  "predicate2": "value 2"
+}


packages/actor-dereference-file/lib/ActorDereferenceFile.ts
```diff
@@ -22,13 +22,17 @@ export class ActorDereferenceFile extends ActorDereference {
   }
 
   public async run({ url }: IActionDereference): Promise<IActorDereferenceOutput> {
+    // E.g., sparql@http://example.org/
```
rubensworks, line 25: If I remember correctly, these tagged URLs will never occur at this stage, they should already be stripped away at an earlier stage.

So I think we only have to check for the presence of a scheme, so just checking for `.indexOf(':') > 0` should be sufficient.

packages/actor-dereference-file/lib/ActorDereferenceFile.ts
```diff
@@ -22,13 +22,17 @@ export class ActorDereferenceFile extends ActorDereference {
...
+    const isURI = (str: string): boolean => OptionallyTaggedURIRegex.exec(str) !== null;
+
     const requestTimeStart = Date.now();
     return {
       data: createReadStream(getPath(url)),
       // This should always be after the creation of the read stream
       requestTime: Date.now() - requestTimeStart,
       exists: true,
-      url,
+      url: isURI(url) ? url : pathToFileURL(url).href,
```
rubensworks, line 35: Aha, I didn't know about this function!
I'm wondering now if all that we need is not just `pathToFileURL`?
We may not even need the `isURI`, since `pathToFileURL` may not do anything if the `file:` scheme is already present.

rubensworks (changes requested):
Really elegant solution!

Can you also add a system test in the `query-sparql` engine to make sure that this solves the original issue?

packages/actor-dereference-file/lib/ActorDereferenceFile.ts
andrewzhurov: I had hopes it works this way too! To my surprise it cares little whether it's already a URI or not.. seems it considers a file URI to be a file name, prepending it with `file://absolute-path/`, so we something like `file://absolute-path/file://absolute-path`
If this function becomes of frequent use we could create our own helper fn that works the way we want

andrewzhurov:
> Really elegant solution!

Thank you for your kind words.:)
I love when there turns out to be a simple solution like that, finding it is often challenging, thanks for guiding in its discovery!

> Can you also add a system test in the `query-sparql` engine to make sure that this solves the original issue?

That's a good idea, coming next

packages/actor-dereference-file/lib/ActorDereferenceFile.ts
andrewzhurov: Good point, indeed tags are stripped out at this point.
As much as I'd like to not have a regex in favor of something simpler, it seems if want to be sure that we don't have false positive matches, we better make it as strict to a URI shape check as we can..
E.g., `C:\file.ttl` would pass if we are to check for `:` presence solely
Perhaps we can't escape having a regex here..

andrewzhurov:
Added test that calling comunica-sparql-file with a path to a local .jsonld file runs fine

packages/actor-dereference-file/lib/ActorDereferenceFile.ts
```diff
@@ -22,13 +22,16 @@ export class ActorDereferenceFile extends ActorDereference {
   }
 
   public async run({ url }: IActionDereference): Promise<IActorDereferenceOutput> {
+    const URIRegex = /\w[\w+.-]*:.*/u;
```
rubensworks, line 25: Why are these characters needed in the regex?: `+.-`
I would guess they can be removed?

rubensworks: Also, I suggest moving this regex as constant readonly field on the class, so that we only have to create it at load-time.

packages/actor-dereference-file/test/ActorDereferenceFile-test.ts
```diff
@@ -69,7 +70,22 @@ describe('ActorDereferenceFile', () => {
...
           data: expect.anything(),
           exists: true,
-          url: p,
+          url: `file://${p}`,
+        },
+      );
+    });
+
+    it('should run for relative paths', async() => {
+      // Relative paths are considered to start from the directory that `node` is ran in
```
rubensworks, line 79: We can not always guarantee this. I recommend using `__dirname` instead, and navigating from there.

rubensworks (changes requested):

packages/actor-dereference-file/lib/ActorDereferenceFile.ts
andrewzhurov: > Why are these characters needed in the regex?: +.-

I haven't seen any  URI schemes with `+.-` in them, so it's been to my surprise to learn that they make for a valid scheme, according to [this wiki page](https://en.wikipedia.org/wiki/URL)
image

Do you think we're better off dropping them?

packages/actor-dereference-file/lib/ActorDereferenceFile.ts
```diff
@@ -22,13 +22,16 @@ export class ActorDereferenceFile extends ActorDereference {
...
+    const isURI = (str: string): boolean => URIRegex.exec(str) !== null;
```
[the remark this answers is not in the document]
andrewzhurov: That's a fair point that this function does not give us value by reducing code repetition.
Another value aspect of it is in abstracting away the details. There are not much of details to abstract away, of course, but personally it still feels a bit better to have it, spares a bit of mental effort that's needed to parse `URIRegex.exec(str) !== null` as `isURI(str)`
Up to have it inline if you think it's a better way, you've got a perfect grasp on the codebase style and we'd like to keep it uniform. Please let me know if needed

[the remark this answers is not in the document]
andrewzhurov: > We can probably simplify it then via url.match(regex).

I'd have a slight preference for that notation as well, but for some reason linter been complaining about it and autofixed it this

packages/actor-dereference-file/lib/ActorDereferenceFile.ts
rubensworks: I see, good catch! Let's keep them there then :-)

packages/actor-dereference-file/lib/ActorDereferenceFile.ts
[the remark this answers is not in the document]
rubensworks: Ok, the linter probably has a good reason for it then, so let's follow it's advice.

packages/actor-dereference-file/test/ActorDereferenceFile-test.ts
andrewzhurov: I'm not sure I get it..
Thing is that current implementation (`pathToFileURL`) resolves relative paths from the directory that node is ran in,
this UX is according to user's expectations,
say they run
```
/current/dir/$ ./bin/query.js ./file.jsonld ...
```
Then they would expect `file.jsonld` to be resolved to `/current/dir/file.jsonld`

Same happens when we run tests from `/path/to/comunica`, relative paths get resolved relative to that dir, and not a dir that a currently executed file is in (in our case, `/path/to/comunica/packages/actor-dereference-file/test/`), so, in order to that we handle relative file paths corectly, we're in need to pass in a relative path that starts from `/path/to/comunica`, which's achieved by a bit tricky to grasp following code line

My first intent been to use `__dirname` in tests to point to the asset file
I expected `path.join(__dirname, "./assets/dummy.jsonld")` to resolve to
`file:///path/to/comunica/packages/actor-dereference-file/test/assets/dummy.jsonld`
whereas, to my surprise, it's been getting resolved to
`file:///path/to/comunica/assets/dummy.jsonld`
That led me to realization, that in order to test current implementation correctly we'd need to craft relative paths from the root of the repo, where tests were launched from

Please let me know if understood your concerns right about this trickery or there is something else you meant to point onto

packages/actor-dereference-file/lib/ActorDereferenceFile.ts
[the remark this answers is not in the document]
andrewzhurov: I thought maybe [by making `isURI` a private static method](https://github.com/comunica/comunica/pull/1135/files#diff-84203b62a6d317be93993706ee5a5f3f7169a72d2a8c49a1e9dd9f87e43a762fR24) we'd have both the performance gain from not having to create regex and fn each time `run` is executed and a slight readability gain from having `isURI` as a name

packages/actor-dereference-file/test/ActorDereferenceFile-test.ts
rubensworks: Ok, that makes sense.
Could you just double-check that the test also still passes if it is run from another directory? (I suspect it will)

If so, ping me if this PR is ready for another review.

andrewzhurov: Sure thing!
Just ran a test - test passes no prob.


Tested that it works by running
```
/path/to/comunica/packages/actor-dereference-file/test/$ npx jest@27.5.1 -c ../../jest.config.js
```

PR seems ready for having another look

rubensworks (approved):
Looks perfect, great work @andrewzhurov!

PR #1142, opened 2023-01-18
Fix errors related to "[assignement of] read only property" and unit test when using node v19

### LINKED ISSUE #1141: ActorHttp fail with node.js v19
#### Issue type:

- :bug: Bug 

____
#### Description:

When using node.js `v19.3.0` the `@comunica/bus-http/lib/ActorHttp.js` fail with the error 

```js
TypeError: Cannot assign to read only property 'ReadableStream' of object '[object global]'
at Object.<anonymous> (node_modules/@comunica/bus-http/lib/ActorHttp.js:7:23)
at Object.<anonymous> (node_modules/@comunica/bus-http/lib/index.js:17:14)
```
looking at the code we can see the comment 
 [`// TODO: Remove when targeting NodeJS 18+`](https://github.com/comunica/comunica/blob/e3b938940fa30b10bf2aa68a5f6ac3086399873a/packages/bus-http/lib/ActorHttp.ts#L5) 
(it has to be noted that this is the equivalent line of the `ts` file hence the incongruence with the line of the error that is related to the `js` compiled file)

it might be related to the error
____
#### Environment:

- comunica:  [Latest commit on the publication of this issue](https://github.com/comunica/comunica-feature-link-traversal/tree/4ce3abef94f8c7d4bd38a2e6b27d1a9bf34b81da)
- node.js: v19.3.0
- OS: Manjaro Linux Sikaris -- 22.0.0

#### Crash log:
```js
TypeError: Cannot assign to read only property 'ReadableStream' of object '[object global]'
at Object.<anonymous> (node_modules/@comunica/bus-http/lib/ActorHttp.js:7:23)
at Object.<anonymous> (node_modules/@comunica/bus-http/lib/index.js:17:14)
```

constraintAutomaton:
- global.ReadableStream is only modify when it is undefined
- Unit test related to solid-client-authn are ignore as the package doesn't support it
- Unit test of error modified to take into account the new error string of node v19
    - Before v19: `Invalid POST body with context received ('{"a:"b"}'): Expected ':' after property name in JSON at position 5`
    - v19: `Invalid POST body with context received ('{"a:"b"}'): Unexpected token b in JSON at position 5`

close #1141

### DIFF
--- a/packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
+++ b/packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
@@ -1724,7 +1724,7 @@ describe('HttpServiceSparqlEndpoint', () => {
         httpRequestMock.headers = { 'content-type': 'application/x-www-form-urlencoded' };
 
         return expect(instance.parseBody(httpRequestMock)).rejects
-          .toThrowError(/(Invalid POST body with context received \('\{"a:"b"\}'\):) .+JSON (at position 5)/u);
+          .toThrowError('Invalid POST body with context received');
       });
 
       it('should reject if content-type is not application/[sparql-query|x-www-form-urlencoded]', () => {
--- a/packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
+++ b/packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
@@ -1724,7 +1724,7 @@ describe('HttpServiceSparqlEndpoint', () => {
         httpRequestMock.headers = { 'content-type': 'application/x-www-form-urlencoded' };
 
         return expect(instance.parseBody(httpRequestMock)).rejects
-          .toThrowError(`Invalid POST body with context received ('{"a:"b"}'): Unexpected token b in JSON at position 5`);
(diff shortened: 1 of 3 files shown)


packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -1724,7 +1724,7 @@ describe('HttpServiceSparqlEndpoint', () => {
         httpRequestMock.headers = { 'content-type': 'application/x-www-form-urlencoded' };
 
         return expect(instance.parseBody(httpRequestMock)).rejects
-          .toThrowError(`Invalid POST body with context received ('{"a:"b"}'): Unexpected token b in JSON at position 5`);
+          .toThrowError(/(Invalid POST body with context received \('\{"a:"b"\}'\):) .+JSON (at position 5)/u);
```
rubensworks, line 1727:
```suggestion
          .toThrowError('Invalid POST body with context received');
```

I applaud your effort in finding this regex, but I think we can just simplify it like the above, no need to check for the literal JSON error message :-)

rubensworks (changes requested):
Thanks for this @constraintAutomaton! Just one minor nit.

rubensworks (approved):
Super, great work @constraintAutomaton!

Issue #1143, opened 2023-01-24
No results produced when querying over a string source with blank nodes using multiple patterns


surilindur:
#### Issue type:

- :bug: Bug 

____
#### Description:

Querying a string source using the [string source actor](https://github.com/comunica/comunica/tree/master/packages/actor-rdf-resolve-quad-pattern-hypermedia) with blank node subjects using multiple triple patterns produces no results.

Example data and query to reproduce:

```
@prefix schema: <http://schema.org/>.

_:t0 a schema:Document.
_:t0 a schema:DigitalDocument.
_:t0 schema:title "Example".
```

```
PREFIX schema: <http://schema.org/>

SELECT * WHERE {
  ?document a schema:Document, schema:DigitalDocument ;
    schema:title ?title .
}
```

There are no results produced as it is. However:
* Converting `_:t0` to `<#doc>` in the data does produce results, indicating that the issue is with blank nodes.
* Restricting the query to just one pattern does produce results, indicating that the issue is with multiple patterns.

Following a short conversation, it could be that the string source is parsed separately for each pattern, causing the blank nodes to be assigned different identifiers for every one of them. Maybe the issue will then also occur for blank nodes elsewhere and not just subjects.

It was suggested that some caching be added to the string source actor to ensure blank node identifiers remain the same for a given string source for each pattern that is being resolved.

Reference line for caching investigation [here](https://github.com/comunica/comunica/blob/cca85ecec92a3cdbdf478d86f3195d96b8f03b2a/packages/actor-rdf-resolve-quad-pattern-string-source/lib/ActorRdfResolveQuadPatternStringSource.ts#L64), with the parsed RDF source.

____
#### Environment:

| software         | version
| ---------------- | -------
| Comunica Engine  | 2.5.1
| node             | v19.4.0

surilindur:
This should be fixed now with #1144.

PR #1144, opened 2023-01-26
Add source cache to string source actor

surilindur:
The changes here were inspired by #1143, where the string source actor was parsing the source separately for each pattern, which in turn made the blank node identifiers (after parsing) different for each pattern. Queries with multiple pattern would therefore not produce any results.

The following changes were made to `ActorRdfResolveQuadPatternStringSource`:
* Added a cache with the data source (object) as key and the (promise of) parsed RDF data as value, so any subsequent requests to resolve a pattern based on a string source will use the same parse result for the same source object.
* Added `cacheSize` parameter to configure the cache size, with default 100. When set to 0, the actor will not cache the results and the bug will reappear, but it might be interesting for someone to try it at some point, so I left it like that.
* Added `implements` to the actor definition for its own interface and adjusted some property definitions to match.

The following changes were made to the tests:
* Added a loop to the `run` function test, that attempts to make sure the actor produces expected results when it is run several (default two) times using the same string source object.
* Changed the mock RDF parse and pattern resolve mediators a little to account for that test case.

Also, I noticed that if the string source object is not the same, the previously cached parsing result will not be used. Even if the values would be identical, if the object is not the same, it will end up as a new entry in the cache.

Any feedback would be welcome. Thank you! :slightly_smiling_face:

### DIFF
--- a/packages/actor-rdf-resolve-quad-pattern-string-source/lib/ActorRdfResolveQuadPatternStringSource.ts
+++ b/packages/actor-rdf-resolve-quad-pattern-string-source/lib/ActorRdfResolveQuadPatternStringSource.ts
@@ -8,7 +8,7 @@ import { KeysRdfResolveQuadPattern } from '@comunica/context-entries';
 import type { IActorTest } from '@comunica/core';
 import type { IActionContext, IDataSource, IDataSourceSerialized } from '@comunica/types';
 import type * as RDF from '@rdfjs/types';
-import LRUCache = require('lru-cache');
+import * as LRUCache from 'lru-cache';
 import { storeStream } from 'rdf-store-stream';
 import { Readable } from 'readable-stream';
 
@@ -43,11 +43,17 @@ export class ActorRdfResolveQuadPatternStringSource extends ActorRdfResolveQuadP
 
   public run(action: IActionRdfResolveQuadPattern): Promise<IActorRdfResolveQuadPatternOutput> {
     const source = <IDataSourceSerialized>getContextSource(action.context)!;
-    const rdfSourcePromise: Promise<RDF.Source> = this.cache?.get(source) ?? this.getRdfSource(action.context, source);
-    if (this.cache && !this.cache.has(source)) {
-      this.cache.set(source, rdfSourcePromise);
+    // A source should only be parsed once (see getRdfSource comment), so if it has been parsed,
+    // that earlier result should be used. Note: if the object is identical value-wise, but is not
+    // the same exact object, it will get parsed again and saved in the cache as its own entry!
+    let rdfSourcePromise = this.cache?.get(source);
+    if (!rdfSourcePromise) {
+      rdfSourcePromise = this.getRdfSource(action.context, source);
+      if (this.cache) {
+        this.cache.set(source, rdfSourcePromise);
+      }
     }
-    return new Promise((resolve, reject) => rdfSourcePromise
+    return new Promise((resolve, reject) => rdfSourcePromise!
       .then(rdfSource => {
         const resolveQuadAction: IActionRdfResolveQuadPattern = {
           pattern: action.pattern,
@@ -60,6 +66,15 @@ export class ActorRdfResolveQuadPatternStringSource extends ActorRdfResolveQuadP
       }).catch(reject));
   }
 
+  /**
+   * Parses the string data source through the RDF parse bus, returning the RDF source.
+   * Parsing a source with blank nodes may produce different identifiers for the same nodes
+   * on different parses, for example if the source gets parsed separately for each pattern
+   * in a query. Consequently, a single source should only be parsed once, and the parse result cached.
+   * @param context The run action context
+   * @param source The source from the run action context
+   * @returns Parsed RDF source that can be passed to quad pattern resolve mediator as an rdfjsSource
+   */
   protected async getRdfSource(context: IActionContext, source: IDataSourceSerialized): Promise<RDF.Source> {
     const textStream = new Readable({ objectMode: true });
     /* istanbul ignore next */
@@ -80,9 +95,7 @@ export class ActorRdfResolveQuadPatternStringSource extends ActorRdfResolveQuadP
     };
 
     const parseResult = await this.mediatorRdfParse.mediate(parseAction);
-    const rdfStore = await storeStream(parseResult.handle.data);
-
-    return rdfStore;
+    return await storeStream(parseResult.handle.data);
   }
 
   private isStringSource(datasource: any): datasource is IDataSourceSerialized {
@@ -98,7 +111,7 @@ export class ActorRdfResolveQuadPatternStringSource extends ActorRdfResolveQuadP
 
 export interface IActorRdfResolveQuadPatternStringSourceArgs extends IActorRdfResolveQuadPatternArgs {
   /**
-   * The maximum number of entries in the LRU cache, set to 0 to disable.
+   * The maximum number of parsed stores in the LRU cache, set to 0 to disable.
    * @range {integer}
    * @default {100}
    */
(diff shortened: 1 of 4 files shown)


packages/actor-rdf-resolve-quad-pattern-string-source/lib/ActorRdfResolveQuadPatternStringSource.ts
```diff
@@ -1,29 +1,33 @@
...
 import type {
   IActionRdfResolveQuadPattern, IActorRdfResolveQuadPatternArgs, IActorRdfResolveQuadPatternOutput,
   MediatorRdfResolveQuadPattern,
 } from '@comunica/bus-rdf-resolve-quad-pattern';
 import { KeysRdfResolveQuadPattern } from '@comunica/context-entries';
 import type { IActorTest } from '@comunica/core';
-import type { IDataSourceSerialized } from '@comunica/types';
+import type { IActionContext, IDataSource, IDataSourceSerialized } from '@comunica/types';
 import type * as RDF from '@rdfjs/types';
+import LRUCache = require('lru-cache');
```
rubensworks, line 11: This should be an import.

packages/actor-rdf-resolve-quad-pattern-string-source/package.json
```diff
@@ -36,6 +36,9 @@
     "@comunica/bus-rdf-resolve-quad-pattern": "^2.6.0",
     "@comunica/context-entries": "^2.6.0",
     "@comunica/core": "^2.6.0",
+    "@comunica/types": "^2.6.0",
+    "@rdfjs/types": "^1.1.0",
```
rubensworks, line 40: The CI failure comes from here:

> Error: Resolution not using '*' import for @rdfjs/types in @comunica/actor-rdf-resolve-quad-pattern-string-source

packages/actor-rdf-resolve-quad-pattern-string-source/lib/ActorRdfResolveQuadPatternStringSource.ts
```diff
@@ -37,9 +41,26 @@ export class ActorRdfResolveQuadPatternStringSource extends ActorRdfResolveQuadP
     return true;
   }
 
-  public async run(action: IActionRdfResolveQuadPattern): Promise<IActorRdfResolveQuadPatternOutput> {
-    const source = <IDataSourceSerialized>getContextSource(action.context);
+  public run(action: IActionRdfResolveQuadPattern): Promise<IActorRdfResolveQuadPatternOutput> {
+    const source = <IDataSourceSerialized>getContextSource(action.context)!;
```
rubensworks, line 45: Could you also add a small jsdoc comment explaining why exactly we need to cache the promise?

packages/actor-rdf-resolve-quad-pattern-string-source/lib/ActorRdfResolveQuadPatternStringSource.ts
```diff
@@ -79,12 +96,17 @@ export class ActorRdfResolveQuadPatternStringSource extends ActorRdfResolveQuadP
   }
 }
 
-export interface IActorRdfResolveQuadPatternStringSource extends IActorRdfResolveQuadPatternArgs {
+export interface IActorRdfResolveQuadPatternStringSourceArgs extends IActorRdfResolveQuadPatternArgs {
+  /**
+   * The maximum number of entries in the LRU cache, set to 0 to disable.
```
rubensworks, line 101:
```suggestion
   * The maximum number of parsed stores in the LRU cache, set to 0 to disable.
```
(also in the README)

packages/actor-rdf-resolve-quad-pattern-string-source/lib/ActorRdfResolveQuadPatternStringSource.ts
```diff
@@ -48,24 +69,20 @@ export class ActorRdfResolveQuadPatternStringSource extends ActorRdfResolveQuadP
...
-      pattern: action.pattern,
-      context: action.context.set(KeysRdfResolveQuadPattern.source, {
-        value: <RDF.Source> await storeStream(parserResult.handle.data),
-        type: 'rdfjsSource',
-      }),
     };
-    return this.mediatorRdfResolveQuadPattern.mediate(resolveQuadAction);
+
+    const parseResult = await this.mediatorRdfParse.mediate(parseAction);
+    const rdfStore = await storeStream(parseResult.handle.data);
```
rubensworks, line 83: We can just return this immediately.

packages/actor-rdf-resolve-quad-pattern-string-source/lib/ActorRdfResolveQuadPatternStringSource.ts
```diff
@@ -37,9 +41,26 @@ export class ActorRdfResolveQuadPatternStringSource extends ActorRdfResolveQuadP
     return true;
   }
 
-  public async run(action: IActionRdfResolveQuadPattern): Promise<IActorRdfResolveQuadPatternOutput> {
-    const source = <IDataSourceSerialized>getContextSource(action.context);
+  public run(action: IActionRdfResolveQuadPattern): Promise<IActorRdfResolveQuadPatternOutput> {
+    const source = <IDataSourceSerialized>getContextSource(action.context)!;
+    const rdfSourcePromise: Promise<RDF.Source> = this.cache?.get(source) ?? this.getRdfSource(action.context, source);
+    if (this.cache && !this.cache.has(source)) {
```
rubensworks, line 47: This `this.cache.has()` call shouldn't be needed here.

Something like this should work:
```
let rdfSourcePromise: Promise<RDF.Source> = this.cache?.get(source);
if (!rdfSourcePromise) {
  // call getRdfSource and set in cache.
}
```

rubensworks (changes requested):
Thanks for this @surilindur!

packages/actor-rdf-resolve-quad-pattern-string-source/lib/ActorRdfResolveQuadPatternStringSource.ts
surilindur: I added a comment to tge getRdfSource function, as well as a short note just above the line where the first cache get is located. Do they look okay, or should they be moved somewhere or adjusted?

packages/actor-rdf-resolve-quad-pattern-string-source/lib/ActorRdfResolveQuadPatternStringSource.ts
surilindur: This should be fixed now.

packages/actor-rdf-resolve-quad-pattern-string-source/package.json
surilindur: This should be fixed now, too.

packages/actor-rdf-resolve-quad-pattern-string-source/lib/ActorRdfResolveQuadPatternStringSource.ts
surilindur: Changed this, both in code and readme.

packages/actor-rdf-resolve-quad-pattern-string-source/lib/ActorRdfResolveQuadPatternStringSource.ts
surilindur: Done!

packages/actor-rdf-resolve-quad-pattern-string-source/lib/ActorRdfResolveQuadPatternStringSource.ts
surilindur: This should be a bit better now after the change.

rubensworks (approved):
Excellent work @surilindur, thanks!

Do you need a new release for this?

surilindur:
Not necessarily, thank you, I can update when a new release is out but it should not be needed to make one just for this. :slightly_smiling_face:

PR #1145, opened 2023-01-31
fix: allow shaclc parser to pipe

jeswr:
This is needed to allow it to be used in packages like rdf-parse (see https://github.com/rubensworks/rdf-parse.js/pull/48/files)

### DIFF
--- a/packages/actor-rdf-parse-shaclc/lib/ActorRdfParseShaclc.ts
+++ b/packages/actor-rdf-parse-shaclc/lib/ActorRdfParseShaclc.ts
@@ -7,6 +7,7 @@ import {
   ActorRdfParseFixedMediaTypes,
 } from '@comunica/bus-rdf-parse';
 import type { IActionContext } from '@comunica/types';
+import { Readable } from 'readable-stream';
 import { parse } from 'shaclc-parse';
 import streamToString = require('stream-to-string');
 import { PrefixWrappingIterator } from './PrefixWrappingIterator';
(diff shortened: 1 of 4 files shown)


jeswr:
Note the following failing browser test https://github.com/comunica/comunica/actions/runs/4050746186/jobs/6968424798

packages/actor-rdf-parse-shaclc/lib/ActorRdfParseShaclc.ts
```diff
@@ -32,10 +33,15 @@ export class ActorRdfParseShaclc extends ActorRdfParseFixedMediaTypes {
 
   public async runHandle(action: IActionRdfParse, mediaType: string, context: IActionContext):
   Promise<IActorRdfParseOutput> {
+    const prefixIterator = new PrefixWrappingIterator(
+      streamToString(action.data).then(str => parse(str, { extendedSyntax: mediaType === 'text/shaclc-ext' })),
```
rubensworks, line 37: Not a requirement for this PR, but I just noticed that we're not `catch`-ing the rejection of this promise. So any error events in `action.data` will get lost here.
Could you look into adding a handler for this?

jeswr: @rubensworks - I think this is a non-issue since the [`PrefixWrappingIterator` extends the `WrappingIterator`](https://github.com/comunica/comunica/blob/2b6f38e3cf1446ea59a678ef327934cc74bfe310/packages/actor-rdf-parse-shaclc/lib/PrefixWrappingIterator.ts#L7) which in turn handles promises [by emitting an error event](https://github.com/RubenVerborgh/AsyncIterator/blob/3171521759194b3ce463f989d8253e35978ef690/asynciterator.ts#L2031).

rubensworks: Ok, good, seems like a non-issue indeed then.

rubensworks:
Released in 2.6.2.

PR #1147, opened 2023-02-03
fix: don't cache QPF default graph as variable

jeswr:
We observed where if we execute the query:

```
PREFIX ldp: <http://www.w3.org/ns/ldp#>
    SELECT * WHERE {
      GRAPH ?g {
        ?s ldp:contains ?o .
      }
    }
```

by itself over a QPF endpoint then the correct results are returned; but if we use the same engine to _first_ execute the query:

````
SELECT * WHERE { ?s <http://www.w3.org/ns/ldp#contains> ?o }
````

then no results are returned.

This is because variables, and the default graph have both been given and id of the empty string and so an empty stream of results was being cached when searching over the default graph with `SELECT * WHERE { ?s <http://www.w3.org/ns/ldp#contains> ?o }` and then this cached stream was incorrectly being re-used when executing the query

```
PREFIX ldp: <http://www.w3.org/ns/ldp#>
    SELECT * WHERE {
      GRAPH ?g {
        ?s ldp:contains ?o .
      }
    }
```

### DIFF
--- a/packages/actor-rdf-resolve-hypermedia-qpf/test/RdfSourceQpf-test.ts
+++ b/packages/actor-rdf-resolve-hypermedia-qpf/test/RdfSourceQpf-test.ts
@@ -4,6 +4,7 @@ import type { IActorDereferenceRdfOutput } from '@comunica/bus-dereference-rdf';
 import { ActionContext, Bus } from '@comunica/core';
 import type * as RDF from '@rdfjs/types';
 import arrayifyStream from 'arrayify-stream';
+import { DataFactory as N3DF } from 'n3';
 import { DataFactory } from 'rdf-data-factory';
 import { RdfSourceQpf } from '../lib/RdfSourceQpf';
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jeswr** [comment]
Note the following flakey browser test https://github.com/comunica/comunica/actions/runs/4080935735/jobs/7033870127

File: packages/actor-rdf-resolve-hypermedia-qpf/test/RdfSourceQpf-test.ts [hunk 1]
```diff
@@ -4,6 +4,7 @@ import type { IActorDereferenceRdfOutput } from '@comunica/bus-dereference-rdf';
 import { ActionContext, Bus } from '@comunica/core';
 import type * as RDF from '@rdfjs/types';
 import arrayifyStream from 'arrayify-stream';
+import { DataFactory as N3DF } from 'n3';
```
rubensworks, line 7: Is there a reason why another datafactory is used here?

jeswr: Not  now - it was an artefact of me thinking that rdf-data-factory was setting the value of the default graph term to be "DEFAULT_GRAPH"

rubensworks: It also uses the empty string if I remember correctly. I guess we can keep using the usual factory here then?

Issue #1095, opened 2022-10-07
Optional fails on N3 store


danielbeeke:
#### Issue type:

- :bug: Bug 

____
#### Description:

Given the following two triples I would expect four bindings as the result of the following query.
```
<http://ex.org/Pluto> <http://ex.org/type> <http://ex.org/Dog>
<http://ex.org/Mickey> <http://ex.org/name> "Lorem ipsum"@nl

```

```
  SELECT * WHERE { 
    ?s ?p ?o .

    OPTIONAL {
      ?s <http://ex.org/name> ?name .
      FILTER(lang(?name) = 'nl')
    }
  }
```
_Expectation_
s, p, o and name

_Actual_
s, p and o

When changing the language filter to something that does not exists I get zero bindings.

See http://danielbeeke.nl/comunica-rdfjs-optional-bug/ for a reproduction.
And the console on https://github.com/danielbeeke/comunica-rdfjs-optional-bug.
____
#### Environment:

I am using N3 and Comunica in a web site.
N3 is fetched as an ESM module.
Comunica is included via a script tag.

-----

## Bounty

A [bounty](https://comunica.dev/association/bounties/) has been placed on this issue by:

| |
|-|
| [<img alt="Comunica Association" src="https://comunica.dev/img/comunica_red.svg" height="50">](https://comunica.dev/association/) |
| €544 |

[Click here to learn more](https://comunica.dev/association/bounty_process/) if you're interested in claiming this bounty by resolving this issue.

rubensworks:
Possible a bug in either https://github.com/comunica/comunica/tree/master/packages/actor-rdf-join-optional-bind or https://github.com/comunica/comunica/tree/master/packages/actor-rdf-join-optional-nestedloop

rubensworks:
A bounty has been placed on this issue via the Comunica Association (see original post).

danielbeeke:
I found out where the bindings are dropped.
They are returned in the first place, it goes through all kinds of actors. 
But when almost being returned they are dropped because there is an expression error.

It happens in **ActorQueryOperationLeftJoin**. 
When calling **evaluator.evaluateAsEBV(bindings)** there is an error: **Unbound variable '"?name"'**
So the bug seems to be happening inside **Sparqlee** or **ActorQueryOperationLeftJoin**.

danielbeeke:
Should the bindings object have all the keys but maybe with empty values before being sent to Sparqlee?

rubensworks:
[...]

PR #1158, opened 2023-02-17
Rdfjs optional fails, check on existing keys in bindings and only check those bindings against Sparqlee

danielbeeke:
https://github.com/comunica/comunica/issues/1095


packages/actor-query-operation-leftjoin/lib/ActorQueryOperationLeftJoin.ts
```diff
@@ -32,6 +34,7 @@ export class ActorQueryOperationLeftJoin extends ActorQueryOperationTypedMediate
         output: ActorQueryOperation.getSafeBindings(output),
         operation,
       }));
+    const variables = operationOriginal.expression ? [ ...this.getVariables(operationOriginal.expression) ] : [];
```
rubensworks, line 37: You should be able to find the list of variables in `joined`.

packages/actor-query-operation-leftjoin/lib/ActorQueryOperationLeftJoin.ts
```diff
@@ -42,6 +45,15 @@ export class ActorQueryOperationLeftJoin extends ActorQueryOperationTypedMediate
         .transform({
           autoStart: false,
           transform: async(bindings: Bindings, done: () => void, push: (item: Bindings) => void) => {
+            const keysArefulfilled = variables.length > 0 ?
+              variables.every(variable => bindings.has(variable.value)) :
```
rubensworks, line 49: We need to look only for the variables in the right-hand stream (the one inside the optional block) I think.

danielbeeke: The checking needed to be done in the left hand stream. That is where bindings were skipped.

rubensworks:
Thanks for the PR @danielbeeke!

Unfortunately though, I don't think this is the right way forward, due to the following reasons:
- You're applying the expression on both the left and right streams, while it should only be applied on the right stream.
- You're overwriting the bindingsStreams inside the join entries array, without overriding their corresponding operations, which will cause issues when using the leftjoin-bind actor (which depends on the bindingStream of the left, but the operation of the right).

Instead of going forward with this approach, I think we can keep using the old approach (join first, then filter). But with the change I mentioned before:
> We may have to check if the values for all keys in the right-hand stream are properly defined, and only pass it along to the evaluator. But if they are not all defined, then we can skip the evaluator, and push the bindings directly.

One other alternative could be that we modify the operation of left on line 27, and wrap it in a filter operation there, so that we don't have to do the filtering in this class anymore. But I'm not 100% certain yet that this is correct semantically (might be though).

danielbeeke:
Hi Ruben,

So if I understand it correctly the way forward is to allow all bindings from the left stream and not evaluate anything there? And only to evaluate the right stream.

How can I filter on bindings from the right stream inside a stream that has left and right? After the join. My understanding is that it should happen in a streaming matter, so I can not keep an array of triples to check against.

rubensworks:
> So if I understand it correctly the way forward is to allow all bindings from the left stream and not evaluate anything there? And only to evaluate the right stream.

Correct, that's the end-goal.

> How can I filter on bindings from the right stream inside a stream that has left and right?

This could be done by the way I said before (and based on the original code):

> We may have to check if the values for all keys in the right-hand stream are properly defined, and only pass it along to the evaluator. But if they are not all defined, then we can skip the evaluator, and push the bindings directly.

But, since this is a bit finicky. It might be interesting to have a look at the [suggested alternative](https://github.com/comunica/comunica/pull/1158#issuecomment-1440003297) as well.

danielbeeke:
> You're overwriting the bindingsStreams inside the join entries array, without overriding their corresponding operations, which will cause issues when using the leftjoin-bind actor (which depends on the bindingStream of the left, but the operation of the right).

Would this be fixed by using a ClosableTransformIterator? It seems it is expecting that.

If a ClosableTransformIterator would help, then I could overwrite the right stream with a ClosableTransformIterator and filter that stream for bindings that have all the keys.

rubensworks:
> Would this be fixed by using a ClosableTransformIterator? It seems it is expecting that.

I don't see how that could help. Unless I'm missing something.

This is what ClosableTransformIterator does:

> A TransformIterator with a callback for when this iterator is closed in any way.

danielbeeke:
I was thinking about the ClosableTransformIterator because of https://github.com/comunica/comunica/blob/master/packages/actor-rdf-join-optional-bind/lib/ActorRdfJoinOptionalBind.ts#L31

danielbeeke:
I have tried to validate the right stream only, but the bindings come out in different order and some tests fail:

- An ActorQueryOperationLeftJoin instance › should correctly handle falsy expressions
- An ActorQueryOperationLeftJoin instance › should correctly handle erroring expressions

All tests pass when validating both streams as before, but in the left stream push bindings that have missing keys (current PR code).

How can we validate what is the right approach? I would say that no failing tests would be a big point. The performance penalty of the extra check might be small I don't know.

> You're overwriting the bindingsStreams inside the join entries array, without overriding their corresponding operations, which will cause issues when using the leftjoin-bind actor (which depends on the bindingStream of the left, but the operation of the right).

I do not fully understand this. Would it be a valid approach to overwrite the operation also? What kind of operation would it need?

rubensworks:
> but the bindings come out in different order and some tests fail

Order is undefined in standard join operators, so this is ok.
Order only becomes important once the `ORDER BY` operator is used, but this usually happens higher up in the query.

> Would it be a valid approach to overwrite the operation also? What kind of operation would it need?

Yes, but then you might be easier off with going for the alternative I suggested before.

danielbeeke:
I changed the approach to be:

join first, filtering only on variables from the expression (right) stream.

@rubensworks is this in line with your alternative?

packages/actor-query-operation-leftjoin/lib/ActorQueryOperationLeftJoin.ts
```diff
@@ -36,12 +36,23 @@ export class ActorQueryOperationLeftJoin extends ActorQueryOperationTypedMediate
...
     if (operationOriginal.expression) {
+      const rightMetadata = await entries[1].output.metadata();
+      const expressionVariables = rightMetadata.variables;
       const config = { ...ActorQueryOperation.getAsyncExpressionContext(context) };
       const evaluator = new AsyncEvaluator(operationOriginal.expression, config);
       const bindingsStream = joined.bindingsStream
         .transform({
           autoStart: false,
           transform: async(bindings: Bindings, done: () => void, push: (item: Bindings) => void) => {
+            const keysArefulfilled = expressionVariables.length > 0 ?
```
rubensworks, line 47:
```suggestion
            const optionalBindingsFulfilled = expressionVariables.length > 0 ?
```

packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
```diff
@@ -124,6 +124,82 @@ describe('ActorQueryOperationLeftJoin', () => {
...
+            BF.bindings([[ DF.variable('c'), DF.literal('1') ]]),
+            BF.bindings([[ DF.variable('a'), DF.literal('1') ]]),
+          ], { autoStart: false }),
+          metadata: () => Promise.resolve({ cardinality: 3, canContainUndefs: true, variables: [ DF.variable('a') ]}),
+          operated: arg,
+          type: 'bindings',
+        }),
+      };
+
+      actor = new ActorQueryOperationLeftJoin({ name: 'actor', bus, mediatorQueryOperation, mediatorJoin });
```
rubensworks, line 146: Making a new actor is not needed if you just override `mediatorQueryOperation.mediate` instead of creating a new `mediatorQueryOperation` object.

packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
```diff
@@ -124,6 +124,82 @@ describe('ActorQueryOperationLeftJoin', () => {
       });
     });
 
+    it('should correctly handle left hand bindings that are missing the variables of the expression', async() => {
```
rubensworks, line 127: I don't fully understand this test.
Won't this test also pass on the old code?

I would have expected a distinction between the two calls to `mediatorQueryOperation`, and return different metadata and bindings.
I also would expect an actual expression, which is only evaluated when bindings variables are complete, in which case we should check for the expected expression result.

packages/actor-query-operation-leftjoin/lib/ActorQueryOperationLeftJoin.ts
rubensworks, line 47: I think this length check can go, given the `every` semantics for empty arrays.

packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
```diff
@@ -124,6 +124,82 @@ describe('ActorQueryOperationLeftJoin', () => {
...
+        expect(await output.metadata()).toMatchObject({
+          cardinality: 100,
+          canContainUndefs: true,
+          variables: [ DF.variable('a'), DF.variable('b') ],
+        });
+        expect(output.type).toEqual('bindings');
+      });
+    });
+
+    it('should correctly handle left hand bindings when there are no variables', async() => {
```
rubensworks, line 165: Not sure we need this test.
The variables should be there. If they are not, then there's an underlying bug in some other actor.

packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
danielbeeke: I have written this test to get 100% code coverage.
It checks the code path where there are no variables.
Do you see a better way to check for this code path?

packages/actor-query-operation-leftjoin/lib/ActorQueryOperationLeftJoin.ts
danielbeeke: will change.

danielbeeke: yes, I fixed it.

packages/actor-query-operation-leftjoin/lib/ActorQueryOperationLeftJoin.ts
```diff
@@ -36,12 +36,21 @@ export class ActorQueryOperationLeftJoin extends ActorQueryOperationTypedMediate
...
+      const expressionVariables = rightMetadata.variables;
       const config = { ...ActorQueryOperation.getAsyncExpressionContext(context) };
       const evaluator = new AsyncEvaluator(operationOriginal.expression, config);
       const bindingsStream = joined.bindingsStream
         .transform({
           autoStart: false,
           transform: async(bindings: Bindings, done: () => void, push: (item: Bindings) => void) => {
+            const optionalBindingsFulfilled = expressionVariables.every(variable => bindings.has(variable.value));
+
+            if (!optionalBindingsFulfilled) {
```
rubensworks, line 49: Since the evaluation of this boolean is quite simple, I guess we can inline it into this if-statement.

packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
rubensworks: One way of doing this would be to check the `arg` in `mediatorQueryOperation.mediate`. If it's the left stream (need something to identify this when passing it to `op`), then you return one thing (bindingsStream+metadata+...). If it's the right stream, you return something else.

rubensworks (changes requested):
Great! (also see the comment I just replied to)

Could you also add an integration test to https://github.com/comunica/comunica/blob/master/engines/query-sparql/test/QuerySparql-test.ts, corresponding approximately to your original issue?

packages/actor-query-operation-leftjoin/lib/ActorQueryOperationLeftJoin.ts
danielbeeke: will do

danielbeeke:
Yes, I will create an integration test.

packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
danielbeeke: If I did understand it correctly, the test checks:

- a binding from the left stream
- that does not have the variables from the expression
- so it must be push without being evaluated

I have removed the cruft around the test so now it only checks if it indeed has the binding that did pass through the if statement that checks if it does not contains all the variables.

@rubensworks do you think it is now okay?

packages/actor-query-operation-leftjoin/lib/ActorQueryOperationLeftJoin.ts
```diff
@@ -36,12 +36,19 @@ export class ActorQueryOperationLeftJoin extends ActorQueryOperationTypedMediate
...
     if (operationOriginal.expression) {
+      const rightMetadata = await entries[1].output.metadata();
+      const expressionVariables = rightMetadata.variables;
       const config = { ...ActorQueryOperation.getAsyncExpressionContext(context) };
       const evaluator = new AsyncEvaluator(operationOriginal.expression, config);
       const bindingsStream = joined.bindingsStream
         .transform({
           autoStart: false,
           transform: async(bindings: Bindings, done: () => void, push: (item: Bindings) => void) => {
+            if (!expressionVariables.every(variable => bindings.has(variable.value))) {
```
rubensworks, line 47: Could you add the following comment before this line?:
`// If variables of the right-hand entry are missing, we skip expression evaluation`

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -686,6 +686,34 @@ describe('System test: QuerySparql', () => {
         expect((await arrayifyStream(await result.execute()))).toEqual([]);
       });
     });
+
+    describe('left join with an optional', () => {
+      it('with correct results', async() => {
```
rubensworks, line 691:
```suggestion
      it('should handle filters inside the optional block', async() => {
```

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -686,6 +686,34 @@ describe('System test: QuerySparql', () => {
...
+        const result = <QueryBindings> await engine.query(`
+        SELECT * WHERE { 
+          ?s ?p ?o .
+      
+          OPTIONAL {
+            ?s <http://ex.org/name> ?name .
+            FILTER(lang(?name) = 'nl')
+          }
+        }`, { sources: [ store ]});
+        expect((await arrayifyStream(await result.execute())).length).toBe(2);
```
rubensworks, line 714: Since we just have 2 results here, I guess we can also add an `expect` line to match the exact results we expect.

packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
rubensworks: I mean something like this:
```
mediatorQueryOperation.mediate = (arg: any) => {
  if (arg is of left stream) return x;
  else return y;
}
```

rubensworks (changes requested):

packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
rubensworks: Did you see my last comment above?

danielbeeke: Hi @rubensworks 

Thanks for the heads up, now I have seen it.
I will have a look at it. 

I can not yet piece it together, I feel like the test is already passing through the code.

What would it improve? if we return a different thing depending on the arg?

rubensworks: > What would it improve?

Code clarity.

Indeed, the test is already causing full coverage of the code. But this feels very arbitrary when just reading the test.
By explicitly adding the if-statement in the test, it should become clearer to the code reader what the test is doing exactly.

danielbeeke: okay, thanks, I will implement this

danielbeeke:
@rubensworks could you have a look at the improved test? Is this is what you are thinking about?

rubensworks (approved):
This looks perfect, nice work @danielbeeke!

danielbeeke:
Thanks @rubensworks for all the reviews! Glad it is merged now and up to the required quality :partying_face:

PR #1165, opened 2023-03-08
fix: destroy unused iterators

jeswr:
This is the result of testing which `AsyncIterator`s were left open at the end of tests and destroying them. In most cases they were test iterators that were not fully consumed - but there were also some cases such as performing `minus` operations where iterators were being left open in the component itself.

### DIFF
--- a/packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
+++ b/packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
@@ -248,7 +248,6 @@ describe('ActorQueryOperationLeftJoin', () => {
           // Do nothing
         });
       });
-      // TODO: See if this should be closed on the error event
       output.bindingsStream.destroy();
     });
   });
--- a/packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
+++ b/packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
@@ -248,6 +248,8 @@ describe('ActorQueryOperationLeftJoin', () => {
           // Do nothing
         });
       });
+      // TODO: See if this should be closed on the error event
+      output.bindingsStream.destroy();
     });
   });
 });
--- a/packages/actor-rdf-resolve-quad-pattern-hypermedia/test/MediatedLinkedRdfSourcesAsyncRdfIterator-test.ts
+++ b/packages/actor-rdf-resolve-quad-pattern-hypermedia/test/MediatedLinkedRdfSourcesAsyncRdfIterator-test.ts
@@ -103,7 +103,7 @@ describe('MediatedLinkedRdfSourcesAsyncRdfIterator', () => {
 
     afterEach(() => {
       // This is here since the MediatedLinkedRdfSourcesAsyncRdfIterator is never fully consumed
-      // **THIS IS AN ALARM BELL** do not merge without checking this
+      // This should not be needed, see https://github.com/comunica/comunica/issues/1197
       source.destroy();
     });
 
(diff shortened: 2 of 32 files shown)


packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
```diff
@@ -215,6 +215,8 @@ describe('ActorQueryOperationLeftJoin', () => {
           // Do nothing
         });
       });
+      // TODO: See if this should be closed on the error event
+      output.bindingsStream.destroy();
```
jeswr, line 219: This is a case where it may be better to update the core behavior to close the iterator on an error event.

packages/actor-rdf-resolve-quad-pattern-hypermedia/test/MediatedLinkedRdfSourcesAsyncRdfIterator-test.ts
```diff
@@ -101,6 +101,12 @@ describe('MediatedLinkedRdfSourcesAsyncRdfIterator', () => {
       );
     });
 
+    afterEach(() => {
+      // This is here since the MediatedLinkedRdfSourcesAsyncRdfIterator is never fully consumed
+      // **THIS IS AN ALARM BELL** do not merge without checking this
+      source.destroy();
+    });
+
```
jeswr, line 109: I think there is likely to be a more fundamental problem with the MediatedLinkedRdfSourcesAsyncRdfIterator here because it is always leaving iterators open.

jeswr:
This work takes us from ~300 -> ~50 iterators still being open after all tests are run. Most concerning are the following from the `QuerySparql` test suite which remain open

```
 ● System test: QuerySparql › query › simple SPO on a TPF entrypoint › with results

    expect(received).toEqual(expected) // deep equality

    Expected: 0
    Received: 3

      at Object.<anonymous> (node_modules/asynciterator/dist/asynciterator.js:85:27)

  ● System test: QuerySparql › query › simple SPO on a TPF entrypoint › with filtered results

    expect(received).toEqual(expected) // deep equality

    Expected: 0
    Received: 1

      at Object.<anonymous> (node_modules/asynciterator/dist/asynciterator.js:85:27)

  ● System test: QuerySparql › query › two-pattern query on a TPF entrypoint › with results

    expect(received).toEqual(expected) // deep equality

    Expected: 0
    Received: 4

      at Object.<anonymous> (node_modules/asynciterator/dist/asynciterator.js:85:27)

  ● System test: QuerySparql › query › two-pattern query on a TPF entrypoint › without results

    expect(received).toEqual(expected) // deep equality

    Expected: 0
    Received: 4

      at Object.<anonymous> (node_modules/asynciterator/dist/asynciterator.js:85:27)
```

rubensworks:
Thanks for this @jeswr!

> Perhaps it is worth using the rdf-streaming-store here as well rather than caching the arrays of quads themselves as this would surely reduce what needs to be stored in memory?

That might be a good solution indeed.

---

Wondering if we should add something to AsyncIterator to allow starting it in a debug-mode, so that it prints all un-closed iterators on process exit.

packages/actor-rdf-resolve-quad-pattern-hypermedia/test/MediatedLinkedRdfSourcesAsyncRdfIterator-test.ts
rubensworks: Yeah, this is a complex part of this codebase, so there might be issues with it.
Something you want to look into within the scope of this PR?

jeswr:
@rubensworks - the TODOs have now been converted to issues; and the branch has been rebased. This is ready to merge.

packages/actor-rdf-resolve-quad-pattern-hypermedia/test/MediatedLinkedRdfSourcesAsyncRdfIterator-test.ts
```diff
@@ -101,6 +101,12 @@ describe('MediatedLinkedRdfSourcesAsyncRdfIterator', () => {
       );
     });
 
+    afterEach(() => {
+      // This is here since the MediatedLinkedRdfSourcesAsyncRdfIterator is never fully consumed
+      // **THIS IS AN ALARM BELL** do not merge without checking this
```
rubensworks, line 106:
```suggestion
      // This should not be needed, see https://github.com/comunica/comunica/issues/1197
```

rubensworks:
Thanks! :)

PR #1167, opened 2023-03-08
fix: use `globalThis` rather than `global`

jeswr:
`global` doesn't exist in browser environments.

`globalThis` refers to `global` in node and `window` in browser.

cc @constraintAutomaton

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -33,6 +33,7 @@ module.exports = {
     Headers: false,
     Request: false,
     XMLHttpRequest: false,
+    globalThis: false,
   },
   rules: {
     // Default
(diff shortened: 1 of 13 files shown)


.eslintrc.js
```diff
@@ -33,6 +33,7 @@ module.exports = {
     Headers: false,
     Request: false,
     XMLHttpRequest: false,
+    globalThis: false,
```
rubensworks, line 36: Wondering if we should/could add a linter rule disallowing the use of `global`?
(I did not find a rule for this after a quick search)

jeswr: I wonder if setting global: true here would work.

jeswr: Will try soon

jeswr: Setting them to "off" works (see https://eslint.org/docs/latest/use/configure/language-options#using-configuration-files-1)

rubensworks:
Thanks!

Issue #1169, opened 2023-03-10
Engine `query-sparql-rdfjs` is missing a dependency


karelklima:
#### Issue type:

- :bug: Bug 

____
#### Description:

Engine `query-sparql-rdfjs` is missing a dependency of `@comunica/actor-rdf-serialize-shaclc` which is introduced through compiled engine configuration.

____
#### Steps to reproduce:

- Create a new npm project
- Run `npm install @comunica/query-sparql-rdfjs`
- Run the following script:

```js
import { QueryEngine } from "@comunica/query-sparql-rdfjs";

const engine = new QueryEngine();
```

____
#### Related:

- https://github.com/comunica/comunica/issues/1139
- https://github.com/comunica/comunica/commit/e3b938940fa30b10bf2aa68a5f6ac3086399873a

I'll try to submit a PR to fix this shortly.

karelklima:
Last working version of the `@comunica/query-sparql-rdfjs` seems to be 2.5.2 based on my testing, i.e. pre-shaclc introduction.

PR #1170, opened 2023-03-10
Fix missing dependency in query-sparql-rdfjs

karelklima:
This PR should fix https://github.com/comunica/comunica/issues/1169


rubensworks:
Thanks @karelklima!
Preparing a new patch release now...

Issue #1163, opened 2023-03-02
Malformed QPF with literal subject


langsamu:
#### Issue type:

- :bug: Bug 

____
##### The issue

Comunica translates some SPARQL queries to malformed QPF queries where the `subject` parameter is a literal, though
> The subject _MUST_ either be a variable or an IRI

-- [Triple Pattern Fragments 3.2](https://linkeddatafragments.org/specification/triple-pattern-fragments/#:~:text=The%20subject%20MUST%20either%20be%20a%C2%A0variable%20or%20an%20IRI)
-- [Quad Pattern Fragments 3.2](https://linkeddatafragments.org/specification/quad-pattern-fragments/#:~:text=The%20subject%20MUST%20either%20be%20a%20variable%20or%20an%20IRI)

Some servers ignore such QPF queries but others reject them.
When such a server responds with `400`, Comunica does not recover and query execution fails.

##### What should be done

I think Comunica should not even attempt to send a QPF requests with literal subjects because
- they're malformed,
- they always return empty results and
- they might fail entirely.

Instead I think Comunica should just ignore these requests.
(If Linked Data Fragments ever supported generalized RDF then this would need to change.)

Assuming you agree, I'd gladly try my hand at a first contribution, but I'd need some guidance.
Unfamiliar as I am with the codebase, I naively think this might be as easy as a condition on the term type of the QPF request subject, if any.

##### To reproduce

Assume a QPF endpoint `https://example.org/qpf` that exposes the following data:
```ttl
BASE <http://example.org/>

<s> <p> "o" .
```

Assume a SPARQL query
```sparql
SELECT * {
  GRAPH ?g {
    ?s1 ?p1 ?o1 .
    ?o1 ?p2 ?o2 .
  }
}
```

Comunica translates this query into two calls:
1. `https://example.org/qpf`
2. `https://example.org/qpf?subject=%22o%22`

____
#### Environment:

I'm using 
```
http://rdf.js.org/comunica-browser/versions/v2.3.0/engines/query-sparql/comunica-browser.js
```

jeswr:
> Assuming you agree, I'd gladly try my hand at a first contribution, but I'd need some guidance. Unfamiliar as I am with the codebase, I naively think this might be as easy as a condition on the term type of the QPF request subject, if any.

[...]

PR #1171, opened 2023-03-12
Return no quads for malformed QPF selectors

langsamu:


langsamu:
@jeswr, @rubensworks can I have some feedback.

I think the adjustment is basically fine:
https://github.com/comunica/comunica/blob/ca0ae98ee657a96b638fafd1635222f7814428a7/packages/actor-rdf-resolve-hypermedia-qpf/lib/RdfSourceQpf.ts#L122-L124

However I wanted my tests to **fail before the adjustment** (in addition to covering the new code).
But they [pass](https://github.com/comunica/comunica/actions/runs/4399067583/jobs/7703351934#step:10:27).

In the test, I'm trying to set up a source with 'weird' quads (e.g. with a literal subject) that would **not be empty** when queried with a malformed QPF selector. This is so I can prove they _are_ empty when malformed selectors are ignored.

rubensworks:
Thanks for the PR @langsamu!

As mentioned in https://github.com/comunica/comunica/issues/1163#issuecomment-1453045914, I would prefer a more general approach, as the same problem could also occur for other source types.
So basically, the same logic could be applied in https://github.com/comunica/comunica/blob/master/packages/actor-query-operation-quadpattern/lib/ActorQueryOperationQuadpattern.ts

To still allow the old mode for people relying on the current behaviour, we'll also have to introduce a `generalizedRdf` context option (can default to `false`).

rubensworks:
Closing this PR due to staleness.
Feel free to re-open if there are any updates! 🙂

PR #1186, opened 2023-03-28
fix: use process shim

jeswr:
Uses the process shim rather than a global. It may be possible to remove the polyfills of process after this PR but we need to evaluate if any dependents use process globally (unless there are already tests in place to detect that in Comunica?)

### DIFF
--- a/lerna.js
+++ b/lerna.js
@@ -21,7 +21,7 @@ async function depInfo({ location, name }, log) {
         dependencies = dependencies.filter(dep => dep !== 'process');
       } else {
         // If it is *not* declared in teh dependencies then mark it as missing
-        missing['process'] ??= [];
+        missing['process'] = missing['process'] || [];
       }
   }
 
--- a/lerna.js
+++ b/lerna.js
@@ -12,11 +12,17 @@ async function depInfo({ location, name }, log) {
 
   let {dependencies, devDependencies, missing, using} = await checkDeps(location, { ignorePatterns: ignore }, val => val);
 
-  if (dependencies.includes('process') && Object.values(using).flat().some(file => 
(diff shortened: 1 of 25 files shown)


rubensworks:
I think we'll want to require `process/` instead of `process`, to avoid the problems explained in https://github.com/nodejs/readable-stream/issues/496.

> but we need to evaluate if any dependents use process globally (unless there are already tests in place to detect that in Comunica?)

We don't have a good way of testing this AFAIK.

jeswr:
> I think we'll want to require process/ instead of process, to avoid the problems explained in https://github.com/nodejs/readable-stream/issues/496.

Done @rubensworks :)

lerna.js
```diff
@@ -10,7 +10,22 @@ async function depInfo({ location, name }, log) {
   let ignore = files ? folders.filter(elem => files.every(file => !file.startsWith(elem.name))) : folders;
   ignore = ignore.map(x => x.isDirectory() ? `${x.name}/**` : x.name)
 
-  const {dependencies, devDependencies, missing, using} = await checkDeps(location, { ignorePatterns: ignore }, val => val);
+  let {dependencies, devDependencies, missing, using} = await checkDeps(location, { ignorePatterns: ignore }, val => val);
+
+  if (Object.values(using).flat().some(file => 
+    readFileSync(file, 'utf8').toString().includes('const process: NodeJS.Process = require(\'process/\')') ||
```
rubensworks, line 16: Instead of these 4 cases, I assume we could just check for the presence of `require(\"process/\")`?
This would also allow us to capture the case where `process` is used inline.

rubensworks (changes requested):
Thanks, looks great!

Just one question about the lerna script.

PR #1202, opened 2023-04-16
feat: bump sparqlee version

jitsedesmet:
Bump sparqlee version and add spec tests for sparql 1.2 date support. Depends on: https://github.com/rubensworks/rdf-test-suite.js/pull/91


jitsedesmet:
I'm having a though time with `yarn install` here. Whenever I call it, the preinstall script starts removing dependencies. However, these dependencies are needed.

rubensworks:
> I'm having a though time with yarn install here. Whenever I call it, the preinstall script starts removing dependencies. However, these dependencies are needed.

Prob because you're using Node 18. Can you try with Node 16?

jitsedesmet:
Ah yes, this seemed to be the issue, thank you!

rubensworks:
A new rdf-test-suite version has been released, so bumping this here should fix the CI I guess.

rubensworks:
Nice! Thanks @jitsedesmet.

PR #1204, opened 2023-04-21
Use --skip for rdf test suite

jitsedesmet:
This PR uses the new feature of `rdf-test-suite.js` making us selectively ignore certain tests. I also removed one `-e` because it didn't seem needed?


rubensworks:
Thanks, and good catch on the SPARQL update tests!

Issue #1139, opened 2023-01-11
Add check for query-sparql-* depencies


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

To avoid issues such as https://github.com/comunica/comunica/commit/e3b938940fa30b10bf2aa68a5f6ac3086399873a, we should add a CI check to ensure that all dependencies from the compiled engine configs are available in the package.json file of each engine.
We may be able to tweak our `depcheck` script for this.

@jeswr any thoughts on this?

-----

## Bounty

A [bounty](https://comunica.dev/association/bounties/) has been placed on this issue by:

| |
|-|
| [<img alt="Comunica Association" src="https://comunica.dev/img/comunica_red.svg" height="50">](https://comunica.dev/association/) |
| €136 |

[Click here to learn more](https://comunica.dev/association/bounty_process/) if you're interested in claiming this bounty by resolving this issue.

jeswr:
Definitely worthwhile - we could try running depcheck by ensuring that depcheck points to the bundled engine. However, I seem to remember that doing so did not include all the required dependencies - is there a reason to not just add a CI check to make sure the package includes all the dependencies that `@comunica/packager` prescribes?

rubensworks:
A bounty has been placed on this issue via the Comunica Association (see original post).

jitsedesmet:
> all dependencies from the compiled engine configs are available in the package.json file of each engine.

With "the configs" do you refer to: https://github.com/comunica/comunica/tree/master/engines/query-sparql-file/config ?

rubensworks:
> With "the configs" do you refer to: https://github.com/comunica/comunica/tree/master/engines/query-sparql-file/config ?

Those are indeed the configs.
But those configs are compiled to `engine-default.js` files inside each query engine package.
Inside `engine-default.js`, there will be a bunch of imports, which we will have to check to ensure that they are available as dependency.

Note that this `engine-default.js` is not available on git, as it's generated during `yarn run build`.

PR #1209, opened 2023-05-04
feat: handle engines in depcheck

jitsedesmet:
This PR aims to resolve: https://github.com/comunica/comunica/issues/1139

### DIFF
--- a/engines/query-sparql-file/package.json
+++ b/engines/query-sparql-file/package.json
@@ -163,7 +163,6 @@
     "@comunica/actor-rdf-update-quads-rdfjs-store": "^2.6.9",
     "@comunica/bus-http-invalidate": "^2.6.8",
     "@comunica/bus-query-operation": "^2.6.8",
-    "@comunica/config-query-sparql": "^2.6.0",
     "@comunica/core": "^2.6.8",
     "@comunica/logger-void": "^2.6.8",
     "@comunica/mediator-all": "^2.6.8",
@@ -172,9 +171,11 @@
     "@comunica/mediator-join-coefficients-fixed": "^2.6.8",
     "@comunica/mediator-number": "^2.6.8",
     "@comunica/mediator-race": "^2.6.8",
-    "@comunica/runner": "^2.6.8",
     "@comunica/runner-cli": "^2.6.8",
     "process": "^0.11.10"
+  },
+  "devDependencies": {
+    "@comunica/types": "^2.6.8"
   },
   "scripts": {
     "build": "npm run build:ts",
--- a/engines/query-sparql-file/package.json
+++ b/engines/query-sparql-file/package.json
@@ -163,6 +163,7 @@
     "@comunica/actor-rdf-update-quads-rdfjs-store": "^2.6.9",
     "@comunica/bus-http-invalidate": "^2.6.8",
     "@comunica/bus-query-operation": "^2.6.8",
+    "@comunica/config-query-sparql": "^2.6.0",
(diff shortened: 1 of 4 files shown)


engines/query-sparql-file/package.json
```diff
@@ -163,7 +163,6 @@
     "@comunica/actor-rdf-update-quads-rdfjs-store": "^2.6.9",
     "@comunica/bus-http-invalidate": "^2.6.8",
     "@comunica/bus-query-operation": "^2.6.8",
-    "@comunica/config-query-sparql": "^2.6.0",
```
rubensworks, line 166: This one should definitely be present, as it will be used when instantiating an engine dynamically from the config file using the query engine factory.

engines/query-sparql-file/package.json
```diff
@@ -172,10 +171,12 @@
     "@comunica/mediator-join-coefficients-fixed": "^2.6.8",
     "@comunica/mediator-number": "^2.6.8",
     "@comunica/mediator-race": "^2.6.8",
-    "@comunica/runner": "^2.6.8",
     "@comunica/runner-cli": "^2.6.8",
     "process": "^0.11.10"
   },
+  "devDependencies": {
+    "@comunica/types": "^2.6.8"
```
rubensworks, line 178: Should be a normal dependency. (but is it used even?)

engines/query-sparql-file/package.json
```diff
@@ -172,10 +171,12 @@
     "@comunica/mediator-join-coefficients-fixed": "^2.6.8",
     "@comunica/mediator-number": "^2.6.8",
     "@comunica/mediator-race": "^2.6.8",
-    "@comunica/runner": "^2.6.8",
```
rubensworks, line 175: I think we'll also need this one for the dynamic engine from config to work.

engines/query-sparql-file/package.json
jitsedesmet: It's not present in any js files. The only time `@comunica/config-query-sparql` is mentioned (in this directory) is in the `config-default.json` and `config-browser.json` files.

engines/query-sparql-file/package.json
jitsedesmet: It is used in `lib/QueryEngine.ts` does this not count?

engines/query-sparql-file/package.json
jitsedesmet: This dependency is also never mentioned throughout the whole dir.

engines/query-sparql-file/package.json
jitsedesmet:
```
{
  "@context": [
    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/config-query-sparql/^2.0.0/components/context.jsonld"
  ],
  "import": [
    "ccqs:config/config-file.json"
  ]
}
```

rubensworks: We'll need a workaround for it then, as it will otherwise cause errors during dependency injection.

engines/query-sparql-file/package.json
rubensworks: Typing packages should always be actual dependencies, otherwise users of comunica will see compilation errors due to missing types.

engines/query-sparql-file/package.json
rubensworks: Like the config one, this is needed during dep injection, so we'll also need a workaround for this.

rubensworks:
Awesome work!

PR #1122, opened 2022-11-23
Feat/rdfstar rebase

jeswr:
Supercededs #1048

PR #1222, opened 2023-06-28
Add RDF-star and SPARQL-star support

rubensworks:
This implements full RDF-star and SPARQL-star support according to the RDF-star CG report: https://w3c.github.io/rdf-star/cg-spec/editors_draft.html

All spec tests (SPARQL-syntax and SPARQL-eval) pass.

Note that SPARQL-star support is enabled by default. Given the fact that quoted triples support will land in RDF-1.2, I don't see much value in hiding it behind a feature flag at this stage.

The plan is to update quoted triples support to align with the decisions of the RDF-star WG, but I doubt there will be significant changes wrt to the CG report.

This was partially based on @jeswr's work in #1122.

### DIFF
--- a/packages/actor-query-operation-quadpattern/test/ActorQueryOperationQuadpattern-test.ts
+++ b/packages/actor-query-operation-quadpattern/test/ActorQueryOperationQuadpattern-test.ts
@@ -55,6 +55,14 @@ describe('ActorQueryOperationQuadpattern', () => {
     it('should be false for a named node', () => {
       expect(ActorQueryOperationQuadpattern.isTermVariable(DF.namedNode('n'))).toBeFalsy();
     });
+
+    it('should be false for a quoted triple', () => {
+      expect(ActorQueryOperationQuadpattern.isTermVariable(DF.quad(
+        DF.namedNode('iri'),
+        DF.namedNode('iri'),
+        DF.namedNode('iri'),
+      ))).toBeFalsy();
+    });
   });
 
   describe('getDuplicateElementLinks', () => {
@@ -75,7 +83,7 @@ describe('ActorQueryOperationQuadpattern', () => {
 
     it('should return correctly on patterns with equal subject and predicate variables', () => {
       return expect(ActorQueryOperationQuadpattern.getDuplicateElementLinks(quad('?v1', '?v1', '?v3', '?v4')))
-        .toEqual({ subject: [ 'predicate' ]});
+        .toEqual({ subject: [[ 'predicate' ]]});
     });
 
     it('should ignore patterns with equal subject and predicate blank nodes', () => {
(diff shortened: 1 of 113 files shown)


packages/actor-query-operation-quadpattern/test/ActorQueryOperationQuadpattern-test.ts
```diff
@@ -760,5 +809,152 @@ describe('ActorQueryOperationQuadpattern', () => {
...
+          ]);
+          data.setProperty('metadata', metadataContent);
+          return Promise.resolve({ data });
+        },
+      );
+
+      const operation = AF.createPattern(
+        DF.quad(DF.variable('s'), DF.namedNode('p'), DF.namedNode('o')),
+        DF.variable('p'),
+        DF.namedNode('o'),
```
andrewzhurov, line 878: How does it happen that query matches quads where `o` is `[o1](https://github.com/comunica/comunica/pull/1222/commits/7acc31264c02321900b683850bc639738e0e3de6#diff-5b4a7cc2053820fe543450f4a971853e65a5e68a67855523613b6c352a997e1dR865)`, `o2`, `o3`?

packages/actor-rdf-resolve-quad-pattern-rdfjs-source/lib/IRdfJsSourceExtended.ts
```diff
@@ -1,6 +1,18 @@
...
 
 export interface IRdfJsSourceExtended extends RDF.Source {
+  /**
+   * A record indicating supported features of this source.
+   */
+  features?: {
+    /**
+     * If true, this source supports passing quad patterns with quoted quad patterns in the `match` method.
+     * If false (or if `features` is `undefined`), such quoted quad patterns can not be passed,
+     * and must be replaced by `undefined` and filtered by the caller afterwards.
```
andrewzhurov, line 11: 'filtered' here got me confused, I understood it as 'such quads would need to be removed by the caller afterwards'.

More clear to me would be:
If false (or if `features` is `undefined`), such quoted quad patterns can not be passed.
A new pattern can be derived, where they are replaced by `undefined`, and evaluated agains the store.
Some of the returned quads may not match agains the original pattern and need to be removed by the caller.

andrewzhurov: A side thought.
This approach will work for the case when store supports storing quoted triples but does not supports matching agains them.

For the case when store does not support storing them (and matching agains, as this is a feature on top), all of the resulting quads, matched agains altered quad patterns will be non-matching agains the original pattern, which would add overhead for the store to resolve and return them and for the client to match and remove.
Perhaps to spare the overhead we could have an additional feature `quotedTripleStoring` or smth, and when it's set and `quotedTripleFiltering` is not set - simply remove quad pattern altogether?

andrewzhurov: Also, is it 'quoted triples' or 'quoted quads'? (`quotedQuadsFiltering`?)

andrewzhurov:
Thanks for adding -star support! :fire: Amazing how blazing fast it got implemented! :muscle:

packages/actor-rdf-resolve-quad-pattern-rdfjs-source/lib/RdfJsQuadSource.ts
```diff
@@ -59,19 +84,26 @@ export class RdfJsQuadSource implements IQuadSource {
...
         );
         matches.on('error', reject);
         matches.on('end', () => resolve(i));
         matches.on('data', () => i++);
       });
     }
+
+    // If `match` would require filtering afterwards, our count will be an over-estimate.
+    const pattern = DF.quad(subject, predicate, object, graph);
+    const wouldRequirePostFiltering = (!quotedTripleFiltering &&
```
andrewzhurov, line 100: Could we instead of issuing a separate `source.match()` in order to count pass in args a stream of processed results from our `match()`?

Atm we only pass an iterator, and it's being exhausted on read from it, so we can't use it to count.
Could we craft another readable stream in `match()` and filter results in there? Then we can both wrap it with `wrapAsyncIterator` and handle it over to user and pass it in here, and read it in order to count quads.

That would spare us the need to `source.match()` twice and would allow to provide `exact` estimate.
One concern is that we have `autoStart`: `false` on iterator, I guess that's for a reason, and doing this strategy does start reading in order to count, so I guess results would be eagerly resolved and available in the iterator. Don't know what adverse effects that might have.

PR #1233, opened 2023-07-18
Turn on `strictFunctionTypes`

Peeja:
My project was reporting a type failure in `BindingsFactory.d.ts`, so I dug in to investigate. It was because `fromBindings()` needs to accept any `RDF.Bindings`, not just Comunica `Bindings`. It looks like the reason that wasn't failing in Comunica itself is that `strictFunctionTypes` was off.

I've taken the liberty of turning it on and resolving the type issues it was covering up. My assumption is that it was turned off because one or more of these was proving difficult. I recommend having it on, not only because it makes for sounder code internally, but also because downstream projects can have their own issues if they don't have it turned off (which is what happened here). I'm hoping that now that the leg work is done, you'll be comfortable turning it back on.

### DIFF
--- a/packages/actor-rdf-resolve-quad-pattern-hypermedia/lib/MediatedQuadSource.ts
+++ b/packages/actor-rdf-resolve-quad-pattern-hypermedia/lib/MediatedQuadSource.ts
@@ -43,8 +43,13 @@ export class MediatedQuadSource implements IQuadSource {
     predicate: RDF.Term,
     object: RDF.Term,
     graph: RDF.Term,
-    context: IActionContext,
+    context?: IActionContext,
   ): AsyncIterator<RDF.Quad> {
+    // IQuadSource["context"] should be mandatory in the next major update.
+    if (!context) {
+      throw new Error(`MediatedQuadSource requires a context.`);
+    }
+
     // Optimized match with aggregated store if enabled and started.
     let aggregatedStore: IAggregatedStore | undefined;
     if (this.aggregateStore) {
--- a/packages/actor-http-fetch/lib/ActorHttpFetch.ts
+++ b/packages/actor-http-fetch/lib/ActorHttpFetch.ts
@@ -41,9 +41,9 @@ export class ActorHttpFetch extends ActorHttp {
    * @param retryDelay Time in milliseconds to wait between retries
    * @returns a fetch `Response` object
    */
-  private static async getResponse(
-    fetchFn: (input: RequestInfo | URL, init?: RequestInit | undefined) => Promise<Response>,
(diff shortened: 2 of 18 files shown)


Peeja:
Also: Linting failed during commit in a way I didn't understand, so I force-committed for now. Do you recognize `Unable to resolve path to module '../engine-default.js'`? I don't see a file by that name in my tree.

jeswr:
> Also: Linting failed during commit in a way I didn't understand, so I force-committed for now. Do you recognize Unable to resolve path to module '../engine-default.js'? I don't see a file by that name in my tree.

That file is a build of the Comunica engine that should be generated after you run `yarn run build`.

sunny121li:
a

packages/actor-rdf-resolve-quad-pattern-hypermedia/lib/MediatedQuadSource.ts
```diff
@@ -43,8 +43,13 @@ export class MediatedQuadSource implements IQuadSource {
     predicate: RDF.Term,
     object: RDF.Term,
     graph: RDF.Term,
-    context: IActionContext,
+    context?: IActionContext,
   ): AsyncIterator<RDF.Quad> {
+    // IQuadSource["context"] should be mandatory in the next major update.
+    if (!context) {
+      throw new Error(`MediatedQuadSource requires a context.`);
```
rubensworks, line 50: This change will probably require a new unit test to achieve 100% test coverage.

rubensworks (commented):
Thanks for this change @Peeja!
I definitely agree that having this option turned on is a good idea.
I don't remember why it was turned off in the first place tbh...

I just published a new release of fetch-sparql-endpoint, this this PR should be unblocked now.

Peeja:
Shoot, [there's a deeper issue.](https://github.com/RubenVerborgh/AsyncIterator/issues/103) 😕 This one's causing errors on lines like [this](https://github.com/comunica/comunica/blob/2e197990a9d49dc8b7b500de8ba2ac7ae7b370c5/packages/actor-rdf-join-inner-multi-sequential/test/ActorRdfJoinMultiSequential-test.ts#L161).

rubensworks:
@Peeja Perhaps in those cases we should make the constructor's type explicit, or use casting.

rubensworks:
Hi @Peeja 👋 , just checking in if this PR is still on your roadmap. If not, I would close this PR for now, as not to have stale PRs open.

Peeja:
It's not right now, it looks like this is deeper than I'd hoped. I'll close for now.

Issue #1203, opened 2023-04-17
Consider adding a context to bindings?


surilindur:
#### Issue type:

- :heavy_plus_sign: Feature request

____
#### Description:

There may be use cases where metadata should be maintained per bindings instance, and carried with the bindings all the way to the query engine output (I do not know how to express this properly, sorry). One such example use case could be when multiple sources may contribute to a single result binding, and information about the exact sources contributing to a specific result from the query engine are needed, to help trace back any potentially weird results or filter them as needed.

To help address this use case and potential other ones, perhaps a context could be added to bindings, with functionality similar to the context being passed to the query engine itself and then passed around internally to actors. For example, a `Bindings` instance could have a `context` property where members could be accessed using a `IBindingsContextKey`:

```
function exampleSourceAddition(bindings: Bindings): void {
  const sourcesKey: IBindingsContextKey = new BindingsContextKey('urn:comunica:something:sources')
  const sourcesValue: Set<string> = bindings.context.getSafe<Set<string>>(sourcesKey)
  sourcesValue.add('https://example.com/sparql')
  bindings.context = bindings.context.set(sourcesKey, sourcesValue)
}
```
Maybe the existing context-related work could be generalised, to have an `IContextKey`, `IContext`, etc.? That way, any future aspects of the engine also needing context could be taken into consideration, and it would help reduce the amount of types because there are already a lot of them.

I am just creating this issue to open the discussion around the topic. :slightly_smiling_face:

rubensworks:
We could probably indeed directly use the `IContext` interface, and add it as a `context` field to bindings.

I guess we'll want the context to be only assignable during creation of the bindings? This will require us to go beyond the bindings factory interface of RDF/JS.

Another question is how to handle joins, where 2 bindings are merged (see the merge functions). We may want to require the caller of the merge function to explicitly say how the contexts of 2 bindings should be merged, as there may not be a generic way that is good for all cases.

maartyman:
In the current version of [incremunica](https://github.com/maartyman/incremunica) (an incremental version of comunica that can process both additions and deletions) I added a [diff boolean](https://github.com/maartyman/incremunica/blob/25bb0d0c9ab29ca8012c220c18b6eab0daec3896/packages/bindings-factory/lib/Bindings.ts#L10) to the Bindings class to define if a Binding has been deleted or added. A context field for the bindings as described here could perfectly allow for this functionality without the need to change this core comunica package.

[...]

PR #1240, opened 2023-08-14
Add context to bindings and allow specification of functions that merge specific entries of context during joins

RubenEschauzier:
This is a draft PR for adding an `ActionContext` to each binding. This PR would solve issue #1203. This context can contain for example the data source of the specific binding. During the join of one binding with another, we merge the binding contexts based on merge handler functions that need to be specified in the Comunica config.

The merge change is located [here](https://github.com/RubenEschauzier/comunica/blob/520d880bedfe5330172f0d723f48ae65f2b24ae8/packages/bindings-factory/lib/Bindings.ts#L130).

Currently, by default we create the set union function for the key 'source' in the binding context, however, this should probably be removed for default Comunica.
An example of how merge handlers are created can be found [here](https://github.com/RubenEschauzier/comunica/blob/520d880bedfe5330172f0d723f48ae65f2b24ae8/packages/actor-init-query/lib/QueryEngineBase.ts#L189C6-L189C6), with the new bus and an example actor for creating merge handlers [here](https://github.com/RubenEschauzier/comunica/blob/520d880bedfe5330172f0d723f48ae65f2b24ae8/packages/bus-merge-binding-factory/lib/ActorMergeBindingFactory.ts) and [here](https://github.com/RubenEschauzier/comunica/blob/520d880bedfe5330172f0d723f48ae65f2b24ae8/packages/actor-merge-binding-factory-context-union/lib/ActorMergeBindingFactoryContextUnion.ts) respectively.

Because the merge function is performance critical, I benchmarked the new and old merge functions using the following code: [standard](https://github.com/RubenEschauzier/Bindings-merge-standard-benchmark) and [new](https://github.com/RubenEschauzier/Binding-context-merge-benchmark). We measure the number of function executions per second by executing the function 1.000.000 times with 20 replications. Non-overlapping bindings means that the two merged bindings have no common variables. The output is as follows. 
Standard Comunica: 
```
Overlapping bindings: 335079.43 ops/s (18451.50 std)
Non-overlapping bindings: 235695.00 ops/s (9769.74 std)
```

New merge function:
```
Non-overlapping bindings, overlapping context entries: 139930.62 ops/s (4047.23 std)
Overlapping bindings, overlapping context entries: 171073.40 ops/s (7390.75 std)
Non-overlapping bindings, no context: 225660.58 ops/s (6854.23 std)
Overlapping bindings, no context: 327583.13 ops/s (7658.40 std)
```
The performance in the case of no context is especially interesting, as this will be the most common configuration. From the benchmark, we see a 4.35% decrease when merging non-overlapping bindings and a 2.26% decrease for overlapping bindings. In both cases, the difference is statistically significant, though, for the overlapping bindings, the p-value is only about 0.05.

Currently, the implementation is lacking a few things. 
1. We currently don't expose a way to set the context of a binding.
2. The keys are not predefined in Keys.ts, as I'm unsure how to deal with predefined Keys and creating merge handlers in runtime.

I will work on these issues, but figured I should already open a draft PR so possible optimizations can be discussed and if the requirements for the binding context are met.
@maartyman, @surilindur


packages/bindings-factory/lib/Bindings.ts
```diff
@@ -8,12 +12,20 @@ import { bindingsToString } from './bindingsToString';
...
+  public context: IActionContext;
+  private readonly contextMergeHandlers: Record<string, IMergeHandler<any>>;
+
   private readonly dataFactory: RDF.DataFactory;
   private readonly entries: Map<string, RDF.Term>;
 
-  public constructor(dataFactory: RDF.DataFactory, entries: Map<string, RDF.Term>) {
+  public constructor(dataFactory: RDF.DataFactory, entries: Map<string, RDF.Term>,
+    contextMergeHandlers: Record<string, IMergeHandler<any>>,
+    context: IActionContext = new ActionContext()) {
```
rubensworks, line 23: This is here a major contributor to performance decrease.
Instead of defaulting the context to a newly-constructed object, let's allow it to be undefined.
Then we can cheaply do things like `if(this.context)` later in this file.

RubenEschauzier: You're right, I added this to the PR with the implementation of setting, getting, and deleting binding context entries. This does remind me to do a full benchmark to make sure I've not made any mistakes that are not in the merge function. Will get back to you with those results.

rubensworks:
I think you accidentally rebased on `master`, instead of on `comunica:next/major`.
Could you cherry-pick your commits on a new branch starting from `comunica:next/major`?

@jitsedesmet recently did something similar, so he can assist you if needed.

RubenEschauzier:
Yes, this should be ok in the other PR. I guess this one can be closed?

rubensworks:
Closing in favor of #1284

PR #1262, opened 2023-09-12
Make ESLint rule overrides more specific and re-enable a number of rules

surilindur:
Here is a small set of changes to make the ESLint rule overrides more specific, because the overrides are used to disable rules and it might make sense to disable then with the smallest possible scope. I have also added comments for what triggers the rules to make it easier to understand why the overrides exist.

There are also a couple of code changes that help make some rule overrides unnecessary:

* Replacing the use of array joins for string repetition with the repeat function from string allows re-enabling the `unicorn/no-new-array` rule.
* Making sure the imports in two files are at the top of the files allows re-enabling the `import/first` rule. This was done by moving some imports to the top of one file, and by removing an unmock workaround that did not appear to have any impact on the tests in another file.

Any feedback is welcome. I tried to keep the number of changes to a minimum without splitting them up too much, because they are all related to ESLint rules.

Re-enabling `unicorn/prefer-string-replace-all` will require changing the tsconfig to bump the target to es2021, so I will make another pull request for that later.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,7 +6,7 @@ module.exports = {
     project: ['./tsconfig.eslint.json'],
   },
   extends: [
-      '@rubensworks'
+    '@rubensworks',
   ],
 
   // TODO: Remove this once solid-client-authn supports node 18.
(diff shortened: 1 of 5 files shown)


rubensworks:
Thanks, looks perfect!

PR #1263, opened 2023-09-12
Update actions/cache to v3, use --frozen-lockfile instead of --pure-lockfile, cancel old workflow when new one is queued

surilindur:
Here is a small set of changes that I was thinking might make sense, but that will need some discussion I feel. There are three changes:

1. Update `actions/cache` to v3, because v2 used Node 12 and that has been deprecated. The new one seems to work the same way as the previous one.
2. Use `--frozen-lockfile` instead of `--pure-lockfile`, because the pure option only skips generating a lockfile, whereas the frozen one ensures that the lockfile and package.json are synced by failing if any lockfile changes would result from the installation according to [the documentation](https://classic.yarnpkg.com/lang/en/docs/cli/install/).
3. Always run linter first, then tests, and then everything else. I feel like this would make sense, because linting should be the cheapest and fastest to do, and since linter errors will need to be fixed anyway, it makes sense to catch them first. Making sure the tests pass before building Dockerfiles or running browser tests feels intuitive, as well. Running TypeDoc takes half an hour, so I also feel like it would be better to do it only after the tests have passed.

Any feedback would be welcome, especially with regards to changing the order of the workflow. The changes make sense to me, but I am not sure if everyone would agree. I will make this a draft for now.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,10 @@
 name: "CI"
 on: [push, pull_request]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
 
   test:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@v3
       - name: Load cache
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           path: |
             **/node_modules
@@ -77,7 +77,7 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@v3
       - name: Load cache
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           path: '**/node_modules'
           key: ${{ runner.os }}-lint-modules-${{ secrets.CACHE_VERSION }}-${{ hashFiles('**/yarn.lock') }}
@@ -99,7 +99,7 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@v3
       - name: Load cache
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           path: '**/node_modules'
           key: ${{ runner.os }}-browser-modules-${{ secrets.CACHE_VERSION }}-${{ hashFiles('**/yarn.lock') }}
@@ -126,7 +126,7 @@ jobs:
       - name: Check out repository
         uses: actions/checkout@v3
       - name: Load cache
-        uses: actions/cache@v2
+        uses: actions/cache@v3
         with:
           path: '**/node_modules'
           key: ${{ runner.os }}-docker-modules-${{ secrets.CACHE_VERSION }}-${{ hashFiles('**/yarn.lock') }}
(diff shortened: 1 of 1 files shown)


surilindur:
There is now also an added concurrency section to the workflow that should allow cancelling in-progress workflows when new commits are pushed, according to the [documentation](https://docs.github.com/en/actions/using-jobs/using-concurrency). There is also one in the [CSS](https://github.com/CommunitySolidServer/CommunitySolidServer/blob/main/.github/workflows/main.yml#L11-L13).

surilindur:
It was pointed out to me that the total time taken to finish the CI by having it do everything in sequence will be higher, which is true. But I was thinking it would make sense to avoid starting a long-running jobs (like the docs one) before making sure that the faster checks all pass. I think it would help avoid unnecessary long-running jobs.

The reason being that the docs task, if it is not connected to anything else, will seemingly keep running even when linting or unit tests fail, which feels unnecessary.

.github/workflows/ci.yml
```diff
@@ -1,8 +1,14 @@
...
 on: [push, pull_request]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
 
   test:
+    needs: lint
```
rubensworks, line 11: Let's not wait for that, so we can reduce absolute wait time.

.github/workflows/ci.yml
```diff
@@ -77,19 +83,17 @@ jobs:
...
-        run: yarn install --pure-lockfile
+        run: yarn install --frozen-lockfile
       - name: Run linter
         run: yarn run lint
 
   browser:
-    needs:
-      - test
-      - lint
+    needs: test
```
rubensworks, line 96: This change can then also be undone.

.github/workflows/ci.yml
```diff
@@ -114,9 +118,7 @@ jobs:
           GH_TOKEN: ${{ secrets.PAT }}
 
   docker:
-    needs:
-      - test
-      - lint
+    needs: test
```
rubensworks, line 121: This as well.

.github/workflows/ci.yml
```diff
@@ -144,6 +146,7 @@ jobs:
           DOCKER_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
 
   docs:
+    needs: test
```
rubensworks, line 149: Very important: docs need to run asap, since they are very slow, so we want to kickstart them asap. If one of the jobs fail, docs will be killed anyways.

rubensworks (changes requested):
I agree with some of these changes, but not all :-)

rubensworks:
> The reason being that the docs task, if it is not connected to anything else, will seemingly keep running even when linting or unit tests fail, which feels unnecessary.

GH will kill the docs job once another task fails.

surilindur:
> GH will kill the docs job once another task fails.

If this does happen then it is fine. But taking 30+ minutes to build the docs feels excessive. Maybe there are some ways to make it faster. :thinking: 

I have removed the changes to the ordering of the workflow jobs, so now there are only the ones for yarn flags, action versions and workflow cancellation left.

rubensworks:
> If this does happen then it is fine. But taking 30+ minutes to build the docs feels excessive. Maybe there are some ways to make it faster. 🤔

I agree. It's a consequence of the last tsdoc update. It used to be much faster. I wasn't able to find any ways to make it faster myself, but feel free to give it a shot.

rubensworks (approved):
Looks good!

PR #1267, opened 2023-09-13
test: general concepts of expression evaluator

jitsedesmet:
This PR aims to close the last issue on the sparqlee repo, being https://github.com/comunica/sparqlee/issues/35. I have made a [report of my findings](https://github.com/comunica/sparqlee/issues/35#issuecomment-1717463412) on that issue. 

If we wish to add the tests for the exists and aggregate hooks, we should do so in this PR.

### DIFF
--- a/packages/expression-evaluator/test/integration/functions/op.logicalOr-test.ts
+++ b/packages/expression-evaluator/test/integration/functions/op.logicalOr-test.ts
@@ -23,7 +23,6 @@ describe('evaluation of "||" like', () => {
       true  invalidDateTime = true
       invalidDateTime true  = true
     `,
-    // TODO: why?
     errorTable: `
       false invalidDateTime = 'Cannot coerce term to EBV'
       invalidDateTime false = 'Cannot coerce term to EBV'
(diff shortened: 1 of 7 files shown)


jitsedesmet:
While working on https://github.com/comunica/comunica/pull/1269 I discovered there is already a spec test testing the exists hook. While working on that same PR. I am refactoring things in a way that I don't think the aggregate hook will be needed. I therefore suggest leaving these 2 cases for what they are?

For the aggregate hook: I don't feel like spending more time on figuring out when it could be called since the original documentation also states:
```
We have not found any SPARQL Algebra for which this occurs, but we happen to find any aggregate expressions nested in the expression (or even at the top level), we will call (similarly to EXISTS) an aggregate hook you might have provided.

aggregate?: (expression: Alg.AggregateExpression) => Promise<RDF.Term>;

You can probably ignore this.
```
https://github.com/comunica/comunica/pull/1269 removes sync evaluators, a spilt will thus no longer be needed. Aggregates are also transferred to be actors themselves. A rework on how the hook would work is thus in place.

packages/expression-evaluator/test/integration/functions/op.logicalOr-test.ts
```diff
@@ -22,6 +23,7 @@ describe('evaluation of "||" like', () => {
       true  invalidDateTime = true
       invalidDateTime true  = true
     `,
+    // TODO: why?
```
rubensworks, line 26: TODO

rubensworks (changes requested):
Aside from the TODO, everything looks good!

rubensworks:
> I therefore suggest leaving these 2 cases for what they are?

👌

PR #1268, opened 2023-09-14
GitHub Actions workflow ref test, do not merge

surilindur:
This is to test how the ref works exactly.


surilindur:
So the ref points at this specific pull request and not the branch, awesome. Will close this.

PR #1270, opened 2023-09-14
Fix/allow es module interop flag

RubenEschauzier:
By changing the default imports, Comunica can be compiled (and run) using the esModuleInterop: true flag. This in turn makes linking development versions of Comunica to a development version of, for example, [comunica-feature-link-traversal](https://github.com/comunica/comunica-feature-link-traversal) possible, as the link-traversal engine uses this flag.

(The .yarnrc.yml file again snuck in, let me know if you would rather see it gone)


packages/actor-dereference-file/test/ActorDereferenceFile-test.ts
```diff
@@ -5,9 +5,10 @@ import type { IActorDereferenceOutput } from '@comunica/bus-dereference';
 import { ActorDereference } from '@comunica/bus-dereference';
 import { ActionContext, Bus } from '@comunica/core';
 import type { IActionContext } from '@comunica/types';
-import * as streamToString from 'stream-to-string';
 import { ActorDereferenceFile } from '../lib/ActorDereferenceFile';
 
+const streamToString = require('stream-to-string');
```
rubensworks, line 10: Could you add the following comment to each location that we replaced with `require`?
`// Use require instead of import for default exports, to be compatible with variants of esModuleInterop in tsconfig.`

rubensworks (changes requested):
Looks great!

Could you also add a second build step after [this one](https://github.com/comunica/comunica/blob/master/.github/workflows/ci.yml#L42-L43) in the CI workflow file, so we build once without the `esModuleInterop` flag (default), and once with that flag enabled? (I think you can just pass it as arg to `yarn run build`)

RubenEschauzier:
Right! Forgot about that one, will add both suggestions

packages/actor-init-query/lib/ActorInitQuery.ts
```diff
@@ -4,14 +4,14 @@ import type { IActionInit, IActorOutputInit } from '@comunica/bus-init';
...
 import type { Readable } from 'readable-stream';
-import yargs from 'yargs';
 import type { IActorInitQueryBaseArgs } from './ActorInitQueryBase';
 import { ActorInitQueryBase } from './ActorInitQueryBase';
 import { CliArgsHandlerBase } from './cli/CliArgsHandlerBase';
 import { CliArgsHandlerQuery } from './cli/CliArgsHandlerQuery';
 import { QueryEngineBase } from './QueryEngineBase';
 
 const streamifyString = require('streamify-string');
+const yargs = require('yargs');
```
surilindur, line 14: This change is unnecessary I think, since the error can be fixed by adjusting the constructor argument. It looks like a typing issue, and converting it to require like this only makes the type `any` which works around the issue but loses types. I can make another PR for this.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -13,20 +11,20 @@ import { ActionContext } from '@comunica/core';
...
 // eslint-disable-next-line import/no-useless-path-segments
 import type { IDynamicQueryEngineOptions } from '..';
 import { CliArgsHandlerBase } from './cli/CliArgsHandlerBase';
 import { CliArgsHandlerHttp } from './cli/CliArgsHandlerHttp';
 
+const clusterUntyped = require('cluster');
 const process: NodeJS.Process = require('process/');
-
 const quad = require('rdf-quad');
+const yargs = require('yargs').default;
```
surilindur, line 24: This also causes yargs to lose its types, and can be fixed by adjusting the constructor arguments based on my testing (same as the comment above). Other changes are fine I think.

packages/actor-init-query/lib/ActorInitQuery.ts
rubensworks: What would need to change exactly? If you can share that, @RubenEschauzier can try it on his end, and already include it in this PR.

surilindur: Changing the constructor argument to `[]` seems to fix it, with maybe also bumping the types version to the latest one. The functionality seems to remain unchanged.

rubensworks: Can you try this @RubenEschauzier?

Also important: we don't have unit tests for the CLI tool, so could you have a quick manual check if `engines/query-sparql/bin/query.js` still seems to work after changing this?

surilindur: The error seems to be this:
```
packages/actor-init-query/lib/ActorInitQuery.ts:42:34 - error TS2345: Argument of type '{}' is not assignable to parameter of type 'string | readonly string[] |undefined'.
Type '{}' is not assignable to type 'undefined'.

42     let argumentsBuilder = yargs({});
```

RubenEschauzier: I could add the constructor changes to this PR if that is smarter?

rubensworks: If that works, why not.

RubenEschauzier: The changes are added and I tested it with a manual check and it seems to still work.

RubenEschauzier:
> Looks great!
> 
> Could you also add a second build step after [this one](https://github.com/comunica/comunica/blob/master/.github/workflows/ci.yml#L42-L43) in the CI workflow file, so we build once without the `esModuleInterop` flag (default), and once with that flag enabled? (I think you can pass it as arg to `yarn run build`)

I added the build with the esModuleInterop flag, but I couldn't get it to work as an arg to yarn run build since afaik any argument you pass to it will be appended to the script, which would pass it as an argument to npm run build:components.

surilindur:
The install step in the CI already builds everything with the defaults from tsconfig, as it would do locally, and runs lerna prepare - even the linting job builds everything after installing the dependencies and runs lerna prepare - so the current 'build' step is actually unnecessary. Maybe it could be reused or modified for the esModuleInterop version?

rubensworks:
> but I couldn't get it to work as an arg to yarn run build since afaik any argument you pass to it will be appended to the script,

You can run `yarn run build:ts` with the passed args instead.

And also what @surilindur said.

RubenEschauzier:
Ah right, I'll update it. And if I understand correctly, you mean removing this [step](https://github.com/comunica/comunica/blob/6a1a5feb21857278a08283b9bb26aeff58f5ace1/.github/workflows/ci.yml#L42) and replacing it with the building it with esModuleInterop enabled?

rubensworks:
> And if I understand correctly, you mean removing this [step](https://github.com/comunica/comunica/blob/6a1a5feb21857278a08283b9bb26aeff58f5ace1/.github/workflows/ci.yml#L42) and replacing it with the building it with esModuleInterop enabled?

Yep

rubensworks:
Could you rebase on master? Thanks to @surilindur, we have some merge conflicts :-)

RubenEschauzier:
I think this should do it? Not sure though since it still says conflicts must be resolved.

RubenEschauzier:
Yep, it actually was an unused require, so I removed it.

rubensworks:
Super, thanks!

PR #1271, opened 2023-09-15
Update readable-stream, yargs, lerna, typescript and eslint

surilindur:
I do not know if this is worth its own PR, but to resolve some local linking issues I had to make sure some packages consistently use the latest versions everywhere. This is not urgent and can wait until other more urgent PRs have been merged.

Any feedback is welcome again.


rubensworks:
Not really necessary I think, but also doesn't hurt.

PR #1282, opened 2023-10-23
Enforce `bindingsToString` key ordering with a unit test

surilindur:
The `bindingsToString` function produces a record where the keys are not in sorted order, but rather in the order in which they were assigned to the bindings object. Therefore, a bindings object that has identical keys and values but in different order will serialize into a different string.

This change adds a unit test to make sure that happens consistently and also makes sure it is made explicit by having a test for it - otherwise someone might assume that the function produces the same string for a bindings object with the same keys and values.

Any feedback is welcome. :slightly_smiling_face:


surilindur:
I also added a test to ActorHashBindingsSha1 unit tests to make sure the hash it produces does not depend on the order of the entries in the bindings object.

rubensworks:
Thanks! Additional unit tests are always welcome :-)

Issue #877, opened 2021-10-25
Implement sort-merge join


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

If bindings streams are sorted in (partially) the same manner, then we can use the more efficient sort-merge join algorithm.

Some ideas:
* Return used order in N3Store if applicable
* Implement an N3Store-like storage that provides all indexes? (quadstore may already provide this? /cc @jacopo)
* When querying SPARQL endpoints => use ORDER BY if needed
* When querying HDT files, check returned order
* If possibly, try to preserve order throughout the UNION operator, which can be done by implementing a dedicated union join actor that only applies if join orders are equal, and does a sort-merge join with union semantics that preserves order.
* This will require extensions to the RDF/JS query spec: https://github.com/rdfjs/query-spec/issues/8

For even better performance, we should look into dictionary-encoding terms, as is done in RDF-3X (#873).

PR #1283, opened 2023-10-23
Chore/prepare function bussification

jitsedesmet:
This PR aims to prepare the expression evaluator package for function bussification.


jitsedesmet:
The `OverloadTree` need to get a `E.TermExpression[]` to perform a search. Because of this, only regular functions will be able to use the `overloadTree`  because of this I suggest we just keep a destination between functions that  defining themselves based on the `Overloadtree` and functions that do not do this. (They should both have the same end interface)

jitsedesmet:
I suggest making the functions all classes (they already get transformed to become objects in `functions/index`. We can do the same by creating a class for all of them and just instantiating the classes in the same way as done in `functions/index`.

jitsedesmet:
The latest implementation might have unbounded this in the functions

jitsedesmet:
@rubensworks would we want to keep extensionFunction support like we do now? Maybe we can  do that using a functionActor that just calls that config callback?

jitsedesmet:
Think about util functions and put them in a bus if they are only required in that bus

jitsedesmet:
The extension functions should still work. maybe we can have a "special function" look at the context and just call to it.

jitsedesmet:
We want all a function-actors to be on single bus.

packages/expression-evaluator/lib/evaluators/ContextualizedEvaluator.ts
```diff
@@ -0,0 +1,147 @@
...
+  public readonly extensionFunctionCreator?: AsyncExtensionFunctionCreator;
+  public readonly now: Date;
+  public readonly baseIRI?: string;
+  public readonly functionArgumentsCache: FunctionArgumentsCache;
+  public readonly superTypeProvider: ISuperTypeProvider;
+  public readonly defaultTimeZone: ITimeZoneRepresentation;
+  public readonly actionContext: IActionContext;
+  public readonly mediatorQueryOperation: MediatorQueryOperation;
+  public readonly mediatorTermFunction: TermFunctionBusType;
+  public readonly mediatorFunction: FunctionBusType;
```
jitsedesmet, line 63: @rubensworks I already mentioned for functions like `=` we need the sync version. (I call this a `termFunction` because the function executes over evaluated terms). I suggest having 2 busses like this. Has a subgroup of the actors of the other one. (This means that all functionActors are registered on the `functionBus` and some of them are also registered on the `termFunctionBus`.) Do you think this is okey, or do you have a better way in mind?)

rubensworks: > I already mentioned for functions like = we need the sync version. (I call this a termFunction because the function executes over evaluated terms)

Doesn't both async and sync both evaluate over terms?

> I suggest having 2 busses like this. Has a subgroup of the actors of the other one. (This means that all functionActors are registered on the functionBus and some of them are also registered on the termFunctionBus.) Do you think this is okey, or do you have a better way in mind?)

Sounds a bit confusing.
Would it be doable to have one bus, but let the IAction interface define `sync: boolean` as a parameter?
This would allow some actors (the async ones) to reject actions that require sync evaluation.

This would also solve the naming issue above. All of them would just be functions then.

jitsedesmet: So the difference is whether the arguments still need to be evaluated. Each function has an `apply` function that gets `args: E.Expression[]`. `E.Expression[]` can contain operations that would need an async evaluation. Some functions have a definition on terms (these used to be called `regularFunctions`, and are the ones defined using an overloadTree). Those functions have the same `apply` function and an additional `applyOnTerms` function that takes `args: E.TermExpression[]`. 

That being said, I would not refer to them as being sync or async. But rather also having the option of passing terms directly.

(Note: this requires a shift away from thinking functions get executed on terms. They mostly do, but not all, some need control over the evaluation of their arguments (used to call `specialFunctions`). The ones that don't need control over the evaluation are just a special case of function. The control flow is just way prettier defining functions this way).

jitsedesmet: The name of the flag could be something like `evaluatesOnTerm`? Or something else?

rubensworks: I'm not really following to be honest. Can you link me the code towards these two different function interfaces?

packages/expression-evaluator/lib/evaluators/AggregateEvaluator.ts
```diff
@@ -10,25 +9,21 @@ import { isSubTypeOf } from '../util/TypeHandling';
 import type { ExpressionEvaluator } from './ExpressionEvaluator';
 
 /**
- * Abstract aggregator actor. This is the base class for all aggregator actors.
- * Only the wildcard count aggregator significantly differs from the others.
+ * This is the base class for all aggregators.
+ * NOTE: The wildcard count aggregator significantly differs from the others and overloads parts of this class.
  */
 export abstract class AggregateEvaluator {
```
rubensworks, line 15: I guess we can move this one to the aggregator bus?

packages/expression-evaluator/lib/evaluators/ExpressionEvaluator.ts
```diff
@@ -1,195 +1,21 @@
...
-  baseIRI?: string;
-  typeCache?: TypeCache;
-  getSuperType?: SuperTypeCallback;
-  functionArgumentsCache?: FunctionArgumentsCache;
-  defaultTimeZone?: ITimeZoneRepresentation;
-  actionContext: IActionContext;
-}
+import type { ContextualizedEvaluator } from './ContextualizedEvaluator';
 
 export class ExpressionEvaluator implements IExpressionEvaluator {
```
rubensworks, line 6: This class doesn't seem to be doing much on its own. Is there a reason why `ContextualizedEvaluator` can't just be merged into this one?

packages/expression-evaluator/lib/evaluators/ExpressionEvaluatorFactory.ts
```diff
@@ -1,42 +1,73 @@
...
-      actionContext: context,
-      bnode: (input?: string) => Promise.resolve(new BlankNodeBindingsScoped(input || `BNODE_${bnodeCounter++}`)),
-      exists: ActorQueryOperation.createExistenceResolver(context, this.mediatorQueryOperation),
-      ...legacyContext,
-    });
+  public async createEvaluator(algExpr: Alg.Expression, context: IActionContext,
+    contextEval?: ContextualizedEvaluator,
+    legacyContext: Partial<IAsyncEvaluatorContext> = {}): Promise<IExpressionEvaluator> {
+    if (!contextEval) {
+      const defContextEval = new ContextualizedEvaluator({
```
rubensworks, line 58: I guess this wouldn't be needed anymore if `ContextualizedEvaluator` gets merged into `ExpressionEvaluator`.

packages/expression-evaluator/lib/evaluators/OrderByEvaluator.ts
```diff
@@ -0,0 +1,109 @@
+import type * as RDF from '@rdfjs/types';
+import type * as E from '../expressions';
+import type { RegularFunction } from '../functions';
+import type { IAsyncEvaluatorContext } from './ContextualizedEvaluator';
+import { ContextualizedEvaluator } from './ContextualizedEvaluator';
+
+export class OrderByEvaluator extends ContextualizedEvaluator {
```
rubensworks, line 7: Perhaps we can already design towards bus-ifying it later on?
I remember us discussing there being a comparator factory bus, with this being a term-comparator.

In that sense, _extending_ `ContextualizedEvaluator` might not be a good approach.
Instead, composition might be better, i.e., just saving it as internal field.

rubensworks (changes requested):
As requested, here's a first review of the `evaluators/` directory.

packages/expression-evaluator/lib/evaluators/ContextualizedEvaluator.ts
jitsedesmet: https://github.com/comunica/comunica/pull/1283/files#diff-99c32e329f466d175ad0920b234c18090aeb77c4a82bd7cda67f80070a9f4727R14 and https://github.com/comunica/comunica/pull/1283/files#diff-99c32e329f466d175ad0920b234c18090aeb77c4a82bd7cda67f80070a9f4727R51

packages/expression-evaluator/lib/evaluators/ExpressionEvaluator.ts
jitsedesmet: Yes! So our translation function is now async (translating means we find the correct function definition for an operator, which means we need a mediator call. A constructor can not be async, so I decided to solve that issue this way. Another way could be using an async `initialize` function.

packages/expression-evaluator/lib/evaluators/OrderByEvaluator.ts
jitsedesmet: ah yes! I forgot about that one. Sure!

packages/expression-evaluator/lib/evaluators/ExpressionEvaluator.ts
rubensworks: > So our translation function is now async (translating means we find the correct function definition for an operator, which means we need a mediator call. A constructor can not be async

I don't understand the relationship between this ^ and the `ContextualizedEvaluator`.

Just blindly moving everything from inside of the `ContextualizedEvaluator` towards this class wouldn't change anything, right?

packages/expression-evaluator/lib/evaluators/ContextualizedEvaluator.ts
rubensworks: Can you clarify very precisely why this difference exists? (I can't derive it from the jsdoc)
I see both have `exprEval` and `exprEval`, but only one has a `mapping`.
So is the main difference that one has a mappings context while the other has not?
Should this difference really exist? I.e., would it hurt if the other would _also_ have a mappings context, but just not use it?
If this is really needed, then this mappings context could be an inspiration for a name?

packages/expression-evaluator/lib/evaluators/ExpressionEvaluator.ts
jitsedesmet: Hmm, so before evaluation, you need to translate the `Algebra.Expressions` to `Sparqlee.Expressions`, this means that the operators should be translated. You do not want to translate each time you evaluate. Comunica wants to evaluate the same Algebra but with different bindings. What I try to do here is create an easy API that translates the Algebra within the constructor. (Which wouldn't be possible when just merging the files because constructors can not be async)

packages/expression-evaluator/lib/evaluators/ContextualizedEvaluator.ts
jitsedesmet: The difference is some functions (e.g. `or`) need explicit control over the evaluation of their arguments. This means that they for example ignore exceptions thrown, or see exceptions as "false", or they need a lazy evaluation approach, ....

Other functions do not need this control. The functions that do not need that control can (mostly) be sync. What I suggest is that a function that doesn't care about the evaluation is just a special case of a function that does care. So, in principle both need the mappings, but one is just defined in a way where it doesn't care for argument evaluation and thus in its definition doesn't need the mapping. (And since we hold on to that definition, we can refer to it)

packages/expression-evaluator/lib/evaluators/ExpressionEvaluator.ts
rubensworks: Ok, I understand the need now.

So the main purpose of `ContextualizedEvaluator` is to translate between the two expression types, which is an async operation.
Would it make sense to call it something like `ExpressionTranslator`?

I'm not too sure yet if the following methods really make sense to be in that class:

- `evaluateAsInternal`
- `evalFunction`
- `evalExistence`
- `evalAggregate`

Should they be moved into `ExpressionEvaluator`? Or do they need to be moved to a new class?

packages/expression-evaluator/lib/evaluators/ContextualizedEvaluator.ts
rubensworks: > The functions that do not need that control can (mostly) be sync. 

So not all of them are sync?
Or do they?
If they are, the sync/async distinction still makes sense to me :-)

In any case, I would keep them in the same bus, and use some distinguishing flag in case we require only one type.
This flag name is something we can still discuss :-)

packages/expression-evaluator/lib/evaluators/ExpressionEvaluator.ts
jitsedesmet: > So the main purpose of ContextualizedEvaluator is to translate between the two expression types, which is an async operation.

Euhm, no :sweat_smile: . `ContextualizedEvaluator` is more of an `InternalEvaluator` it is able to evaluate something that has already been translated.

jitsedesmet:
Alright, big refactor again. I think the control flow is as good as final. Names should definitely get fixed up! (It's truly a mess.) The next thing is to fix the tests. If that is done, this PR can be opened. @rubensworks if you ever have too much time on your hand, you are welcome to review the control flow of this PR already. If you come across something, and you feel like it's a bad name, please point it out and if possible, suggest an alternative :sweat_smile: .  Fixing the tests will likely take some time.

packages/actor-bindings-aggregator-factory-average/lib/AverageAggregator.ts
```diff
@@ -33,8 +32,8 @@ export class AverageAggregator extends AggregateEvaluator implements IBindingsAg
       this.state = { sum, count: 1 };
     } else {
       const internalTerm = this.termToNumericOrError(term);
-      this.state.sum = <E.NumericLiteral> this.summer.apply([ this.state.sum, internalTerm ],
-        <ExpressionEvaluator> this.evaluator);
+      this.state.sum = <E.NumericLiteral> this.additionFunction.applyOnTerms([ this.state.sum, internalTerm ],
+        (<ExpressionEvaluator> this.evaluator).internalizedExpressionEvaluator);
```
rubensworks, line 36: This feels a bit icky. Would it be feasible to avoid the cast? Perhaps we can just pass the IEvaluator interface?

packages/actor-bindings-aggregator-factory-average/test/AverageAggregator-test.ts
```diff
@@ -13,6 +13,27 @@ async function runAggregator(aggregator: IBindingsAggregator, input: RDF.Binding
...
 
+async function createAggregator({ expressionEvaluatorFactory, context, distinct }: {
+  expressionEvaluatorFactory: IExpressionEvaluatorFactory;
+  context: IActionContext;
+  distinct: boolean;
+}): Promise<AverageAggregator> {
+  return new AverageAggregator(
+    await expressionEvaluatorFactory.createEvaluator(makeAggregate('avg', distinct).expression, context),
+    distinct,
+    <ITermFunction> await expressionEvaluatorFactory.createFunction({
```
rubensworks, line 24: Isn't there a way to return `ITermFunction` from `createFunction` so the cast is not needed?

packages/expression-evaluator/lib/evaluators/ExpressionEvaluatorFactory.ts
```diff
@@ -47,6 +96,13 @@ export class ExpressionEvaluatorFactory implements IExpressionEvaluatorFactory {
       context,
     })).aggregator;
   }
+
+  public createFunction = this.functionsBus;
+
+  public async createOrderByEvaluator(orderAction: IOrderByBusActionContext):
```
rubensworks, line 102: `createTermComparator`?

packages/expression-evaluator/lib/evaluators/ContextualizedEvaluator.ts
```diff
@@ -0,0 +1,136 @@
...
+  public readonly now: Date;
+  public readonly baseIRI?: string;
+  public readonly functionArgumentsCache: FunctionArgumentsCache;
+  public readonly superTypeProvider: ISuperTypeProvider;
+  public readonly defaultTimeZone: ITimeZoneRepresentation;
+  public readonly actionContext: IActionContext;
+  public readonly mediatorQueryOperation: MediatorQueryOperation;
+  public readonly mediatorFunction: FunctionBusType;
+
+  public constructor(context: IAsyncEvaluatorContext) {
```
rubensworks, line 61: Since this class essentially materialized the evaluator context, shall we rename it to `MaterializedEvaluatorContext`?

packages/expression-evaluator/lib/evaluators/ExpressionEvaluatorFactory.ts
```diff
@@ -1,42 +1,91 @@
...
-      now: context.get(KeysInitQuery.queryTimestamp),
-      baseIRI: context.get(KeysInitQuery.baseIRI),
-      functionArgumentsCache: context.get(KeysInitQuery.functionArgumentsCache),
-      actionContext: context,
-      bnode: (input?: string) => Promise.resolve(new BlankNodeBindingsScoped(input || `BNODE_${bnodeCounter++}`)),
-      exists: ActorQueryOperation.createExistenceResolver(context, this.mediatorQueryOperation),
-      ...legacyContext,
-    });
+  public async createEvaluator(algExpr: Alg.Expression, context: IActionContext,
+    contextEval?: ContextualizedEvaluator,
```
rubensworks, line 74: Can/should we enforce `contextEval` to be defined?
So we can avoid cases where we are continuously recreating `ContextualizedEvaluator` due to an implementation oversight?

packages/types/lib/ExpressionEvaluator.ts
```diff
@@ -57,11 +65,56 @@ export interface IExpressionEvaluator {
...
+export interface IExpressionFunction {
+  definitionType: 'onExpression';
+  apply: (evalContext: IEvalContext) => Promise<E.TermExpression>;
+  /**
+   * Makes you able to error in the termTransformer.
+   */
+  checkArity: (args: E.Expression[]) => boolean;
+}
+
+export interface ITermFunction {
```
rubensworks, line 99: Since an `ITermFunction` also appears to be an `IExpressionFunction`, perhaps we can let the former extend the latter?
The latter could then just have a boolean field `supportsTermExpressions: true`.

packages/types/lib/ExpressionEvaluator.ts
```diff
@@ -57,11 +65,56 @@ export interface IExpressionEvaluator {
...
+   * Makes you able to error in the termTransformer.
+   */
+  checkArity: (args: E.Expression[]) => boolean;
+  applyOnTerms: (args: E.TermExpression[], exprEval: ContextualizedEvaluator) => E.TermExpression;
+}
+
+export interface IFunctionBusActionContext {
+  functionName: string;
+  arguments?: Alg.Expression[];
+  definitionType?: 'onExpression' | 'onTerm';
```
rubensworks, line 112: Related to above, we could make this an optional flag: `requireTermExpression`

packages/actor-bindings-aggregator-factory-average/lib/AverageAggregator.ts
jitsedesmet: The more I think abut how to solve this, the less I feel like we should have an interface like `IExpressionEvaluator`. We need to access to context here, if I describe that context in an interface, we lose all generality the interface would bring? At that point, we just have the added complexity of an interface without any gains?

packages/expression-evaluator/lib/evaluators/ExpressionEvaluatorFactory.ts
jitsedesmet: I don't know why I had that argument anymore. It wasn't used. It's not expensive to recreate either.

jitsedesmet:
I think these first 2 if statements are either redundant, or we have a bug: https://github.com/jitsedesmet/comunica/blob/chore/prepare-function-bussification/packages/expression-evaluator/lib/functions/RegularFunctions.ts#L329-L336.

The statements are also included in the spec: https://w3c.github.io/rdf-star/cg-spec/2021-12-17.html#sparql-compare.

The statements call the `<` operator on subjects and predicates. These are however [blank nodes or IRIs](https://www.w3.org/TR/rdf11-concepts/#dfn-subject). Since sparql `<` can not be evaluated on a named node comunica throws an exception. NOTE that this behavior even has a test: https://github.com/jitsedesmet/comunica/blob/chore/prepare-function-bussification/packages/expression-evaluator/test/integration/functions/op.lesserThan-test.ts#L153.

So either:
1. This behavior is correct, and an error should be thrown (this means we can remove the [undefined check](https://github.com/jitsedesmet/comunica/blob/chore/prepare-function-bussification/packages/expression-evaluator/lib/functions/RegularFunctions.ts#L329-L336.))
2. This behavior is wrong, we want to be able to compare two the IRIs (I don't know for sure, but maybe [type promotion](https://www.w3.org/TR/xpath20/#promotion) should happen here? I don't know, since I don't think I understand the distiction between a named node and a literal that has type `xs:anyURI`)

rubensworks:
@jitsedesmet If I remember correctly, some of the RDF-star spec tests check ordering of things like `<ex:a ex:b "a">` and `<ex:a ex:b "b">`, which requires us to check the IRIs `ex:a` and `ex:b` first (to see they are equal), after which the non-equal objects can be compared.

So based on my understanding, I believe the current implementation is correct. (I could be wrong though!)

jitsedesmet:
@rubensworks yes, but since a predicate is always a named node in comunica, we can not compare it using `<`. Therefore, the [undefined check ](https://github.com/jitsedesmet/comunica/blob/chore/prepare-function-bussification/packages/expression-evaluator/lib/functions/RegularFunctions.ts#L334-L336) would not be needed? (I am looking for a way to get code coverage for that line, but don't think the case can happen) I suggest we calculate the equality there (since it might throw an exception is some cases), but ignoring the result, since it would always be `undefined`?

rubensworks:
@jitsedesmet True, in strict RDF, predicates are always IRIs. However, Comunica aims to be compatible with [generalized RDF](https://www.w3.org/TR/rdf11-concepts/#section-generalized-rdf), in which all triple components can be of any type. So to cope with these edge cases, we need the undefined check I think.

packages/actor-bindings-aggregator-factory-average/lib/AverageAggregator.ts
rubensworks: Based on our last chat, I think bus-ifying this may solve this problem.

jitsedesmet: Fixed by extracting an ActionConfig.

packages/expression-evaluator/lib/evaluators/ExpressionEvaluator.ts
jitsedesmet: @rubensworks I now renamed it to `InternalEvaluator` and extracted the config in ActionContextKeys. From what I can see, that's much prettier. If you agree you can resolve this conversation :)

rubensworks: Looks good!

packages/expression-evaluator/test/integration/evaluators/TermComparator-test.ts
```diff
@@ -0,0 +1,184 @@
...
+    await genericOrderTestLower(decimal('2'), decimal('11'));
+  });
+  it('float type comparison', async() => {
+    await genericOrderTestLower(float('2'), float('11'));
+  });
+  it('dateTime type comparison', async() => {
+    await genericOrderTestLower(dateTime('2000-01-01T00:00:00Z'), dateTime('2001-01-01T00:00:00Z'));
+  });
+  it('langString type comparison', async() => {
+    // Spec does not say anything about order of langStrings, but we use Operator Extensibility to define it.
```
jitsedesmet, line 123: This is a false note! We do not have any operator extensibility in place for languageString `<`! 
I wonder whether we want this? @rubensworks What do you think? Right now (and since a long time), this test passes by accident.

rubensworks: I would keep the same behaviour as we have now as introduced in https://github.com/comunica/sparqlee/pull/139
Do you think this is reasonable?

rubensworks: Don't we want ORDER BY to be equivalent to the `<` operator?
Or do you have a reason in mind why we might not want that?

rubensworks: I'll leave the choice up to you, as long as no current behaviour is broken :-)

jitsedesmet: Okay! Then I won't implement it since I don't think it makes sense in `<`. And for the order by case, there are tests in place, so that should be alright too. Thank you!

jitsedesmet:
Looks like the spec and integration tests work (both excluding a single test, that I have disabled because (I think) they have only been added at V3)

jitsedesmet:
@rubensworks only 2 tests still fail. One seems to always fail on my machine (I wrote it myself, but the mock seems to hate me): 
https://github.com/jitsedesmet/comunica/blob/chore/prepare-function-bussification/packages/expression-evaluator/test/unit/transformers/TermTransformer-test.ts

The other one, I don't think I touched that code and I have no clue what the code does. Maybe you know it and could help me? 
https://github.com/jitsedesmet/comunica/blob/chore/prepare-function-bussification/packages/actor-rdf-resolve-quad-pattern-hypermedia/test/MediatedQuadSource-test.ts#L270

rubensworks:
> The other one, I don't think I touched that code and I have no clue what the code does. Maybe you know it and could help me?

Does it fail locally or also during CI?

packages/bus-functions/lib/ActorFunctions.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  public abstract run<T extends IActionFunctions>(action: T):
+  Promise<T extends { requireTermExpression: true } ? IActorFunctionsOutputTerm : IActorFunctionsOutput>;
+}
+
+export interface IExpressionFunction {
+  apply: (evalContext: IEvalContext) => Promise<E.TermExpression>;
+  /**
+   * Makes you able to error in the termTransformer.
```
rubensworks, line 34: Not sure I understand what this means.
Could you expand this description?

packages/bus-term-comparator-factory/lib/ActorTermComparatorFactory.ts
```diff
@@ -0,0 +1,50 @@
...
+
+}
+
+export interface ITermComparator {
+  /**
+   * Orders two RDF terms according to: https://www.w3.org/TR/sparql11-query/#modOrderBy
+   * @param termA the first term
+   * @param termB the second term
+   */
+  orderTypes: (termA: RDF.Term | undefined, termB: RDF.Term | undefined) => -1 | 0 | 1;
```
rubensworks, line 38: Awesome! We'll be able to use this very nicely within https://github.com/comunica/comunica/issues/877

packages/expression-evaluator/lib/types.ts
```diff
@@ -0,0 +1,23 @@
...
+  evaluate: (mapping: RDF.Bindings) => Promise<RDF.Term>;
+
+  /**
+   * Evaluates the provided bindings in terms of the context the evaluator was created,
+   * returning the effective boolean value.
+   * @param mapping the RDF bindings to evaluate against.
+   */
+  evaluateAsEBV: (mapping: RDF.Bindings) => Promise<boolean>;
+
+  evaluateAsInternal: (mapping: RDF.Bindings) => Promise<E.Expression>;
```
rubensworks, line 22: If it's only supposed to be used internally, is it needed on this interface?

rubensworks (changes requested):
Very nice work!

No major comments from my end.

packages/expression-evaluator/lib/types.ts
jitsedesmet: Apparently not anymore :)

packages/actor-term-comparator-factory-inequality-functions-based/README.md
```diff
@@ -0,0 +1,41 @@
+# Comunica Inequality Functions Based Term Comparator Factory Actor
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-term-comparator-factory-inequality-functions-based.svg)](https://www.npmjs.com/package/@comunica/actor-term-comparator-factory-inequality-functions-based)
+
+A comunica Inequality Functions Based Term Comparator Factory Actor.
```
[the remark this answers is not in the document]
jitsedesmet: Would this suffice?
`The actor orders types based on the inequality functions (`<` and `=`) of the expression evaluator.`

packages/actor-query-operation-orderby-sparqlee/lib/ActorQueryOperationOrderBySparqlee.ts
```diff
@@ -41,12 +42,15 @@ export class ActorQueryOperationOrderBySparqlee extends ActorQueryOperationTyped
...
     // Sorting backwards since the first one is the most important therefore should be ordered last.
+    const orderByEvaluator = await this.expressionEvaluatorFactory.createTermComparator({ context });
+
     for (let i = operation.expressions.length - 1; i >= 0; i--) {
       let expr = operation.expressions[i];
       const isAscending = this.isAscending(expr);
       expr = this.extractSortExpression(expr);
       // Transform the stream by annotating it with the expr result
-      const evaluator = this.expressionEvaluatorFactory.createEvaluator(expr, context);
+      const evaluator = (await this.expressionEvaluatorFactory
```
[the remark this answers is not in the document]
jitsedesmet: Why not?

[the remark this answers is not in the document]
jitsedesmet: The orderTypes expects an RDF.Term right now. Would you want it to receive a Binding just like evaluators do? So it would evaluate and then return a result?

packages/actor-expression-evaluator-factory-base/lib/ActorExpressionEvaluatorFactoryBase.ts
```diff
@@ -0,0 +1,131 @@
...
+import { AlgebraTransformer } from './AlgebraTransformer';
+import { ExpressionEvaluator } from './ExpressionEvaluator';
+
+export function prepareEvaluatorActionContext(orgContext: IActionContext): IActionContext {
+  let context = orgContext;
+
+  context =
+    context.set(KeysExpressionEvaluator.now, context.get(KeysInitQuery.queryTimestamp) || new Date(Date.now()));
+
+  context = context.set(KeysExpressionEvaluator.baseIRI, context.get(KeysInitQuery.baseIRI));
```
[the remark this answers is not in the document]
jitsedesmet: Good question. I think this replaces the getBaseExpressionContext of ActorQueryOperation.ts: https://github.com/comunica/comunica/blob/master/packages/bus-query-operation/lib/ActorQueryOperation.ts#L104-L124 . Do you suggest something else? 
(I did now remove ` context = context.set(KeysExpressionEvaluator.baseIRI, context.get(KeysInitQuery.baseIRI));`).

packages/expression-evaluator/lib/types.ts
jitsedesmet: Okey, This PR has been open for too long. I think the internal is a wrong name. I's more like the thing this returns has been parsed by the expression evaluator. So maybe `evaluateAsEvaluatorExpression` would be better?

packages/actor-expression-evaluator-factory-base/lib/ActorExpressionEvaluatorFactoryBase.ts
```diff
@@ -0,0 +1,131 @@
+import type {
```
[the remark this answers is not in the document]
jitsedesmet: Aha, I was thinking like an hour about it hahah, yes! I think it used to be because you could use the function helpers when using the package. But that might not be the case anymore. Since functions is it's own bus already :)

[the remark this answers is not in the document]
jitsedesmet: It's not really a base actor I think. It's a true implementation of the actor. (It's used by the engines) I think we should look for a more suitable name. Can I call it `ExpressionEvaluatorFactorySparqlee`?

packages/actor-bindings-aggregator-factory-average/lib/ActorBindingsAggregatorFactoryAverage.ts
```diff
@@ -3,15 +3,24 @@ import type {
...
+export interface IActorBindingsAggregatorFactoryAverageArgs extends IActorBindingsAggregatorFactoryArgs {
+  factory: ActorExpressionEvaluatorFactory;
+}
+
 /**
  * A comunica Average Expression Evaluator Aggregate Actor.
  */
 export class ActorBindingsAggregatorFactoryAverage extends ActorBindingsAggregatorFactory {
-  public constructor(args: IActorBindingsAggregatorFactoryArgs) {
+  private readonly factory: ActorExpressionEvaluatorFactory;
```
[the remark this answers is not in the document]
jitsedesmet: I don't think I've thought about that actually. Using a mediator would be prettier. And It should then be added in this PR.

packages/actor-expression-evaluator-factory-base/lib/ActorExpressionEvaluatorFactoryBase.ts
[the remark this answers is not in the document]
jitsedesmet: Oh! I understand what you mean now. SPARQLEE used to have their own context items for this. It might be better to use the original context keys like `KeysInitQuery.baseIRI` and ` KeysInitQuery.queryTimestamp`?

engines/query-sparql/package.json
```diff
@@ -195,7 +197,7 @@
     "browser": "npm run prepare && \"../../node_modules/webpack/bin/webpack.js\" --config webpack.config.js --mode production",
     "browser-profile": "\"../../node_modules/webpack/bin/webpack.js\" --config webpack.config.js --profile --mode production --json > stats.json && webpack-bundle-analyzer stats.json && rm stats.json",
     "spec:base": "node ../../node_modules/rdf-test-suite/bin/Runner.js spec/sparql-engine.js http://w3c.github.io/rdf-tests/sparql/sparql11/manifest-all.ttl -c ../../.rdf-test-suite-cache/",
-    "spec:query": "yarn run spec:base -s http://www.w3.org/TR/sparql11-query/",
+    "spec:query": "yarn run spec:base -s http://www.w3.org/TR/sparql11-query/ --skip 'manifest#agg-groupconcat-04$'",
```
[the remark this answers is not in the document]
jitsedesmet: Maybe best in the merge PR?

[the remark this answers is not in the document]
rubensworks: Yup, just reminding :-)

packages/actor-expression-evaluator-factory-base/lib/ActorExpressionEvaluatorFactoryBase.ts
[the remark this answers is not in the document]
rubensworks: Indeed, if the semantics of the context entries remain identical (which should be possible for all/most?), then I'd reuse the same key.

packages/actor-query-operation-orderby-sparqlee/lib/ActorQueryOperationOrderBySparqlee.ts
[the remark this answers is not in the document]
rubensworks: Nvm, I thought that object was not being used, but it is.

packages/expression-evaluator/lib/types.ts
rubensworks: That works for me!

packages/actor-expression-evaluator-factory-base/lib/ActorExpressionEvaluatorFactoryBase.ts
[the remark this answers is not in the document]
rubensworks: But would it hurt to move it to the bus package, just to avoid the need for another package that most actors on that bus will depend on anyway?

Regarding the name, I'd avoid "SPARQL" in the name, as we aim to be more generic than SPARQL.

packages/actor-term-comparator-factory-inequality-functions-based/README.md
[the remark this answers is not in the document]
rubensworks: That's very clear to me!

Perhaps we can name the actor `actor-term-comparator-factory-expressions-evaluator`?

packages/actor-expression-evaluator-factory-base/lib/ActorExpressionEvaluatorFactoryBase.ts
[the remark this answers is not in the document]
jitsedesmet: Since the actor is an actual actor. I will rename it to `actor-expression-evaluator-factory-default`.

packages/actor-expression-evaluator-factory-base/lib/ActorExpressionEvaluatorFactoryBase.ts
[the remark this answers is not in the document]
jitsedesmet: I was required to replace it with:
```
context = context.setDefault(KeysInitQuery.queryTimestamp, new Date(Date.now()));
context = context.setDefault(KeysInitQuery.functionArgumentsCache, {});
```
 is this oke?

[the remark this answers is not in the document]
jitsedesmet: They were no needed

.github/typedoc-json.sh
```diff
@@ -2,7 +2,7 @@
 
 typedocjson='{
   "extends": [
-    "../../typedoc.base.json"
+    "../../typedoc.default.json"
```
rubensworks, line 5: I think this one is not needed?

engines/config-query-sparql/config/term-comparator-factory/actors/inequality-functions-based.json
```diff
@@ -0,0 +1,17 @@
...
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/runner/^2.0.0/components/context.jsonld",
+
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-term-comparator-factory-expression-evaluator/^1.0.0/components/context.jsonld"
+  ],
+  "@id": "urn:comunica:default:Runner",
+  "@type": "Runner",
+  "actors": [
+    {
+      "@id": "urn:comunica:default:term-comparator-factory/actors#inequality-functions-based",
+      "@type": "ActorTermComparatorFactoryExpressionEvaluator",
```
rubensworks, line 12: Can we rename this "@id" and filename to reflect the new actor name?

packages/bus-term-comparator-factory/lib/ActorTermComparatorFactory.ts
```diff
@@ -0,0 +1,50 @@
...
+import type { MediatorQueryOperation } from '@comunica/bus-query-operation';
+import type { IAction, IActorArgs, IActorOutput, IActorTest, Mediate } from '@comunica/core';
+import { Actor } from '@comunica/core';
+import type * as RDF from '@rdfjs/types';
+
+/**
+ * A comunica actor for term-comparator-factory events.
+ *
+ * Actor types:
+ * * Input:  IActionTermComparatorFactory:      TODO: fill in.
```
rubensworks, line 11: TODO (may also exist in other files)

packages/bus-functions/lib/ActorFunctions.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export interface ITermFunction extends IExpressionFunction {
+  supportsTermExpressions: true;
+  applyOnTerms: (args: E.TermExpression[], exprEval: IInternalEvaluator) => E.TermExpression;
+}
+export interface IActionFunctions extends IAction {
+  functionName: string;
+  arguments?: Alg.Expression[];
+  requireTermExpression?: boolean;
```
rubensworks, line 47: Can you document this?

packages/bus-functions/lib/ActorFunctions.ts
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * Actor types:
+ * * Input:  IActionFunctions:      TODO: fill in.
+ * * Test:   <none>
+ * * Output: IActorFunctionsOutput: TODO: fill in.
+ *
+ * @see IActionFunctions
+ * @see IActorFunctionsOutput
+ */
+export abstract class ActorFunctions extends
```
rubensworks, line 18: As discussed, we can rename this to `bus-function-factory`.

packages/actor-term-comparator-factory-expression-evaluator/lib/InequalityFunctionBasedComparator.ts
```diff
@@ -0,0 +1,110 @@
+import type { InternalEvaluator } from '@comunica/actor-expression-evaluator-factory-default/lib/InternalEvaluator';
+import type { ITermFunction } from '@comunica/bus-functions';
+import type { ITermComparator } from '@comunica/bus-term-comparator-factory';
+import type * as E from '@comunica/expression-evaluator/lib/expressions';
+import type * as RDF from '@rdfjs/types';
+
+export class InequalityFunctionBasedComparator implements ITermComparator {
```
rubensworks, line 7: `TermComparatorExpressionEvaluator`?

rubensworks (commented):
Just some minor nits, should be good to go after this review round :-)

.github/typedoc-json.sh
jitsedesmet: oh oh. that's not good... :/

rubensworks:
Nice!! 🚀 🚀

PR #1284, opened 2023-10-24
Feature/action context bindings

RubenEschauzier:
This is a draft PR for adding an `ActionContext` to each binding. This PR would solve issue #1203. This context can contain for example the data source of the specific binding. During the join of one binding with another, we merge the binding contexts based on merge handler functions that need to be specified in the Comunica config.

The merge change is located [here](https://github.com/RubenEschauzier/comunica/blob/520d880bedfe5330172f0d723f48ae65f2b24ae8/packages/bindings-factory/lib/Bindings.ts#L130).

Currently, by default we create the set union function for the key 'source' in the binding context, however, this should probably be removed for default Comunica.
An example of how merge handlers are created can be found [here](https://github.com/RubenEschauzier/comunica/blob/520d880bedfe5330172f0d723f48ae65f2b24ae8/packages/actor-init-query/lib/QueryEngineBase.ts#L189C6-L189C6), with the new bus and an example actor for creating merge handlers [here](https://github.com/RubenEschauzier/comunica/blob/520d880bedfe5330172f0d723f48ae65f2b24ae8/packages/bus-merge-binding-factory/lib/ActorMergeBindingFactory.ts) and [here](https://github.com/RubenEschauzier/comunica/blob/520d880bedfe5330172f0d723f48ae65f2b24ae8/packages/actor-merge-binding-factory-context-union/lib/ActorMergeBindingFactoryContextUnion.ts) respectively.

Because the merge function is performance-critical, I benchmarked the new and old merge functions using the following code: [standard](https://github.com/RubenEschauzier/Bindings-merge-standard-benchmark) and [new](https://github.com/RubenEschauzier/Binding-context-merge-benchmark). We measure the number of function executions per second by executing the function 1,000,000 times with 20 replications. Non-overlapping bindings means that the two merged bindings have no common variables. The output is as follows. 
Standard Comunica: 
```
Overlapping bindings: 335079.43 ops/s (18451.50 std)
Non-overlapping bindings: 235695.00 ops/s (9769.74 std)
```

New merge function:
```
Non-overlapping bindings, overlapping context entries: 139930.62 ops/s (4047.23 std)
Overlapping bindings, overlapping context entries: 171073.40 ops/s (7390.75 std)
Non-overlapping bindings, no context: 225660.58 ops/s (6854.23 std)
Overlapping bindings, no context: 327583.13 ops/s (7658.40 std)
```
The performance in the case of no context is especially interesting, as this will be the most common configuration. From the benchmark, we see a 4.35% decrease when merging non-overlapping bindings and a 2.26% decrease for overlapping bindings. In both cases, the difference is statistically significant, though, for the overlapping bindings, the p-value is only about 0.05.

A more extensive benchmark on WatDiv will follow.
(This is the same draft PR with fixed history)


RubenEschauzier:
As promised, here are the benchmark results comparing the standard Comunica implementation and the one from the PR. Note that for the standard Comunica, I used the version of the next/major branch.

benchmark_results_all

Here we split the queries due to significantly higher execution times of the complex queries.
benchmark_results_complex
benchmark_results_non_complex

Let me know what you think.

RubenEschauzier:
On the topic of the failing docker build, I'm not sure what is going wrong. I tried it on my PC and it works fine. Is it possibly a problem with newer versions being available on npm?

packages/actor-abstract-path/lib/ActorAbstractPath.ts
```diff
@@ -83,6 +82,7 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
     predicate: Algebra.PropertyPathSymbol,
     graph: RDF.Variable,
     context: IActionContext,
+    BF: BindingsFactory,
```
rubensworks, line 85: This is a nit-picky comment, but could we write out the parameter name as `bindingsFactory`? Just to be consistent with the coding style of other methods params.
(this comment also applies to the rest of this PR)

packages/actor-abstract-path/test/PathVariableObjectIterator-test.ts
```diff
@@ -10,7 +10,7 @@ import { Factory } from 'sparqlalgebrajs';
 import { PathVariableObjectIterator } from '../lib';
 
 const DF = new DataFactory();
-const BF = new BindingsFactory();
+const BF = new BindingsFactory({});
```
rubensworks, line 13: Could we make the merge handlers optional in the constructor?
Then we don't need changes in places like this.

packages/types/lib/IQueryContext.ts
```diff
@@ -24,7 +23,6 @@ export interface IQueryContextCommon {
   // Types of these entries should be aligned with contextKeyShortcuts in ActorInitQueryBase
   // and Keys in @comunica/context-entries
 
-  source?: IDataSource;
```
rubensworks, line 27: Why can't we just keep this instead of `sourceBinding`?

packages/bus-merge-binding-factory/lib/ActorMergeBindingFactory.ts
```diff
@@ -0,0 +1,45 @@
...
+  mergeHandlers: Record<string, IMergeHandler<any>>;
+}
+
+export type IActorMergeBindingFactoryArgs = IActorArgs<
+IActionMergeBindingFactory, IActorTest, IActorMergeBindingFactoryOutput>;
+
+export type MediatorMergeBindingFactory = Mediate<
+IActionMergeBindingFactory, IActorMergeBindingFactoryOutput>;
+
+export interface IMergeHandler<V> {
```
rubensworks, line 42: Suggestion for a more specific name: `IBindingsContextMergeHandler`.

Also add some jsdoc here explaining what it is and does.
Make sure to add in jsdoc that `V` is the type corresponding to the context key.

packages/bus-merge-binding-factory/lib/ActorMergeBindingFactory.ts
```diff
@@ -0,0 +1,45 @@
+import type { IAction, IActorArgs, IActorOutput, IActorTest, Mediate } from '@comunica/core';
+import { Actor } from '@comunica/core';
+import type { IActionContext } from '@comunica/types';
+
```
rubensworks, line 4: This file needs some more docs on what each class and param does (have a look at other buses to see examples).

packages/bus-merge-binding-factory/lib/ActorMergeBindingFactory.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Actor types:
+ * * Input:  IActionMergeBindingFactory:      TODO: fill in.
+ * * Test:   <none>
+ * * Output: IActorMergeBindingFactoryOutput: TODO: fill in.
+ *
+ * @see IActionMergeBindingFactory
+ * @see IActorMergeBindingFactoryOutput
+ */
+
+export abstract class ActorMergeBindingFactory extends Actor<IActionMergeBindingFactory,
```
rubensworks, line 17: I know this is an annoying request, but I suggest renaming the bus to `bus-merge-bindings-context`

packages/bus-merge-binding-factory/lib/ActorMergeBindingFactory.ts
```diff
@@ -0,0 +1,45 @@
...
+  /**
+  * @param args - @defaultNested {<default_bus> a <cc:components/Bus.jsonld#Bus>} bus
+  */
+  public constructor(args: IActorMergeBindingFactoryArgs) {
+    super(args);
+  }
+}
+
+export interface IActionMergeBindingFactory extends IAction {
+  context: IActionContext;
```
rubensworks, line 29: Not sure if that would be useful at this stage, but we could also consider passing the whole `Bindings` object here. WDYT?

packages/bus-rdf-join/lib/ActorRdfJoin.ts
```diff
@@ -36,6 +36,9 @@ export abstract class ActorRdfJoin
    * If this actor will be logged in the debugger and physical query plan logger
    */
   public includeInLogs = true;
+  /**
+   * Merge handlers for context of bindings, this will be initialised once in first execution of this actor
```
rubensworks, line 40: Wrong docs?

packages/bindings-factory/lib/Bindings.ts
```diff
@@ -138,13 +191,93 @@ export class Bindings implements RDF.Bindings {
...
+      }
+      return new Bindings(this.dataFactory, Map(entries), this.contextMergeHandlers, mergedContext);
+    }
+    const castOther = <Bindings> other;
+    return new Bindings(this.dataFactory, Map(entries), this.contextMergeHandlers, castOther.context);
+  }
+
+  private mergeContext(context: IActionContext, otherContext: IActionContext): IActionContext {
+    // Get Set of all keys present in either of the bindings
+    const keysContext = this.uniqueKeys([ ...context.keys(),
```
rubensworks, line 208: Can't we just put them in a `Set`? Then we can simplify `uniqueKeys`.

packages/bindings-factory/lib/Bindings.ts
```diff
@@ -138,13 +191,93 @@ export class Bindings implements RDF.Bindings {
...
+      ...otherContext.keys() ]);
+
+    // Get Set of all keys present in both bindings
+    const keysBothContext = context.keys().filter(
+      element => otherContext.keys().some(({ name }) => element.name === name),
+    );
+    // Merge context based on supplied mergeHandlers
+    for (const key of keysContext) {
+      const keyString = key.name;
+      const occursInBoth = keysBothContext.some(x => x.name === key.name);
```
rubensworks, line 218: If `keysBothContext` would be a set, we can very cheaply call `.includes()` on it.

packages/actor-rdf-resolve-hypermedia-sparql/lib/ActorRdfResolveHypermediaSparql.ts
```diff
@@ -53,6 +62,10 @@ export interface IActorRdfResolveHypermediaSparqlArgs extends IActorRdfResolveHy
    * @default {false}
    */
   forceHttpGet: boolean;
+  /**
+   * A mediator for creating binding context merge handlers
+   */
+  mediatorMergeHandlers: MediatorMergeBindingFactory;
```
rubensworks, line 68: Can we rename this mediator to `mediatorMergeBindingsContext`? (also in other places)

rubensworks (changes requested):
Really great work @RubenEschauzier!

This review contains mainly some high-level comments.
Once you had a chance to look at those, I'll do another (possibly final) pass.

packages/types/lib/IQueryContext.ts
RubenEschauzier: Mainly because IDataSource means it can be 4 different types. This will make the mergeHandler function more complex as making a Set of objects won't work properly (as far as I know). It is possible, but will require more type checks and might be slightly more expensive.

packages/bindings-factory/lib/Bindings.ts
RubenEschauzier: The problem is the Set will not consider two key objects with the same name as the equal. If we have
```
const key1 = { name: "key" }; const key2 = { name: "key" }
const set = new Set([key1, key2]);
```
Then set will contain both ```key1``` and ```key2```. 
I could use immutable.js though, if that would be ok here.

packages/actor-rdf-resolve-hypermedia-sparql/lib/RdfSourceSparql.ts
```diff
@@ -136,10 +138,10 @@ export class RdfSourceSparql implements IQuadSource {
    * @param {string} query A SPARQL query string.
    * @return {BindingsStream} A stream of bindings.
    */
-  public queryBindings(endpoint: string, query: string): BindingsStream {
+  public async queryBindings(endpoint: string, query: string): Promise<BindingsStream> {
```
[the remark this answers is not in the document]
RubenEschauzier: It's changed back to sync and no apparent problems occurred. Not sure why I changed it to async at some point.

packages/types/lib/IQueryContext.ts
rubensworks: Sets of objects should work fine AFAIK, as Sets will just hash the objects.

The problem with removing that source key is that it will break many other things that make use of that key. So I'd recommend keeping it as-is. And if we really need a different type, then I'd add a different key.

packages/bindings-factory/lib/Bindings.ts
rubensworks: But in our case, context keys are singletons, so they will always refer to the same entity.
So in this case, the Set should deduplicate them AFAIK.

RubenEschauzier: Right, because you use the constant defined in keys.ts? That should work. In the tests, I had them as new objects, which is probably why it didn't work.

packages/bus-merge-binding-factory/lib/ActorMergeBindingFactory.ts
RubenEschauzier: I don't know how that would work or the benefit. Would you then call the merge-bindings-context mediator each time you need to merge bindings?

packages/types/lib/IQueryContext.ts
RubenEschauzier: I think it's the same 'problem' as with the keys, if you get two objects describing the same data source, then taking the set of them wouldn't deduplicate them. Unless you know for sure that if two bindings have the same data source object in content, they also reference the same object.

In any case, I don't remember why I removed the source entry in the context. So that should be fixed for sure.

rubensworks: > I think it's the same 'problem' as with the keys, if you get two objects describing the same data source, then taking the set of them wouldn't deduplicate them.

Indeed, it's the same as with the keys. Sources will be equal by identity. So equal sources refer to the same objects, so Sets will dedupe them.

rubensworks:
Merged! Thanks for the great work @RubenEschauzier! :-)

PR #1288, opened 2023-11-11
Fix `BindingsFactory`'s `fromBindings` type

Peeja:
My project was reporting a type failure in `BindingsFactory.d.ts`, so I dug in to investigate. It was because `fromBindings()` needs to accept any `RDF.Bindings`, not just Comunica `Bindings`.

It looks like the reason that wasn't failing in Comunica itself is that `strictFunctionTypes` was off. #1233 tried to turn that back on and address various issues it was covering up. But as noted there, it wasn't quite that simple to turn on. For now, this addresses the original type issue I hit.


rubensworks (approved):
Looks great, thanks @Peeja!

Issue #618, opened 2020-03-12
Add option to disable caching


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request

____
#### Description:

In cases where caching is not wanted, we should allow this to be disabled completely.

For this, we can add a new context entry, which is checked within the relevant actors that do caching.

From the CLI, it should be possible to pass the option `--no-cache`, which defines this context entry.

Optionally, we may also want this feature to cause the `Cache-Control: no-cache` header to be sent in requests.

(In some cases, caching may only be desired within a certain query exec, but not across multiple query execs)

rubensworks:
Thanks for sharing @mrkvon!

An alternative solution would be to call `engine.invalidateHttpCache()` after/before each query execution.

mrkvon:
So, the issue is open only to keep track of the desired `--no-cache` option for CLI, isn't it?

After reading the issue, i wrongly assumed that no cache invalidation is implemented, and didn't even bother looking into documentation... Anyways, thanks for clarifying @rubensworks!

rubensworks:
> So, the issue is open only to keep track of the desired --no-cache option for CLI?

Also that, but also to disable caching to happening in the first place.
This should save a bit on memory consumption during query execution compared to calling `invalidateHttpCache()` after each query exec.

PR #1297, opened 2023-12-14
Add option to disable caching

simonvbrae:
- Added a context key to disable caching
- Implemented cache disabling via command line argument
- Didn't disable request counting in ActionObserverHTTP because it shouldn't cause overhead

Types of caching disabled:
 - Caching MediatedQuadSource with URL as key
 - Caching empty patterns

Related to https://github.com/comunica/comunica/issues/618


git add .
```diff
@@ -0,0 +1,83 @@
+[1mdiff --git a/packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts b/packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts[m
```
rubensworks, line 1: This is a mistake?

packages/context-entries/lib/Keys.ts
```diff
@@ -173,6 +173,7 @@ export const KeysInitQuery = {
    * A JSON-LD context
    */
   jsonLdContext: new ActionContextKey<any>('@context'),
+  disableCaching: new ActionContextKey<boolean>('@comunica/actor-init-query:disableCaching'),
```
rubensworks, line 176:
```suggestion
  noCache: new ActionContextKey<boolean>('@comunica/actor-init-query:noCache'),
```

Could you also document this briefly?

packages/actor-rdf-resolve-quad-pattern-hypermedia/lib/MediatedLinkedRdfSourcesAsyncRdfIterator.ts
```diff
@@ -121,7 +122,7 @@ export class MediatedLinkedRdfSourcesAsyncRdfIterator extends LinkedRdfSourcesAs
   }
 
   protected shouldStoreSourcesStates(): boolean {
-    return this.aggregatedStore === undefined;
+    return this.aggregatedStore === undefined && !this.context.get(KeysInitQuery.disableCaching);
```
rubensworks, line 125: I don't think we need it here, since aggregate stores always only exist in the context of a single query exec.

packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts
```diff
@@ -251,6 +251,11 @@ export class CliArgsHandlerBase implements ICliArgsHandler {
     if (args.unionDefaultGraph) {
       context[KeysQueryOperation.unionDefaultGraph.name] = true;
     }
+
+    // Define union default graph
```
rubensworks, line 255: Wrong jsdoc.

rubensworks: Option also needs to be added here: https://github.com/comunica/comunica/blob/master/packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts#L66

rubensworks: And also here: https://github.com/comunica/comunica/blob/master/packages/actor-init-query/lib/ActorInitQueryBase.ts#L124

rubensworks (changes requested):
In general, this looks good.

rubensworks:
Ah, I also test coverage is not 100% yet.

simonvbrae:
I've implemented your comments and suggestions, including test coverage.

rubensworks:
Great, thanks! Do you have any [insights on performance](https://github.com/comunica/comunica/pull/1297#pullrequestreview-1811253307)?

simonvbrae:
> Great, thanks! Do you have any [insights on performance](https://github.com/comunica/comunica/pull/1297#pullrequestreview-1811253307)?

Performance testing showed that this approach causes a decrease in performance, with some queries displaying a nearly doubled execution time. This is likely caused by the repeated occurrence of the same request within one query execution, making it unproductive to completely disable caches.
The proposed solution to this was to re-enable caching and instead call `engine.invalidateHttpCache()` after each query execution.
I have implemented this solution. As a result of this, the following list of operations is executed after every query:
- ´ActorRdfUpdateQuadsHypermedia´ has a  ´LRUCache<string, Promise<IQuadDestination>>´ which is cleared via ´.clear()´
- `ActorRdfResolveQuadPatternHypermedia` has a `LRUCache<string, MediatedQuadSource>` which is cleared via ´.clear()´
- The `ActionObserverHttp` instance in `ActorQueryResultSerializeSparqlJson` has it's `this.requests` counter reset to 0
- The `ActionObserverHttp` instance in `ActorQueryResultSerializeStats` has it's `this.requests` counter reset to 0

packages/actor-init-query/test/ActorInitQuery-test.ts
```diff
@@ -772,6 +772,22 @@ LIMIT 100
         });
       });
 
+      it('handles the --cache flag', async() => {
+        const stdout = await stringifyStream(<any> (await actor.run({
+          argv: [ sourceHypermedia, '-q', queryString, '--cache', 'false' ],
```
rubensworks, line 777: I guess this should be `--noCache`?

packages/actor-init-query/test/ActorInitQuery-test.ts
```diff
@@ -772,6 +772,22 @@ LIMIT 100
...
+          env: {},
+          stdin: <Readable><any> new PassThrough(),
+          context,
+        })).stdout);
+        expect(stdout).toContain(`{"a":"triple"}`);
+        expect(spyQueryOrExplain).toHaveBeenCalledWith(queryString, {
```
rubensworks, line 783: I suspect we should be expecting the `noCache` context key here.

packages/actor-rdf-resolve-quad-pattern-federated/lib/FederatedQuadSource.ts
```diff
@@ -53,8 +53,8 @@ export class FederatedQuadSource implements IQuadSource {
     this.skipEmptyPatterns = skipEmptyPatterns;
     this.algebraFactory = new Factory();
 
-    // Initialize sources in the emptyPatterns datastructure
```
rubensworks, line 56: This change can be undone.

packages/actor-rdf-resolve-quad-pattern-hypermedia/lib/ActorRdfResolveQuadPatternHypermedia.ts
```diff
@@ -94,6 +94,7 @@ export class ActorRdfResolveQuadPatternHypermedia extends ActorRdfResolveQuadPat
         },
       );
 
+      // If cache is not disabled
```
rubensworks, line 97: This change can be undone.

packages/actor-init-query/lib/QueryEngineBase.ts
```diff
@@ -249,6 +249,11 @@ implements IQueryEngine<QueryContext, QueryStringContextInner, QueryAlgebraConte
       };
     }
 
+    // Invalidate caches if cache argument is set to false
+    if (actionContext.get(KeysInitQuery.noCache)) {
```
rubensworks, line 253: While reading this code, I'm realizing that this will only invalidate cache for the next query execution.

If a query execution was executed before without noCache, then the cache will still be filled when executing this query, and will only be invalidated afterwards.

I think we may want to move this block so that it's the very first thing that happens before actual query execution.

What do you think?

rubensworks (changes requested):
Thanks for the changes!

Could you re-do the performance tests after applying my suggestions?

packages/actor-init-query/test/ActorInitQuery-test.ts
simonvbrae: Ok, changed the argument to `--noCache` instead of `--cache false`

packages/actor-init-query/lib/QueryEngineBase.ts
simonvbrae: I think this is a good suggestion. In light of the case where the user wants their execution to not be influenced by what the cache contains it makes sense to clear the cache before execution. This should have an impact on performance, I'll test it out.

In the case where the user wants to avoid memory overhead it makes sense to also clear the cache after query execution. So I would look at the impact on performance of clearing the caches before and after.

And there's the case where the user wants one of their queries to have no influence on the cache, also leaving them intact. In that case clearing the caches works against what the user intends. Do you think this use case is common?

rubensworks: Maybe clearing cache before and after exec is indeed what we want, if performance impact is not too high.

simonvbrae: Impact on performance seems very limited when clearing the cache before and after execution, so let's continue with this solution.
What about the use case of wanting to do a query that leaves cache state intact?

rubensworks: > What about the use case of wanting to do a query that leaves cache state intact?

I would leave this out of scope for this PR.

rubensworks (approved):
Looks good to me, nice work!

Could you do a quick performance test and share the results here?

simonvbrae:
> Could you do a quick performance test and share the results here?
I can't measure a significant impact on the performance.

|query|cache enabled|cache disabled|
|-|-|-|
|q1|6.068s|5.876s|
|q2|17.788s|16.372s|
|q3|6.110s|6.068s|

Queries: https://pastebin.com/U7MWKFUE

rubensworks:
Wait, so with cache disabled, things are faster? That's surprising.
Could you try averaging your results across multiple executions to rule out outliers?

rubensworks:
Could you share your test code?

simonvbrae:
The previous results were a fluke. I ran the tests again and got these results, which to me confirms that the difference is too small to be measured like this.

|query|cache enabled|cache disabled|
|-|-|-|
|q1|6.179s|6.281s|
|q2|16.240s|17.122s|
|q3|6.256s|6.192s|

simonvbrae:
> Could you share your test code?

https://pastebin.com/ydXpCwGZ

rubensworks:
Nice work, thanks @simonvbrae!

PR #1299, opened 2023-12-21
Switch to `@smessie/readable-web-to-node-stream@^3.0.3`

smessie:
Hopefully closes https://github.com/comunica/comunica-feature-link-traversal/issues/121

This fork uses the latest version of the readable-stream dependency.


rubensworks:
Looks good, thanks!
For reference, once we drop Node 16 support, we may be able to use https://nodejs.org/api/stream.html#streamreadablefromwebreadablestream-options

PR #1321, opened 2024-04-10
Allow the link queue wrapper to wrap a generic type

RubenEschauzier:
PR to allow the wrapped link queue type to be generic (which defaults to the current implementation of the link queue wrapper). This will enable users to wrap other types of link queues, like a priority queue.

Sidenote: I thought the linter would fix the weird change log of ActorRdfJoin, but it didn't. If this is a problem let me know.

Credit to @joachimvh, as he suggested this implementation.

### DIFF
--- a/packages/bus-rdf-join/lib/ActorRdfJoin.ts
+++ b/packages/bus-rdf-join/lib/ActorRdfJoin.ts
@@ -117,6 +117,7 @@ export abstract class ActorRdfJoin
     if (bindings.length === 1) {
       return bindings[0];
     }
+
     let acc: Bindings = bindings[0];
     for (const binding of bindings.slice(1)) {
       const merged = acc.merge(binding);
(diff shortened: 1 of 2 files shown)


packages/bus-rdf-join/lib/ActorRdfJoin.ts
```diff
@@ -279,7 +279,7 @@ export abstract class ActorRdfJoin
       parentPhysicalQueryPlanNode = action.context.get(KeysInitQuery.physicalQueryPlanNode);
       action.context = action.context.set(KeysInitQuery.physicalQueryPlanNode, action);
     }
-
```
rubensworks, line 282: Unneeded changes?

joachimvh:
Had a look since I got tagged. Depending on how generic you want to go and what it is exactly that you're having issues with due to the fixed typing, you could also put this on the interface level. E.g., `export interface ILinkQueue<T extends ILink = ILink>` and then `export class LinkQueueWrapper<T extends ILink = ILink> implements ILinkQueue<T>`. That way you could also have generics in the response types of your functions.

packages/bus-rdf-join/lib/ActorRdfJoin.ts
RubenEschauzier: Yep, I noticed my whole code base is a bit messed up. I'll open a new pull request which is cleaner hopefully

RubenEschauzier:
That would be more flexible, but it seems to me that because the interface is called LinkQueue, you should expect to only queue links. You can already pass different types of links to the queue by extending the ILink interface.

RubenEschauzier:
Closed in favor of the cleaner branch

PR #1322, opened 2024-04-10
Wrap generic type with LinkQueueWrapper

RubenEschauzier:
(This is the same as the previous PR, however the previous branch had weird issues due to rebasing comunica v3 on it.)
PR to allow the wrapped link queue type to be generic (which defaults to the current implementation of the link queue wrapper). This will enable users to wrap other types of link queues, like a priority queue.

Sidenote: I thought the linter would fix the weird change log of ActorRdfJoin, but it didn't. If this is a problem let me know.

Credit to @joachimvh, as he suggested this implementation.


rubensworks:
@RubenEschauzier The linter seems to fail :-(

RubenEschauzier:
Yep my bad, this should do it.

rubensworks:
Thanks @RubenEschauzier!
Do you urgently need a new release for this?

RubenEschauzier:
It's not urgent, as I still have to update my local link traversal repos to V3, but it would be nice sometime soon.

PR #1325, opened 2024-04-17
Source attribution in binding context

RubenEschauzier:
To track what bindings are produced by what sources (which includes sources obtained by traversing links) this PR introduces the `ActorQuerySourceIdentifyHypermediaAnnotateSource` actor that wraps hypermedia sources and adds the source URL of produced bindings to their context. 

During joins these sources will be unioned and added to the context of the resulting joined binding. Thus tracking what sources were needed to produce the result

In addition, this PR allows the user to optionally return this source attribution as an additional binding when serializing the bindings to JSON in `ActorQueryResultSerializeSparqlJson`. This is turned off by default.

This PR currently has the actors in the context by default, mainly to show a usage example. If this is breaking (which it probably is) I'll remove the config entries and give them as an example in the README.md.


packages/actor-query-source-identify-hypermedia-annotate-source/lib/ActorQuerySourceIdentifyHypermediaAnnotateSource.ts
```diff
@@ -0,0 +1,60 @@
...
+export class ActorQuerySourceIdentifyHypermediaAnnotateSource extends ActorQuerySourceIdentifyHypermedia {
+  public readonly mediatorMergeBindingsContext: MediatorMergeBindingsContext;
+  public readonly mediatorQuerySourceIdentifyHypermedia: MediatorQuerySourceIdentifyHypermedia;
+
+  public constructor(args: IActorQuerySourceIdentifyHypermediaAnnotateSourceArgs) {
+    super(args, 'file');
+  }
+
+  public async testMetadata(
+    _action: IActionQuerySourceIdentifyHypermedia,
```
rubensworks, line 25:
```suggestion
    action: IActionQuerySourceIdentifyHypermedia,
```

packages/actor-query-source-identify-hypermedia-annotate-source/lib/ActorQuerySourceIdentifyHypermediaAnnotateSource.ts
```diff
@@ -0,0 +1,60 @@
...
+  ): Promise<IActorQuerySourceIdentifyHypermediaTest> {
+    if (_action.context.get(KEY_CONTEXT_WRAPPED)) {
+      throw new Error('Unable to wrap link queues multiple times');
+    }
+    return { filterFactor: 0 };
+  }
+
+  public async run(action: IActionQuerySourceIdentifyHypermedia): Promise<IActorQuerySourceIdentifyHypermediaOutput> {
+    this.logInfo(action.context, `Wrapped source with provenance: ${action.url}`);
```
rubensworks, line 34: Not sure if we really need this logger here.
If you feel strongly about it, I'd lower it to the debug-level.

packages/actor-query-source-identify-hypermedia-annotate-source/lib/ActorQuerySourceIdentifyHypermediaAnnotateSource.ts
```diff
@@ -0,0 +1,60 @@
...
+    const context = action.context.set(KEY_CONTEXT_WRAPPED, true);
+    action.context = context;
+
+    const { source, dataset } = await this.mediatorQuerySourceIdentifyHypermedia.mediate(action);
+    source.referenceValue = action.url;
+
+    const wrappedSource = new QuerySourceAddSourceAttribution(source);
+    wrappedSource.toString = () => `QuerySourceWrappedAnnotateSource(${action.url})`;
```
rubensworks, line 42:
```suggestion
    wrappedSource.toString = () => `QuerySourceAddSourceAttribution(${action.url})`;
```

packages/context-entries/lib/Keys.ts
```diff
@@ -281,6 +281,13 @@ export const KeysRdfUpdateQuads = {
   destination: new ActionContextKey<IDataDestination>('@comunica/bus-rdf-update-quads:destination'),
 };
 
+export const KeysBindingContext = {
```
rubensworks, line 284:
```suggestion
export const KeysMergeBindingsContext = {
```

packages/types/lib/IQueryContext.ts
```diff
@@ -48,4 +48,6 @@ export interface IQueryContextCommon {
   extensionFunctions?: Record<string, (args: RDF.Term[]) => Promise<RDF.Term>>;
   explain?: QueryExplainMode;
   recoverBrokenLinks?: boolean;
+  // The specific source of a binding object, for determining provenance of bindings
+  sourceBinding?: string;
```
rubensworks, line 52: Is this one needed here?

rubensworks (changes requested):
This is really nice!

No major comments about the code itself.

But as you said, not sure if we want to enable this by default.
Could you do some quick manual perf checks (nothing too fancy) to see if queries become significantly slower?
I suppose it will, in which case we may want to only enable this in a separate config.

packages/actor-query-result-serialize-sparql-json/lib/ActorQueryResultSerializeSparqlJson.ts
```diff
@@ -114,6 +117,14 @@ export class ActorQueryResultSerializeSparqlJson extends ActorQueryResultSeriali
         // JSON SPARQL results spec does not allow unbound variables and blank node bindings
         const bindingsJson = Object.fromEntries([ ...bindings ]
           .map(([ key, value ]) => [ key.value, ActorQueryResultSerializeSparqlJson.bindingToJsonBindings(value) ]));
+
+        if (this.addSourceAttributionToBinding) {
+          bindingsJson._sourceAttribution = {
+            value: bindings.getContextEntry(KeysBindingContext.sourceBinding),
+            type: 'literal',
```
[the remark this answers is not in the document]
RubenEschauzier: In the context of link traversal, they will likely be an IRI as the source has to be dereferenced to get data from it. In 'normal' federated querying a source could be a file. I'm unsure if that matters though, what would the alternative to an IRI be?

packages/actor-query-source-identify-hypermedia-annotate-source/lib/ActorQuerySourceIdentifyHypermediaAnnotateSource.ts
RubenEschauzier: I just copied it from another hypermedia actor and figured I would keep it. I'll remove it

packages/actor-query-result-serialize-sparql-json/lib/ActorQueryResultSerializeSparqlJson.ts
```diff
@@ -114,6 +117,14 @@ export class ActorQueryResultSerializeSparqlJson extends ActorQueryResultSeriali
         // JSON SPARQL results spec does not allow unbound variables and blank node bindings
         const bindingsJson = Object.fromEntries([ ...bindings ]
           .map(([ key, value ]) => [ key.value, ActorQueryResultSerializeSparqlJson.bindingToJsonBindings(value) ]));
+
+        if (this.addSourceAttributionToBinding) {
+          bindingsJson._sourceAttribution = {
```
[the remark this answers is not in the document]
RubenEschauzier: _source works for me.

packages/actor-query-result-serialize-sparql-json/lib/ActorQueryResultSerializeSparqlJson.ts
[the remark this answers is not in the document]
rubensworks: I'd go for iri then :-)

Also, we may want to ensure (in the future) that the `referenceValue` of sources is always an IRI.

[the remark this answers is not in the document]
RubenEschauzier: Ah, I misread the comment. I agree with going for IRI. To be sure, that means replacing `type: 'literal'` with `type: 'NamedNode'`?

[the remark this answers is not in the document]
rubensworks: According to the spec, it should be `uri`: https://www.w3.org/TR/sparql11-results-json/#select-encode-terms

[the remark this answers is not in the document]
RubenEschauzier: Oh right, should be fixed now.

RubenEschauzier:
Here are some micro-benchmark results on the first six solidbench queries (over HTTP). The queries are run 4 times and averaged.
First, the plot of execution time in seconds:
image

Second, the log of execution times to see the difference for the faster queries better
image

Finally, the raw averages:
```
    with_attribution =    [3.76, 2.92, 5.17, 2.45, 1.93, 237.37 ]
    without_attribution = [3.98, 2.87, 3.83, 2.36, 1.97, 117.29 ]

```

The results show that source attribution has significant overhead for primarily longer-running queries. So, disabling it by default is probably the best way to go.

rubensworks:
Ok, thanks for checking!

Could you remove it from the default config then?
I guess you mainly would use it for traversal, so I guess we can create a dedicated config for it there, which would then be auto-published as web client.

RubenEschauzier:
Yep, will do that. However, I'm not 100% sure the source attribution will be displayed on the web client as is. It depends on what serialization is used there.

RubenEschauzier:
I've added a query process wrapper that takes the sources in the binding context and binds them to a binding variable called _source. When this is serialized you thus get the source information regardless of serialization. All unit tests succeeded when source attribution was included in the config, however, the spec tests didn't (as adding variables to bindings is not spec-compliant).

I've now once again removed it from the default configs. Instead, I'll make a PR for [Link-Traversal](https://github.com/comunica/comunica-feature-link-traversal) that includes the config that enables source attribution when this gets merged.

engines/config-query-sparql/config/query-source-identify-hypermedia/actors.json
```diff
@@ -4,7 +4,9 @@
 
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-query-source-identify-hypermedia-none/^3.0.0/components/context.jsonld",
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-query-source-identify-hypermedia-qpf/^3.0.0/components/context.jsonld",
-    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-query-source-identify-hypermedia-sparql/^3.0.0/components/context.jsonld"
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-query-source-identify-hypermedia-sparql/^3.0.0/components/context.jsonld",
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-query-source-identify-hypermedia-annotate-source/^3.0.0/components/context.jsonld"
```
rubensworks, line 8: Here's your problem.

packages/actor-query-source-identify-hypermedia-annotate-source/lib/QuerySourceAddSourceAttribution.ts
```diff
@@ -0,0 +1,83 @@
...
+  public queryBindings(
+    operation: Algebra.Operation,
+    context: IActionContext,
+    options: IQueryBindingsOptions | undefined,
+  ): BindingsStream {
+    return this.addSourceUrlToBindingContext(this.innerSource.queryBindings(operation, context, options));
+  }
+
+  public addSourceUrlToBindingContext(iterator: BindingsStream): BindingsStream {
+    const ret = iterator.transform({
```
rubensworks, line 41: You can also use `iterator.map` here.

packages/context-entries/lib/Keys.ts
```diff
@@ -281,6 +281,13 @@ export const KeysRdfUpdateQuads = {
   destination: new ActionContextKey<IDataDestination>('@comunica/bus-rdf-update-quads:destination'),
 };
 
+export const KeysMergeBindingsContext = {
+  /**
+   * The data source of the binding
```
rubensworks, line 286: Name and comment is a bit inaccurate, since we have a `string[]`, so multiple sources.

rubensworks (changes requested):
Looks good!
I can go over it again once the CI passes.

packages/context-entries/lib/Keys.ts
```diff
@@ -281,6 +281,13 @@ export const KeysRdfUpdateQuads = {
   destination: new ActionContextKey<IDataDestination>('@comunica/bus-rdf-update-quads:destination'),
 };
 
+export const KeysMergeBindingsContext = {
+  /**
+   * The data sources required to produce the binding
+   */
+  sourceBinding: new ActionContextKey<string[]>('@comunica/bus-merge-bindings-context:sourceBinding'),
```
rubensworks, line 288:
```suggestion
  sourcesBinding: new ActionContextKey<string[]>('@comunica/bus-merge-bindings-context:sourcesBinding'),
```

rubensworks (changes requested):
Just one minor thing; then it's good to go!

rubensworks:
Nice, thanks @RubenEschauzier!

PR #1326, opened 2024-04-17
Fix DELETE/INSERT update operation order

### LINKED ISSUE #1301: DELETE/INSERT is deleting the quad but not updating the same value in the data store.
#### Issue type:

- :bug: Bug 

____
#### Description:

```

  await myEngine.queryVoid(
    `
    WHERE
    {
    }`,
    {
      sources: [store],
    }
  );
```

```

  await myEngine.queryVoid(
    `
    WHERE
    {
    }`,
    {
      sources: [store],
    }
  );

```
Why is it so?

____
#### Environment:

`node v18.6.0`
`"n3": "^1.17.2"`

#### Crash log:

-----

## Bounty

| |
|-|
| €272 |

karelklima:
Resolves #1301.

I swapped the order of operations of insert and delete streams to first delete quads and then insert. After that, updating a quad that has already been in the data source works as expected, i.e. it is not deleted.

### DIFF
--- a/packages/bus-rdf-update-quads/lib/IQuadDestination.ts
+++ b/packages/bus-rdf-update-quads/lib/IQuadDestination.ts
@@ -22,6 +22,12 @@ export interface IQuadDestination {
    */
   delete: (quads: AsyncIterator<RDF.Quad>) => Promise<void>;
   /**
+   * Updates destination with quad stream to insert and quad stream to delete.
+   * @param quadStreamInsert The quads to insert.
+   * @param quadStreamDelete The quads to delete.
+   */
+  update: (quadStreamInsert: AsyncIterator<RDF.Quad>, quadStreamDelete: AsyncIterator<RDF.Quad>) => Promise<void>;
+  /**
    * Graphs that should be deleted.
    * @param graphs The graph(s) in which all triples must be removed.
    * @param requireExistence If true, and any of the graphs does not exist, an error must be emitted.
--- a/packages/actor-rdf-update-hypermedia-patch-sparql-update/lib/QuadDestinationPatchSparqlUpdate.ts
+++ b/packages/actor-rdf-update-hypermedia-patch-sparql-update/lib/QuadDestinationPatchSparqlUpdate.ts
@@ -5,6 +5,7 @@ import { validateHttpResponse } from '@comunica/bus-rdf-update-quads';
 import type { IActionContext } from '@comunica/types';
 import type * as RDF from '@rdfjs/types';
 import type { AsyncIterator } from 'asynciterator';
+import { EmptyIterator } from 'asynciterator';
 import { Headers } from 'cross-fetch';
 import { termToString } from 'rdf-string-ttl';
 import { Readable } from 'readable-stream';
@@ -29,31 +30,28 @@ export class QuadDestinationPatchSparqlUpdate implements IQuadDestination {
   }
 
   public insert(quads: AsyncIterator<RDF.Quad>): Promise<void> {
-    return this.wrapSparqlUpdateRequest('INSERT', quads);
+    return this.wrapSparqlUpdateRequest(quads);
   }
 
-  public async delete(quads: AsyncIterator<RDF.Quad>): Promise<void> {
-    return this.wrapSparqlUpdateRequest('DELETE', quads);
+  public delete(quads: AsyncIterator<RDF.Quad>): Promise<void> {
+    return this.wrapSparqlUpdateRequest(undefined, quads);
   }
 
-  public async wrapSparqlUpdateRequest(type: 'INSERT' | 'DELETE', quads: AsyncIterator<RDF.Quad>): Promise<void> {
-    // Wrap triples in DATA block
-    const dataWrapped = quads
-      .map((quad: RDF.Quad) => {
-        let stringQuad = `${termToString(quad.subject)} ${termToString(quad.predicate)} ${termToString(quad.object)} .`;
-        if (quad.graph.termType === 'DefaultGraph') {
-          stringQuad = `  ${stringQuad}\n`;
-        } else {
-          stringQuad = `  GRAPH ${termToString(quad.graph)} { ${stringQuad} }\n`;
-        }
-        return stringQuad;
-      })
-      .prepend([ `${type} DATA {\n` ])
-      .append([ '}' ]);
+  public update(quadsToInsert: AsyncIterator<RDF.Quad>, quadsToDelete: AsyncIterator<RDF.Quad>): Promise<void> {
+    return this.wrapSparqlUpdateRequest(quadsToInsert, quadsToDelete);
+  }
+
+  private async wrapSparqlUpdateRequest(
+    quadsToInsert?: AsyncIterator<RDF.Quad>,
+    quadsToDelete?: AsyncIterator<RDF.Quad>,
+  ): Promise<void> {
+    // Create combined query stream with quads to insert and delete
+    const queryStream = this.createCombinedQuadsQuery(quadsToInsert, quadsToDelete);
+
     const readable = new Readable();
     readable._read = () => true;
-    dataWrapped.on('data', (quad: RDF.Quad) => readable.push(quad));
-    dataWrapped.on('end', () => readable.push(null));
+    queryStream.on('data', (chunk: string) => readable.push(chunk));
+    queryStream.on('end', () => readable.push(null));
 
     // Send data in PUT request
     const headers: Headers = new Headers({ 'content-type': 'application/sparql-update' });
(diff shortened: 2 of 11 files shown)


rubensworks (approved):
Thanks @karelklima, looks like a perfect solution to me!

@jitsedesmet could you have a look at this one as well, to check if it resolves your problem as well?

jitsedesmet:
Hmm, I checked and for me, it does not solve the issue. Interestingly, though, For my use case it seems to only add the resource... :/

I am serving a minimal resource like:
```n-triples
<http://localhost:3000/pods/00000000000000000096/posts/2024-05-08#416608218494388> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://localhost:3000/www.ldbc.eu/ldbc_socialnet/1.0/vocabulary/Post> .
<http://localhost:3000/pods/00000000000000000096/posts/2024-05-08#416608218494388> <http://localhost:3000/www.ldbc.eu/ldbc_socialnet/1.0/vocabulary/id> "416608218494388"^^<http://www.w3.org/2001/XMLSchema#long> .
```

A query:
```sparql
prefix ns1: <http://localhost:3000/www.ldbc.eu/ldbc_socialnet/1.0/vocabulary/>
prefix xsd: <http://www.w3.org/2001/XMLSchema#>

DELETE {
    ?id ns1:id "416608218494388"^^xsd:long .
} INSERT {
      ?id ns1:id "416608218494389"^^xsd:long .
} where {
    ?id ns1:id "416608218494388"^^xsd:long .
}
```

I would expect the result
```n-triples
<http://localhost:3000/pods/00000000000000000096/posts/2024-05-08#416608218494388> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://localhost:3000/www.ldbc.eu/ldbc_socialnet/1.0/vocabulary/Post> .
<http://localhost:3000/pods/00000000000000000096/posts/2024-05-08#416608218494388> <http://localhost:3000/www.ldbc.eu/ldbc_socialnet/1.0/vocabulary/id> "416608218494389"^^<http://www.w3.org/2001/XMLSchema#long> .
```

On master I only see the remove, resulting in:
```n-triples
<http://localhost:3000/pods/00000000000000000096/posts/2024-05-08#416608218494388> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://localhost:3000/www.ldbc.eu/ldbc_socialnet/1.0/vocabulary/Post> .
```

Using the solution provided by @karelklima I get:
```n-triples
<http://localhost:3000/pods/00000000000000000096/posts/2024-05-08#416608218494388> <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://localhost:3000/www.ldbc.eu/ldbc_socialnet/1.0/vocabulary/Post> .
<http://localhost:3000/pods/00000000000000000096/posts/2024-05-08#416608218494388> <http://localhost:3000/www.ldbc.eu/ldbc_socialnet/1.0/vocabulary/id> "416608218494388"^^<http://www.w3.org/2001/XMLSchema#long> .
<http://localhost:3000/pods/00000000000000000096/posts/2024-05-08#416608218494388> <http://localhost:3000/www.ldbc.eu/ldbc_socialnet/1.0/vocabulary/id> "416608218494389"^^<http://www.w3.org/2001/XMLSchema#long> .
```
Seems like here, only the delete works.

Full context: I am using the Community Solid Server to serve the resources, so the issue might be a compatibility issue.

rubensworks:
@jitsedesmet I also suspect we may be running into another problem there. Could you inspect the produced SPARQL queries that are sent to the CSS server? And in what order they are sent? That will probably allow us to determine whether it's the same problem or a different one.

karelklima:
I have done testing using CSS server **WITH my patch**.

I am serving resource `http://localhost/test` with content:
```sparql
<https://x/s> <https://x/p> <https://x/v1> .
<https://x/s> <https://x/p> <https://x/v2> .
```

Running the following update query:
```sparql
DELETE {
    <https://x/s> <https://x/p> <https://x/v2> .
}
INSERT {
    <https://x/s> <https://x/p> <https://x/v3> .
}
WHERE {
    <https://x/s> <https://x/p> <https://x/v2> .
}
```

Resulting in:
```sparql
<https://x/s> <https://x/p> <https://x/v1> .
<https://x/s> <https://x/p> <https://x/v3> .
```

Seems like it behaves as expected in this case, as the query correctly replaced value `v2` with `v3`.

When I run the code without the patch, the `v2` triple is just gone, which is consistent with my findings.

The script I am running: https://gist.github.com/karelklima/283f119629da41283b92c916922c0bba

One thing I noticed though - after the DELETE/INSERT query, when I query the pod file using CONSTRUCT to list all triples, Comunica does not give me updated data, but rather the original content - I assume that's because of some caching going on somewhere.

karelklima:
I also ran the test more similar to what @jitsedesmet tried:

Input:
```
<https://x/s> <https://x/p> <https://x/v1> .
```

Query:
```
DELETE {
    <https://x/s> <https://x/p> <https://x/v1> .
}
INSERT {
    <https://x/s> <https://x/p> <https://x/v2> .
}
WHERE {
    <https://x/s> <https://x/p> <https://x/v1> .
}
```

Result:
```
<https://x/s> <https://x/p> <https://x/v2> .
```

Seems fine to me. Any idea what may be the issue?

rubensworks:
Thanks for checking @karelklima!

@jitsedesmet Is it possible you are executing the query differently somehow?

> I assume that's because of some caching going on somewhere.

Indeed, that can be fixed (for now) by calling `engine.invalidateCache()`.

jitsedesmet:
I just tried again, and now it seems to work :/ (Sad smiley because I'm convinced I am doing the same thing.)
Since It's wrapped in a `Promise.all` I guess the promise handling is not always the same? I can't really explain it since I cannot recreate the problem. I am wondering whether we could just do one http request instead of awaiting two? (But that might be out of scope for this PR.)

(The only thing I changed in my system was installing `nodemon` and `node-ts` on my machine, but that could not possibly change the behaviour, right?)

karelklima:
@jitsedesmet I think you are right! I completely missed that `Promise.all` part. It seems like we are dealing with a race condition here. The execution of the promises is semi-deterministic here, depends on the destination. For `N3.Store` it behaves stable enough it seems, as there are no I/O lags and Promises evaluate right away, but in general the order of execution is non-deterministic.

Long story short - the `destination.insert` may start before `destination.delete` is finished. It works perfectly fine if the quads are distinct, but if there is an intersection then the result is basically random.

I propose the following: let's split the `Promise.all` so that we delete first, and insert new data only after the delete is finished. That is (as far as I can tell) the only way to ensure consistency.

Something like this:
```ts
const execute = (): Promise<void> => Promise.all([
    action.quadStreamDelete ? destination.delete(action.quadStreamDelete) : Promise.resolve(),
    action.deleteGraphs ?
      destination.deleteGraphs(
        action.deleteGraphs.graphs,
        action.deleteGraphs.requireExistence,
        action.deleteGraphs.dropGraphs,
      ) :
      Promise.resolve(),
  ]).then(() =>Promise.all([
    action.quadStreamInsert ? destination.insert(action.quadStreamInsert) : Promise.resolve(),
    action.createGraphs ?
      destination.createGraphs(action.createGraphs.graphs, action.createGraphs.requireNonExistence) :
      Promise.resolve(),
  ]));
```

What do you think?

jitsedesmet:
I agree with you, but I'd suggest waiting on Rubens confirmation because awaiting the deletion will also be a burden on IO delays. I suspect that we could resolve this issue in a way that requires only a single HTTP request., thus not burdening us with an increasing IO delay... :)

karelklima:
Sounds good. In order to do that, we will likely need to add a method to `IQuadDestination` interface. For example:
```typescript
interface IQuadDestination {
  update: (quadsToInsert: AsyncIterator<RDF.Quad>, quadsToDelete: AsyncIterator<RDF.Quad>) => Promise<void>;
}
```
Then we will be able to handle both fetch SPARQL and PATCH SPARQL update in one HTTP request.

karelklima:
I have implemented (hopefully) a more correct version of the DELETE/INSERT operations for various data destinations. That is by no means a final version, but rather a ground for a follow up discussion.

1) I am not sure whether the PATCH sparql update conforms to specification
2) There is a lot of duplicate code for regular SPARQL query and PATCH variant - that is by design as I expect more discrepancies in the final version
3) I am sure that the use of the async iterators can be improved

What do you think?

My original one-liner grew quite a bit :-)

packages/bus-rdf-update-quads/lib/IQuadDestination.ts
```diff
@@ -21,6 +21,12 @@ export interface IQuadDestination {
    * @return {AsyncIterator<RDF.Quad>} The deleted quad stream.
    */
   delete: (quads: AsyncIterator<RDF.Quad>) => Promise<void>;
+  /**
+   * Updates destination with quad stream to insert and quad stream to delete.
+   * @param quadStreamInsert The quads to insert.
+   * @param quadStreamDelete The quads to delete.
+   */
+  update: (quadStreamInsert: AsyncIterator<RDF.Quad>, quadStreamDelete: AsyncIterator<RDF.Quad>) => Promise<void>;
```
rubensworks, line 29: Perhaps we can even remove the `insert` and `delete` methods, and specify `update` as having insert and delete streams each as optional.

I would also propose being able to call it as follows: `update({ insert: streamA, delete: streamB })`. (just so the order of args is not arbitrary)

rubensworks (changes requested):
Thanks @karelklima!

While the proposed changes look great (aside from the minor comments below), they are breaking changes to the API, so we can only merge it for the next major update, which might take a while.

So I would propose the following:

1. To create a PR on the `master` branch with the removal of Promise.all, and instead run the insert/delete promises in sequence. This will be slower of course, but at least won't cause race conditions.
2. Create a separate PR on the `next/major` branch with this breaking change (including my comments). As soon as we're ready for the next major update (probably later this year when @jitsedesmet is ready with refactoring expressions), we can include this breaking change.

Does this sound ok to you @karelklima?

rubensworks:
> Could you please elaborate how these changes are breaking to the API? If we keep the insert and delete methods on the destination interface, then we are not removing anything. Or is the addition of the update method a concern?

Indeed, the addition of `update` is the problem, as other (external) `IQuadDestination` impls would not adhere anymore.
We could also make the method optional, but that would require some unclean workarounds.

karelklima:
Closing this PR, as it is effectively replaced by #1328 and #1331.

PR #1328, opened 2024-04-23
Hotfix DELETE/INSERT update operation order

### LINKED ISSUE #1301: DELETE/INSERT is deleting the quad but not updating the same value in the data store.

karelklima:
Resolves #1301.

This is a follow up work to #1326. It includes a hotfix to prevent race condition when inserting and deleting quads to/from destination. Delete and update operations are now being run in sequence.

### DIFF
--- a/packages/bus-rdf-update-quads/lib/ActorRdfUpdateQuadsDestination.ts
+++ b/packages/bus-rdf-update-quads/lib/ActorRdfUpdateQuadsDestination.ts
@@ -51,23 +51,20 @@ export abstract class ActorRdfUpdateQuadsDestination extends ActorRdfUpdateQuads
     destination: IQuadDestination,
     action: IActionRdfUpdateQuads,
   ): Promise<IActorRdfUpdateQuadsOutput> {
-    const execute = (): Promise<void> => Promise.all([
-      action.quadStreamDelete ? destination.delete(action.quadStreamDelete) : Promise.resolve(),
-      action.deleteGraphs ?
+    const execute = async(): Promise<void> => {
+      await (action.quadStreamDelete ? destination.delete(action.quadStreamDelete) : Promise.resolve());
+      await (action.deleteGraphs ?
         destination.deleteGraphs(
           action.deleteGraphs.graphs,
           action.deleteGraphs.requireExistence,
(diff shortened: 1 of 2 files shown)


packages/bus-rdf-update-quads/lib/ActorRdfUpdateQuadsDestination.ts
```diff
@@ -52,7 +52,6 @@ export abstract class ActorRdfUpdateQuadsDestination extends ActorRdfUpdateQuads
     action: IActionRdfUpdateQuads,
   ): Promise<IActorRdfUpdateQuadsOutput> {
     const execute = (): Promise<void> => Promise.all([
-      action.quadStreamInsert ? destination.insert(action.quadStreamInsert) : Promise.resolve(),
```
rubensworks, line 55: Since the race condition can probably also occur between `quadStreamDelete` and `deleteGraphs` (and also inserts), we may just want to await each operation separately.

Something like this should work:
```
const execute = async (): Promise<void> => {
  await action.quadStreamDelete ? destination.delete(action.quadStreamDelete);
  await action.deleteGraphs ?
        destination.deleteGraphs(
          action.deleteGraphs.graphs,
          action.deleteGraphs.requireExistence,
          action.deleteGraphs.dropGraphs,
        );
  ...
}
```

karelklima: I changed it as you suggested, although I am not sure about the exact order of operations that we should perform. I went with `delete -> deleteGraphs -> createGraphs -> insert`. When we introduce `update`, we might have a problem.

rubensworks (approved):
Looks great, thanks @karelklima!

PR #1329, opened 2024-04-23
Migrate from readable-web-to-node-stream to readable-from-web

surilindur:
Here is a small set of changes to migrate from `@smessie/readable-web-to-node-stream` to `readable-from-web`.

I had to adjust the tests a little, as well, including one `ActorHttpWayback` test that was missing the `closed` promise and the `cancel` function from its mock implementation of WHATWG ReadableStream.

The `@smessie/readable-web-to-node-stream` package is still in yarn.lock because it is required by 2.x versions of Comunica, which are required by `rdf-parse`, which are required by `componentsjs`.

Is there something that is not obvious, but that should also be checked or adjusted with this change? :thinking:


rubensworks:
Thanks @surilindur, this is great!

PR #1331, opened 2024-04-23
Fix DELETE/INSERT update operation order [next/major]

### LINKED ISSUE #1301: DELETE/INSERT is deleting the quad but not updating the same value in the data store.

karelklima:
Resolves #1301 properly.

This is a follow up work to #1326 and #1328.

It includes a proper fix to prevent race condition when inserting and deleting quads to/from destination. Delete and update operations are now being run in sequence and batched into single SPARQL queries when applicable for extra performance gain.

This PR introduces breaking changes in `IQuadsDestination` interface.

### DIFF
--- a/packages/actor-rdf-update-hypermedia-patch-sparql-update/lib/QuadDestinationPatchSparqlUpdate.ts
+++ b/packages/actor-rdf-update-hypermedia-patch-sparql-update/lib/QuadDestinationPatchSparqlUpdate.ts
@@ -5,6 +5,7 @@ import { validateHttpResponse } from '@comunica/bus-rdf-update-quads';
 import type { IActionContext } from '@comunica/types';
 import type * as RDF from '@rdfjs/types';
 import type { AsyncIterator } from 'asynciterator';
+import { ArrayIterator, EmptyIterator } from 'asynciterator';
 import { Headers } from 'cross-fetch';
 import { termToString } from 'rdf-string-ttl';
 import { Readable } from 'readable-stream';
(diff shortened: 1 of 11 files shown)


packages/actor-rdf-update-hypermedia-patch-sparql-update/lib/QuadDestinationPatchSparqlUpdate.ts
```diff
@@ -5,6 +5,7 @@ import { validateHttpResponse } from '@comunica/bus-rdf-update-quads';
 import type { IActionContext } from '@comunica/types';
 import type * as RDF from '@rdfjs/types';
 import type { AsyncIterator } from 'asynciterator';
+import { ArrayIterator, EmptyIterator } from 'asynciterator';
```
rubensworks, line 8: `EmptyIterator` is not safe to use (with comunica's enforcement on `autoStart: false`).
Instead, we should always use `ArrayIterator([], { autoStart: false })`.
See https://github.com/RubenVerborgh/AsyncIterator/issues/25

packages/actor-rdf-update-hypermedia-patch-sparql-update/lib/QuadDestinationPatchSparqlUpdate.ts
```diff
@@ -28,17 +29,30 @@ export class QuadDestinationPatchSparqlUpdate implements IQuadDestination {
...
+    await this.wrapSparqlUpdateRequest(queryStream);
   }
 
-  public async delete(quads: AsyncIterator<RDF.Quad>): Promise<void> {
-    return this.wrapSparqlUpdateRequest('DELETE', quads);
+  private createCombinedQuadsQuery(
+    quadsToInsert?: AsyncIterator<RDF.Quad>,
+    quadsToDelete?: AsyncIterator<RDF.Quad>,
+  ): AsyncIterator<string> {
+    return new EmptyIterator<string>()
```
rubensworks, line 44: Here as well.

packages/actor-rdf-update-hypermedia-sparql/lib/QuadDestinationSparql.ts
```diff
@@ -35,17 +36,29 @@ export class QuadDestinationSparql implements IQuadDestination {
...
+    await this.wrapSparqlUpdateRequest(queryStream);
   }
 
-  public async delete(quads: AsyncIterator<RDF.Quad>): Promise<void> {
-    return this.wrapSparqlUpdateRequest('DELETE', quads);
+  private createCombinedQuadsQuery(
+    quadsToInsert?: AsyncIterator<RDF.Quad>,
+    quadsToDelete?: AsyncIterator<RDF.Quad>,
+  ): AsyncIterator<string> {
+    return new EmptyIterator<string>()
```
rubensworks, line 51: Here as well.

rubensworks (changes requested):
Thanks @karelklima, just some minor comments regarding the use of `EmptyIterator`.
Should be good to merge after that.

rubensworks:
Super, thanks @karelklima!

Once we get a reply from OCE, we can get your bounty on #1301 sorted out. So it's definitely yours to claim :-)

PR #1332, opened 2024-04-24
Update fetch-sparql-endpoint to v5

surilindur:
This is just a small change to bump the `fetch-sparql-endpoint` library to version 5 now that it has been released.

Any feedback is welcome.


rubensworks:
Just like #1330, could you target `next/major`?

PR #1333, opened 2024-04-24
Tooling idea: use yarn over npm CLI where possible

surilindur:
This is an idea that I made a PR for to get some feedback, hence the title and the draft status.

Essentially, not everyone will likely have the NPM CLI installed (like myself), so making the basic scripts use Yarn (build, lint, pre-commit hooks) would help those people avoid having to adjust the package.json file(s) between cloning and installing. Yarn is required for the workspaces features in either case, so it is reasonable to expect it to be present.

Since `pre-commit` expects NPM CLI to also be present, that could be replaced with `husky`.

There are some publish scripts that still have npm in them, because the publishing features between the two are apparently not fully identical and the idea is not to break anything.


rubensworks:
Makes sense indeed!
Would it make sense to also immediately enable corepack so that people installing the comunica dev environment immediately have yarn installed as well?

> There are some publish scripts that still have npm in them, because the publishing features between the two are apparently not fully identical and the idea is not to break anything.

Yeah, I agree.

rubensworks:
I was referring to this: https://github.com/LinkedSoftwareDependencies/Components-Generator.js/blob/master/package.json#L4

surilindur:
Oh right, yes, that is definitely doable! But it only works if people have corepack enabled already, I think. I will look into it when I re-target the PR.

surilindur:
Was it okay for this one to be merged into master? I think I mixed up some branches and re-targeted this by accident. :sweat:

rubensworks:
Thanks!

Could you also update the generator templates in the same way? https://github.com/comunica/generate-comunica

Issue #1312, opened 2024-03-12
Query with multiple `COUNT` clauses returns incorrect result depending on order in projection


pmaria:
#### Issue type:

- :bug: Bug 

____
#### Description:

Given the following input
```trig
@prefix ex: <http://example.org/> .

<http://foo.org/id/graph/foo> {
  <http://foo.org/id/object/foo>
    ex:foo ex:Foo ;
    ex:bar [] ;
    ex:baz "baz1", "baz2", "baz3" .
}
```

running this  query:
```sparql
SELECT (COUNT(DISTINCT ?s) AS ?subjects) (COUNT(DISTINCT ?o) as ?objects) ?p
FROM <http://foo.org/id/graph/foo>
{
  ?s ?p ?o .
}
GROUP BY ?p
```

results in

![image](https://github.com/comunica/comunica/assets/12606875/f61c867e-c4d0-45f9-9334-71dcbf4db839)

while running this query (object count switched with subject count in projection):

```sparql
SELECT (COUNT(DISTINCT ?o) as ?objects) (COUNT(DISTINCT ?s) AS ?subjects) ?p
FROM <http://foo.org/id/graph/foo>
{
  ?s ?p ?o .
}
GROUP BY ?p
```

results in the expected

![image](https://github.com/comunica/comunica/assets/12606875/9421f0d4-e028-494b-b9d6-3a28cef3e3f5)
____
#### Environment:

| software         | version
| ---------------- | -------
| Comunica Engine  | 2.10.2 
| node             | v21.6.1
| npm              | 10.2.4
| yarn             | _Yarn is unavailable_
| Operating System | linux (Linux 5.15.133.1-microsoft-standard-WSL2)

**NOTE** I'm using `npx comunica-sparql-file-http` on the input file above.

I've tried this on several other SPARQL implementations and these do not show this behavior.

#### Crash log:

none

simonvbrae:
As discussed with @rubensworks I will work on this issue.

simonvbrae:
@jitsedesmet I was told to ping you for my questions related to the expressions code.

One thing that stuck out to me in the case where counts are incorrect is that startTerm [(code)](https://github.com/comunica/comunica/blob/052ac7d24db46a2056b0b7d694d2e8c20b2ede6d/packages/expression-evaluator/lib/evaluators/AsyncAggregateEvaluator.ts#L28) equals the subject of the current binding, rather than the predicate which we are supposed to be counting.

My question is what does startTerm mean?

[...]

PR #1335, opened 2024-05-05
Fixed a bug where the order of count operations influenced the result

simonvbrae:
Fixed a bug where the order of count operations influenced the result. See https://github.com/comunica/comunica/issues/1312

The problem was that we avoided counting a binding twice, even though it was for different aggregate expressions.
I solved it by hashing the binding with the expression variable and checking that this combination doesn't occur twice.

I'm not sure about the best practices when it comes to hashing, so it would be great if someone could update me on that.

### DIFF
--- a/engines/query-sparql/test/QuerySparql-test.ts
+++ b/engines/query-sparql/test/QuerySparql-test.ts
@@ -798,29 +798,44 @@ SELECT ?obsId {
           ],
         };
 
-        const received = (await arrayifyStream(await engine.queryBindings(`
-        SELECT (COUNT(DISTINCT ?s) AS ?subjects) (COUNT(DISTINCT ?o) as ?objects) ?p
-        FROM <http://foo.org/id/graph/foo>
-        {
-          ?s ?p ?o .
-        }
-        GROUP BY ?p
-        `, context)));
+        const expectedResult = [
+          [
+            [ DF.variable('objects'), DF.literal('1', DF.namedNode('http://www.w3.org/2001/XMLSchema#integer')) ],
+            [ DF.variable('p'), DF.namedNode('http://example.org/foo') ],
+            [ DF.variable('subjects'), DF.literal('1', DF.namedNode('http://www.w3.org/2001/XMLSchema#integer')) ],
+          ],
+          [
+            [ DF.variable('objects'), DF.literal('1', DF.namedNode('http://www.w3.org/2001/XMLSchema#integer')) ],
+            [ DF.variable('p'), DF.namedNode('http://example.org/bar') ],
+            [ DF.variable('subjects'), DF.literal('1', DF.namedNode('http://www.w3.org/2001/XMLSchema#integer')) ],
+          ],
+          [
+            [ DF.variable('objects'), DF.literal('3', DF.namedNode('http://www.w3.org/2001/XMLSchema#integer')) ],
+            [ DF.variable('p'), DF.namedNode('http://example.org/baz') ],
+            [ DF.variable('subjects'), DF.literal('1', DF.namedNode('http://www.w3.org/2001/XMLSchema#integer')) ],
+          ],
+        ];
 
-        const expected = (await arrayifyStream(await engine.queryBindings(`
+        const bindings1 = (await arrayifyStream(await engine.queryBindings(`
         SELECT (COUNT(DISTINCT ?o) as ?objects) (COUNT(DISTINCT ?s) AS ?subjects) ?p
         FROM <http://foo.org/id/graph/foo>
         {
           ?s ?p ?o .
         }
         GROUP BY ?p
-        `, context)));
+        `, context))).map(binding => [ ...binding ].sort(([ var1, _c1 ], [ var2, _c2 ]) => var1.value.localeCompare(var2.value)));
 
-        expect(received).toHaveLength(expected.length);
-        for (const [ index, receivedBindings ] of received.entries()) {
-          expect([ ...receivedBindings.entries ].sort((a, b) => a[0].localeCompare(b[0])))
-            .toEqual([ ...expected[index].entries ].sort((a, b) => a[0].localeCompare(b[0])));
+        const bindings2 = (await arrayifyStream(await engine.queryBindings(`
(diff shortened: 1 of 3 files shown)


packages/actor-query-operation-group/lib/GroupsState.ts
```diff
@@ -86,11 +87,13 @@ export class GroupsState {
         await Promise.all(this.pattern.aggregates.map(async(aggregate) => {
           // If distinct, check first whether we have inserted these values already
           if (aggregate.distinct) {
-            const hash = this.hashBindings(bindings);
-            if (this.distinctHashes!.get(groupHash)!.has(hash)) {
+            const bindingsHash = this.hashBindings(bindings);
+            const hashBindingsAndAggregate = createHash('sha256')
```
rubensworks, line 91: We should avoid directly hashing in this actor.
All hashing should be delegated to `bus-hash-bindings`, so downstream users can configure different hashing methods.

packages/actor-query-operation-group/lib/GroupsState.ts
```diff
@@ -86,11 +87,13 @@ export class GroupsState {
...
+              .update(aggregate.variable.value).update(bindingsHash).digest().toString();
```
rubensworks, line 92: Instead of including the aggregate variable name, wouldn't it also be possible to just scope the hashes to each variable separately. E.g. `distinctHashes: Record<variable name, Map<BindingsHash, Set<BindingsHash>>>`

rubensworks (changes requested):
Could you also add a unit test for this?

Also asking @jitsedesmet to have a look at this one.

packages/actor-query-operation-group/lib/GroupsState.ts
```diff
@@ -86,11 +87,13 @@ export class GroupsState {
         await Promise.all(this.pattern.aggregates.map(async(aggregate) => {
           // If distinct, check first whether we have inserted these values already
           if (aggregate.distinct) {
```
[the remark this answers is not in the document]
simonvbrae: Thank you for this. Distinct handling in the expression evaluator does work as expected even though we don't supply it a variable to group by. That's because the GroupsState makes sure we use a seperate aggregator per aggregate.variable.

The new commit I'm pushing implements this solution, it solves the issue I referenced above.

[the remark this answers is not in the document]
jitsedesmet: Amazing! Tnx!

rubensworks:
@simonvbrae Just pinging you again regarding the unit test in case you missed my previous comment.

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -775,6 +775,54 @@ SELECT ?obsId {
...
+        };
+
+        const received = (await arrayifyStream(await engine.queryBindings(`
+        SELECT (COUNT(DISTINCT ?s) AS ?subjects) (COUNT(DISTINCT ?o) as ?objects) ?p
+        FROM <http://foo.org/id/graph/foo>
+        {
+          ?s ?p ?o .
+        }
+        GROUP BY ?p
+        `, context))).sort((a, b) => a - b);
```
rubensworks, line 808: Does this sort do anything? a and b will be bindings objects I suspect?
(also below)

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -775,6 +775,54 @@ SELECT ?obsId {
...
+        SELECT (COUNT(DISTINCT ?o) as ?objects) (COUNT(DISTINCT ?s) AS ?subjects) ?p
+        FROM <http://foo.org/id/graph/foo>
+        {
+          ?s ?p ?o .
+        }
+        GROUP BY ?p
+        `, context))).sort((a, b) => a - b);
+
+        expect(received).toHaveLength(expected.length);
+        for (const [ index, receivedBindings ] of received.entries()) {
```
rubensworks, line 820: I'm not really sure I understand what this code does. Are you still working with bindings objects?
I'm not convinced this is actually correct 😅 

I think we can just manually do something like `expect(expected[0]).toEqual(received[0]);` and so on.

engines/query-sparql/test/QuerySparql-test.ts
simonvbrae: I meant to use the default sorting function which converts to string and compares strings. I've updated it.

engines/query-sparql/test/QuerySparql-test.ts
simonvbrae: expected and received are indeed bindings objects
I wanted to make sure the order in which these bindings objects are returned doesn't matter within the test, which is why I tried to sort the array of bindings objects in `received` and `expected`. For now the test succeeds without sorting as well.
(I sorted the entries in each bindings object according to the keys `subject` `object` and `predicate` because this order differs between these queries.)

engines/query-sparql/test/QuerySparql-test.ts
simonvbrae: Thank you for spotting my mistake, I meant to use the default algorithm for '.sort()' which converts to strings but the linter required an explicit sort algorithm to be passed. Now I'm not sure: should I leave out the sort (which succeeds) or explicitly convert to string and sort the strings?

engines/query-sparql/test/QuerySparql-test.ts
rubensworks: Let me attempt to clarify my issue with this test:
At the moment it is difficult to intuitively understand the expected output from this test.
What IMO would clarify things, is to just directly test all expected bindings (there should only be a few I guess) separately to precise values.
This would allow the code reader to see exactly what the expected output should be.

simonvbrae: Sorry for the misunderstanding. I've added an explicit test result, I wanted to compare both queries to the same variable, that way it would be clear at first glance that both queries in the test should yield the same result. So that explains why I sorted the query results.

rubensworks:
Excellent, thanks @simonvbrae!

PR #1341, opened 2024-05-15
feat: implement opt plus

### LINKED ISSUE #825: Implement OPT+ operator
#### Issue type:

____
#### Description:

-----

## Bounty

| |
|-|
| €1088 |

jeswr:
Closes #825.

This is a working implementation of #825. I have the following design questions, once answered by @rubensworks, I will cleanup the implementation and add more extensive testing.

1. Is the use of the `union` between the `left` stream and `inner`  join a satisfactory way of implementing `OPT+` (I believe so, and don't expect a custom iterator to improve performance much).
2. Are you happy having the logic performing the `OPT+` operation inline in the left join, or would you like this to be an actor of its own, noting that it is difficult for this implementation to implement the `ActorRdfJoin#getJoinCoefficients` method due to the fact that all of the work of joining is delegated.
3. Is the CLI arg name `--optplus` desired for enabling the `@comunica/bus-query-operation:optPlus` context key.

### DIFF
--- a/engines/query-sparql/test/QuerySparql-test.ts
+++ b/engines/query-sparql/test/QuerySparql-test.ts
@@ -1,14 +1,14 @@
 /** @jest-environment setup-polly-jest/jest-environment-node */
 
 import { QuerySourceSkolemized } from '@comunica/actor-context-preprocess-query-source-skolemize';
-import { KeysHttpWayback, KeysQuerySourceIdentify } from '@comunica/context-entries';
+import { KeysHttpWayback, KeysQueryOperation, KeysQuerySourceIdentify } from '@comunica/context-entries';
 import { BlankNodeScoped } from '@comunica/data-factory';
 import type { QueryBindings, QueryStringContext } from '@comunica/types';
 import type * as RDF from '@rdfjs/types';
 import 'jest-rdf';
 import arrayifyStream from 'arrayify-stream';
 import { Store } from 'n3';
-import { DataFactory } from 'rdf-data-factory';
+import { DataFactory, NamedNode, Variable } from 'rdf-data-factory';
 import { Factory } from 'sparqlalgebrajs';
 import { QueryEngine } from '../lib/QueryEngine';
 import { usePolly } from './util';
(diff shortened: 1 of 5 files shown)


rubensworks:
Thanks @jeswr!

> Is the use of the union between the left stream and inner join a satisfactory way of implementing OPT+ (I believe so, and don't expect a custom iterator to improve performance much).

I'm afraid not, as the paper says:

> Implementing the idea of the OPT+ operator by simply using the
semantically equivalent expressions with AND and UNION is
not an efficient approach

As mentioned in #825, we'll need the modified nested loop join algorithm from the paper (mNLJ+: section 6.3 of the paper).
This one performs the best according to experiments.

> Are you happy having the logic performing the OPT+ operation inline in the left join, or would you like this to be an actor of its own, noting that it is difficult for this implementation to implement the ActorRdfJoin#getJoinCoefficients method due to the fact that all of the work of joining is delegated.

Since mNLJ+ is quite different semantically, I'd put this in a separate actor.

> Is the CLI arg name --optplus desired for enabling the @comunica/bus-query-operation:optPlus context key.

If we move this to a separate actor, we don't need a new context key.

jeswr:
>  This one performs the best according to experiments.

It is my understanding that when NLJ is used for the inner join this implementation is equivalent to `mNLJ+` due to the way our AsyncIterators operate internally.  In particular, looking at the description `mNLJ+` we see

"mappings from the input iterator are materialized into a list. In addition to appending each such mapping to this list, the mapping is returned immediately as an output mapping"

this is what is happening under the hood assuming that the Nested Loop Join is selected to perform the join because the `#clone()` call is causing the iterator to implicitly build this array that is used for the outer iteration in the inner join whilst emitting the relevant elements.

Moreover, I believe if the goal is to optimise for having the most streaming results output as early as possible, then Comunica is actually going to do better if this implementation is used with the symmetric hash join selected because if `I_r` and `I_l` are both large remote streams then `OPT+` can return whatever of `I_r` has arrived + whatever joins between `I_r` and `I_l` are possible without either of `I_r` or `I_l` having to have completed. I think for cases like link traversal this is the best behavior.

Given that is the case my suggestion for how I proceed is that I implement 2 `OPT+` actors - one which takes the above union approach, one with the custom `mNLJ+` iterator. This would allow you to then benchmark these for different properties on link traversal use cases if you wished to do so.

rubensworks:
> This statement is made because they are assuming that results will be fetched from the remote source twice for I_l, this is not the case for the above implementation because of the use of #clone()

Aha, I see. Ok, let's go for your current approach then!

Shall we move it to a separate actor then? Not sure if we want this as a separate `actor-query-operation-leftjoin-optplus`, or on the join bus. Maybe the former makes more sense.

Once this is merged and released, it would be great to have a separate config in the link traversal repo that has this actor enabled by default, so we can see its impact on our example web client link traversal queries. Not required for the bounty, but just nice to have.

Issue #759, opened 2020-11-12
Ensure backpressure is maintained in streams


rubensworks:
#### Issue type:

- :bug: Bug 

____
#### Description:

There seem to be a couple of places where we are not adhering to Node's backpressure conventions with regards to streams.

Concretely, we seem to be creating new streams (such as `Readable`, `PassThrough` and `Transform`), and `push`-ing into them (via a `'data'` handler on another stream). Node handles backpressuring via the return value of `push`, which we are ignoring in this manner.

https://github.com/search?q=repo%3Acomunica%2Fcomunica%20on(%27data&type=code

A better solution would be to simply `pipe` instead of calling `push` on each data element.

Related to https://github.com/rubensworks/rdf-parse.js/commit/269c757935c54b388e1bde076dc29c2afc2e8e7b

-----

## Bounty

A [bounty](https://comunica.dev/association/bounties/) has been placed on this issue by:

| |
|-|
| [<img alt="Comunica Association" src="https://comunica.dev/img/comunica_red.svg" height="50">](https://comunica.dev/association/) |
| €1088 |

[Click here to learn more](https://comunica.dev/association/bounty_process/) if you're interested in claiming this bounty by resolving this issue.

jeswr:
As discussed with @rubensworks , I will work on this issue.

jacoscaz:
@jeswr @rubensworks insofar as possible, I would suggest making all internal interfaces between actors use `asynciterator` rather than Node and/or Web streams, which (IMHO) should only be used in public interfaces.

rubensworks:
> I would suggest making all internal interfaces between actors use asynciterator rather than Node and/or Web streams, which (IMHO) should only be used in public interfaces.

I wouldn't do this just yet, since this would cause a breaking change in the API.
Things like RDF parsers and serializers are fine to remain plain streams IMO.

PR #1345, opened 2024-05-17
backpressure: fix serialisation backpressure issues

jeswr:
This PR resolves backpressure issues in the serialisers (partly resolving #759).

It also makes it easy to migrate to having serialisers return `AsyncIterator`s rather than `Readable`s as we are now largely just wrapping `Readable`s around `AsyncIterator`s. If there is demand for consuming the serialised information as `AsyncIterator`s then one option would be to extend the output interface to be

```ts
export interface IActorQueryResultSerializeOutput extends IActorOutput {
  /**
   * A readable string stream in a certain SPARQL serialization that was serialized.
   */
  data: NodeJS.ReadableStream;
  
  iterator: AsyncIterator<string>;
}
```

and implement it such that the `data` key is a getter which only creates a readable stream wrapping the asynciterator when the `data` key is retrieved.

cc @rubensworks @jacoscaz

### DIFF
--- a/packages/actor-query-result-serialize-stats/lib/ActorQueryResultSerializeStats.ts
+++ b/packages/actor-query-result-serialize-stats/lib/ActorQueryResultSerializeStats.ts
@@ -9,6 +9,7 @@ import type {
   IQueryOperationResultBindings,
   IQueryOperationResultQuads,
 } from '@comunica/types';
+import { wrap } from 'asynciterator';
 import { Readable } from 'readable-stream';
 import type { ActionObserverHttp } from './ActionObserverHttp';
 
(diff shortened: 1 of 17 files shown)


packages/actor-query-result-serialize-stats/lib/ActorQueryResultSerializeStats.ts
```diff
@@ -45,37 +46,54 @@ export class ActorQueryResultSerializeStats extends ActorQueryResultSerializeFix
     data.push(`${header}\n`);
   }
 
+  /**
+   * @deprecated Use {@link createStat} instead.
+   */
+  /* istanbul ignore next */
```
rubensworks, line 52: I'd prefer not to have `istanbul ignore next` in the codebase.
I think it's safe to remove these methods, as they are not really meant to be part of the public API.
But if we keep them, we need to have a unit test for them.

rubensworks (changes requested):
Looks very good!

Since everything is an asynciterator now, could you open a PR to `next/major` after this PR has been merged to include this commit, and then also change the signature of the actor API to use `AsyncIterator` instead of `Readable`?

packages/actor-query-result-serialize-stats/lib/ActorQueryResultSerializeStats.ts
jeswr: This is resolved, will PR to next/major once this is merged

rubensworks:
Nice work, thanks @jeswr!
Feel free to claim the bounty by submitting an invoice through our open collective page.

PR #1346, opened 2024-05-17
HTTP cache actor

iMac7:
Issue #166

### DIFF
--- a/engines/config-query-sparql/config/http/actors-fetch.json
+++ b/engines/config-query-sparql/config/http/actors-fetch.json
@@ -3,7 +3,8 @@
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/runner/^3.0.0/components/context.jsonld",
 
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-http-fetch/^3.0.0/components/context.jsonld",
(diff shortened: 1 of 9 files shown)


rubensworks:
@iMac7 The idea would be that the user can supply a `Cache` object via the query context, which the query engine can then use during execution. This allows the user to be in full control over the cache.
When the user does not supply a cache object, the engine should create one internally, and the `http-cache-storage-lru` approach feels like a clean solution to me. It's important that this cache is internally _cached_ so that it's reused across multiple executions within the same query engine.

> I tested this locally by running requests on the terminal

This will create separate Node environments with separate query engine instances. It's not in-scope of this issue to handle caching across those instances. The main goal is that we can achieve caching across the same query engine instance, or with a user-supplied cache.

> Should the response sent from the cache actor be json or a readablestream or other format?

This must be a stream, so that comunica can keep on processing everything in a streaming manner.



> >   A simple LRUCache-based Cache implementation should be placed here if the user provided none in the context

> When testing requests, that linked part of the code appears to not be reached by logs, and I'm not sure what it does.

This is the new place: https://github.com/comunica/comunica/blob/master/packages/actor-context-preprocess-set-defaults/lib/ActorContextPreprocessSetDefaults.ts

rubensworks:
Hi @iMac7, just checking in to see how things are going, and if you're still up for completing this bounty.

rubensworks:
@iMac7 Ok, done! Thanks for the effort so far.

PR #1347, opened 2024-05-17
backpressure/internals

jeswr:
- **fix: fix backpressure on more internal actors**
- **chore: fix lint errors**

### DIFF
--- a/packages/actor-rdf-update-hypermedia-patch-sparql-update/lib/QuadDestinationPatchSparqlUpdate.ts
+++ b/packages/actor-rdf-update-hypermedia-patch-sparql-update/lib/QuadDestinationPatchSparqlUpdate.ts
@@ -50,9 +50,7 @@ export class QuadDestinationPatchSparqlUpdate implements IQuadDestination {
       .prepend([ `${type} DATA {\n` ])
       .append([ '}' ]);
     const readable = new Readable();
-    readable._read = () => true;
-    dataWrapped.on('data', (quad: RDF.Quad) => readable.push(quad));
-    dataWrapped.on('end', () => readable.push(null));
+    readable.wrap(<any> dataWrapped);
 
     // Send data in PUT request
     const headers: Headers = new Headers({ 'content-type': 'application/sparql-update' });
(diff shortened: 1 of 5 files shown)


packages/actor-rdf-update-hypermedia-patch-sparql-update/lib/QuadDestinationPatchSparqlUpdate.ts
```diff
@@ -50,9 +50,7 @@ export class QuadDestinationPatchSparqlUpdate implements IQuadDestination {
       .prepend([ `${type} DATA {\n` ])
       .append([ '}' ]);
     const readable = new Readable();
-    readable._read = () => true;
-    dataWrapped.on('data', (quad: RDF.Quad) => readable.push(quad));
-    dataWrapped.on('end', () => readable.push(null));
+    readable.wrap(<any> dataWrapped);
```
rubensworks, line 53: Are we sure `Readable.wrap` correctly handles backpressure? (I can't find much about it, which is why I'm asking)

rubensworks (approved):
Just one question, looks good to go then!

packages/actor-rdf-update-hypermedia-patch-sparql-update/lib/QuadDestinationPatchSparqlUpdate.ts
jeswr: Strictly speaking `Readable.wrap` is designed to be used for wrapping old-style stream. [As described in the docs](https://nodejs.org/api/stream.html#readablewrapstream) one of the required methods for `#wrap` to consume an iterator is `.pause()` which indicates that it is controlling the flow of the stream to handle backpressure.

This backpressure handling can be seen in the codebase https://github.com/nodejs/node/blob/d57af10d1be390f4e33e6efe596133de9da36c86/lib/internal/js_stream_socket.js#L70-L71

rubensworks: > [As described in the docs](https://nodejs.org/api/stream.html#readablewrapstream) one of the required methods for #wrap to consume an iterator is .pause() which indicates that it is controlling the flow of the stream to handle backpressure.

Ok, that should be enough indeed.

> Alternatively I could publish a micro-library implementing the function toReadable as follows, and use that instead of #wrap.

Do you think this would be a better approach? (I don't think it's needed myself, but I may be missing something)

rubensworks: Ok, let's keep `Readable.wrap` then.

rubensworks:
Thanks @jeswr!

PR #1353, opened 2024-05-23
fix: cancel http stream in wayback

jeswr:


rubensworks:
Thanks! Could you add a unit test?

Issue #1354, opened 2024-05-24
Stats writer produces wrong totals


rubensworks:
#### Issue type:

- :bug: Bug 

____
#### Description:

As of https://github.com/comunica/comunica/commit/0d7538cb154284af91a5027f89772bb3cc09f6fa, the stats writer produces wrong total times and result counts.

For example:
```
...
25809,16858.99227797985,45
25810,16859.01178896427,45
25811,16859.174924969673,45
25812,16859.219471931458,45
25813,16859.261418938637,45
25814,16859.264099001884,45
25815,16859.31561398506,45
25816,16859.519886016846,45
TOTAL,0.07797801494598389,21
```

Total time and request count is lower than those of the last result, which is impossible of course.

@jeswr, since this seems to be caused by your commit, could you look into fixing this? I suspect it will be a minor thing.

rubensworks:
I suspect the problem will also occur in other places, such as here: https://github.com/comunica/comunica/commit/0d7538cb154284af91a5027f89772bb3cc09f6fa#diff-3abd93e4d79e1224fa9a2ae3a4b3d3accc08c21c674ba1a6cce1bb4aa670070bR107

jeswr:
> I suspect the problem will also occur in other places, such as here: [0d7538c#diff-3abd93e4d79e1224fa9a2ae3a4b3d3accc08c21c674ba1a6cce1bb4aa670070bR107](https://github.com/comunica/comunica/commit/0d7538cb154284af91a5027f89772bb3cc09f6fa#diff-3abd93e4d79e1224fa9a2ae3a4b3d3accc08c21c674ba1a6cce1bb4aa670070bR107)
> 
> And _maybe_ even here: [0d7538c#diff-d14c48a122bbde2e2eebd2681ddabd7cd3124d5d42c2e80d1676b1cf3debdfceR111](https://github.com/comunica/comunica/commit/0d7538cb154284af91a5027f89772bb3cc09f6fa#diff-d14c48a122bbde2e2eebd2681ddabd7cd3124d5d42c2e80d1676b1cf3debdfceR111)

Right, so to check my understanding here `httpObserver.requests` is not a static value and thus we can only retrieve the value once we have consumed all of the incoming stream?

PR #1356, opened 2024-05-24
chore: cleanup integration tests and add serialisation test

jeswr:
Done in the process of trying to work out how to write tests to catch #1354. @rubensworks could you please advise on how to programatically get stats for the purpose of testing?

### DIFF
--- a/engines/query-sparql/test/QuerySparql-test.ts
+++ b/engines/query-sparql/test/QuerySparql-test.ts
@@ -23,7 +23,7 @@ describe('System test: QuerySparql', () => {
   usePolly();
 
   let engine: QueryEngine;
-  beforeEach(() => {
+  beforeAll(() => {
     engine = new QueryEngine();
   });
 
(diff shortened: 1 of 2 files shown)


engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -75,6 +75,11 @@ describe('System test: QuerySparql', () => {
           .length).toBeGreaterThan(100);
         expect((await arrayifyStream(await (<QueryBindings> await engine.query(query, context)).execute()))
           .length).toBeGreaterThan(100);
+
```
rubensworks, line 78: Why was this added? Does the same as above?

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -92,81 +97,61 @@ describe('System test: QuerySparql', () => {
           .toEqual([]);
         await expect((arrayifyStream(await (<QueryBindings> await engine.query(query, context)).execute()))).resolves
           .toEqual([]);
+
```
rubensworks, line 100: Same question here.

rubensworks (commented):
> @rubensworks could you please advise on how to programatically get stats for the purpose of testing?

Via the CLI it's via `-t stats`. For unit tests, you could should also be able to just call `resultsToString` with `stats`.

PR #1359, opened 2024-05-26
chore: reduce use of transform

jeswr:
`#map` is much more efficient when `#transform` due to https://github.com/RubenVerborgh/AsyncIterator/pull/75.


rubensworks:
Nice, thanks @jeswr!

PR #1360, opened 2024-05-26
feat: bump CI workflow versions

KasperZutterman:
# What
Bump CI Workflow Versions

# Why
https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

# Testing
[CI Annotations](https://github.com/comunica/comunica/actions/runs/9243577720?pr=1360)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,15 +17,15 @@ jobs:
           - 20.x
     steps:
       - name: Use Node.js ${{ matrix.node-version }}
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node-version }}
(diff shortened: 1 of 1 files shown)


jeswr:
Might be worth adding a dependabot config to bump these automatically going forward https://github.com/jeswr/template-typescript/blob/53bd36e3df4e7ee9ccdd79f63e5a7f44ce90c928/.github/dependabot.yml#L14-L19

rubensworks:
We used to do this automatically via renovate, but I disabled it as it caused too much commit spam: https://github.com/rubensworks/renovate-presets/blob/master/js.json#L63C3-L65C4
But since we only seem to be using major version references in the worflow, we can probably re-enable it again though.

PR #1361, opened 2024-05-26
feat: opt plus

### LINKED ISSUE #825: Implement OPT+ operator

jeswr:
Supercedes  #1341 and closes #825


packages/actor-query-operation-leftjoin/lib/ActorQueryOperationLeftJoin.ts
```diff
@@ -2,16 +2,20 @@ import { BindingsFactory } from '@comunica/bindings-factory';
...
+  IJoinEntry,
+  IQueryOperationResultBindings,
+} from '@comunica/types';
 import type { Algebra } from 'sparqlalgebrajs';
 
-/**
- * A comunica LeftJoin Query Operation Actor.
- */
-export class ActorQueryOperationLeftJoin extends ActorQueryOperationTypedMediated<Algebra.LeftJoin> {
+export abstract class ActorQueryOperationLeftJoinAbstract
```
rubensworks, line 17: Hrmm, introducing new abstract actors to achieve common logic feels icky.
Comunica's architecture has been designed to achieve this through composition instead of extension.

This actually would be a good reason to move the optplus logic to the rdf-join bus. Then we don't have to duplicate the common join logic.
Do you agree?

jeswr: >  Do you agree?

That was my original intuition as well, but it doesn't work well in practice because the `getJoinCoefficients` method cannot be implemented when the actual `join` is being delegated to another actor - because there is no public API for getting the coefficients of the actor selected by the mediator.

rubensworks: > but it doesn't work well in practice because the getJoinCoefficients method cannot be implemented when the actual join is being delegated to another actor 

Right, I see the issue.

But, since this new join actor should be the only optional join actor in the bus, the join coefficients don't matter that much. So it may be ok here to just have dummy coefficients in there (`0` for all, and perhaps `n+m` for `iterations`).

Do you agree with this approach?

jeswr: Done in #1371.

PR #1362, opened 2024-05-26
fix: fix stats

### LINKED ISSUE #1354: Stats writer produces wrong totals

jeswr:
Closes #1354

### DIFF
--- a/packages/actor-query-result-serialize-stats/lib/ActorQueryResultSerializeStats.ts
+++ b/packages/actor-query-result-serialize-stats/lib/ActorQueryResultSerializeStats.ts
@@ -68,9 +68,14 @@ export class ActorQueryResultSerializeStats extends ActorQueryResultSerializeFix
 
     const startTime = this.now();
     let result = 1;
+    
+    const createFooter = (startTime: number) => this.createFooter(startTime);
+    function *end() {
+      yield createFooter(startTime)
+    }
     const stream = wrap(resultStream)
       .map(() => this.createStat(startTime, result++))
-      .append([ this.createFooter(startTime) ]);
+      .append(wrap(end()));
 
     this.pushHeader(data);
     data.wrap(<any> stream);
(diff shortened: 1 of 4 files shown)


packages/actor-query-result-serialize-stats/lib/ActorQueryResultSerializeStats.ts
```diff
@@ -68,9 +68,13 @@ export class ActorQueryResultSerializeStats extends ActorQueryResultSerializeFix
...
     const startTime = this.now();
     let result = 1;
+
+    function* end(cb: () => string): Generator<string> {
+      yield cb();
+    }
     const stream = wrap(resultStream)
       .map(() => this.createStat(startTime, result++))
-      .append([ this.createFooter(startTime) ]);
+      .append(wrap(end(() => this.createFooter(startTime))));
```
rubensworks, line 77: Have you manually confirmed that this produces the expected output now?

Because I would expect that `wrap` creates an asynciterator with `autoStart: true`, and therefore eagerly creates the footer before it's actually needed.
But I could be incorrect of course.

jeswr: It should use the [`WrappingIterator`](https://github.com/RubenVerborgh/AsyncIterator/blob/266e5deb20df6d9033f01c137bcafee6721b1215/asynciterator.ts#L2223) which has no buffering and thus no autostart, instead pulling elements on-demand.

Will double check with a `stats` integration test.

rubensworks: Aha, then I will probably be wrong indeed.

jeswr:
https://github.com/comunica/comunica/pull/1362/commits/89c7878b87be16310eab8a52c93c86027389f55f supercedes https://github.com/comunica/comunica/pull/1356 and removes the lines commented upon.

rubensworks:
Thanks! :-)

PR #1363, opened 2024-05-26
chore: enable windows and mac ci

### LINKED ISSUE #798: Run CI on windows and mac
#### Issue type:

____
#### Description:

As we have migrated to GitHub Actions to run our CI, it should be easy to run on Windows and Mac as well.
Essentially, we need to add `macos-latest` and `windows-latest` to the test job in our CI.
However, [tests seem to fail](https://github.com/comunica/comunica/runs/2174042973?check_suite_focus=true), which is why it isn't enabled yet.

May be related to #713 and #616.

-----

## Bounty

| |
|-|
| €272 |

jeswr:
Closes #798


.github/workflows/ci.yml
```diff
@@ -59,9 +94,11 @@ jobs:
...
           parallel: true
       - name: Run spec tests
-        run: npx lerna run spec --concurrency 1
-      - name: Run integration tests
-        run: npx lerna run integration --concurrency 1
+        run: cd engines/query-sparql && yarn run spec
+      - run: cd engines/query-sparql && yarn run integration
+      - run: cd engines/query-sparql-file && yarn run integration
+      - run: cd engines/query-sparql-rdfjs && yarn run integration
+      - run: cd engines/query-sparql-rdfjs-lite && yarn run integration
```
jeswr, line 101: I've done this due to there seeming to be an issue with the way `lerna run` plays with the `windows-latest` CI https://github.com/lerna/lerna/issues/4014.

.github/workflows/ci.yml
```diff
@@ -7,11 +7,41 @@ concurrency:
 
 jobs:
 
+  build:
```
rubensworks, line 10: While I see this speeds up builds, this skips building on windows (and mac), which is something we'll want to check as well.

.github/workflows/ci.yml
rubensworks: Could you add a TODO comment in the workflow file referring to the lerna issue together with the old `lerna run` command?

rubensworks (changes requested):

engines/query-sparql/test/util.ts
```diff
@@ -1,42 +1,26 @@
...
-    persisterOptions: { fs: { recordingsDir }},
-    recordFailedRequests: true,
-    matchRequestsBy: {
-      headers: {
-        exclude: [ 'user-agent' ],
-      },
-    },
-  });
-}
+export async function fetch(...args: Parameters<typeof fetchFn>): ReturnType<typeof fetchFn> {
```
rubensworks, line 8: I would prefer to stick with Polly if possible.
I'm afraid we'll introduce other problems with our custom script that Polly has had years to stabilize over.

If the problem is related to long filenames, then I suspect Polly will have a way to somehow shortening them.

.github/workflows/ci.yml
jeswr, line 101:
```suggestion
      #  TODO: Reintroduce `npx lerna run` command when https://github.com/lerna/lerna/issues/4014 is resolved
      #  run: npx lerna run spec --concurrency 1
        run: cd engines/query-sparql && yarn run spec
      #  TODO: Reintroduce `npx lerna run` command when https://github.com/lerna/lerna/issues/4014 is resolved
      #  npx lerna run integration --concurrency 1
      - run: cd engines/query-sparql && yarn run integration
      - run: cd engines/query-sparql-file && yarn run integration
      - run: cd engines/query-sparql-rdfjs && yarn run integration
      - run: cd engines/query-sparql-rdfjs-lite && yarn run integration
```

engines/query-sparql/test/util.ts
jeswr: > I had a go at extending this persistor to make my own custom persistor but it was non-trivial. I will briefly continue with that attempt.

In starting to look into this I've noticed that Polly is not caching all responses needed in e2e tests [and 23 tests fail when tests are run offline](https://pastebin.com/K9N8Yw75). This means that Comunica tests will start failing if any of fragments.dbpedia.org, xmlns.com or www.rubensworks.net are down and to me serves as a good argument for updating this mechanism anyway.

.github/workflows/ci.yml
jeswr: Note to self - it appears that the lerna issue is from the fact that there is a file called `lerna.js` in the root and so on windows `lerna ...` trys to open that file rather than run the `lerna` command.

PR #1371, opened 2024-06-02
feat: implement optplus as a join actor

### LINKED ISSUE #825: Implement OPT+ operator

jeswr:
Supercedes #1361 and closes #825


engines/config-query-sparql/config/rdf-join/actors.json
```diff
@@ -3,8 +3,7 @@
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/config-query-sparql/^3.0.0/components/context.jsonld"
   ],
   "import": [
-    "ccqs:config/rdf-join/actors-inner.json",
-    "ccqs:config/rdf-join/actors-minus.json",
+    "ccqs:config/rdf-join/actors-base.json",
```
rubensworks, line 6: I would not introduce an actors-base file, it's clearer IMO to just have the imports directly.

engines/config-query-sparql/config/rdf-join/actors-streaming.json
```diff
@@ -0,0 +1,9 @@
+{
```
rubensworks, line 1: All of our operators are streaming, so the filename is a bit of a misnomer.
"monotonic" would be better.

But maybe we don't need this file anyways. I suspect this is here for usage in the link traversal repo,
but we won't be able to import it because link traversal has a custom rdf-join actors file.

engines/config-query-sparql/config/config-default.json
```diff
@@ -3,59 +3,7 @@
...
-    "ccqs:config/rdf-resolve-hypermedia-links/mediators.json",
-    "ccqs:config/rdf-resolve-hypermedia-links-queue/actors.json",
-    "ccqs:config/rdf-resolve-hypermedia-links-queue/mediators.json",
-    "ccqs:config/rdf-serialize/actors.json",
-    "ccqs:config/rdf-serialize/mediators.json",
-    "ccqs:config/rdf-update-hypermedia/actors.json",
-    "ccqs:config/rdf-update-hypermedia/mediators.json",
-    "ccqs:config/rdf-update-quads/actors.json",
-    "ccqs:config/rdf-update-quads/mediators.json"
+    "ccqs:config/config-base.json",
```
rubensworks, line 6: See my comment in rdf-join, I don't think we need any changes to the `config-` files in this repo.

rubensworks (changes requested):
Just some minor config comments, everything else looks good!

rubensworks:
Super, this is perfect, thanks @jeswr!

PR #1377, opened 2024-06-16
chore: remove actor-http-native

jeswr:
It is already stated in the readme that this package is deprecated, so I don't see why it needs to keep having updated releases.

I am opening this PR to delete this now as this package use the `parse-link-header` package which is not ESM compatible, and so this package cannot be updated to ESM unless we do maintenance to change out this dependency as I am doing with other packages.


rubensworks:
This package used to have better performance than the fetch-based one.
It's been a while since I tested this, so this may be different now.
If it's not too much work to replace the `parse-link-header` dep in this package with another one, I'd rather keep it, for performance comparison reasons.
If it _is_ too much work, then we need a performance check first to confirm that the fetch-based impl is faster than this one.

jeswr:
It would appear that the difference is negligible on average

```bash
$ npx ts-node test.ts 
[class ActorHttpFetch extends ActorHttp] http://example.org
103.52
1.48
1.96
[class ActorHttpFetch extends ActorHttp] https://www.rubensworks.net/
97.87
1.15
1.46
[class ActorHttpNative extends ActorHttp] http://example.org
97.13
1.54
1.96
[class ActorHttpNative extends ActorHttp] https://www.rubensworks.net/
99.34
1.45
1.77
$ npx ts-node test.ts 
[class ActorHttpNative extends ActorHttp] http://example.org
90.09
1.74
2.24
[class ActorHttpNative extends ActorHttp] https://www.rubensworks.net/
86.54
1.43
1.78
[class ActorHttpFetch extends ActorHttp] http://example.org
87.36
1.39
1.93
[class ActorHttpFetch extends ActorHttp] https://www.rubensworks.net/
88.25
1.4
1.85
```

Results were obtained running
```ts
import { Bus, ActionContext } from '@comunica/core';
import { ActorHttpNative } from './packages/actor-http-native/lib/index';
import { ActorHttpFetch } from './packages/actor-http-fetch/lib/index';
import { ActorHttp } from '@comunica/bus-http';

async function run() {
    for (const Actor of [ActorHttpNative, ActorHttpFetch]) {
        for (const url of ['http://example.org', 'https://www.rubensworks.net/']) {
            const times: [number, number, number][] = [];
            for (let i = 0; i < 100; i++) {
                const fetch = new ActorHttpFetch({
                    name: 'actor-http',
                    bus: new Bus({
                        name: 'bus',
                    }),
                });
        
                let tim: [number, number, number] = [0, 0, 0]
                times.push(tim);
            
                const startTime = Date.now();
                const res = await fetch.run({ input: 'http://example.org', context: new ActionContext() });
                const endTime = Date.now();
                tim[0] = endTime - startTime;
            
                ActorHttp.toNodeReadable(res.body).on('data', (data: any) => {
                    tim[1] = Date.now() - endTime;
                }).on('end', () => {
                    tim[2] = Date.now() - endTime;
                    // console.log(tim);
                });
            }
        
            setTimeout(() => {
                console.log(Actor, url)
                console.log(times.map(t => t[0]).reduce((a, b) => a + b) / times.length);
                console.log(times.map(t => t[1]).reduce((a, b) => a + b) / times.length);
                console.log(times.map(t => t[2]).reduce((a, b) => a + b) / times.length);
            }, 500);
            }
        }
}

run();
```

rubensworks:
Ok, that's great!
There may also be a difference in browser performance, do you have insights on that?

jeswr:
So I had some variables wrong in the above experiment, turns out `ActorHttpFetch` is about 2x faster on node on average (c.f. https://github.com/jeswr/comunica-perf/actions/runs/9927282637/job/27422034531#step:6:11); which provides even more support to remove `actor-http-native`.

Trying to run browser tests but running into webpack issues that I'll create an issue for now.

rubensworks:
> So I had some variables wrong in the above experiment, turns out ActorHttpFetch is about 2x faster on node on average 

That's really good to hear!

> Trying to run browser tests but running into webpack issues that I'll create an issue for now.

https://www.npmjs.com/package/node-polyfill-webpack-plugin should resolve these issues.
If you already tried this, I would prefer if we could keep `actor-http-native`, but somehow exclude it from ESM-ification. Would this be easy to achieve?

jeswr:
> https://www.npmjs.com/package/node-polyfill-webpack-plugin should resolve these issues.

I've already tried that on [this branch](https://github.com/jeswr/comunica-perf/tree/webpack-attempt).

> If you already tried this, I would prefer if we could keep actor-http-native, but somehow exclude it from ESM-ification. Because we don't know what the impact on browsers is atm. Would this be easy to achieve?

It should be feasible to exclude. That said, I'm not sure I see the benefit of keeping it given that `actor-http-native` is not used in any configurations in this repo and I'm also not sure would successfully bundle in any other projects due to the webpack issues demonstrated above.

rubensworks:
> It should be feasible to exclude.

Let's exclude it then.

> That said, I'm not sure I see the benefit of keeping it given that actor-http-native is not used in any configurations in this repo and I'm also not sure would successfully bundle in any other projects due to the webpack issues demonstrated above.

I'm sure the issues can be fixed (but this is definitely not required for this PR and bounty).
I just want to keep the package around for future browser-based experiments (especially for TPF/QPF, as we used the native actor for them in the past). Once we have proof that fetch is faster in browsers as well, we can safely remove the actor.

PR #1378, opened 2024-06-16
BREAKING CHANGE: remove cross fetch

jeswr:


surilindur:
I do not know if this removal is related, but `cross-fetch` has been removed in `next/major` (due to being a breaking change), or at least it was removed, in case it has been reverted since (#1330). Just dropping this comment here since I noticed the PR.

rubensworks:
Indeed, this is already included in `next/major`.

@jeswr is this a strong requirement for the ESM stuff? If so, then it might be better to rebase your work onto `next/major`. (in that case, we may first want to merge `master` to it again)

@surilindur Can you double-check that all of the changes in this PR are also in `next/major` already?
If so, we can close this PR.

surilindur:
I went through all the changes and they all seem to have been in that other PR already. The wording of the readme in actor-http-fetch is different between the two, however, so I think that should be checked. The version from here seems to be better.

rubensworks:
Ok, thanks for checking @surilindur!

PR #1382, opened 2024-06-24
Bindings merge optimization

maartyman:
## Summary by CodeRabbit

- **Bug Fixes**
	- Updated test assertions for binding comparisons to ensure logical equivalence.
  
- **Tests**
	- Restructured test cases for merging bindings to enhance clarity and coverage.
	- Improved organization of test cases for both RDF and native bindings.
  
- **Refactor**
	- Enhanced merging logic in the Bindings class for improved efficiency and clarity.

### DIFF
--- a/engines/query-sparql/test/QuerySparql-test.ts
+++ b/engines/query-sparql/test/QuerySparql-test.ts
@@ -1,12 +1,15 @@
 /** @jest-environment setup-polly-jest/jest-environment-node */
 
 import { QuerySourceSkolemized } from '@comunica/actor-context-preprocess-query-source-skolemize';
+import type { Bindings } from '@comunica/bindings-factory';
+import { BindingsFactory } from '@comunica/bindings-factory';
 import { KeysHttpWayback, KeysQuerySourceIdentify } from '@comunica/context-entries';
 import { BlankNodeScoped } from '@comunica/data-factory';
 import type { QueryBindings, QueryStringContext } from '@comunica/types';
 import type * as RDF from '@rdfjs/types';
 import arrayifyStream from 'arrayify-stream';
 import 'jest-rdf';
+import '@comunica/jest';
 import { Store } from 'n3';
 import { DataFactory } from 'rdf-data-factory';
 import { Factory } from 'sparqlalgebrajs';
@@ -17,6 +20,7 @@ import { fetch as cachedFetch } from './util';
 const stringifyStream = require('stream-to-string');
 
 const DF = new DataFactory();
+const BF = new BindingsFactory();
 const factory = new Factory();
 
 globalThis.fetch = cachedFetch;
@@ -656,24 +660,22 @@ describe('System test: QuerySparql', () => {
             FILTER(lang(?name) = 'nl')
           }
         }`, { sources: [ store ]});
-        const results = await arrayifyStream(await result.execute());
 
-        const expectedResult = [
-          [
+        const expectedResult: Bindings[] = [
+          BF.bindings([
             [ DF.variable('s'), DF.namedNode('http://ex.org/Pluto') ],
             [ DF.variable('p'), DF.namedNode('http://ex.org/type') ],
             [ DF.variable('o'), DF.namedNode('http://ex.org/Dog') ],
-          ],
-          [
+          ]),
+          BF.bindings([
             [ DF.variable('name'), DF.literal('Lorem ipsum', 'nl') ],
             [ DF.variable('s'), DF.namedNode('http://ex.org/Mickey') ],
             [ DF.variable('p'), DF.namedNode('http://ex.org/name') ],
             [ DF.variable('o'), DF.literal('Lorem ipsum', 'nl') ],
-          ],
+          ]),
         ];
 
-        const bindings = results.map(binding => [ ...binding ]);
-        expect(bindings).toMatchObject(expectedResult);
+        await expect(result.execute()).resolves.toEqualBindingsStream(expectedResult);
       });
     });
 
--- a/packages/bindings-factory/lib/Bindings.ts
+++ b/packages/bindings-factory/lib/Bindings.ts
@@ -107,55 +107,84 @@ export class Bindings implements RDF.Bindings {
   }
 
   public merge(other: RDF.Bindings | Bindings): Bindings | undefined {
-    // Collect entries
-    const entries: [string, RDF.Term][] = [];
+    let entries = this.entries;
     let nonCompatibleElements = false;
-    // Benchmarking this function showed that foreach was faster than for...of
-    // eslint-disable-next-line unicorn/no-array-for-each
-    other.forEach((right, variable) => {
-      if (nonCompatibleElements) {
-        return;
-      }
-      const left = this.entries.get(variable.value)!;
-      if (left && !left.equals(right)) {
-        // Set nonCompatibleElements to true to prevent further processing
-        nonCompatibleElements = true;
-      } else {
-        entries.push([ variable.value, right ]);
-      }
-    });
+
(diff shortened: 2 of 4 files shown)


maartyman:
I'm still optimizing the function for when the bindings are of type `Bindings` instead of `RDF.Bindings`.

rubensworks:
@maartyman I see tests are failing due to the last commit.

maartyman:
@rubensworks, yes, see the review message I added.

maartyman:
As you see in the CI, some ActorRdfJoin.joinBindings function tests fail, this is because I switched from making an array of entries `const entries: [string, RDF.Term][] = [];` to using the immutable.js features `let entries = this.entries;` to build the entries of the new bindings. This causes the entries to not have the "ownerID": {} field. I'm not sure what to do, and would like your input on this @rubensworks.

rubensworks:
Maybe we shouldn't use `toEqual` for those tests, and instead use https://github.com/comunica/comunica/blob/master/packages/bindings-factory/lib/Bindings.ts#L76

maartyman:
Okay, this pr is ready. Some Noes: I do only use the largest bindings as a base if that binding is of the native comunica bindings type. Because of this change, I had to change the way one of the integration tests checks equality between results. I also added some tests to evaluate these optimizations.

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -1,12 +1,15 @@
...
 import { QuerySourceSkolemized } from '@comunica/actor-context-preprocess-query-source-skolemize';
+import type { Bindings } from '@comunica/bindings-factory';
+import { BindingsFactory } from '@comunica/bindings-factory';
 import { KeysHttpWayback, KeysQuerySourceIdentify } from '@comunica/context-entries';
 import { BlankNodeScoped } from '@comunica/data-factory';
 import type { QueryBindings, QueryStringContext } from '@comunica/types';
 import type * as RDF from '@rdfjs/types';
 import arrayifyStream from 'arrayify-stream';
 import 'jest-rdf';
+import '@comunica/jest';
```
rubensworks, line 12: I'm surprised this works correctly.
We deliberately excluded this package, as this file is also tested in browsers using karma.
But the CI seems to pass, so I guess it's fine...

To be sure, can you double-check this by running the browser tests locally? (see the CI file on how to do this)

rubensworks (commented):
Looks great!

For posterity reasons, could you share the performance results here in this PR?

engines/query-sparql/test/QuerySparql-test.ts
maartyman: To be double sure, I can just copy the logic of toEqualBindingsStream and test it that way, your call?

rubensworks: Well, if it works correctly in the browser environment, then I guess we can keep the import there.

rubensworks: @maartyman Could you check locally if the browser tests run fine for you?
Because I'd love to merge this!

maartyman: Will do! I was recently checking the performance and saw that the difference between `for` and `foreach` is actually pretty negligible. As the linter prefers `for`, I will change this and post the benchmarking results. This is the reason I haven't finalized this pr yet...

rubensworks: Sounds good!

The reason I'm asking is because I was doing some profiling with a flame graph, and the (old) merge method is a significant bottleneck for one of our WatDiv queries. When using your version, the flame graph shows a reduction of 50% for that part of the graph. So that's quite something.

maartyman:
I measured the time it took to merge 2 bindings in a few scenarios with different sizes of bindings. The "amount of overlap between bindings" is how many variables are both in the first and second bindings. I included the `for` and `foreach` results (the `for` was eventually chosen due to similar results with `foreach` and the preference of the linter).

Scenarios: (size of first binding, size of second binding, amount of overlap between bindings)
S1:  5,  2,  1
S2:  20, 2,  1
S3:  40, 2,  1
S4:  2,  5,  1
S5:  2,  20, 1
S6:  2,  40, 1
S7:  5,  5,  0
S8:  5,  5,  1
S9:  5,  5,  5
S10: 20, 20, 5
S11: 20, 20, 10
S12: 20, 20, 20

results: (original, with a for loop, with foreach loop)
S1:  8353,  2339,  2378
S2:  22130, 3790,  3584
S3:  33578, 4096,  3789
S4:  7370,  1596,  1336
S5:  21530, 3739,  3551
S6:  36340, 3991,  3970
S7:  26958, 19318, 18287
S8:  23080, 15446, 15175
S9:  8586,  3613,  3488
S10: 31309, 17661, 16766
S11: 30566, 19696, 19189
S12: 26719, 16411, 16812

rubensworks:
That looks great!

For the sake of reference in case we need it in the future, could you share the script you used to measure this? (can be just a gist)

maartyman:
And I can confirm that the browser tests pass on my local machine.

maartyman:
And this is the script for benchmarking:
```
let BindingsFactory = require('../comunica/packages/bindings-factory').BindingsFactory;
let DataFactory = require('rdf-data-factory').DataFactory;
const fs = require('fs');

const DF = new DataFactory();
const BF = new BindingsFactory();

function makeBindings(size, random, start = 0) {
    let entries = [];
    for (let i = start; i < size + start; i++) {
        entries.push([DF.variable('a' + random + i), DF.namedNode('ex:a' + random + i) ])
    }
    return BF.bindings(entries);
}

let sizes = [
    {size1: 5, size2: 2, overlap: 1},
    {size1: 20, size2: 2, overlap: 1},
    {size1: 40, size2: 2, overlap: 1},
    {size1: 2, size2: 5, overlap: 1},
    {size1: 2, size2: 20, overlap: 1},
    {size1: 2, size2: 40, overlap: 1},
    {size1: 5, size2: 5, overlap: 0},
    {size1: 5, size2: 5, overlap: 1},
    {size1: 5, size2: 5, overlap: 5},
    {size1: 20, size2: 20, overlap: 5},
    {size1: 20, size2: 20, overlap: 10},
    {size1: 20, size2: 20, overlap: 20},
]

//warmup
async function warmup() {
    for (let size of sizes) {
        for (let i = 0; i < 100; i++) {
            let random = Math.random();
            let bindings1 = makeBindings(size.size1, random);
            let bindings2 = makeBindings(size.size2, random, size.size1 - size.overlap);
            bindings1.merge(bindings2);
            await new Promise(resolve => setTimeout(resolve, 0));
        }
        await new Promise(resolve => setTimeout(resolve, 1));
    }
}
warmup().then(() => {
    let allTimes = [];
    let repeat = 10_000;
    for (let size of sizes) {
        let times = 0;
        for (let i = 0; i < repeat; i++) {
            let random = Math.random();
            let bindings1 = makeBindings(size.size1, random);
            let bindings2 = makeBindings(size.size2, random, size.size1 - size.overlap);

            let start = process.hrtime();
            bindings1.merge(bindings2);
            let end = process.hrtime(start);
            times += end[1];
        }
        allTimes.push(times/repeat)
    }

    let csvContent = ""
    for (const size of sizes) {
        csvContent += JSON.stringify(size) + ";"
    }
    csvContent += "\n"
    for (const time of allTimes) {
        csvContent += time + ";"
    }
    csvContent += "\n"

    fs.writeFile('merge-benchmark/foreach.csv', csvContent, 'utf8', (err) => {
        if (err) {
            console.error('Error writing to CSV file', err);
        } else {
            console.log('CSV file has been saved.');
        }
    });
})
```

rubensworks:
Super, thanks @maartyman! 🚀

PR #1386, opened 2024-06-29
Extract function operators to actors

jitsedesmet:
This PR aims to extract all function operators to actors using a script. I will track the script in this PR so it is easy to validate the steps takes to achieve the extracted actors.

### DIFF
--- a/packages/actor-function-factory-not/lib/ActorFunctionFactoryNot.ts
+++ b/packages/actor-function-factory-not/lib/ActorFunctionFactoryNot.ts
@@ -1,0 +1,23 @@
+import type { IActionFunctionFactory, IActorFunctionFactoryOutput, IActorFunctionFactoryArgs } from '@comunica/bus-function-factory';
+import { ActorFunctionFactory } from '@comunica/bus-function-factory';
+import type { IActorTest } from '@comunica/core';
+import * as C from '@comunica/expression-evaluator/lib/util/Consts';
+import { Not } from './NotFunction';
+
+/**
+ * A comunica Not Function Factory Actor.
+ */
+export class ActorFunctionFactoryNot extends ActorFunctionFactory {
+  public constructor(args: IActorFunctionFactoryArgs) {
+    super(args);
+  }
+
+  public async test(action: IActionFunctionFactory): Promise<IActorTest> {
+    // Does support action.requireTermExpression, so no need to check for that.
+    return action.functionName === C.RegularOperator.NOT;
+  }
+
+  public async run(action: IActionFunctionFactory): Promise<IActorFunctionFactoryOutput> {
+    return new Not();
+  }
+}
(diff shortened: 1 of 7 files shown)


packages/actor-function-factory-not/lib/ActorFunctionFactoryNot.ts
```diff
@@ -0,0 +1,23 @@
...
+ * A comunica Not Function Factory Actor.
+ */
+export class ActorFunctionFactoryNot extends ActorFunctionFactory {
+  public constructor(args: IActorFunctionFactoryArgs) {
+    super(args);
+  }
+
+  public async test(action: IActionFunctionFactory): Promise<IActorTest> {
+    // Does support action.requireTermExpression, so no need to check for that.
+    return action.functionName === C.RegularOperator.NOT;
```
rubensworks, line 17: This should throw an error for unsupported operators. Merely a boolean is not enough.

packages/actor-function-factory-equality/lib/Equality.ts
```diff
@@ -0,0 +1,104 @@
...
+  RDFEqualTypeError,
+} from '@comunica/expression-evaluator';
+
+/**
+ * https://www.w3.org/TR/sparql11-query/#func-RDFterm-equal
+ */
+export class Equality extends RegularFunction {
+  protected arity = 2;
+
+  public operator = RegularOperator.EQUAL;
```
jitsedesmet, line 32: Currently, each  operat

jitsedesmet: >  Not necessarily in this PR, but we need a plan, which this PR needs to be compatible with.

I agree. I also agree with the union of string. But I also think we should no longer differentiate between regular and special operators.

rubensworks: > But I also think we should no longer differentiate between regular and special operators.

If that's feasible, that's ok for me!

jitsedesmet: I think it should be feasable. This whole PR a a breaking change so I don't think it'll matter :))

jitsedesmet:
To be considered: the "exists" function. :)

packages/expression-evaluator/lib/expressions/SpecialOperator.ts
```diff
@@ -1,14 +0,0 @@
-import type { FunctionApplication } from '../functions/OverloadTree';
-import type { Expression, SpecialOperatorExpression } from './Expressions';
-import {
-  ExpressionType,
-} from './Expressions';
-
-export class SpecialOperator implements SpecialOperatorExpression {
```
jitsedesmet, line 7: @rubensworks I was able to delete this expression type as a result of our previous discussion.

I think a similar approach can be taken for the namedFunctions: https://github.com/jitsedesmet/comunica/blob/chore/bussify-all-function/packages/expression-evaluator/lib/expressions/Named.ts

The difference is that namedFunctions also have a `name` presented as a namedNode.
SparqlOperator obviously do not have this, but they could contain a name as a string (like `=`). I would like to also generalize named functions to a string (so that internally, they are approached just like a function with a URI as name). I think this will make for cleaner code. I also don't see an issue with this generalization because we want to look beyond SPARQL anyway.

PR #1387, opened 2024-07-03
Feature/statistics tracking

RubenEschauzier:
This PR introduces a system to track statistics during query execution and exposes these statistics to other components using event emitters.
The process works as follows:

1.	The `actor-context-preprocess-set-defaults` adds an empty `Map<IActionContextKey<any>, any>` object to the context.
2.	Other `actor-context-preprocess` actors can add statistics trackers to this map.

These statistics trackers must either implement an interface that extends `IStatistic` or implement `IStatisticAggregate`, which should be defined in `@comunica/types/lib/IStatistic.ts.`
The `IStatistic` interface requires an `eventEmitter` and a function to retrieve this emitter. This setup allows actors with access to the context to add listeners to these emitters, gaining access to data produced by the statistics trackers.
The `IAggregateStatistic` interface is designed for trackers that need to combine data from multiple statistics trackers, such as tracking queried topology combined with intermediate result statistics. This interface requires the attribute `toAggregate: Record<string, IStatistic>`, which defines the statistics the object will aggregate. As `IAggregateStatistic` extends the base `IStatistic` interface, it also exposes an `eventEmitter`.
This is a draft PR as I’m unsure of several design choices:
1.	Interface Genericity:

- How generic should the interfaces be?
- Should an interface be restricted to one type of event logged in Comunica?
- The advantage of having a specific interface for each log is that when a statistic is output in Comunica, the function generating the statistic can call a single function with all the data available, delegating processing to the (possibly user-defined) statistic tracker. However, this might reduce the interface's generality.

2.	Multiplicity of Statistic Instances:

- Should the `statisticMap` allow multiple different instances of the same base interface (e.g., IStatisticDiscoveredLinks)?

3.	Scope of Implementation:

- Currently, I have implemented trackers for discover and dereference events, which are mainly useful for link traversal. Which parts of the PR should be in Comunica, and which parts should be in link traversal?

4.	Typing of Events:

- I am not sure how to properly type the events. Currently, events are typed as `any`. Ideally, a statistics interface would also specify the type of data to be emitted, but this might limit the interface's generality.


packages/types/lib/IStatistic.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export interface IStatistic {
+  /**
+   * Query string of tracked execution
+   */
+  query: string, 
+
+  statisticEvents: EventEmitter;
+
+  getEmitter: () => EventEmitter;
```
rubensworks, line 16: I did a quick search here, and it looks like you're only emitting on-data events.
So for type-safety, I would propose just adding the following two methods instead:
```
addListener(cb: () => void): void;
emit(data: T): void;
```
For the `T`, you'll have to add a generic type to the `IStatistic` to indicate the type of data that is emitted.

packages/types/lib/IStatistic.ts
```diff
@@ -0,0 +1,101 @@
...
+import type { ILink } from '@comunica/bus-rdf-resolve-hypermedia-links';
+import type { ActionContextKey } from '@comunica/core';
+import type { IActionContext } from './IActionContext';
+import type { IQuerySource } from './IQuerySource';
+
+export interface IStatistic {
+  /**
+   * Query string of tracked execution
+   */
+  query: string,
```
rubensworks, line 12: This seems odd to have here. Can it be removed?

packages/actor-context-preprocess-set-defaults/lib/ActorContextPreprocessSetDefaults.ts
```diff
@@ -35,8 +39,21 @@ export class ActorContextPreprocessSetDefaults extends ActorContextPreprocess {
         .setDefault(KeysQuerySourceIdentify.sourceIds, new Map())
         .setDefault(KeysCore.log, this.logger)
         .setDefault(KeysInitQuery.functionArgumentsCache, this.defaultFunctionArgumentsCache)
-        .setDefault(KeysQuerySourceIdentify.hypermediaSourcesAggregatedStores, new Map());
+        .setDefault(KeysQuerySourceIdentify.hypermediaSourcesAggregatedStores, new Map())
+        .setDefault(KeysStatisticsTracker.statistics, new Map());
 
+      if (context.get(KeysStatisticsTracker.statisticsSaveLocation)) {
```
rubensworks, line 45: Not sure if this really belongs in Comunica (it's not browser-safe). Could be a post-processing step outside of Comunica I guess?

packages/context-entries/lib/Keys.ts
```diff
@@ -294,3 +299,42 @@ export const KeysRdfJoin = {
    */
   lastPhysicalJoin: new ActionContextKey<string>('@comunica/bus-rdf-join:lastPhysicalJoin'),
 };
+
+export const KeysStatisticsTracker = {
+  /**
+   * Map holding all statistics being tracked, filled with keys from KeysTrackableStatistics
+   */
+  statistics: new ActionContextKey<Map<IActionContextKey<any>, any>>(
```
rubensworks, line 307: I would introduce a new class/type for these statistics, as `Map<IActionContextKey<any>, any>` is not that nice, and not type-safe.

packages/types/lib/IStatistic.ts
```diff
@@ -0,0 +1,101 @@
...
+   * Query string of tracked execution
+   */
+  query: string, 
+
+  statisticEvents: EventEmitter;
+
+  getEmitter: () => EventEmitter;
+}
+
+export interface IAggregateStatistic extends IStatistic {
```
rubensworks, line 19: I wonder if we really need a dedicated interface for this?
The stats that want to do aggregation can just do it? There's nowhere where an `IAggregateStatistic` is explicitly required I guess?

rubensworks (commented):
Below some comments in-line, and some answers to your questions:

> Should an interface be restricted to one type of event logged in Comunica?

I would say yes.

> Should the statisticMap allow multiple different instances of the same base interface (e.g., IStatisticDiscoveredLinks)?

I would say no.

> Currently, I have implemented trackers for discover and dereference events, which are mainly useful for link traversal. Which parts of the PR should be in Comunica, and which parts should be in link traversal?

TPF queries also traverse in base comunica, so I guess all of these would be applicable here?

> I am not sure how to properly type the events. Currently, events are typed as any. Ideally, a statistics interface would also specify the type of data to be emitted, but this might limit the interface's generality.

Yeah, we don't want no `any` :-)

packages/types/lib/IStatistic.ts
RubenEschauzier: I agree that it's odd, my reasoning was if an engine instance is used for multiple queries in a row and the user wants to track the statistics computed by comunica, a statistics logger might want to output the query associated with the statistic. However, this should be removed if we're not using the logger.

packages/types/lib/IStatistic.ts
RubenEschauzier: Right, that does make more sense.

packages/context-entries/lib/Keys.ts
RubenEschauzier: It might be overkill, but could it also be an action context?

packages/types/lib/IStatistic.ts
RubenEschauzier: No, I guess not, if every type of statistic tracking will have a dedicated interface. It might be needed to define a `getTrackedStatistic(): any` function in the IStatistic interface then, as you would want all statistic trackers to expose their information to other parts of Comunica.

packages/actor-context-preprocess-set-defaults/lib/ActorContextPreprocessSetDefaults.ts
rubensworks: I also don't think CLI access is super useful.

packages/context-entries/lib/Keys.ts
rubensworks: Probably won't work, as that one is immutable, whereas we do want something mutable here.
But something similar type-wise could work.

packages/types/lib/IStatistic.ts
RubenEschauzier: On second thought it's not needed, as the emitters already expose all data.

RubenEschauzier:
I see tests fail due to me removing the `networkCache/` contents and adding it to .gitignore. How does this work, were the contents  supposed to be tracked?

rubensworks:
> I see tests fail due to me removing the networkCache/ contents and adding it to .gitignore. How does this work, were the contents supposed to be tracked?

These files must definitely not be ignored, and are to be checked into git.

packages/actor-context-preprocess-statistic-link-discovery/lib/StatisticLinkDiscovery.ts
```diff
@@ -0,0 +1,84 @@
...
+    this.log = log;
+  }
+
+  public setDiscoveredLink(link: ILink, parent: ILink): boolean {
+    // If self-edge or duplicate edge we don't track.
+    if (link.url === parent.url || this.edgeList.has(JSON.stringify([ parent.url, link.url ]))) {
+      return false;
+    }
+
+    this.edgeList.add(JSON.stringify([ parent.url, link.url ]));
```
rubensworks, line 41: Why the stringify? Can't we keep the original objects?

packages/actor-context-preprocess-statistic-link-discovery/lib/StatisticLinkDiscovery.ts
```diff
@@ -0,0 +1,84 @@
...
+  public edgeList: Set<string>;
+
+  // Metadata is saved as follows: First key indicates what url this metadata belongs to, while value is
+  // a list of all metadata objects recorded. This list can contain multiple Records as links can be discovered from
+  // multiple data sources.
+  public metadata: Record<string, Record<any, any>[]>;
+
+  // Emitter that indicates when a new discover statistic is recorded. This can be used downstream for, e.g, aggregate
+  // statistic trackers
+  public statisticEvents: EventEmitter;
```
rubensworks, line 19: We probably don't need to make this public. (applies to all statistics)

packages/actor-context-preprocess-statistic-link-discovery/lib/StatisticLinkDiscovery.ts
```diff
@@ -0,0 +1,84 @@
+// eslint-disable-next-line import/no-nodejs-modules
+import { EventEmitter } from 'events';
```
rubensworks, line 2: Since the name of this package clashes with a node built-in (it's missing from your package.json's), we'll always have linter issues with it.
Maybe it's not worth it to use it, and just implement a basic listener system ourselves.
We can probably implement a `StatisticBase` or something similar for this.

packages/actor-context-preprocess-statistic-link-discovery/lib/StatisticLinkDiscovery.ts
```diff
@@ -0,0 +1,84 @@
...
+    const discoverEventData: IDiscoverEventData = {
+      edge: [ parent.url, link.url ],
+      metadataChild: this.metadata[link.url],
+      metadataParent: this.metadata[parent.url],
+    };
+
+    this.emit(discoverEventData);
+
+    // Log info if available
+    this.log('Discover Event', () => ({
```
rubensworks, line 65: Is log still needed if we already have emit?
I would suspect logging can be implemented on top of emit.

packages/actor-context-preprocess-aggregate-statistic-traversed-topology/lib/AggregateStatisticTraversedTopology.ts
```diff
@@ -0,0 +1,152 @@
...
+  IStatistic,
+  IStatisticDereferencedLinks,
+  IStatisticDiscoveredLinks,
+  ITopologyEventNotification,
+} from '@comunica/types';
+
+export class AggregateStatisticTraversedTopology implements IStatistic<ITopologyEventNotification> {
+  // Emitter that indicates when an update to the aggregate statistic happened
+  public statisticEvents: EventEmitter;
+  public toAggregate: StatisticsHolder;
```
rubensworks, line 17: This one doesn't seem really required.

packages/actor-context-preprocess-aggregate-statistic-traversed-topology/lib/AggregateStatisticTraversedTopology.ts
```diff
@@ -0,0 +1,152 @@
...
+
+    return true;
+  }
+
+  public addListener(cb: (arg0: ITopologyEventNotification) => void): void {
+    this.statisticEvents.addListener('data', cb);
+  }
+
+  public processDiscoverEvent(data: IDiscoverEventData): void {
+    this.edgeList.add(JSON.stringify(data.edge));
```
rubensworks, line 58: Stringify not needed?

rubensworks (changes requested):
Here's some initial feedback. I can do a more detailed pass after these comments are resolved.

packages/actor-context-preprocess-set-defaults/lib/index.ts
```diff
@@ -1 +1,2 @@
 export * from './ActorContextPreprocessSetDefaults';
+export * from './StatisticsHolder';
```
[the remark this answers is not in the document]
RubenEschauzier: The `StatisticsHolder` class is used for typing the `StatisticsHolder` object obtained from the action context. I think it should be exported then?

eslint.config.js
```diff
@@ -15,6 +15,7 @@ module.exports = config([
       // Default
       'unicorn/consistent-destructuring': 'off',
       'unicorn/no-array-callback-reference': 'off',
+      'unicorn/prefer-node-protocol': 'off',
```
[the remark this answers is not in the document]
rubensworks: > I can remove this line and instead only use // eslint-disable-next-line import/no-nodejs-modules if you want?

That would be better indeed. But given one of my later comments, maybe we don't need EventEmitter at all.

packages/actor-context-preprocess-set-defaults/lib/index.ts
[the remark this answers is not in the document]
rubensworks: Yes, that's right.

packages/actor-context-preprocess-statistic-link-discovery/lib/StatisticLinkDiscovery.ts
RubenEschauzier: Right, which could also have typed events, as I believe is done in CSS. Should I create a separate package for this or where do you prefer to put it?

rubensworks: I would start by implementing it here, and if needed, we can move it to a separate dedicated package.
We don't need anything too complicated here I think.

RubenEschauzier: This actor only tracks discover statistics, so we would create a between-actor dependency for other statistics (like link dereferences). I can start with that for now, but it would need to be refactored in the future.

RubenEschauzier: I implemented the StatisticLinkDiscovery with the StatisticBase you proposed. Is this what you meant? If so I can place the StatisticBase in `bus-context-preprocess` as all statistic tracker actors will use this package.

rubensworks: I have some suggestions on class/method naming, but the idea looks good to me!

> If so I can place the StatisticBase in bus-context-preprocess as all statistic tracker actors will use this package.

This bus might not be the best suited place for this. But I don't have a better suggestion atm, so let's start with that for now, and move later if good inspiration arrives.

packages/actor-context-preprocess-statistic-link-discovery/lib/StatisticLinkDiscovery.ts
RubenEschauzier: Do you mean by passing a logger to the emit function as an optional parameter?

rubensworks: No, not really.
What we can do, is remove this logger code here.
And if the user still requires the logging of this information, the user can add a listener to this statistic, and manually call the logger from the statistic listener callback.
Do you think this make sense?

RubenEschauzier: That makes sense, though this would require the user to make an actor. As the statistics are only available in the context, which is inaccessible to the user.

rubensworks: > As the statistics are only available in the context, which is inaccessible to the user.

Won't they be available to the user that invoked the query (through the JS API)?
Because the user can provide an input context, and can also obtain the output context after starting execution.

RubenEschauzier: They can obtain the output context? Oops never knew, then I completely agree with removing the log function

packages/actor-context-preprocess-query-source-identify/lib/ActorContextPreprocessQuerySourceIdentify.ts
```diff
@@ -54,6 +58,26 @@ export class ActorContextPreprocessQuerySourceIdentify extends ActorContextPrepr
         .map(querySource => this.expandSource(querySource)));
       const querySources: IQuerySourceWrapper[] = await Promise.all(querySourcesUnidentifiedExpanded
         .map(async querySourceUnidentified => this.identifySource(querySourceUnidentified, action.context)));
+
+      /**
+       * When identifying sources in preprocess actor, we record this as a dereference seed document event
+       */
+      const statistics: IStatisticsHolder = action.context.get(KeysInitQuery.statistics)!;
```
rubensworks, line 65: You can use `context.getSafe` here. Probably applies to other places as well.

packages/actor-context-preprocess-query-source-identify/lib/ActorContextPreprocessQuerySourceIdentify.ts
```diff
@@ -54,6 +58,26 @@ export class ActorContextPreprocessQuerySourceIdentify extends ActorContextPrepr
...
+      const statisticDereferenceLinks = <IStatisticBase<ILink>> statistics
```
rubensworks, line 66: This should also work:
```suggestion
      const statisticDereferenceLinks: IStatisticBase<ILink> = statistics
```

packages/actor-context-preprocess-statistic-link-dereference/lib/ActorContextPreprocessStatisticLinkDereference.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  public async test(_action: IActionContextPreprocess): Promise<IActorTest> {
+    return true;
+  }
+
+  public async run(action: IActionContextPreprocess): Promise<IActorContextPreprocessOutput> {
+    const statisticsHolder: StatisticsHolder = action.context.get(KeysInitQuery.statistics)!;
+
+    // Create StatisticLinkDereference with bound log function with context entry already filled in
```
rubensworks, line 27: Comment can go?

packages/actor-context-preprocess-statistic-link-dereference/lib/ActorContextPreprocessStatisticLinkDereference.ts
```diff
@@ -0,0 +1,34 @@
...
+    const statisticLinkDereference: StatisticLinkDereference = new StatisticLinkDereference();
```
rubensworks, line 28: I'd just inline this to line 30.

packages/actor-context-preprocess-statistic-link-dereference/lib/StatisticLinkDereference.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export class StatisticLinkDereference extends StatisticBase<ILink> {
+  public constructor() {
+    super();
+  }
+
+  public updateStatistic(link: ILink, source: IQuerySource): boolean {
+    const metadata: Record<string, any> = {
+      type: source.constructor.name,
+      dereferencedTimestamp: Date.now(),
```
rubensworks, line 12: Open question: do we want low-resolution timestamps (like Date.now) or high-resolution timestamps (like Performance.now). The latter is more precise and more modern. https://developer.mozilla.org/en-US/docs/Web/API/Performance/now#performance.now_vs._date.now

packages/actor-context-preprocess-statistic-link-dereference/lib/StatisticLinkDereference.ts
```diff
@@ -0,0 +1,27 @@
+import { StatisticBase } from '@comunica/bus-context-preprocess';
+import type { ILink, IQuerySource } from '@comunica/types';
+
+export class StatisticLinkDereference extends StatisticBase<ILink> {
+  public constructor() {
```
rubensworks, line 5: I think this ctor can be removed, as it just calls super.

packages/actor-context-preprocess-statistic-link-dereference/lib/StatisticLinkDereference.ts
```diff
@@ -0,0 +1,27 @@
...
+    };
+
+    const dereferencedLink: ILink = {
+      url: link.url,
+      metadata,
+      context: link.context,
+      transform: link.transform,
+    };
+
+    this.emit(dereferencedLink);
```
rubensworks, line 23: I'd inline all of the above into here.

packages/actor-context-preprocess-statistic-link-discovery/lib/ActorContextPreprocessStatisticLinkDiscovery.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  public async test(_action: IActionContextPreprocess): Promise<IActorTest> {
+    return true;
+  }
+
+  public async run(action: IActionContextPreprocess): Promise<IActorContextPreprocessOutput> {
+    const statisticsHolder: StatisticsHolder = action.context.get(KeysInitQuery.statistics)!;
+
+    const statisticLinkDiscovery: StatisticLinkDiscovery = new StatisticLinkDiscovery();
+    statisticsHolder.set(KeysTrackableStatistics.discoveredLinks, statisticLinkDiscovery);
```
rubensworks, line 25: Let's inline here as well, to reduce the LOC.
The blank lines can then also go, since they don't add much.
Blank lines are mainly useful between complex code blocks. But if a method is this simple, there's no need IMO.

packages/actor-context-preprocess-statistic-link-discovery/lib/StatisticLinkDiscovery.ts
```diff
@@ -0,0 +1,61 @@
+import { StatisticBase } from '@comunica/bus-context-preprocess';
+import type { ILink } from '@comunica/types';
+
+export class StatisticLinkDiscovery extends StatisticBase<IDiscoverEventData> {
+  // Number of discover events tracked
+  public discoverEvents: number;
```
rubensworks, line 6:
```suggestion
  public events: number;
```
Would this still make sense?

packages/actor-context-preprocess-statistic-link-discovery/lib/StatisticLinkDiscovery.ts
```diff
@@ -0,0 +1,61 @@
...
+    };
+
+    if (discoveredLinkMetadata) {
+      // Retain previous metadata if this link has already been discovered, and add any metadata in the passed link
+      this.metadata[link.url] = this.metadata[link.url] ?
+          [ ...this.metadata[link.url], discoveredLinkMetadata ] :
+          [ discoveredLinkMetadata ];
+    }
+
+    const discoverEventData: IDiscoverEventData = {
```
rubensworks, line 33: Let's inline this to line 39.

packages/actor-context-preprocess-statistic-link-discovery/lib/StatisticLinkDiscovery.ts
```diff
@@ -0,0 +1,61 @@
...
+      edge: [ parent.url, link.url ],
+      metadataChild: this.metadata[link.url],
+      metadataParent: this.metadata[parent.url],
+    };
+
+    this.emit(discoverEventData);
+
+    // Increment number of discover events to track discover order
+    this.discoverEvents += 1;
```
rubensworks, line 42:
```suggestion
    this.discoverEvents++;
```

packages/actor-context-preprocess-statistic-link-discovery/lib/StatisticLinkDiscovery.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    const discoverEventData: IDiscoverEventData = {
+      edge: [ parent.url, link.url ],
+      metadataChild: this.metadata[link.url],
+      metadataParent: this.metadata[parent.url],
+    };
+
+    this.emit(discoverEventData);
+
+    // Increment number of discover events to track discover order
```
rubensworks, line 41: Comment can go.

packages/actor-query-source-identify-hypermedia/lib/LinkedRdfSourcesAsyncRdfIterator.ts
```diff
@@ -1,7 +1,15 @@
-import type { ILink } from '@comunica/bus-rdf-resolve-hypermedia-links';
+import type { StatisticsHolder } from '@comunica/actor-context-preprocess-set-defaults';
```
rubensworks, line 1: I guess we should be using StatisticsHolder instead of IStatisticsHolder?

packages/actor-query-source-identify-hypermedia/lib/MediatedLinkedRdfSourcesAsyncRdfIterator.ts
```diff
@@ -144,9 +153,22 @@ export class MediatedLinkedRdfSourcesAsyncRdfIterator extends LinkedRdfSourcesAs
     return this.linkQueue;
   }
 
-  protected async getSourceLinks(metadata: Record<string, any>): Promise<ILink[]> {
+  protected async getSourceLinks(metadata: Record<string, any>, startSource: ISourceState): Promise<ILink[]> {
```
rubensworks, line 156: Is keeping track of this `startSource` really needed, given that we have `this.firstUrl`?

packages/bus-context-preprocess/lib/StatisticBase.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export abstract class StatisticBase<TEventData> implements IStatisticBase<TEventData> {
+  private readonly listeners: ((data: TEventData) => any)[] = [];
+
+  public on(fn: (data: TEventData) => any): StatisticBase<TEventData> {
+    this.listeners.push(fn);
+    return this;
+  }
+
+  public off(fn: (data: TEventData) => any): StatisticBase<TEventData> {
```
rubensworks, line 15:
```suggestion
  public removeListener(fn: (data: TEventData) => any): StatisticBase<TEventData> {
```

packages/bus-context-preprocess/lib/StatisticBase.ts
```diff
@@ -0,0 +1,42 @@
+import type { IStatisticBase } from '@comunica/types';
+
+/**
+ * Base class for statistics with event emitter logic implemented. Each statistic tracker implementation
+ * should only define their updateStatistic function.
+ */
+export abstract class StatisticBase<TEventData> implements IStatisticBase<TEventData> {
+  private readonly listeners: ((data: TEventData) => any)[] = [];
+
+  public on(fn: (data: TEventData) => any): StatisticBase<TEventData> {
```
rubensworks, line 10:
```suggestion
  public on(listener: (data: TEventData) => any): StatisticBase<TEventData> {
```
(also below)

packages/bus-rdf-resolve-hypermedia-links/lib/ActorRdfResolveHypermediaLinks.ts
```diff
@@ -38,6 +38,7 @@ export interface IActorRdfResolveHypermediaLinksOutput extends IActorOutput {
   links: ILink[];
 }
 
+// TODO deprecate this in next major version
```
rubensworks, line 41: Good!
But let's just re-export the ILink from the types here, to make sure they are always identical.

```suggestion
// TODO remove this in next major version
```

packages/context-entries/lib/Keys.ts
```diff
@@ -12,10 +12,13 @@ import type {
   QuerySourceUnidentified,
   IQuerySourceWrapper,
   QuerySourceReference,
+  IStatisticsHolder,
 } from '@comunica/types';
 import type * as RDF from '@rdfjs/types';
 import type { IDocumentLoader } from 'jsonld-context-parser';
 import type { Algebra } from 'sparqlalgebrajs';
+import type { StatisticLinkDereference } from '../../actor-context-preprocess-statistic-link-dereference/lib';
```
rubensworks, line 20: This and the next should import by package name.

packages/context-entries/lib/Keys.ts
```diff
@@ -298,3 +307,18 @@ export const KeysRdfJoin = {
    */
   lastPhysicalJoin: new ActionContextKey<string>('@comunica/bus-rdf-join:lastPhysicalJoin'),
 };
+
+export const KeysTrackableStatistics = {
+  /**
+   * All discovered links during traversal (not only the relevant ones) in form of parent, child
```
rubensworks, line 313:
```suggestion
   * All discovered links during query execution. Not all of them will necessarily be dereferenced.
```

(the reader should be able to find additional information in `StatisticLinkDiscovery`)

packages/types/lib/IStatistic.ts
```diff
@@ -0,0 +1,49 @@
+import type { IActionContextKey } from './IActionContext';
+
+/**
+ * Base interface for statistic trackers that defines all method a statistic should have.
```
rubensworks, line 4:
```suggestion
 * Statistics are able to track specific types of information during query execution.
 * It is possible to listen to events that are produced by each specific statistic.
```

packages/types/lib/IStatistic.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+export interface IStatisticBase<T> {
```
rubensworks, line 6: Let's move this interface and the next one to their own dedicated files, corresponding to the interface name.

rubensworks (changes requested):
Nice work! No major comments remaining atm from my end.

packages/actor-context-preprocess-statistic-link-dereference/lib/StatisticLinkDereference.ts
RubenEschauzier: I wasn't aware of Performance.now(), given it's more accurate, using it is probably better.

packages/actor-context-preprocess-statistic-link-discovery/lib/StatisticLinkDiscovery.ts
RubenEschauzier: Yes, but it would lose some of its meaning. It annotates a discovery event with how many links were discovered before it. Would you think `numDiscoveredLinks` or something like this would be better? Or do you think `events` is clear enough given the name of the class?

rubensworks: > Or do you think events is clear enough given the name of the class?

This was my reasoning indeed.

rubensworks: Or just `count` could event work.

packages/actor-query-source-identify-hypermedia/lib/LinkedRdfSourcesAsyncRdfIterator.ts
RubenEschauzier: I think it was a mistake because we don't want to make a dependency between `set-defaults` and `identify-hypermedia` right?

rubensworks: Ideally not indeed.

packages/actor-query-source-identify-hypermedia/lib/MediatedLinkedRdfSourcesAsyncRdfIterator.ts
RubenEschauzier: Yes, as `this.firstUrl` will always refer to the first URL that is dereferenced and not the link responsible (the parent of the found links in this case) for the call to getSourceLinks.

packages/context-entries/lib/Keys.ts
```diff
@@ -298,3 +301,18 @@ export const KeysRdfJoin = {
    */
   lastPhysicalJoin: new ActionContextKey<string>('@comunica/bus-rdf-join:lastPhysicalJoin'),
 };
+
+export const KeysTrackableStatistics = {
```
rubensworks, line 305: Perhaps this can now just become `KeysStatistics` ?

RubenEschauzier: Agreed, this should be done now

packages/actor-query-source-identify-hypermedia/lib/MediatedLinkedRdfSourcesAsyncRdfIterator.ts
RubenEschauzier: I believe it's because during link traversal `MediatedLinkedRdfSourcesAsyncRdfIterator` is only made once. The class attribute `this.firstUrl` is then set to the seed document. During traversal a URL in the queue is dereferenced [here](https://github.com/comunica/comunica/blob/721bdf88a407e769e57a25102df6151f1021b0f0/packages/actor-query-source-identify-hypermedia/lib/LinkedRdfSourcesAsyncRdfIterator.ts#L301). From this URL, new URLs are discovered [here](https://github.com/comunica/comunica/blob/721bdf88a407e769e57a25102df6151f1021b0f0/packages/actor-query-source-identify-hypermedia/lib/LinkedRdfSourcesAsyncRdfIterator.ts#L256) and the next iterator in the queue is started [here](https://github.com/comunica/comunica/blob/721bdf88a407e769e57a25102df6151f1021b0f0/packages/actor-query-source-identify-hypermedia/lib/LinkedRdfSourcesAsyncRdfIterator.ts#L270).  In this process, no new `MediatedLinkedRdfSourcesAsyncRdfIterator` is made, so `this.firstUrl` does not (most of the time) correspond to the parent URL of newly discovered URLs.

rubensworks: Ah nvm, I misread the diff.
I thought this change was in `getSourceLinks`, but it's in `_read`.

Looks good in that case!

packages/actor-context-preprocess-query-source-identify/lib/ActorContextPreprocessQuerySourceIdentify.ts
```diff
@@ -54,6 +57,24 @@ export class ActorContextPreprocessQuerySourceIdentify extends ActorContextPrepr
         .map(querySource => this.expandSource(querySource)));
       const querySources: IQuerySourceWrapper[] = await Promise.all(querySourcesUnidentifiedExpanded
         .map(async querySourceUnidentified => this.identifySource(querySourceUnidentified, action.context)));
+
+      /**
+       * When identifying sources in preprocess actor, we record this as a dereference seed document event
+       */
```
rubensworks, line 63:
```suggestion
      // When identifying sources in preprocess actor, we record this as a dereference seed document event
```

packages/actor-context-preprocess-query-source-identify/lib/ActorContextPreprocessQuerySourceIdentify.ts
```diff
@@ -54,6 +57,24 @@ export class ActorContextPreprocessQuerySourceIdentify extends ActorContextPrepr
...
+      const statisticDereferenceLinks: IStatisticBase<ILink> | undefined = action.context
+        .get(KeysStatistics.dereferencedLinks);
+
+      if (statisticDereferenceLinks) {
```
rubensworks, line 67:
```suggestion
      if (statisticDereferenceLinks) {
```

packages/actor-query-source-identify-hypermedia/lib/LinkedRdfSourcesAsyncRdfIterator.ts
```diff
@@ -308,6 +315,22 @@ export abstract class LinkedRdfSourcesAsyncRdfIterator extends BufferedIterator<
             this.iteratorsPendingCreation++;
             this.sourceStateGetter(nextLink, handledDatasets)
               .then((nextSourceState) => {
+                // Find any statistics tracking dereference events
+                const statisticDereferenceLinks: IStatisticBase<ILink> | undefined = this.context.get(
+                  KeysStatistics.dereferencedLinks,
+                );
+
+                // If we find a statistic tracking dereference events we emit the relevant data
+                if (statisticDereferenceLinks) {
```
rubensworks, line 324:
```suggestion
                // If we find a statistic tracking dereference events we emit the relevant data
                const statisticDereferenceLinks: IStatisticBase<ILink> | undefined = this.context.get(
                  KeysStatistics.dereferencedLinks,
                );
                if (statisticDereferenceLinks) {
```

packages/actor-query-source-identify-hypermedia/lib/MediatedLinkedRdfSourcesAsyncRdfIterator.ts
```diff
@@ -144,9 +152,18 @@ export class MediatedLinkedRdfSourcesAsyncRdfIterator extends LinkedRdfSourcesAs
...
     try {
       const { links } = await this.mediatorRdfResolveHypermediaLinks.mediate({ context: this.context, metadata });
+      const traversalTracker: IStatisticBase<IDiscoverEventData> | undefined =
+        this.context.get(KeysStatistics.discoveredLinks);
+
+      // Update discovery event statistic if available
+      if (traversalTracker) {
```
rubensworks, line 162:
```suggestion
      // Update discovery event statistic if available
      const traversalTracker: IStatisticBase<IDiscoverEventData> | undefined =
        this.context.get(KeysStatistics.discoveredLinks);
      if (traversalTracker) {
```

packages/statistic-base/lib/StatisticBase.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (!this.listeners || this.listeners.length === 0) {
+      return false;
+    }
+    for (const listener of this.listeners) {
+      listener(data);
+    }
+    return true;
+  }
+
+  public rawListeners(): ((data: TEventData) => any)[] {
```
rubensworks, line 37: I guess we can just rename this to `getListeners()`?

packages/statistic-link-dereference/README.md
```diff
@@ -0,0 +1,23 @@
...
+```bash
+$ yarn add @comunica/statistic-link-dereference
+```
+
+## Adding to the context
+
+```javascript
+    let context = new ActionContext();
+    const statisticTracker = new StatisticLinkDereference();
+    context = context.set(statisticTracker.key, statisticTracker);
```
rubensworks, line 22: It may be more helpful to add an example where a query engine is invoked and a context is passed.

We don't need an actual query though, something like `'...'` may be sufficient.

rubensworks (changes requested):
Looking good!

Just some minor comments.
Should be good to go after that!

rubensworks:
Cool, thanks @RubenEschauzier!

PR #1398, opened 2024-07-20
chore: migrate stream-to-string

jeswr:
In preparation for ESM support, migrates from `stream-to-string` to `@jeswr/stream-to-string` and removes `streamify-string`


jeswr:
@rubensworks the CI error here is the same CI error as the main branch - this should be ready to go once that error is fixed on main and the changes are merged here.

rubensworks:
The error on master looks weird. Was the CI rerun at some point? Because the original run from two weeks ago worked fine? https://github.com/comunica/comunica/actions/runs/9868889647
Looks related to https://stackoverflow.com/questions/42940954/cannot-find-namespace-nodejs-after-webpack-upgrade, perhaps due to some upstream type changes.

jeswr:
> Was the CI re run at some point?

I just manually re-ran it (https://github.com/comunica/comunica/actions/runs/10025192122/job/27708443421) but, previous to that there was still an error on main (https://github.com/comunica/comunica/actions/runs/10025192122/job/27708109659) on a run which I did not manually trigger.

jeswr:
> The only obvious environment difference that I can see from the last 2 weeks is that the node version for the 22.x workflow has gone from 22.4.0 -> 22.5.0. However I am running node 22.5.1 locally and not seeing these errors.

... and it looks like 22.5.0 is indeed buggy with issues around yarn module resolution (https://x.com/nodejs/status/1814112932650701195); so looks like we just need to wait for the CI to use 22.5.1

rubensworks:
Oof, a relief that it's not a problem on our end 😅

joachimvh:
FYI, this is the issue that needs to be resolved: https://github.com/actions/setup-node/issues/1112

rubensworks:
I had a quick look at `@jeswr/stream-to-string`, and I saw you're using a default export for the `stringify`.
I would consider making that a named export, as I've seen other libs (with dual-packaging) that had default exports migrate to purely named exports due to issues that arise from semantic differences between default exports in CJS/ESM.
I suspect that as long as there's only a single default export and nothing else, that these problems should not occur, but I'm not 100% certain about that.

jeswr:
> I would consider making that a named export

Done :)

PR #1400, opened 2024-07-21
chore: migrate parse-link-header to http-link-header

jeswr:
migrate `parse-link-header` to `http-link-header` in preparation for ESM support

### DIFF
--- a/packages/actor-http-memento/package.json
+++ b/packages/actor-http-memento/package.json
@@ -40,10 +40,8 @@
     "@comunica/bus-http": "^3.2.0",
     "@comunica/context-entries": "^3.2.0",
     "@comunica/core": "^3.2.0",
+    "@types/http-link-header": "^1.0.7",
     "cross-fetch": "^4.0.0",
     "http-link-header": "^1.1.3"
-  },
-  "devDependencies": {
-    "@types/http-link-header": "^1.0.7"
   }
 }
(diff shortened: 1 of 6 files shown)


jeswr:
@rubensworks the CI error here is the same CI error as the main branch - this should be ready to go once that error is fixed on main and the changes are merged here.

jeswr:
CI is failing on `@comunica/actor-http-native` which is removed in #1377 ; so lets get that merged first.

packages/actor-http-memento/package.json
```diff
@@ -40,8 +40,10 @@
     "@comunica/bus-http": "^3.2.0",
     "@comunica/context-entries": "^3.2.0",
     "@comunica/core": "^3.2.0",
-    "@types/parse-link-header": "^2.0.0",
     "cross-fetch": "^4.0.0",
-    "parse-link-header": "^2.0.0"
+    "http-link-header": "^1.1.3"
+  },
+  "devDependencies": {
+    "@types/http-link-header": "^1.0.7"
```
rubensworks, line 47: This can be an actual dep.

jeswr: I don't believe that is necessary in this case none of the types from `@types/http-link-header` are re-exported.

rubensworks: AFAIK, all of our types packages are non-dev deps (some probably also don't re-export types like here), so it would be good to remain consistent with the rest of the codebase.
Also, if this actor would somehow expose `http-link-header` types in the future, we may introduce a bug if we're not careful.
These packages are tiny anyways, and do not increase bundle size.

rubensworks:
Nice, thanks @jeswr!

PR #1419, opened 2024-08-23
Chore: extract function busses

jitsedesmet:


packages/actor-function-factory-not/README.md
```diff
@@ -0,0 +1,41 @@
...
+    ...
+    {
+      "@id": "urn:comunica:default:function-factory/actors#not",
+      "@type": "ActorFunctionFactoryNot"
+    }
+  ]
+}
+```
+
+### Config Parameters
```
rubensworks, line 37: The script can probably be changed to remove this section in the readme.

jitsedesmet: Yes! Is good. Is that all? :D

rubensworks: As far as I see atm, yes, this looks good.
I do think the linter will complain about some things still, but if those are fixed, this looks ok to me.

packages/actor-function-factory-not/lib/Not.ts
```diff
@@ -0,0 +1,17 @@
+import { TermSparqlFunction } from '@comunica/bus-function-factory';
+
+import {
+  bool,
+  declare,
+  SparqlOperator,
+} from '@comunica/expression-evaluator';
+
+export class Not extends TermSparqlFunction {
+  protected arity = 1;
```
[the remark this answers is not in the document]
jitsedesmet: I don't think so, it does remind me that these should be readonly. The fields are just there to describe the declared function better. 
(And they are used by the abstract class to generate proper errors)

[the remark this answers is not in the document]
rubensworks: > it does remind me that these should be readonly

This might be an indication that this approach is not strict enough.

I was thinking something like this:
```
const function = new TermSparqlFunction({
  arity: 1,
  operator: SparqlOperator.NOT,
  overloads: declare(SparqlOperator.NOT)
    .onTerm1(() => val => bool(!val.coerceEBV()))
    .collect()
});
```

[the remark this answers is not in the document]
jitsedesmet: Ah yes, might be better than using an abstract class

packages/actor-function-factory-term-function-not/lib/ActorFunctionFactoryTermFunctionNot.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * A comunica TermFunctionNot Function Factory Actor.
+ */
+export class ActorFunctionFactoryTermFunctionNot extends ActorFunctionFactory {
+  public constructor(args: IActorFunctionFactoryArgs) {
+    super(args);
+  }
+
+  public async test(action: IActionFunctionFactory): Promise<IActorTest> {
+    // Does support action.requireTermExpression, so no need to check for that.
```
rubensworks, line 23: I think this comment can go.

packages/actor-function-factory-term-function-not/lib/TermFunctionNot.ts
```diff
@@ -0,0 +1,19 @@
+import { TermFunctionBase } from '@comunica/bus-function-factory';
+
```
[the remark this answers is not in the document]
jitsedesmet: Hmm, seems like the linter places that there when it removes the unused imports :/

If this is an issue, we'll need to remove it manually :/ (or find a linter rule for this)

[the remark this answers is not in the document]
rubensworks: Not a big deal, can stay if too much work.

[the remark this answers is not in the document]
jitsedesmet: ^(import.*\n)\n -> $1

engines/config-query-sparql/config/function-factory/actors.json
```diff
@@ -3,6 +3,90 @@
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/config-query-sparql/^3.0.0/components/context.jsonld"
   ],
   "import": [
+    "ccqs:config/function-factory/actors/term-function-xsd-to-year-month-duration.json",
```
rubensworks, line 6: Could we get these sorted somehow?
Just so they are shown in the same order as when looking on the filesystem.

engines/config-query-sparql/config/function-factory/actors/expression-function-bnode.json
```diff
@@ -0,0 +1,15 @@
...
+  ],
+  "@id": "urn:comunica:default:Runner",
+  "@type": "Runner",
+  "actors": [
+    {
+      "@id": "urn:comunica:default:function-factory/actors#expression-function-bnode",
+      "@type": "ActorFunctionFactoryExpressionFunctionBnode"
+    }
+  ]
+}
```
rubensworks, line 15: These files seem to have a missing line break at the end.

packages/actor-function-factory-expression-function-bnode/README.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+## Configure
+
+After installing, this package can be added to your engine's configuration as follows:
+```text
+{
+  "@context": [
+    ...
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-function-factory-expression-function-bnode/^1.0.0/components/context.jsonld"
```
rubensworks, line 25: This should probably use 3.0.0?

packages/actor-function-factory-expression-function-bnode/lib/ActorFunctionFactoryExpressionFunctionBnode.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * A comunica ExpressionFunctionBnode Function Factory Actor.
+ */
+export class ActorFunctionFactoryExpressionFunctionBnode extends ActorFunctionFactory {
+  public constructor(args: IActorFunctionFactoryArgs) {
+    super(args);
+  }
+
+  public async test(action: IActionFunctionFactory): Promise<IActorTest> {
+    if (action.functionName === SparqlOperator.BNODE || action.requireTermExpression) {
```
rubensworks, line 23: Is this correct? Should this become an `&&`?

jitsedesmet: You're right! Wow, logic escaped me last night! It should be `action.functionName === SparqlOperator.BNODE && !action.requireTermExpression`

engines/config-query-sparql/config/function-factory/actors.json
```diff
@@ -3,6 +3,90 @@
...
+    "ccqs:config/function-factory/actors/term-function-lcase.json",
+    "ccqs:config/function-factory/actors/term-function-lesser-than-equal.json",
+    "ccqs:config/function-factory/actors/term-function-lesser-than.json",
+    "ccqs:config/function-factory/actors/term-function-md5.json",
+    "ccqs:config/function-factory/actors/term-function-minutes.json",
+    "ccqs:config/function-factory/actors/term-function-month.json",
+    "ccqs:config/function-factory/actors/term-function-multiplication.json",
+    "ccqs:config/function-factory/actors/term-function-not.json",
+    "ccqs:config/function-factory/actors/term-function-now.json",
+    "ccqs:config/function-factory/actors/term-function-object-sparql-function.json",
```
rubensworks, line 47: `-sparql-function` should not be there?

packages/actor-function-factory-expression-function-coalesce/test/coalesce01-spec-test.ts
```diff
@@ -1,6 +1,8 @@
+import { ActorFunctionFactoryTermFunctionDivision } from '@comunica/actor-function-factory-term-function-division';
 import { int, decimal } from '@comunica/expression-evaluator/test/util/Aliases';
 import { Notation } from '@comunica/expression-evaluator/test/util/TestTable';
-import { runFuncTestTable } from '../util';
+import { runFuncTestTable } from '../../bus-function-factory/test/util';
```
rubensworks, line 4: I guess something like this will also work here (and possible other places):
```suggestion
import { runFuncTestTable } from '@comnunica/bus-function-factory/test/util';
```

rubensworks (changes requested):
I quickly skimmed everything, and all looks really good to me!

Once the CI is fixed, and that one minor comment is resolved, I will merge this.

packages/actor-function-factory-expression-function-extensions/lib/ActorFunctionFactoryExpressionFunctionExtensions.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  public async test({ context, functionName }: IActionFunctionFactory): Promise<IActorTest> {
+    const extensionFinder: AsyncExtensionFunctionCreator =
+      context.getSafe(KeysExpressionEvaluator.extensionFunctionCreator);
+    const definition = await extensionFinder(new DataFactory<RDF.Quad>().namedNode(functionName));
+    if (definition) {
+      return true;
+    }
+    throw new Error('no');
```
rubensworks, line 32: This is maybe not the most descriptive error message.

PR #1420, opened 2024-08-26
Fix source attribution for qpf and SPARQL endpoints

RubenEschauzier:
This fix should solve the problems encountered in #1396. It prioritizes the source attribution wrapper by setting the filter factor to positive infinity (as higher is better in this case). Furthermore, it adds Invalidation states to binding stream metadata obtained from SPARQL endpoints and QPF. 

Question:
Some tests are failing as the QPF and SPARQL endpoint tests don't expect a state in the metadata of these bindings. As these tests explicitly omit the invalidation states I'm not confident these states weren't omitted for a reason. Were they? If so, doesn't the metadata interface require a state?

### DIFF
--- a/packages/actor-query-source-identify-hypermedia-annotate-source/lib/ActorQuerySourceIdentifyHypermediaAnnotateSource.ts
+++ b/packages/actor-query-source-identify-hypermedia-annotate-source/lib/ActorQuerySourceIdentifyHypermediaAnnotateSource.ts
@@ -27,7 +27,7 @@ export class ActorQuerySourceIdentifyHypermediaAnnotateSource extends ActorQuery
     if (action.context.get(KEY_CONTEXT_WRAPPED)) {
       throw new Error('Unable to wrap query source multiple times');
     }
-    return { filterFactor: 0 };
+    return { filterFactor: Number.POSITIVE_INFINITY };
   }
 
   public async run(action: IActionQuerySourceIdentifyHypermedia): Promise<IActorQuerySourceIdentifyHypermediaOutput> {
--- a/packages/actor-query-source-identify-hypermedia-annotate-source/test/ActorQuerySourceIdentifyHypermediaAnnotateSource-test.ts
+++ b/packages/actor-query-source-identify-hypermedia-annotate-source/test/ActorQuerySourceIdentifyHypermediaAnnotateSource-test.ts
@@ -47,7 +47,7 @@ describe('ActorQuerySourceIdentifyHypermediaSourceAttribution', () => {
 
     it('should test', async() => {
       await expect(actor.test({ metadata: <any> null, quads: <any> null, url: '', context }))
-        .resolves.toEqual({ filterFactor: 0 });
+        .resolves.toEqual({ filterFactor: Number.POSITIVE_INFINITY });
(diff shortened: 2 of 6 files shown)


rubensworks:
The filter factor changes look good to me, but I'm not sure about the metadata state changes.

`actor-query-source-identify-hypermedia` is responsible for setting metadata states, and since this actor is always invoked as parent of the QPF and SPARQL actors, it should always set the metadata states properly.
So I'm not sure this change in those actors is needed.
Perhaps something else is going wrong in the hypermedia actor that causes the metadata state to not be set properly for those sub-sources.
Could you look into what could be going wrong here?

RubenEschauzier:
I looked into it and I believe the state is never set in the `actor-query-source-identify-hypermedia`, as for the rdfjs source (which is the one that worked in my testing)  the following [line](https://github.com/comunica/comunica/blob/721bdf88a407e769e57a25102df6151f1021b0f0/packages/actor-query-source-identify-rdfjs/lib/QuerySourceRdfJs.ts#L83) sets the metadata (if no metadata is found). In this function, the state is set [here](https://github.com/comunica/comunica/blob/721bdf88a407e769e57a25102df6151f1021b0f0/packages/actor-query-source-identify-rdfjs/lib/QuerySourceRdfJs.ts#L134). 

As a fix, I can add the state to the metadata passed [here](https://github.com/comunica/comunica/blob/721bdf88a407e769e57a25102df6151f1021b0f0/packages/actor-query-source-identify-hypermedia/lib/QuerySourceHypermedia.ts#L213).

However, in the `qpf source` when `query bindings` is called multiple times, the metadata for the bindings is constructed from the metadata extract mediator. As far as I can tell, no metadata extract actor will add a state to the metadata, so there will be no state in this metadata object.

I can probably add an actor that adds a state to the metadata. This would add a state to all metadata objects.

rubensworks:
> as for the rdfjs source (which is the one that worked in my testing) the following [line](https://github.com/comunica/comunica/blob/721bdf88a407e769e57a25102df6151f1021b0f0/packages/actor-query-source-identify-rdfjs/lib/QuerySourceRdfJs.ts#L83) sets the metadata (if no metadata is found). In this function, the state is set [here](https://github.com/comunica/comunica/blob/721bdf88a407e769e57a25102df6151f1021b0f0/packages/actor-query-source-identify-rdfjs/lib/QuerySourceRdfJs.ts#L134).

Ah ok, I see. In that case, your current solution looks like the right one. So let's keep things as-is!

Is this PR good to go for you? If so, I'll merge it.

rubensworks:
> Question:
Some tests are failing as the QPF and SPARQL endpoint tests don't expect a state in the metadata of these bindings. As these tests explicitly omit the invalidation states I'm not confident these states weren't omitted for a reason. Were they? If so, doesn't the metadata interface require a state?

We should probably add a state check in the tests indeed.
The check can be in this form: https://github.com/comunica/comunica/blob/721bdf88a407e769e57a25102df6151f1021b0f0/packages/bus-rdf-join/test/ActorRdfJoin-test.ts#L570

rubensworks:
Great, thanks @RubenEschauzier!

PR #1421, opened 2024-09-11
Refactor HTTP fetch and retry logic, adjust User-Agent creation and browser detection

surilindur:
This is a PR of all the work so far, with the following changes:

**Migrated HTTP request retry logic into its own `ActorHttpRetry` actor**
* This will enable retries to work on everything in the HTTP bus.
* The new actor also handles HTTP status codes 429 and 503 with `Retry-After` header, and delays subsequent requests to the associated host by the time instructed by the server.
* The retry logic with regards to server responses has also been adjusted, and the client now skips retries when the error code is in the 400 or 500 ranges (except those two exceptions mentioned earlier), because retrying a bad request or retrying against a failing server does not make any sense.

**Simplified `ActorHttpFetch` considerably**
* Now that the retry logic is elsewhere, this actor could be simplified a lot.
* Moved the Inrupt Solid client workaround to only occur once in the main actor itself, instead of being duplicated in the init preprocessor.
* Removed the `node-fetch` workaround in the init preprocessor, due to `node-fetch` not being used anymore.

**Adjusted browser detection logic**
* The current version of Comunica detects Deno as a browser, according to my hobby server nginx logs. This means that someone is running Comunica on Deno, after all, but it also means the current browser detection is not working.
* Here, I have changed the code to check for the existence of `globalThis.window.document`, which should be unlikely to exist outside browsers. This new check is located in a static method of `ActorHttp` since it is used more than once.

**Adjusted HTTP `User-Agent` header**
* The header value in Node/Deno/others will now be of the form `Comunica/<major_ver>.0 (<operating_system>; <system_architecture>) <actor_name>/<actor_version> <runtime_user_agenr>`, for example `Comunica/3.0 (linux; x64) ActorHttpFetch/3.2.1 Node.js/22`.
* This follows the example set by some browsers, bots and others, and better reflects the configuration of Comunica used.

**Other small clean-up**
* Removed `KeysHttp.retryOnServerError` as unused after the retry refactoring.
* Removed `web-streams-ponyfill` as unneeded now that Node 18 is the minimum supported version.

Any feedback would be welcome. :slightly_smiling_face: The reason all of this is in the same PR is that the changes would be a pain to make on their own and then update the unit tests every time for each iteration.

Edit: I will make a separate PR for the dynamic request rate limiting later, since that is independent of these, although it relies on some of the changes here.

## Summary by CodeRabbit

- **New Features**
	- Introduced `ActorHttpRetry` for enhanced HTTP request resilience with retry logic on transient errors.
	- Updated configuration to support the new retry actor in HTTP handling.

- **Bug Fixes**
	- Removed the `httpRetryOnServerError` option, streamlining error handling for server errors.

- **Documentation**
	- Added comprehensive README for the `@comunica/actor-http-retry` package detailing installation and configuration.
	- Enhanced clarity in existing documentation for HTTP fetch actors.

- **Tests**
	- Expanded test coverage for `ActorHttpRetry` and `ActorHttpFetch`, ensuring robust handling of various HTTP scenarios.


rubensworks:
@coderabbitai review

packages/actor-http-fetch/lib/ActorHttpFetch.ts
```diff
@@ -4,187 +4,93 @@ import { KeysHttp } from '@comunica/context-entries';
 import type { TestResult } from '@comunica/core';
 import { passTest } from '@comunica/core';
 import type { IMediatorTypeTime } from '@comunica/mediatortype-time';
-import type { Readable } from 'readable-stream';
+
+// eslint-disable-next-line import/extensions
+import { version as actorVersion } from '../package.json';
```
rubensworks, line 9: This may not work in browsers. Can you test?

packages/actor-http-fetch/lib/ActorHttpFetch.ts
```diff
@@ -4,187 +4,93 @@ import { KeysHttp } from '@comunica/context-entries';
...
+    const fetchFunction = action.context.get<Fetch>(KeysHttp.fetch) ?? fetch;
+    const requestInit = await this.fetchInitPreprocessor.handle(init);
+
+    let timeoutCallback: () => void;
+    let timeoutHandle: NodeJS.Timeout | undefined;
+
+    if (httpTimeout) {
+      const abortController = new AbortController();
+      requestInit.signal = abortController.signal;
+      timeoutCallback = () => abortController.abort(new Error(`Dereferencing timed out for ${ActorHttp.getInputUrl(action.input).href} after ${httpTimeout} ms`));
```
rubensworks, line 60:
```suggestion
      timeoutCallback = () => abortController.abort(new Error(`Fetch timed out for ${ActorHttp.getInputUrl(action.input).href} after ${httpTimeout} ms`));
```

packages/bus-http/lib/ActorHttp.ts
```diff
@@ -67,6 +62,40 @@ export abstract class ActorHttp extends Actor<IActionHttp, IActorTest, IActorHtt
...
+  }
+
+  /**
+   * Creates an appropriate User-Agent header string for Node.js or browser environment.
+   * Within browser environments, the browser agent header is used as-is to avoid problems.
+   * @returns {string} User agent string
+   */
+  public static createUserAgent(actorName: string, actorVersion: string): string {
+    if (ActorHttp.isBrowser()) {
+      return globalThis.navigator.userAgent;
```
rubensworks, line 82: If I understand correctly, this will still lead to the user agent to be defined in the fetch input within a browser environment.
I suspect this will lead to CORS errors in browser environments, so we may want to just return `undefined` here, and then not set this in the request input down the line.

packages/actor-http-fetch/lib/FetchInitPreprocessor.ts
```diff
@@ -17,22 +18,14 @@ export class FetchInitPreprocessor implements IFetchInitPreprocessor {
   }
 
   public async handle(init: RequestInit): Promise<RequestInit> {
-    // Convert body Web stream to Node stream, as node-fetch does not support Web streams
-    let halfDuplex = false;
-    if (init.body && typeof init.body !== 'string' && 'getReader' in <any> init.body) {
-      init.body = <any> ActorHttp.toNodeReadable(<any> init.body);
```
rubensworks, line 23: Are you certain this is not required anymore?

packages/actor-http-retry/lib/ActorHttpRetry.ts
```diff
@@ -0,0 +1,154 @@
...
+      const response = await this.mediatorHttp.mediate({
+        ...action,
+        context: action.context.set(ActorHttpRetry.keyWrapped, true),
+      });
+
+      if (response.ok) {
+        return response;
+      }
+
+      if (response.status === 429 || response.status === 503) {
```
rubensworks, line 65: There may be some more 5xx range codes we may want to consider in addition.

For example, 500 and 504 could make sense as well.

I wonder if we even may want to consider making this list configurable, either through the config or the context.

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#5xx_server_errors

rubensworks: I think this was the original reason for introducing `httpRetryOnServerError`, which you removed.
I would consider keeping it, or at least a variant by which the same can be achieved.

rubensworks (changes requested):
Nice work @surilindur! Good and clean code!

> The reason all of this is in the same PR is that the changes would be a pain to make on their own and then update the unit tests every time for each iteration.

That's fine by me, as long as the commits remain clean.

packages/bus-http/lib/ActorHttp.ts
surilindur: That is actually a great idea!

packages/actor-http-fetch/lib/FetchInitPreprocessor.ts
surilindur: I can test it, but the comment mentions that the workaround was for `node-fetch`, which is not used anymore. I thought the CI also tests it, but I think I misremembered how the CI works here. I will look into it!

packages/actor-http-retry/lib/ActorHttpRetry.ts
surilindur: Retrying on 504 might make sense, not sure about 500. I can make it configurable via the context, I think that would be the cleanest solution, and would allow any response status to trigger a retry.

packages/actor-http-fetch/lib/FetchInitPreprocessor.ts
rubensworks: Undici is now used everywhere indeed. But I don't know if it fully supports Web streams.

packages/actor-http-retry/lib/ActorHttpRetry.ts
surilindur: I have added a new `httpRetryStatusCodes` context entry, which holds a list of status codes (as numbers), and can be used to trigger a retry for any status code and bypass the normal handling of that code (for example, to trigger a retry on 500 or even 404, or any other code that is not 'ok').

The retry actor now also retries on 504, because that one definitely does make sense.

rubensworks:
I noticed this one is still open? Is it still a WIP?

packages/actor-http-fetch/lib/FetchInitPreprocessor.ts
surilindur: It should, according to the documentation: https://undici.nodejs.org/#/?id=requestbody

surilindur:
No, I just need to rebase it again.

packages/actor-http-fetch/lib/FetchInitPreprocessor.ts
rubensworks: But did you test it?

packages/actor-http-retry/lib/ActorHttpRetry.ts
```diff
@@ -0,0 +1,178 @@
...
+        }));
+        continue;
+      }
+
+      if (response.status === 429 || response.status === 503) {
+        // When the server reports a rate limit or temporary unavailability,
+        // then wait for the amount of time or for the specific date/time specified in Retry-After,
+        // or wait for the default user-specified time if this header is not provided by the server
+        const retryAfter = ActorHttpRetry.parseRetryAfterHeader(
+          response.headers.get('retry-after') ?? retryDelay.toString(10),
```
rubensworks, line 92: We need 2 additional checks here:

- A sanity check to make sure the delay makes sense. As to not make the query engine wait for literal ages in the case of server misconfigurations. Maybe we need an upper limit of what we accept for this value.
- A try-catch around this parsing, which we emit as a warning. As to not make the engine crash if the server emits some illegal value.

packages/actor-http-retry/lib/ActorHttpRetry.ts
```diff
@@ -0,0 +1,178 @@
...
+  // Expression that matches dates expressed in the HTTP Date header format
+  // eslint-disable-next-line max-len
+  private static readonly dateRegex = /^(Mon|Tue|Wed|Thu|Fri|Sat|Sun), [0-9]{2} (Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec) [0-9]{4} [0-9]{2}:[0-9]{2}:[0-9]{2} GMT$/u;
+  private static readonly numberRegex = /^[0-9]+$/u;
+  private static readonly keyWrapped = new ActionContextKey<boolean>('urn:comunica:actor-http-retry#wrapped');
+
+  public constructor(args: IActorHttpQueueArgs) {
+    super(args);
+    this.mediatorHttp = args.mediatorHttp;
+    this.activeDelays = {};
```
rubensworks, line 21: We'll need a http cache invalidation observer on here, to reset all delays if we forcefully invalidate comunica's cache.

packages/actor-http-retry/lib/ActorHttpRetry.ts
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  public async test(action: IActionHttp): Promise<TestResult<IMediatorTypeTime>> {
+    if (action.context.has(ActorHttpRetry.keyWrapped)) {
+      return failTest(`${this.name} can only wrap a request once`);
+    }
+    const retryCount = action.context.get<number>(KeysHttp.httpRetryCount);
```
rubensworks, line 28: This should now work:

```suggestion
    const retryCount = action.context.get(KeysHttp.httpRetryCount);
```

rubensworks (changes requested):

packages/actor-http-retry/lib/ActorHttpRetry.ts
surilindur: There is now an invalidation observer, which is also being tested by the unit tests.

packages/actor-http-retry/lib/ActorHttpRetry.ts
surilindur: These have been incorporated now:
* `httpRetryDelayLimit` is the upper limit for the wait time
* the actor will ignore the server-provided header and log a warning when the value is bad

packages/actor-http-fetch/lib/FetchInitPreprocessor.ts
surilindur: Yes, it works on my machine<sup>TM</sup>

surilindur:
This is now ready for a review! Based on some manual testing, it seems to work as expected. The Wikidata SPARQL endpoint still cuts requests because of the load limits, but that should hopefully be solved by the rate limiter later.

rubensworks:
Nice work!

PR #1422, opened 2024-09-13
Allow transformations on intermediate bindings streams produced during query execution

RubenEschauzier:
This PR aims to introduce a system that allows the creation of `process-iterator` actors that apply a transform on `streams` produced by actors subscribed to `bus-query-operation` and `bus-rdf-join`. Wrapper actors for `rdf-join` and `query-operation` apply the transform and can be enabled in the config.

An example usage pattern is in the PR. The `actor-process-iterator-record-intermediate-results` wraps around inner join streams and updates a `statistic-intermediate-results` with the produced intermediate result and any available metadata. Users can then hook into this statistic to determine what intermediate results are produced.

Currently, this PR is a draft because it has not been unit-tested. Initial experiments show it works for simple queries though. If possible, I would appreciate feedback on whether this is the correct way to implement this and if you have any high-level suggestions.

(Also configs are currently with it enabled, will update to disable by default).

## Summary by CodeRabbit

- **New Features**
	- Introduced multiple new configuration files for the Comunica framework, enhancing the capabilities of iterator transformations and query operations.
	- Added new actors and mediators to facilitate advanced RDF join operations and query processing, including `ActorQueryOperationWrapStream` and `ActorRdfJoinWrapStream`.
	- New `ActorIteratorTransformRecordIntermediateResults` class to update intermediate result statistics during transformations.
	- Added a new type definition `PartialResult` for flexible representation of partial results in RDF data handling.

- **Documentation**
	- Added a README file for the `@comunica/actor-process-iterator-record-intermediate-results` package, detailing installation and configuration.

- **Tests**
	- Comprehensive test suites added for various new and updated actors, ensuring functionality and reliability across different scenarios, including tests for `ActorRdfJoinWrapStream` and `ActorQueryOperationWrapStream`.

- **Chores**
	- New package.json files created for several packages to manage dependencies and metadata effectively.

### DIFF
--- a/engines/config-query-sparql/config/query-operation/actors/query/wrap-stream.json
+++ b/engines/config-query-sparql/config/query-operation/actors/query/wrap-stream.json
@@ -1,0 +1,17 @@
+{
+  "@context": [
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/runner/^3.0.0/components/context.jsonld",
+
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-query-operation-wrap-stream/^3.0.0/components/context.jsonld"
+  ],
+  "@id": "urn:comunica:default:Runner",
+  "@type": "Runner",
+  "actors": [
+    {
+      "@id": "urn:comunica:default:query-operation/actors#wrap-stream",
+      "@type": "ActorQueryOperationWrapStream",
+      "mediatorQueryOperation": { "@id": "urn:comunica:default:query-operation/mediators#main" },
+      "mediatorIteratorTransform": { "@id": "urn:comunica:default:iterator-transform/mediators#main" }
+    }
+  ]
+}
--- a/engines/config-query-sparql/config/query-operation/actors/query/wrap-stream.json
+++ b/engines/config-query-sparql/config/query-operation/actors/query/wrap-stream.json
@@ -1,8 +1,8 @@
 {
   "@context": [
-    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/runner/^3.0.0/components/context.jsonld",
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/runner/^4.0.0/components/context.jsonld",
 
-    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-query-operation-wrap-stream/^3.0.0/components/context.jsonld"
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-query-operation-wrap-stream/^4.0.0/components/context.jsonld"
   ],
   "@id": "urn:comunica:default:Runner",
   "@type": "Runner",
--- a/packages/actor-iterator-transform-record-intermediate-results/README.md
+++ b/packages/actor-iterator-transform-record-intermediate-results/README.md
@@ -2,9 +2,9 @@
 
 [![npm version](https://badge.fury.io/js/%40comunica%2Factor-process-iterator-record-intermediate-results.svg)](https://www.npmjs.com/package/@comunica/actor-process-iterator-record-intermediate-results)
 
-A comunica Record Intermediate Results Process Iterator Actor. This actor wraps (intermediate) iterators produced during query execution. This wrapper
+A Comunica Record Intermediate Results Process Iterator Actor. This actor wraps (intermediate) iterators produced during query execution. This wrapper
 calls the `updateStatistic` function of `@comunica/statistic-intermediate-results` which should be present in the context. This statistic should be
-added to the context by the user. This is possible when using Comunica in a javascript [application](https://comunica.dev/docs/query/getting_started/query_app/)
+added to the context by the user. This is possible when using Comunica in a javascript [application](https://comunica.dev/docs/query/getting_started/query_app/) or by implementing your own `actor-context-preprocess` actor that adds this statistic to the context.
 
 This module is part of the [Comunica framework](https://github.com/comunica/comunica),
 and should only be used by [developers that want to build their own query engine](https://comunica.dev/docs/modify/).
--- a/packages/actor-iterator-transform-record-intermediate-results/README.md
+++ b/packages/actor-iterator-transform-record-intermediate-results/README.md
@@ -1,6 +1,6 @@
 # Comunica Count Intermediate Results Process Iterator Actor
 
-[![npm version](https://badge.fury.io/js/%40comunica%2Factor-process-iterator-count-intermediate-results.svg)](https://www.npmjs.com/package/@comunica/actor-process-iterator-count-intermediate-results)
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-process-iterator-record-intermediate-results.svg)](https://www.npmjs.com/package/@comunica/actor-process-iterator-record-intermediate-results)
 
 A comunica Count Intermediate Results Process Iterator Actor. This actor wraps (intermediate) iterators produced during query execution. This wrapper
 calls the `updateStatistic` function of `@comunica/statistic-intermediate-results` which should be present in the context. This statistic should be
@@ -14,7 +14,7 @@ and should only be used by [developers that want to build their own query engine
 ## Install
 
 ```bash
-$ yarn add @comunica/actor-process-iterator-count-intermediate-results
+$ yarn add @comunica/actor-process-iterator-record-intermediate-results
 ```
 
 ## Configure
@@ -24,12 +24,12 @@ After installing, this package can be added to your engine's configuration as fo
 {
   "@context": [
     ...
-    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-process-iterator-count-intermediate-results/^1.0.0/components/context.jsonld"
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-process-iterator-record-intermediate-results/^1.0.0/components/context.jsonld"
   ],
   "actors": [
     ...
     {
-      "@id": "urn:comunica:default:process-iterator/actors#count-intermediate-results",
+      "@id": "urn:comunica:default:process-iterator/actors#record-intermediate-results",
       "@type": "ActorProcessIteratorCountIntermediateResults"
     }
   ]
--- a/packages/actor-iterator-transform-record-intermediate-results/README.md
+++ b/packages/actor-iterator-transform-record-intermediate-results/README.md
@@ -1,8 +1,8 @@
-# Comunica Count Intermediate Results Process Iterator Actor
+# Comunica Record Intermediate Results Process Iterator Actor
 
 [![npm version](https://badge.fury.io/js/%40comunica%2Factor-process-iterator-record-intermediate-results.svg)](https://www.npmjs.com/package/@comunica/actor-process-iterator-record-intermediate-results)
 
-A comunica Count Intermediate Results Process Iterator Actor. This actor wraps (intermediate) iterators produced during query execution. This wrapper
+A comunica Record Intermediate Results Process Iterator Actor. This actor wraps (intermediate) iterators produced during query execution. This wrapper
 calls the `updateStatistic` function of `@comunica/statistic-intermediate-results` which should be present in the context. This statistic should be
 added to the context by the user. This is possible when using Comunica in a javascript [application](https://comunica.dev/docs/query/getting_started/query_app/)
 
@@ -30,7 +30,7 @@ After installing, this package can be added to your engine's configuration as fo
     ...
     {
       "@id": "urn:comunica:default:process-iterator/actors#record-intermediate-results",
-      "@type": "ActorProcessIteratorCountIntermediateResults"
+      "@type": "ActorProcessIteratorRecordIntermediateResults"
     }
   ]
 }
--- a/packages/actor-iterator-transform-record-intermediate-results/README.md
+++ b/packages/actor-iterator-transform-record-intermediate-results/README.md
@@ -1,0 +1,40 @@
+# Comunica Count Intermediate Results Process Iterator Actor
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-process-iterator-count-intermediate-results.svg)](https://www.npmjs.com/package/@comunica/actor-process-iterator-count-intermediate-results)
+
+A comunica Count Intermediate Results Process Iterator Actor. This actor wraps (intermediate) iterators produced during query execution. This wrapper
+calls the `updateStatistic` function of `@comunica/statistic-intermediate-results` which should be present in the context. This statistic should be
+added to the context by the user. This is possible when using Comunica in a javascript [application](https://comunica.dev/docs/query/getting_started/query_app/)
+
+This module is part of the [Comunica framework](https://github.com/comunica/comunica),
+and should only be used by [developers that want to build their own query engine](https://comunica.dev/docs/modify/).
+
+[Click here if you just want to query with Comunica](https://comunica.dev/docs/query/).
+
+## Install
+
+```bash
+$ yarn add @comunica/actor-process-iterator-count-intermediate-results
+```
+
+## Configure
+
+After installing, this package can be added to your engine's configuration as follows:
+```text
+{
+  "@context": [
+    ...
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-process-iterator-count-intermediate-results/^1.0.0/components/context.jsonld"
+  ],
+  "actors": [
+    ...
+    {
+      "@id": "urn:comunica:default:process-iterator/actors#count-intermediate-results",
+      "@type": "ActorProcessIteratorCountIntermediateResults"
+    }
+  ]
+}
+```
+
+### Config Parameters
+* `wraps`: String representations of the operations that this actor should wrap. If `undefined` this actor will wrap all operations.
--- a/packages/actor-iterator-transform-record-intermediate-results/lib/ActorIteratorTransformRecordIntermediateResults.ts
+++ b/packages/actor-iterator-transform-record-intermediate-results/lib/ActorIteratorTransformRecordIntermediateResults.ts
@@ -22,7 +22,7 @@ export class ActorIteratorTransformRecordIntermediateResults extends ActorIterat
 
   public async transformIteratorBindings(action: IActionIteratorTransformBindings):
   Promise<ITransformIteratorOutput<AsyncIterator<RDF.Bindings>, MetadataBindings>> {
-    const statisticIntermediateResults: StatisticIntermediateResults = action.context
+    const statisticIntermediateResults = <StatisticIntermediateResults> action.context
       .getSafe(KeysStatistics.intermediateResults);
     const output = action.stream.map((data) => {
       statisticIntermediateResults.updateStatistic(
@@ -42,7 +42,7 @@ export class ActorIteratorTransformRecordIntermediateResults extends ActorIterat
 
   public async transformIteratorQuads(action: IActionIteratorTransformQuads):
   Promise<ITransformIteratorOutput<AsyncIterator<RDF.Quad>, MetadataQuads>> {
-    const statisticIntermediateResults: StatisticIntermediateResults = action.context
+    const statisticIntermediateResults = <StatisticIntermediateResults> action.context
       .getSafe(KeysStatistics.intermediateResults);
     const output = action.stream.map((data) => {
       statisticIntermediateResults.updateStatistic(
--- a/packages/actor-iterator-transform-record-intermediate-results/lib/ActorIteratorTransformRecordIntermediateResults.ts
+++ b/packages/actor-iterator-transform-record-intermediate-results/lib/ActorIteratorTransformRecordIntermediateResults.ts
@@ -1,4 +1,5 @@
 import type {
+  ActionIteratorTransform,
   IActionIteratorTransformBindings,
   IActionIteratorTransformQuads,
   IActorIteratorTransformArgs,
@@ -6,6 +7,8 @@ import type {
   from '@comunica/bus-iterator-transform';
 import { ActorIteratorTransform } from '@comunica/bus-iterator-transform';
 import { KeysStatistics } from '@comunica/context-entries';
+import type { IActorTest, TestResult } from '@comunica/core';
+import { passTestVoid } from '@comunica/core';
 import type { StatisticIntermediateResults } from '@comunica/statistic-intermediate-results';
 import type { MetadataBindings, MetadataQuads } from '@comunica/types';
 import type * as RDF from '@rdfjs/types';
@@ -58,5 +61,10 @@ export class ActorIteratorTransformRecordIntermediateResults extends ActorIterat
       return data;
     });
     return { stream: output, metadata: action.metadata };
+  }
+
+  public async testIteratorTransform(_action: ActionIteratorTransform):
+  Promise<TestResult<IActorTest>> {
+    return passTestVoid();
   }
 }
--- a/packages/actor-iterator-transform-record-intermediate-results/lib/ActorIteratorTransformRecordIntermediateResults.ts
+++ b/packages/actor-iterator-transform-record-intermediate-results/lib/ActorIteratorTransformRecordIntermediateResults.ts
@@ -1,0 +1,62 @@
+import type {
+  IActionIteratorTransformBindings,
+  IActionIteratorTransformQuads,
+  IActorIteratorTransformArgs,
+  ITransformIteratorOutput }
+  from '@comunica/bus-iterator-transform';
+import { ActorIteratorTransform } from '@comunica/bus-iterator-transform';
+import { KeysStatistics } from '@comunica/context-entries';
+import type { StatisticIntermediateResults } from '@comunica/statistic-intermediate-results';
+import type { MetadataBindings, MetadataQuads } from '@comunica/types';
+import type * as RDF from '@rdfjs/types';
+import type { AsyncIterator } from 'asynciterator';
+
+/**
+ * A comunica Record Intermediate Results Iterator Transform Actor.
+ * This actor updates the intermediate result statistic when an intermediate result is produced.
+ */
+export class ActorIteratorTransformRecordIntermediateResults extends ActorIteratorTransform {
+  public constructor(args: IActorIteratorTransformArgs) {
+    super(args);
+  }
+
+  public async transformIteratorBindings(action: IActionIteratorTransformBindings):
+  Promise<ITransformIteratorOutput<AsyncIterator<RDF.Bindings>, MetadataBindings>> {
+    const statisticIntermediateResults: StatisticIntermediateResults = action.context
+      .getSafe(KeysStatistics.intermediateResults);
+    const output = action.stream.map((data) => {
+      statisticIntermediateResults.updateStatistic(
+        {
+          type: action.type,
+          data,
+          metadata: {
+            operation: action.operation,
+            metadata: action.metadata,
+          },
+        },
+      );
+      return data;
+    });
+    return { stream: output, metadata: action.metadata };
+  }
+
+  public async transformIteratorQuads(action: IActionIteratorTransformQuads):
+  Promise<ITransformIteratorOutput<AsyncIterator<RDF.Quad>, MetadataQuads>> {
+    const statisticIntermediateResults: StatisticIntermediateResults = action.context
+      .getSafe(KeysStatistics.intermediateResults);
+    const output = action.stream.map((data) => {
+      statisticIntermediateResults.updateStatistic(
+        {
+          type: action.type,
+          data,
+          metadata: {
+            operation: action.operation,
+            metadata: action.metadata,
+          },
+        },
+      );
+      return data;
+    });
+    return { stream: output, metadata: action.metadata };
+  }
+}
--- a/packages/actor-iterator-transform-record-intermediate-results/package.json
+++ b/packages/actor-iterator-transform-record-intermediate-results/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@comunica/actor-iterator-transform-record-intermediate-results",
-  "version": "3.2.1",
+  "version": "4.0.1",
   "description": "A record-intermediate-results iterator-transform actor",
   "lsd:module": true,
   "license": "MIT",
@@ -37,10 +37,9 @@
     "build:components": "componentsjs-generator"
   },
   "dependencies": {
-    "@comunica/bus-iterator-transform": "^3.2.0",
-    "@comunica/context-entries": "^3.2.0",
-    "@comunica/statistic-intermediate-results": "^3.2.1",
-    "@comunica/types": "^3.2.1",
+    "@comunica/bus-iterator-transform": "^4.0.1",
+    "@comunica/context-entries": "^4.0.1",
+    "@comunica/types": "^4.0.1",
     "@rdfjs/types": "*",
     "asynciterator": "^3.9.0"
   }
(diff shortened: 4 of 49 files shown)


rubensworks:
@coderabbitai review

packages/actor-query-operation-wrap-stream/lib/ActorQueryOperationWrapStream.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  public async run(action: IActionQueryOperation): Promise<IQueryOperationResult> {
+    // Prevent infinite recursion. In consequent query operation calls this key should be set to false
+    // To allow the operation to wrap ALL query operation runs
+    action.context = ActorQueryOperation.setContextWrapped(action.context, true);
+    const output: IQueryOperationResult = await this.mediatorQueryOperation.mediate(action);
+    switch (output.type) {
+      case 'bindings':
+        output.bindingsStream = <AsyncIterator<RDF.Bindings>>
```
rubensworks, line 39: I wonder if we also should allow the metadata to be optionally processed?

This could be useful if someone were to add a process iterator actor that does some filtering, and thereby could reduce the cardinality and update it in the metadata.

packages/bus-process-iterator/lib/ActorProcessIterator.ts
```diff
@@ -0,0 +1,89 @@
+import type { LogicalJoinType } from '@comunica/bus-rdf-join';
+import type { IAction, IActorArgs, IActorOutput, IActorTest, Mediate } from '@comunica/core';
+import { Actor } from '@comunica/core';
+import type { IActionContext } from '@comunica/types';
+import type * as RDF from '@rdfjs/types';
+import type { AsyncIterator } from 'asynciterator';
+import type { types } from 'sparqlalgebrajs/lib/algebra';
+
+/**
+ * A comunica actor for process-iterator events.
```
rubensworks, line 10: Just thinking about naming.

When I read this, my first thought also goes to a process at OS-level. Like a Node.js process, or a Firefox process.
But maybe this is just me. But if you agree that naming could be improved, something like `transform-iterator` could be an alternative.

rubensworks (changes requested):
The overall approach looks good to me!

RubenEschauzier:
Thanks for the review! I changed the name and added the metadata field to the iterator-transform bus. I'll start writing unit tests to finish the PR and ping you when it's ready for another (more in-depth) review.

engines/config-query-sparql/config/process-iterator/actors.json
```diff
@@ -0,0 +1,15 @@
...
+  ],
+  "@id": "urn:comunica:default:Runner",
+  "@type": "Runner",
+  "actors": [
+    {
+      "@id": "urn:comunica:default:iterator-transform/actors#record-intermediate-results",
+      "@type": "ActorIteratorTransformRecordIntermediateResults",
+      "wraps": ["inner", "minus", "optional", "project"]
+    }
+  ]
```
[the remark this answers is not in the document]
rubensworks: No, what @RubenEschauzier did seems fine to me.

engines/config-query-sparql/config/query-operation/actors/query/wrap-stream.json
```diff
@@ -0,0 +1,17 @@
+{
+  "@context": [
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/runner/^3.0.0/components/context.jsonld",
+
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-query-operation-wrap-stream/^3.0.0/components/context.jsonld"
+  ],
```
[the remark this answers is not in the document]
rubensworks: No, this seems fine.

packages/bus-query-operation/lib/ActorQueryOperation.ts
```diff
@@ -255,6 +255,10 @@ export abstract class ActorQueryOperation extends Actor<IActionQueryOperation, I
     }
     return shape.operation.pattern.type === operation.type;
   }
+
+  public static setContextWrapped(context: IActionContext, value: boolean): IActionContext {
```
rubensworks, line 259: I understand the need for calling this function.
However, it feels error-prone, as forgetting to call it may lead to difficult bugs that are difficult to debug.

An alternative approach could be to make `KEY_CONTEXT_WRAPPED_QUERY_OPERATION` refer not to a boolean, but the actual operation object that is being wrapped. Then you can just check in the actor's test if the wrapped operation is the same as current or not to decide if the test should pass.

packages/actor-query-operation-wrap-stream/README.md
```diff
@@ -0,0 +1,42 @@
+# Comunica Wrap Stream Query Operation Actor
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-query-operation-wrap-stream.svg)](https://www.npmjs.com/package/@comunica/actor-query-operation-wrap-stream)
+
+A comunica Wrap Stream Query Operation Actor. This actor calls the `bus-iterator-transform` for all iterators produced by `query-operation` actors.
+This actor should only be included if you require the functionality of `bus-iterator-transform` as it will likely slow down query execution.
```
rubensworks, line 6:
```suggestion
This actor should only be included if you require the functionality of `bus-iterator-transform` as it may slow down query execution.
```

packages/actor-query-operation-wrap-stream/lib/ActorQueryOperationWrapStream.ts
```diff
@@ -0,0 +1,93 @@
...
+      case 'bindings': {
+        const bindingIteratorTransformed = await this.mediatorIteratorTransform.mediate(
+          {
+            operation: action.operation.type,
+            stream: output.bindingsStream,
+            streamMetadata: output.metadata,
+            context: action.context,
+            metadata: {
+              type: output.type,
+              ...await output.metadata(),
```
rubensworks, line 47: Can't we just pass `output.metadata` directly without the `await`?

packages/actor-query-operation-wrap-stream/lib/ActorQueryOperationWrapStream.ts
```diff
@@ -0,0 +1,93 @@
...
+          },
+        );
+        output.quadStream = <AsyncIterator<RDF.Quad>> iteratorTransformed.stream;
+        output.metadata = <() => Promise<MetadataQuads>> iteratorTransformed.streamMetadata;
+        break;
+      }
+      default: {
+        break;
+      }
+    }
```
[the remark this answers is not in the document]
rubensworks: No, the current approach is good.

packages/actor-rdf-join-wrap-stream/README.md
```diff
@@ -0,0 +1,40 @@
+# Comunica Wrap Stream RDF Join Actor
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-rdf-join-wrap-stream.svg)](https://www.npmjs.com/package/@comunica/actor-rdf-join-wrap-stream)
+
+A comunica Wrap Stream RDF Join Actor. This actor calls the `bus-iterator-transform` for all iterators produced by `query-operation` actors.
+This actor should only be included if you require the functionality of `bus-iterator-transform` as it will likely slow down query execution.
```
rubensworks, line 6:
```suggestion
This actor should only be included if you require the functionality of `bus-iterator-transform` as it may slow down query execution.
```

packages/actor-rdf-join-wrap-stream/lib/ActorRdfJoinWrapStream.ts
```diff
@@ -0,0 +1,90 @@
...
+        stream: result.bindingsStream,
+        streamMetadata: result.metadata,
+        context: action.context,
+        metadata: {
+          joinEntries: action.entries.length,
+          ...await result.metadata(),
+          resultContext: result.context,
+        },
+      },
+    ));
```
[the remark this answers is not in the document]
rubensworks: Not needed.

packages/actor-rdf-join-wrap-stream/lib/ActorRdfJoinWrapStream.ts
```diff
@@ -0,0 +1,90 @@
...
+    // Prevent infinite recursion. In consequent query operation calls this key is set to false
+    // To allow the operation to wrap ALL rdf-join runs
+    action.context = ActorRdfJoin.setContextWrapped(action.context, true);
+    const result: IQueryOperationResultBindings = await this.mediatorJoin.mediate(action);
+
+    const { stream, streamMetadata } = (await this.mediatorIteratorTransform.mediate(
+      {
+        operation: action.type,
+        stream: result.bindingsStream,
+        streamMetadata: result.metadata,
```
rubensworks, line 51: Why both metadata and streamMetadata?

packages/bus-iterator-transform/lib/ActorIteratorTransform.ts
```diff
@@ -0,0 +1,133 @@
...
+   */
+  streamMetadata: () => Promise<M>;
+  /**
+   * Optional metadata
+   */
+  metadata?: Record<string, any>;
+}
+
+/**
+ * TODO If this fails for multiple transforms its because of lack of context, SO CHECK IT
```
rubensworks, line 82: TODO

packages/context-entries/lib/Keys.ts
```diff
@@ -233,6 +234,12 @@ export const KeysQueryOperation = {
    * The sources to query over.
    */
   querySources: new ActionContextKey<IQuerySourceWrapper[]>('@comunica/bus-query-operation:querySources'),
+  /**
+   * The process-iterator mediator that will wrap operation outputs of query-operation and rdf-join actors
+   * to allow users to define processors over Asynciterators
+   */
+  mediatorProcessIterator: new ActionContextKey<any>
```
rubensworks, line 241: This one is not needed (anymore)?

packages/statistic-intermediate-results/lib/StatisticIntermediateResults.ts
```diff
@@ -0,0 +1,19 @@
...
+  public constructor() {
+    super();
+    this.key = KeysStatistics.intermediateResults;
+  }
+
+  public updateStatistic(intermediateResult: IPartialResult): boolean {
+    intermediateResult.metadata.time = performance.now();
+    this.emit(intermediateResult);
+    return true;
+  }
```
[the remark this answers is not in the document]
rubensworks: No.

packages/types/lib/statistics/IPartialResult.ts
```diff
@@ -0,0 +1,6 @@
+import type * as RDF from '@rdfjs/types';
+
+export interface IPartialResult {
+  data: RDF.Bindings | RDF.Quad;
+  metadata: Record<string, any>;
```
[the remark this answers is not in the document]
rubensworks: No.

packages/types/lib/statistics/IPartialResult.ts
```diff
@@ -0,0 +1,6 @@
+import type * as RDF from '@rdfjs/types';
+
+export interface IPartialResult {
+  data: RDF.Bindings | RDF.Quad;
```
rubensworks, line 4: To enable more easy type-checking when using these partial results, I propose something like this:

```
export type PartialResult = {
  type: 'bindings';
  data: RDF.Bindings;
} | { type: 'quad'; ... };
```

rubensworks (changes requested):
Good stuff!

I closed the bot's comments I disagree with.

packages/statistic-intermediate-results/README.md
```diff
@@ -0,0 +1,45 @@
...
+    } LIMIT 10`, context);
+    bindingsStream.on('data', (binding) => {
+        console.log(binding.toString());
+    });
+```
+
+### Allowing Comunica to call the tracker
+Comunica does not call the `updateStatistic` function by default. To enable this users should first add the following actors to the config:
+`actor-query-operation-wrap-stream` and `actor-rdf-join-wrap-stream`. This will call the `bus-iterator-transform` for each intermediate iterator produced
+by Comunica. Users should then add `actor-iterator-transform-record-intermediate-results` to the config, which will call the `updateStatistic` method of `statistic-intermediate-result` for each produced intermediate result.
```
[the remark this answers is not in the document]
RubenEschauzier: Do we need the extra fields? I can understand the usefulness of writing down the output format, although that will be pretty long and depend on the intermediate result type. But the last three don't seem useful.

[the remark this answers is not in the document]
rubensworks: Yeah, let's not do that here. Maybe something for the website docs in the future.

packages/bus-query-operation/test/ActorQueryOperationTyped-test.ts
```diff
@@ -81,6 +81,7 @@ describe('ActorQueryOperationTyped', () => {
         [KeysInitQuery.physicalQueryPlanLogger.name]: logger,
         [KeysInitQuery.physicalQueryPlanNode.name]: operation,
         [KeysQueryOperation.operation.name]: operation,
+        [KEY_CONTEXT_WRAPPED_QUERY_OPERATION.name]: false,
```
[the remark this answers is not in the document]
RubenEschauzier: I don't think the new key should be added here, as it is only added during query execution or does not interact with these tests.

packages/actor-query-operation-update-load/test/ActorQueryOperationLoad-test.ts
```diff
@@ -115,6 +116,7 @@ describe('ActorQueryOperationLoad', () => {
         operation: expect.anything(),
         context: new ActionContext({
           '@comunica/bus-query-operation:operation': expect.anything(),
+          [KEY_CONTEXT_WRAPPED_QUERY_OPERATION.name]: false,
```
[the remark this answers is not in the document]
RubenEschauzier: This would probably make things more complex

packages/bus-iterator-transform/lib/ActorIteratorTransform.ts
```diff
@@ -0,0 +1,133 @@
...
+M extends MetadataBindings | MetadataQuads,
+>
+  extends Actor<IActionIteratorTransform<T, M>, IActorTest, IActorIteratorTransformOutput<T, M>> {
+  public wraps: possibleOperationTypes[];
+  /**
+   * @param args - @defaultNested {<default_bus> a <cc:components/Bus.jsonld#Bus>} bus
+   */
+  public constructor(args: IActorIteratorTransformArgs) {
+    super(args);
+  }
```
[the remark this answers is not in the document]
RubenEschauzier: Are we sure this should be removed? Other actors keep this 'useless constructor' too

packages/actor-query-operation-wrap-stream/lib/ActorQueryOperationWrapStream.ts
RubenEschauzier: I don't think so, as the metadata function returns a promise.

packages/actor-rdf-join-wrap-stream/lib/ActorRdfJoinWrapStream.ts
RubenEschauzier: The intention is to allow users to change the `streamMetadata` (the metadata function given by the join result) in iterator transform actors. The `metadata` field is intended as general metadata including more than just the stream metadata. If that makes sense.

packages/bus-iterator-transform/lib/ActorIteratorTransform.ts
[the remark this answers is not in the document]
rubensworks: Yeah, the ctor must remain in this case.

packages/actor-rdf-join-wrap-stream/lib/ActorRdfJoinWrapStream.ts
rubensworks: I may be misunderstanding something, but if users want to change the metadata, this could be done in the return value of the mediator, right?
I suspect we may only need `streamMetadata`.

packages/types/lib/statistics/IPartialResult.ts
RubenEschauzier: So I checked to see if this works, but the problem is that when you use such an approach you have to determine when your input is a `binding` or `quad`. I could make a mapping from operation to quad or binding but that seems like a lot of unnecessary work. How do you see this work?

packages/actor-rdf-join-wrap-stream/lib/ActorRdfJoinWrapStream.ts
RubenEschauzier: I think the intention was to allow actors on `bus-transform-iterator` to apply a transformation on the intermediate iterator's `metadata` function. For example, if you filter some results and want to update the estimated cardinality. In this case `streamMetadata` is the `metadata` function of the intermediate iterator (which may be somewhat confusingly named).

packages/types/lib/statistics/IPartialResult.ts
rubensworks: It can be used as follows:
```
const partialResult: PartialResult = ...;
if (partialResult.type === 'bindings') {
  // process data as bindings
}
```

rubensworks: Or are you referring to using this interface in your transformer?

Perhaps we need different transform operations for the different types to simplify handling?

packages/actor-rdf-join-wrap-stream/lib/ActorRdfJoinWrapStream.ts
rubensworks: Ok, but then `streamMetadata` is all we need, and `metadata` can be removed (and perhaps then `streamMetadata` can be renamed to `metadata`)?

RubenEschauzier: But some metadata could be useful to pass through to the `iterator-transform` actors that are not in `streamMetadata`, like in `rdf-join-wrap-stream`, the number of entries in the wrapped action shows how many joins were already performed. Where would you suggest this be added?

rubensworks: Aha, now I understand you.
I didn't understand the need for two types of metadata fields, but now I do.

Instead of passing custom fields such as the number of join entries in `metadata`, why not omit this custom metadata, and instead directly pass the original `action` via e.g. `originalAction`, to allow the user to derive the number of join entries like that?

RubenEschauzier: Right, that works!

packages/types/lib/statistics/IPartialResult.ts
RubenEschauzier: I was referring to 'creating' the input to `updateStatistic`, but I see now that it shouldn't be too difficult.

rubensworks:
@RubenEschauzier Before I forget; since we want to include this in the upcoming major release, could you retarget this PR to `next/major`?

packages/types/lib/statistics/IPartialResult.ts
RubenEschauzier: I've changed the typings around (a lot) following advice from Joachim. I'm not sure the typings are better this way so please let me know what you think!

packages/actor-iterator-transform-record-intermediate-results/lib/ActorIteratorTransformRecordIntermediateResults.ts
```diff
@@ -0,0 +1,52 @@
...
+ * A comunica Record Intermediate Results Iterator Transform Actor.
+ * This actor simply updates the intermediate result statistic when an intermediate result is produced.
+ */
+export class ActorIteratorTransformRecordIntermediateResults
+  extends ActorIteratorTransform<
+    AsyncIterator<RDF.Bindings> | AsyncIterator<RDF.Quad>,
+    MetadataBindings | MetadataQuads
+> {
+
+  public async transformIterator(action: IActionIteratorTransformQuad):
```
rubensworks, line 26: I'm not sure I understand the need for these method overrides.
AFAICS, they are not actually being used anymore?
If so, the signature from line 32 should be sufficient I guess?

packages/actor-iterator-transform-record-intermediate-results/lib/ActorIteratorTransformRecordIntermediateResults.ts
```diff
@@ -0,0 +1,57 @@
...
+          metadata: {
+            operation: action.operation,
+            ...action.metadata,
+          },
+        },
+      );
+      return data;
+    });
+    return { stream: output, metadata: <() => Promise<MetadataQuads>> action.metadata };
+  }
```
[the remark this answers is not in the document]
RubenEschauzier: This does not work.

packages/bus-iterator-transform/lib/ActorIteratorTransform.ts
```diff
@@ -0,0 +1,171 @@
...
+    const { stream, metadata } = await this.transformIteratorQuad(action);
+    return {
+      type: action.type,
+      operation: action.operation,
+      stream,
+      metadata,
+      originalAction: action.originalAction,
+      context: action.context,
+    };
+  }
```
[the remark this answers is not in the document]
RubenEschauzier: This also does not work

RubenEschauzier:
I updated the code to a ready state. The tests still fail due to changed function names and types, but this will be resolved once the general approach seems good.
I also changed some configs for local testing, this will also be reverted once the approach is good to go.

packages/actor-query-source-identify-hypermedia-annotate-source/test/ActorQuerySourceIdentifyHypermediaAnnotateSource-test.ts
```diff
@@ -86,7 +86,7 @@ describe('ActorQuerySourceIdentifyHypermediaSourceAttribution', () => {
         quad('s1', 'p1', 'o1'),
         quad('s2', 'p2', 'o2'),
       ]);
-      context = context.set(KEY_CONTEXT_WRAPPED, true);
+      context = context.set(KEY_CONTEXT_WRAPPED, actor);
```
rubensworks, line 89: This looks strange; is it correct?

packages/actor-rdf-join-wrap-stream/lib/ActorRdfJoinWrapStream.ts
```diff
@@ -0,0 +1,87 @@
...
+      physicalName: 'wrap-stream',
+      limitEntries: 0,
+      limitEntriesMin: true,
+      canHandleUndefs: true,
+      isLeaf: false,
+    });
+  }
+
+  public override async test(action: IActionRdfJoin): Promise<IMediatorTypeJoinCoefficients> {
+    if (action.context.get(KEY_CONTEXT_WRAPPED_RDF_JOIN) === this) {
```
rubensworks, line 33: I would advise against storing the actor inside the context.
Instead, you can store the operation that is being wrapped, and test based on that.

packages/bus-iterator-transform/lib/ActorIteratorTransform.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  public async test(
+    action: ActionIteratorTransform,
+  ): Promise<IActorTest> {
+    if (!(this.wraps === undefined) && !this.wraps.includes(action.operation)) {
```
rubensworks, line 58: I think something like this will also work:
```suggestion
    if (!this.wraps?.includes(action.operation)) {
```

RubenEschauzier: It doesn't seem so, it returns true when `this.wraps` is undefined, while that should return false, as `this.wraps === undefined` should make the actor wrap all operations.

packages/actor-iterator-transform-record-intermediate-results/lib/ActorIteratorTransformRecordIntermediateResults.ts
```diff
@@ -0,0 +1,62 @@
...
+          metadata: {
+            operation: action.operation,
+            metadata: action.metadata,
+          },
+        },
+      );
+      return data;
+    });
+    return { stream: output, metadata: action.metadata };
+  }
```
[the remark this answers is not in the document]
RubenEschauzier: @rubensworks I'm not sure about this one. Should we leave the `metadata` call to the statistic or do it in the `transformIterator` call? I'm leaning towards calling it in the statistic, as statistics might not use the metadata so it would increase overhead for no reason.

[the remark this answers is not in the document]
rubensworks: I agree with your opinion @RubenEschauzier, let's leave calling it to the statistic itself.

packages/actor-iterator-transform-record-intermediate-results/README.md
```diff
@@ -0,0 +1,40 @@
+# Comunica Count Intermediate Results Process Iterator Actor
```
rubensworks, line 1: The `.npmignore` files can be removed since v4.

packages/actor-iterator-transform-record-intermediate-results/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@comunica/actor-iterator-transform-record-intermediate-results",
+  "version": "4.0.1",
+  "description": "A record-intermediate-results iterator-transform actor",
+  "lsd:module": true,
+  "license": "MIT",
+  "homepage": "https://comunica.dev/",
```
rubensworks, line 7: Can you add the `funding` entry to all package.json's which are now present in all packages since v4?

packages/actor-query-operation-source/lib/ActorQueryOperationSource.ts
```diff
@@ -44,8 +44,8 @@ export class ActorQueryOperationSource extends ActorQueryOperation {
     }
 
     const sourceWrapper: IQuerySourceWrapper = getOperationSource(action.operation)!;
-    const mergedContext = sourceWrapper.context ? action.context.merge(sourceWrapper.context) : action.context;
-
+    let mergedContext = sourceWrapper.context ? action.context.merge(sourceWrapper.context) : action.context;
+    mergedContext = setContextWrapped(action.operation, mergedContext);
```
rubensworks, line 48: Are you sure this is needed here?
I would expect this context entry to only have to be set within the actor doing the actual wrapping.

packages/bus-iterator-transform/lib/ActorIteratorTransform.ts
rubensworks: The explicit `undefined` check can surely go, or be replaced by a falsy check.

packages/bus-query-operation/lib/ActorQueryOperationTyped.ts
```diff
@@ -56,6 +57,8 @@ TS = undefined,
       );
       action.context = action.context.set(KeysInitQuery.physicalQueryPlanNode, action.operation);
     }
+    // Set wrapped to false to allow recursive calls to query operation to also be wrapped
+    action.context = setContextWrapped(action.operation, action.context);
```
rubensworks, line 61: I suspect this also to not be required.

packages/bus-rdf-join/lib/ActorRdfJoin.ts
```diff
@@ -433,6 +433,8 @@ TS
       );
     }
 
+    action.context = this.setContextWrapped(action, action.context);
```
rubensworks, line 436: Same here.

packages/bus-rdf-join/lib/ActorRdfJoin.ts
```diff
@@ -544,11 +556,6 @@ export interface IActorRdfJoinInternalOptions {
   requiresVariableOverlap?: boolean;
 }
 
-/**
- * Represents a logical join type.
- */
-export type LogicalJoinType = 'inner' | 'optional' | 'minus';
```
rubensworks, line 550: Be careful: this will be a breaking change.

packages/types/lib/statistics/IPartialResult.ts
```diff
@@ -0,0 +1,7 @@
+import type * as RDF from '@rdfjs/types';
+
+export type PartialResult = {
+  type: 'bindings';
+  data: RDF.Bindings;
+  metadata: Record<string, any>;
+} | { type: 'quads'; data: RDF.Quad; metadata: Record<string, any> };
```
rubensworks, line 7: Let's put these on separate lines as well.

rubensworks (changes requested):

packages/bus-rdf-join/lib/ActorRdfJoin.ts
RubenEschauzier: Right, I added the following line: `// TODO remove this in next major version
export { LogicalJoinType } from '@comunica/types';`.
Is that ok? I figured moving it would be better as it it also used by the `iterator-transform` actors.

rubensworks: Yep, something like that is ok!
I don't see it in the diff yet though, so I assume it still needs to be pushed?

RubenEschauzier: I added it at line 593, is that ok?

RubenEschauzier: It is [here](https://github.com/comunica/comunica/blob/332ad8b1ce2f9ccf5f9ff1fb24815b88b4b1356c/packages/bus-rdf-join/lib/ActorRdfJoin.ts#L593), not sure how to post the permalink to the diff, but it does show up for me.

rubensworks: Looks perfect!

packages/utils-query-operation/lib/Utils.ts
```diff
@@ -110,3 +110,20 @@ export function removeOperationSource(operation: Algebra.Operation): void {
...
   }
 }
+
+/**
+ * Sets KEY_CONTEXT_WRAPPED_QUERY_OPERATION to the operation being executed.
+ * @param operation The query operation.
+ * @param context The current action context.
+ * @returns A new action context with the operation marked as wrapped.
+ */
+export function setContextWrapped(operation: Algebra.Operation, context: IActionContext): IActionContext {
```
rubensworks, line 120: Just a question, as you probably have a better view on this than me:
Is this util function (and context key) required here?
Can we not move it to the wrapper actor itself?
Because I suspect that only that wrapper actor will ever need to call this and use this context key?

The same applies to the util function in `ActorRdfJoin`.

rubensworks (changes requested):
Just a minor comment.
Other than that, this looks good to go!

Just double-checking; this is fully backwards-compatible, right?
(AFAICS, it is, since only new actors are added, and the new actors are not added to the default config)

packages/utils-query-operation/lib/Utils.ts
RubenEschauzier: Yep, I agree.

rubensworks:
@RubenEschauzier Is anything else needed for this PR? (also see my last question above regarding backwards-compat)

RubenEschauzier:
It should be backward compatible. The only place I've changed the existing code is in `packages/context-entries/lib/Keys.ts`, where I've changed the underlying name of a key. Now that I think of it, if someone were to rely on the name of the key instead of the object, this could cause issues. But that is an edge case.

rubensworks:
Very nice work @RubenEschauzier!
Do make sure to document how to use this on the website in the near future.
We'll probably need a new page dedicated to this in the advanced modification docs.

PR #1423, opened 2024-09-16
Feature/indexed bus for functions

jitsedesmet:
This PR aims to create an indexed bus for the FunctionCreator bus.
I started by investigating the existing BusIndexed, but found that they are not 100% compatible.
Right now, the bus I implemented does not inherit from BusIndexed. I do think that it is possible to make them share more code if I can also change the code of BusIndexed. What do you think about that @rubensworks ?

### DIFF
--- a/engines/query-sparql-file/package.json
+++ b/engines/query-sparql-file/package.json
@@ -282,6 +282,7 @@
     "@comunica/actor-rdf-update-quads-hypermedia": "^3.2.1",
     "@comunica/actor-rdf-update-quads-rdfjs-store": "^3.2.1",
     "@comunica/actor-term-comparator-factory-expression-evaluator": "^3.2.1",
+    "@comunica/bus-function-factory": "^3.2.1",
     "@comunica/bus-http-invalidate": "^3.2.1",
     "@comunica/bus-query-operation": "^3.2.2",
     "@comunica/config-query-sparql": "^3.2.1",
--- a/packages/bus-function-factory/lib/BusFunctionFactory.ts
+++ b/packages/bus-function-factory/lib/BusFunctionFactory.ts
@@ -1,0 +1,81 @@
+import type { IActorReply, IActorTest, IBusIndexedArgs } from '@comunica/core';
+import { Bus } from '@comunica/core';
+import type { ActorFunctionFactory, IActionFunctionFactory, IActorFunctionFactoryOutput } from './ActorFunctionFactory';
+
+/**
+ * Bus inspired by BusIndexed but specific for function factory.
+ *
+ * The implementation differs. In BusIndexed, each actor is indexed only once.
+ * Here, a single actor can be indexed multiple times (max 2).
+ */
+export class BusFunctionFactory
+  extends Bus<ActorFunctionFactory, IActionFunctionFactory, IActorTest, IActorFunctionFactoryOutput> {
+  protected readonly actorsIndex: Record<string, ActorFunctionFactory[]> = {};
+  protected readonly actorIdentifierFields: string[];
+  protected readonly actionIdentifierFields: string[];
+
+  public constructor(args: IBusIndexedArgs) {
+    super(args);
+  }
+
+  public override subscribe(actor: ActorFunctionFactory): void {
+    for (const actorId of this.getActorIdentifiers(actor)) {
+      let actors = this.actorsIndex[actorId];
+      if (!actors) {
+        actors = this.actorsIndex[actorId] = [];
+      }
+      actors.push(actor);
+      super.subscribe(actor);
+    }
+  }
+
+  public override unsubscribe(actor: ActorFunctionFactory): boolean {
+    let unsubscribed = true;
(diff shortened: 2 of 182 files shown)


packages/bus-function-factory/lib/BusFunctionFactory.ts
```diff
@@ -0,0 +1,81 @@
...
+    if (actionId) {
+      const actors = [ ...this.actorsIndex[actionId] || [], ...this.actorsIndex._undefined_ || [] ];
+      return actors.map((actor: ActorFunctionFactory):
+      IActorReply<ActorFunctionFactory, IActionFunctionFactory, IActorTest, IActorFunctionFactoryOutput> =>
+        ({ actor, reply: actor.test(action) }));
+    }
+    return super.publish(action);
+  }
+
+  protected getActorIdentifiers(actor: any): string[] {
```
rubensworks, line 62: Since this is hot code, we need to make sure this is highly optimized.

For example, the isArray check might slow down things, as it may lead the V8 optimizing compiler to see the variable as untyped, which makes it more difficult for V8 to optimize this into a C struct.

The string conversion and concat may slow down things as well. Keeping things in pure object-form will probably be faster.

rubensworks (changes requested):
I like the idea!

And you're sure that the BusIndexed can not be used in its current form?
If not, could it be tweaked (without perf impact) to make it compatible?
If not, then a dedicated new bus sounds good to me.

packages/bus-function-factory/lib/BusFunctionFactory.ts
jitsedesmet: You're right! High optimization is key here. If I think about it like that, I think that it might be better not to index the boolean, instead, let's only index the different function names. All actors returned from the index are still tested, and I don't think we can get a speed-up on the boolean check by doing it in the index itself.

Additionally, this allows the indexed bus and function factory bus to share more code.

rubensworks: 👍

packages/core/lib/BusIndexed.ts
```diff
@@ -69,8 +75,13 @@ export class BusIndexed<A extends Actor<I, T, O>, I extends IAction, T extends I
...
 
-  protected getActorIdentifier(actor: A): string {
-    return this.actorIdentifierFields.reduce((object: any, field): A => object[field], actor);
+  protected getActorIdentifiers(actor: A): string[] | undefined {
+    const identifierValue = <string | string[] | undefined> this.actorIdentifierFields
+      .reduce((object: any, field): A => object[field], actor);
+    if (!identifierValue) {
+      return;
+    }
+    return Array.isArray(identifierValue) ? identifierValue : [ identifierValue ];
```
rubensworks, line 84: Is it really required for this being possibly an array?

Because this logic will make it so that V8 can not make assumptions about the type of this value, which may slow down things.

engines/query-sparql-file/package.json
```diff
@@ -282,6 +282,7 @@
     "@comunica/actor-rdf-update-quads-hypermedia": "^3.2.1",
     "@comunica/actor-rdf-update-quads-rdfjs-store": "^3.2.1",
     "@comunica/actor-term-comparator-factory-expression-evaluator": "^3.2.1",
+    "@comunica/bus-function-factory": "^3.2.1",
```
rubensworks, line 285: Are you sure these deps are required here? (they might, just checking to make sure)

packages/core/lib/BusIndexed.ts
jitsedesmet: I think it does not matter here since this code is only run when an actor subscribes to the bus. It is thus only a startup issue?

engines/query-sparql-file/package.json
jitsedesmet: If they are not here, depcheck is not happy. I think it's necessary since the actors inherit from the dedicated actor which contains code?

packages/core/lib/BusIndexed.ts
rubensworks: Ah wait, there's both `getActionIdentifier` and `getActionIdentifiers`.
The first one is used at runtime for finding actors.
The second one during registration.
If my understanding is correct, then this comment can be closed :-)

engines/query-sparql-file/package.json
rubensworks: But then I'd expect that only the actors would need a dependency on this bus.

But if the depcheck says it needs to be there, it's probably right.

packages/core/lib/BusIndexed.ts
jitsedesmet: Correct

engines/query-sparql-file/package.json
jitsedesmet: No clue, but the same holds for bus query operation. I suspect it's fine.

rubensworks: Ok!

Issue #122, opened 2018-03-26
Invalid .bin files from repo packages


joachimvh:
#### Issue type:

_Only keep the issue types that are applicable._

- :bug: Bug

#### Description:
The generated .bin files of other packages in the monorep seem te be unusable.
E.g. `.bin/comunica-run` looks as follows:
```bash
"$basedir/../../../runner-cli/bin/run.js"   "$@"
exit $?
```
while a normal generated .bin file looks something like this:

```bash
#!/bin/sh
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")

case `uname` in
    *CYGWIN*) basedir=`cygpath -w "$basedir"`;;
esac

if [ -x "$basedir/node" ]; then
  "$basedir/node"  "$basedir/../acorn/bin/acorn" "$@"
  ret=$?
else 
  node  "$basedir/../acorn/bin/acorn" "$@"
  ret=$?
fi
exit $ret
```

Might be a Lerna thing, but it causes me to be unable to use those .bin files.
____
#### Environment:

_Exact versions of the software in your environment, not just *latest*. Only needed for :bug: and :snail:._

| software            | version(s)
| ------------------- | -------
| Comunica Init Actor | `@comunica/actor-init-hello-world` 1.0.0
| Comunica            | 1.0.0
| node                | 8.4.0
| npm                 | 5.7.1
| Operating System    | Windows

rubensworks:
Not sure if I'll be able to help with that, probably a Windows-only issue, as my bins are just symlinks.

joachimvh:
I found out about it because Arne had problems with the bin stuff, so I assumed there's also a problem on Unix. Also, symlinking to the file is different behaviour from what the bin files are supposed to do. As you can see in the example above, that bin file actually calls `node FILE_NAME_HERE` instead of just calling the file as a command. 

Also this is totally low priority, this issue is mostly here so we can remember it when the problem pops up :P

joachimvh:
Update: this is apparently only a problem with comunca-run, all the other .bin files are fine

PR #1432, opened 2024-09-27
Workflow and tooling updates

surilindur:
Here are the CI workflow and tooling (Webpack, Karma, Jest) updates from my working branch. The following major changes are included:
* Updated the Yarn version to 4.5.0 due to the speed increases. The postinstall script is then used for the Husky hook.
* Jest configuration is split into different 'projects' to allow testing the packages separately from the engines. The combined tests are run by default, however, to reach the coverage targets.
* The Webpack configuration is cleaned up and moved to the workspace root.
* The Karma browser test setup is updated to include Webkit.
* The workspace package scripts are now sorted, and named appropriately, with the build ones starting with `build:`, tests starting with `test:`, etc.
* The CI workflow is updated, and now runs almost everything in parallel.
* Documentation is now published using the official upload action from GitHub directly, rather than using the legacy branch approach. This might require changing the publish method in the repository settings. The allowed branches can be edited in the environment settings for pages, also in repository settings.

Any feedback is welcome. :slightly_smiling_face:


surilindur:
GitHub says I 'deployed it to github-pages' when, in reality, the actual deploy step was skipped (which can be seen from the documentation step of the actions). That is curious. In any case, it was not deployed, and would probably be blocked by the github-pages environment protection rules anyway. At least it was on my fork, before I added the branch name following the [documentation](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment).

.github/workflows/ci.yml
```diff
@@ -5,143 +5,314 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env:
+  DEFAULT_NODE_VERSION: 22.x
+  DOCUMENTATION_BRANCH: feat/simplify-extension-ignores
```
rubensworks, line 10: Is this correct?

.github/workflows/ci.yml
```diff
@@ -5,143 +5,314 @@ concurrency:
...
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Ensure line endings are consistent
+        run: git config --global core.autocrlf input
+      - name: Use Node.js ${{ env.DEFAULT_NODE_VERSION }}
+        uses: actions/setup-node@v4
+        with:
+          node-version: ${{ env.DEFAULT_NODE_VERSION }}
+      - name: Enable Corepack
```
rubensworks, line 24: Let's not use corepack, as Node will remove it: https://thenewstack.io/node-removes-corepack-bun-runs-native-c-from-javascript/

.github/workflows/ci.yml
```diff
@@ -5,143 +5,314 @@ concurrency:
...
-          key: ${{ runner.os }}-test-modules-${{ secrets.CACHE_VERSION }}-${{ hashFiles('**/yarn.lock') }}
-      - name: Setup firefox
-        uses: browser-actions/setup-firefox@latest
-        with:
-          firefox-version: latest
-      - name: Install dependencies
-        run: yarn install --frozen-lockfile
-      - name: Build project
-        run: yarn run build
-      - name: Build project -- esModuleInterop true
```
rubensworks, line 44: This must not be removed.

.github/workflows/ci.yml
```diff
@@ -5,143 +5,314 @@ concurrency:
...
-        run: yarn run build:ts --esModuleInterop true
-      - name: Run depcheck
-        run: yarn run depcheck
-      - name: Run tests
-        run: yarn run test:ci
-      - name: Run browser tests
```
rubensworks, line 50: Please keep the retry.

.github/workflows/ci.yml
```diff
@@ -5,143 +5,314 @@ concurrency:
...
-        if: ${{ matrix.os == 'ubuntu-latest' }}
-        uses: nick-fields/retry@v3
-        with:
-          timeout_minutes: 2
-          max_attempts: 3
-          command: yarn run test:browser
+          key: ${{ runner.os }}-yarn-${{ hashFiles('yarn.lock') }}
+      - name: Install project dependencies
+        run: yarn install --immutable
+      - name: Compile TypeScript
```
rubensworks, line 76: This and the following steps are not necessary, as they should be invoked automatically upon calling `yarn install`.

engines/query-sparql-file/package.json
```diff
@@ -54,7 +53,6 @@
     "build": "yarn run build:ts",
     "build:ts": "node \"../../node_modules/typescript/bin/tsc\"",
     "prepare": "comunica-compile-config config/config-default.json > engine-default.js && comunica-compile-config config/config-browser.json > engine-browser.js",
-    "browser": "yarn run prepare && \"../../node_modules/webpack/bin/webpack.js\" --config webpack.config.js --mode production",
```
rubensworks, line 57: Why was this one removed?

engines/query-sparql-file/package.json
```diff
@@ -204,9 +202,5 @@
     "@comunica/runner-cli": "^3.2.1",
     "@comunica/types": "^3.2.1",
     "process": "^0.11.10"
-  },
-  "browser": {
```
rubensworks, line 208: We also need to keep this.

engines/query-sparql-rdfjs-lite/webpack.config.ts
```diff
@@ -0,0 +1,10 @@
+import { createConfig } from '../../webpack.config';
```
rubensworks, line 1: Given the `.ts` rename, are you sure our compiler will always compile this down to `.js`? (since it's not in `lib/`)

jest.config.ts
```diff
@@ -0,0 +1,85 @@
+import type { Config } from '@jest/types';
```
rubensworks, line 1: Do we really need to rename these to `.ts`?
This probably means that we first need to build before we can run the tests.

package.json
```diff
@@ -1,46 +1,47 @@
...
   "workspaces": [
     "engines/*",
     "packages/*",
     "performance/*"
   ],
   "scripts": {
-    "test-changed": "lerna run test --since HEAD",
-    "build-changed": "lerna run build --since HEAD",
-    "test": "jest",
-    "test:ci": "jest --ci --maxWorkers=4 --coverage",
```
rubensworks, line 14: Why was this removed?

package.json
```diff
@@ -1,46 +1,47 @@
...
-    "doc": "packages/utils-monorepo/sh/typedoc-json.sh create && typedoc && packages/utils-monorepo/sh/typedoc-json.sh remove",
-    "preinstall": "node -e \"if (!fs.existsSync('packages/runner-cli/bin/run.js')) { var os = fs.createWriteStream('packages/runner-cli/bin/run.js'); os.write('#!/usr/bin/env node\\nconsole.error(\\'Temporary runner script during lerna installation. If you see this, please run yarn install again.\\');\\nprocess.exit(1);'); os.end(); }\" && node -e \"if (!fs.existsSync('packages/runner/bin/compile-config.js')) { var os = fs.createWriteStream('packages/runner/bin/compile-config.js'); os.write('#!/usr/bin/env node\\nconsole.error(\\'Temporary compile-config script during lerna installation. If you see this, please run yarn install again.\\');\\nprocess.exit(1);'); os.end(); }\"",
```
rubensworks, line 29: Why was this removed?

package.json
```diff
@@ -88,9 +90,11 @@
     "streamify-array": "^1.0.1",
     "ts-jest": "^29.0.5",
     "ts-loader": "^9.4.1",
+    "ts-node": "^10.9.2",
```
rubensworks, line 93: Not necessary?

packages/actor-init-query/package.json
```diff
@@ -60,9 +60,6 @@
     "readable-stream": "^4.5.2",
     "yargs": "^17.7.2"
   },
-  "optionalDependencies": {
```
rubensworks, line 63: Why was this removed?

setup-jest.js
```diff
@@ -1,3 +0,0 @@
-jest.setTimeout(20000);
```
rubensworks, line 1: Why was this removed?

.github/workflows/ci.yml
rubensworks: Also applies to all other jobs in this workflow.

rubensworks (changes requested):
Thanks for this @surilindur.

I do notice that you removed a lot of things that probably should not be removed.
Besides all my comments, can you have a detailed look at this PR's diff, to make sure all removals are really required?

.github/workflows/ci.yml
surilindur: The Node version yes, the docs branch should not be even used anywhere anymore, it is a remainder from my WIP branch. I need to double-check this, thank you for spotting it! :grimacing:

.github/workflows/ci.yml
surilindur: Oh, that is sad news. I think bumping the Yarn version without Corepack available would be really annoying for anyone working on the repo in that case, because manually setting up the correct version for everything would be confusing. :frowning:

.github/workflows/ci.yml
surilindur: True, I completely forgot this existed! Thank you for pointing it out.

.github/workflows/ci.yml
surilindur: I actually adjusted the package `postinstall` script to not build everything by default, because it was doing it in the CI, as well, which is unnecessary for the lint task, hence the splitting up also here. Now it explicitly builds when the user instructs it to build. I thought it would be clearer this way, and would help better illustrate the need to build after changes. :slightly_smiling_face:

package.json
surilindur: These scripts were renamed, and the CI test now uses the same script as the normal tests. I changed the normal test script to use `--runInBand` because it is surprisingly faster that way (easy to test locally, it took around 90 seconds without it, around 55 with it, if I remember correctly).

.github/workflows/ci.yml
rubensworks: IMO, the dev environment should be fully ready after running `install`.
Requiring users to still execute a `build` script is counter-intuitive IMO, so I would prefer if we keep this.

package.json
surilindur: It appears to just print a warning that tells the user to do yarn install again, when the packages in question have not been built. This implicitly relies on the `postinstall` hook to build everything, which is not told to the user anyway. I have never seen the script get called, and the expectation would be that everyone runs build after their changes in a TypeScript project.

So in short, I think it was an unnecessary and unused means to display a message that did not actually say what the problem is, while also relying on the `postinstall` script that I changed to not build everything upon install. I have had situations where I just want an install without doing a build, knowing the build would fail anyway, for example.

package.json
surilindur: It was needed for either Webpack or Jest to be able to parse the TS config files.

packages/actor-init-query/package.json
surilindur: I could not get depcheck to stop complaining about a missing `process` dependency when it was located in optional ones, so I moved it to non-optional. :sweat:

setup-jest.js
surilindur: The value is now set in the main Jest config file as `testTimeout`.

package.json
rubensworks: We need to keep this, because jest does some things differently when run in a CI: https://jestjs.io/docs/cli#--ci

You probably have some caching going on, because `runInBand` takes almost 4 minutes on my system.

package.json
rubensworks: Please check the git blame history on this.
I don't remember the precise reason, but there is an important reason for this.

package.json
rubensworks: This then ties into my other comment on if we really need to migrate all these js config files to ts. So let's continue this discussion there.

engines/query-sparql-file/package.json
surilindur: Because file-based querying is unlikely to work in the browser, so I thought building a browser bundle for it was kind of unnecessary. But maybe I am wrong.

engines/query-sparql-file/package.json
surilindur: (see the comment above on the `browser` build script)

packages/actor-init-query/package.json
rubensworks: But it's totally gone now?

Also, something probably changed in depcheck then due to your changes.

setup-jest.js
rubensworks: Ok 👍

engines/query-sparql-file/package.json
rubensworks: It most definitely works in the browser, for querying (only) Linked Data documents.

jest.config.ts
surilindur: They do not need to be built beforehand, Jest can use TS config files and the setup script seems to be handled as it is, as well.

engines/query-sparql-file/package.json
surilindur: Ah okay, I thought it was specifically for files, not online documents. The name was a bit confusing.

packages/actor-init-query/package.json
surilindur: It is actually still in `dependencies` above, but it was there already previously. This is strange. :thinking:

jest.config.ts
surilindur: Having the config files as TS makes it easier to do typing in them, and helps set them apart from compiler-generated JS files. That was my reasoning behind the change, but maybe it is not enough to justify the changes, actually.

package.json
surilindur: It seems to have been related to lerna bootstrapping in 83f1feea648760a2739f698b3249a313d3ee1670, but the repo no longer uses lerna bootstrapping as far as I know? It seems to have been something before Yarn workspaces were introduced. The CI also runs fine on Windows without this workaround now (something discussed in #122), so I would imagine it is not needed anymore.

package.json
surilindur: It seems you were right, I ran all three versions locally to test, with the `--ci` flag, on this branch:
* `jest --selectProjects combined --ci --runInBand` ~60...70s
* `jest --selectProjects combined --ci --maxWorkers 4` ~30s
* `jest --selectProjects combined --ci` ~42s

I think I must have used different branches between the two or something.

surilindur:
After going through the changes again, most of the things here would need to be reverted anyway for one reason or another, and the remaining changes would not be significant enough to warrant that change. I was not aware of all the small details or reasons around why things are done the way they are. So I will close this, and make smaller PRs later if relevant. Thank you for all the feedback, and sorry about taking your time! :sweat:

package.json
rubensworks: Ok, thanks for checking. Looks like this can be removed indeed now.

PR #1437, opened 2024-09-30
Support Brotli by adding `Accept-Encoding` header with `br,gzip,deflate` if not set

smessie:


packages/actor-http-fetch/lib/FetchInitPreprocessor.ts
```diff
@@ -27,6 +27,13 @@ export class FetchInitPreprocessor implements IFetchInitPreprocessor {
       halfDuplex = true;
     }
 
+    // Add 'Accept-Encoding' headers
+    const headers = new Headers(init.headers);
+    if (!headers.has('Accept-Encoding')) {
+      headers.set('Accept-Encoding', 'br,gzip,deflate');
+    }
+    init = { ...init, headers };
```
rubensworks, line 35: This line can move inside the if-statement.

rubensworks (changes requested):
Thanks @smessie!

PR #1442, opened 2024-10-04
Actors do not create an expression evaluator instance in their test function

jitsedesmet:


engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -375,15 +375,15 @@ describe('System test: QuerySparql', () => {
           const context = <any> { sources: [ store ]};
           context.extensionFunctions = baseFunctions;
           await expect(engine.query(baseQuery('nonExist'), context)).rejects.toThrow(
-            `Query operation processing failed: none of the configured actors were able to handle the operation type filter`,
+            `ActorFunctionFactory_default_bus failed to handle an action`,
```
rubensworks, line 378: That reminds me; can you add a test to https://github.com/comunica/comunica/blob/42146e155f4fe77acbcd1171263342eb746f4b6b/engines/query-sparql/test/Mediators-test.ts with the new buses that you introduced?
These aim to test the high-level behaviour of the configured mediator on that bus, and its error message when no actors on that bus can answer the action.
Can be a separate PR.

rubensworks (approved):

PR #1445, opened 2024-10-08
Chore: rename expression evaluator functions dirs

jitsedesmet:


rubensworks:
I guess the package.json's and README's will also have to be updated.

PR #1458, opened 2024-11-20
Adjust SPARQL Service Description parsing

surilindur:
Here is a small set of changes to adjust the parsing of SPARQL Service Descriptions. Specifically, this does the following:
* Add parsing of `unionDefaultGraph`, because it can be later passed on to `QuerySourceSparql` and used for cardinality estimation inside that class.
* Add parsing of `defaultDataset` as it could be interesting, and is usually used as an intermediate step from service -> default dataset -> default graph. This also helps ensure the correct default graph is chosen, in case there are multiple datasets with their own default graphs (unlikely to happen, though).

I will keep this as a draft until the remaining bits of SPARQL SD / VoID / cardinality things are done.


packages/actor-rdf-metadata-extract-sparql-service/package.json
```diff
@@ -44,5 +44,8 @@
     "@comunica/bus-rdf-metadata-extract": "^4.0.2",
     "@comunica/core": "^4.0.2",
     "relative-to-absolute-iri": "^1.0.7"
+  },
+  "devDependencies": {
+    "@rdfjs/types": "*"
```
rubensworks, line 49: Must be a regular dep.

surilindur:
I adjusted the implementation to also extract `basicFederatedQuery` (flag when this feature is declared), as well as the following lists of values when they are available: `inputFormats`, `resultFormats` and `supportedLanguages`. The `@rdfjs/types` dependency is also a non-dev one now.

packages/actor-rdf-metadata-extract-sparql-service/lib/ActorRdfMetadataExtractSparqlService.ts
```diff
@@ -27,28 +28,63 @@ export class ActorRdfMetadataExtractSparqlService extends ActorRdfMetadataExtrac
       // Forward errors
       action.metadata.on('error', reject);
 
-      // Immediately resolve when a SPARQL service endpoint URL has been found
-      const metadata: any = {};
-      action.metadata.on('data', (quad) => {
-        if (quad.predicate.value === 'http://www.w3.org/ns/sparql-service-description#endpoint' &&
-          (quad.subject.termType === 'BlankNode' || quad.subject.value === action.url)) {
-          metadata.sparqlService = quad.object.termType === 'Literal' ?
```
rubensworks, line 35: This literal check seems to be gone

rubensworks (changes requested):

packages/actor-rdf-metadata-extract-sparql-service/lib/ActorRdfMetadataExtractSparqlService.ts
surilindur: Yes, I figured it would not be needed, if a named node value would also happen to be relative and could be resolved/handled with `resolveIri`? If that is not the case, then I can restore the literal check.

rubensworks: Named nodes are always absolute.
Otherwise something would be horribly broken in the parser.

rubensworks (changes requested):
Besides that one comment, everything looks good to me!

packages/actor-rdf-metadata-extract-sparql-service/lib/ActorRdfMetadataExtractSparqlService.ts
surilindur: This is fixed now!

PR #1459, opened 2024-11-21
Add VoID description metadata extract actor

surilindur:
This is a draft PR for the VoID description metadata extract actor, rewritten with performance in mind, because initial parsing of the entire Wikidata VoID description was taking too long. Now the extractor just stores all the data in an RDF/JS store, and queries the relevant parts when needed.

I will keep this as a draft until I have done more testing with and without SPARQL endpoints. :slightly_smiling_face:


packages/actor-rdf-metadata-extract-void/lib/ActorRdfMetadataExtractVoid.ts
```diff
@@ -0,0 +1,547 @@
...
+  private readonly bindingsCache: LRUCache<string, RDF.Bindings[]>;
+
+  public static readonly VOID = 'http://rdfs.org/ns/void#';
+  public static readonly SPARQLSD = 'http://www.w3.org/ns/sparql-service-description#';
+
+  public constructor(args: IActorRdfMetadataExtractVoidArgs) {
+    super(args);
+    this.queryEngine = new QueryEngineBase(args.actorInitQuery);
+    this.inferUriRegexPattern = args.inferUriRegexPattern;
+    this.bindingsCache = new LRUCache<string, RDF.Bindings[]>({ max: args.bindingsCacheSize });
```
rubensworks, line 32: Is this cache really needed?
If really so, then we need to listen to cache invalidation.

packages/actor-rdf-metadata-extract-void/lib/ActorRdfMetadataExtractVoid.ts
```diff
@@ -0,0 +1,547 @@
...
+  source: string;
+  /**
+   * The identifier of the dataset, either an IRI or a blank node.
+   */
+  identifier: RDF.NamedNode | RDF.BlankNode;
+  /**
+   * The RDF/JS store containing the data for this dataset description.
+   * Note that the store might also contain data for other datasets defined in the dame VoID description.
+   */
+  store: RDF.Store;
```
rubensworks, line 529: Do we need to expose this?
Might be safer to hide it wrt future breaking changes?

rubensworks (commented):
Nice stuff!

packages/actor-rdf-metadata-extract-void/lib/ActorRdfMetadataExtractVoid.ts
surilindur: It is probably not strictly needed, but I figured it would make sense, since a lot of the formulae end up pulling the same properties from the VoID descriptions for different purposes (multipliers, divisors, etc.). I think I will remove this cache thing, and if performance becomes an issue later, it can be added back.

packages/actor-rdf-metadata-extract-void/lib/ActorRdfMetadataExtractVoid.ts
surilindur: That is a great point! I did not realise it can be hidden. Thank you. :slightly_smiling_face:

packages/actor-rdf-metadata-extract-void/lib/ActorRdfMetadataExtractVoid.ts
surilindur: There is now a per-dataset cache, that is used on the dataset metadata object itself (stored in the metadata stream), which will be deleted whenever the metadata itself is deleted, and does not need actor-specific invalidation handling, like other metadata.

packages/actor-query-source-identify-hypermedia-sparql/lib/QuerySourceSparql.ts
```diff
@@ -180,10 +190,17 @@ export class QuerySourceSparql implements IQuerySource {
         }));
 
         const cachedCardinality = this.cache?.get(countQuery);
-        if (cachedCardinality !== undefined) {
```
rubensworks, line 183: Please restore this to what it was. This is needed because cardinalities can be zero.

packages/actor-rdf-metadata-extract-void/lib/ActorRdfMetadataExtractVoid.ts
```diff
@@ -0,0 +1,220 @@
...
+
+        # Exclude union default graphs that merely combine every other graph
+        FILTER NOT EXISTS {
+          ?var4 sd:feature sd:UnionDefaultGraph .
+          ?var4 sd:defaultDataset/sd:defaultGraph ?identifier .
+        }
+
+        # By definition, sd:Graph and sd:Dataset are both also void:Datasets, however
+        # sd:Dataset represents the dataset and not the graph, so it can likely be ignored.
+        FILTER(?type IN (sd:Graph,void:Dataset))
```
rubensworks, line 115: Could you try with a VALUES clause instead? Will probably be faster.

rubensworks (changes requested):
Code looks very clean, I liked it @surilindur!

> There is also no cardinality estimation for property paths

What will happen exactly in this case? Incorrect estimations? Or fallback to sending COUNT queries?

engines/config-query-sparql/config/rdf-metadata-extract/actors.json
```diff
@@ -45,6 +46,10 @@
     {
       "@id": "urn:comunica:default:rdf-metadata-extract/actors#sparql-service",
       "@type": "ActorRdfMetadataExtractSparqlService"
+    },
+    {
```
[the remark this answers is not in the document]
surilindur: Ah okay, then I think leaving it out of the config makes sense, to make it easier to set up the benchmarks before the next major is released. Is it okay to have the packages in there, still, before the next major?

packages/actor-query-source-identify-hypermedia-sparql/lib/QuerySourceSparql.ts
surilindur: The `cachedCardinality` value is an `RDF.QueryResultCardinality` object or `undefined`, so the new type of checking should also work. I guess it was previously a number, which is why the check was implemented that way.

packages/actor-rdf-metadata-extract-void/lib/VoidDataset.ts
```diff
@@ -0,0 +1,366 @@
...
+
+  public constructor(args: IVoidDatasetArgs) {
+    this.triples = args.triples;
+    this.source = args.source;
+    this.store = args.store;
+    this.identifier = args.identifier;
+    this.vocabularies = args.vocabularies;
+    this.queryEngine = args.queryEngine;
+    this.resourceUriPattern = args.resourceUriPattern;
+    this.bindingsCache = new LRUCache({ max: args.bindingsCacheSize });
```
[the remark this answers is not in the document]
surilindur: The cache is on an object in the metadata, so unless all other metadata also gets purged when cache is invalidated, I think this does not need a listener.

The cache simply exists because parsing all the metadata into an object when it is discovered takes too long, but if it were faster, the metadata entry would instead be an object with a lot of fields that would not need to be reset upon cache invalidation, either. That is my reasoning, at least. :thinking:

surilindur:
> What will happen exactly in this case? Incorrect estimations? Or fallback to sending COUNT queries?

It will return infinity, which is not optimal, for sure. But after checking some more queries with property paths in them, I think Comunica does the translation into a set of joins before it sends them for estimation, so there do not seem to be any `Algebra.Path` objects getting into the estimator function. Maybe in some edge cases there will be, however.

engines/config-query-sparql/config/rdf-metadata-extract/actors.json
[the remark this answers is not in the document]
rubensworks: Sure, the packages can be in there.

packages/actor-rdf-metadata-extract-void/lib/VoidDataset.ts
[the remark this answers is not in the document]
rubensworks: That sounds reasonable indeed.

rubensworks:
> Maybe in some edge cases there will be, however.

Yeah, pretty sure there will be some cases that don't do prior translation.
For those, it might be good to fallback to sending the count?

surilindur:
All the feedback should now have been addressed, and the code will fall back to sending a count query if a local estimate cannot be produced or it ends up being infinity. The packages are there, but they are not installed as dependencies to the query engines even.

Tomorrow I will go through some of the queries I tested with and add the ones that work without local cardinality estimation into the web performance benchmark. It will probably be a separate pull request, because it is not strictly tied to this one.

packages/actor-rdf-metadata-extract-void/lib/VoidDataset.ts
```diff
@@ -0,0 +1,426 @@
...
```
rubensworks, line 33: Could you double-check my earlier comment regarding these variable names? (would also apply to the actor)

packages/actor-rdf-metadata-extract-void/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+After installing, this package can be added to your engine's configuration as follows:
+```json
+{
+  "@context": [
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-rdf-metadata-extract-void/^4.0.0/components/context.jsonld"
+  ],
+  "actors": [
+    {
+      "@id": "urn:comunica:default:rdf-metadata-extract/actors#void",
```
rubensworks, line 35: If I understand correctly, this is the only actor that needs to be added for all of this to work, right?

If so, then I may have an idea for doing this in a backwards-compat manner.
(Because it would be good to have this in the default config asap)
But I will only try it after this PR is merged.

rubensworks (changes requested):
Just two minor things, then this should be good to go!

rubensworks:
Nice work @surilindur! 🚀

PR #1460, opened 2024-11-21
Remove separate engine and package projects from Jest config

surilindur:
This is just to remove the separate engine and package test setups from Jest, since the unit tests will probably not be separated from the engine ones in the near future coverage-wise, and having the projects in there seems to interfere with running only specific tests locally.

Any feedback would be welcome, though.

## Summary by CodeRabbit

- **Chores**
	- Simplified Jest configuration by consolidating multiple project settings into a single configuration object.
	- Updated the `test:node` script in `package.json` to streamline test execution by removing project selection.

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -1,16 +1,13 @@
 import type { Config } from '@jest/types';
 
-const moduleFileExtensions: string[] = [ 'ts', 'js' ];
-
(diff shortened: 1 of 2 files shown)


rubensworks:
Thanks!

PR #1462, opened 2024-11-25
Add HTTP rate limit actor

surilindur:
Here is a draft PR for the simple rate limit actor, after a lot of revisions back and forth.
* The actor current keeps track of the *n* previous request durations as a sliding list/window (not sure what the correct term it), measured from when the request was sent, to when the server replied, excluding the time taken to read the request body. Based on these previous request durations, it spaces out future requests in a way that avoids request spikes.
* The actor also maintains a concurrent request limit per host. When a request to a host is successful, and there is a considerable number of requests in queue for it (currently >10x the concurrent request limit), the actor will increase the concurrent request limit by 1. When a request fails, it will halve the limit.

The actor is really simple in the way it functions, because after multiple revisions, I decided that it is better to keep it simple. For example, tracking a host-specific processing time budget client-side (such as the Wikidata 60s of processing every 60s) would make the implementation unnecessarily complicated, and would limit the concurrent request count to 1 because the estimated running time for a query would have to be infinite to avoid budget overruns. While this would work for SPARQL endpoints, it would be unnecessarily slow for document-based servers.

Any feedback would be welcome. :slightly_smiling_face: When I find the time, I will try to do some more benchmarks on this to see the impact. From what I have noticed, it does not unnecessarily slow down things, but the history length does impact how fast it will ramp up the requests per second initially.

### DIFF
--- a/packages/actor-http-limit-rate/README.md
+++ b/packages/actor-http-limit-rate/README.md
@@ -1,0 +1,46 @@
+# Comunica HTTP Rate Limit Actor
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-http-limit-rate.svg)](https://www.npmjs.com/package/@comunica/actor-http-limit-rate)
+
+An [HTTP](https://github.com/comunica/comunica/tree/master/packages/bus-http) actor that performs rate limiting,
+by spacing out future requests based on past request durations,
+in an attempt to match the number of requests sent per second to the number of responses served per second by the server.
+By default, the actor waits for the first request to fail for a given host prior to spacing out requests.
+
+This module is part of the [Comunica framework](https://github.com/comunica/comunica),
+and should only be used by [developers that want to build their own query engine](https://comunica.dev/docs/modify/).
+
+[Click here if you just want to query with Comunica](https://comunica.dev/docs/query/).
+
+## Install
+
+```bash
+$ yarn add @comunica/actor-http-limit-rate
+```
+
+## Configure
+
+After installing, this package can be added to your engine's configuration as follows:
+
+```json
+{
+  "@context": [
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-http-limit-rate/^4.0.0/components/context.jsonld"
+  ],
+  "actors": [
+    {
+      "@id": "urn:comunica:default:http/actors#limit-rate",
+      "@type": "ActorHttpLimitRate"
+    }
+  ]
+}
+```
+
+### Config Parameters
+
+* `mediatorHttp`: A mediator over the [HTTP bus](https://github.com/comunica/comunica/tree/master/packages/bus-http).
+* `httpInvalidator`: A mediator over the [HTTP invalidate bus](https://github.com/comunica/comunica/tree/master/packages/bus-http-invalidate).
+* `correctionMultiplier`: How aggressively the request interval should follow the latest response time. Defaults to `0.1`.
+* `failureMultiplier`: The response time of a failed request is taken into account with this multiplier applied. Defaults to `10.0`.
+* `limitByDefault`: Whether the actor should do rate limiting by default, already before a request fails. Defaults to `false`.
+* `allowOverlap`: Whether the actor should allow overlapping requests to be sent to the server when rate limiting is applied. Defaults to `false`.
--- a/packages/actor-http-limit-rate/lib/ActorHttpLimitRate.ts
+++ b/packages/actor-http-limit-rate/lib/ActorHttpLimitRate.ts
@@ -1,0 +1,174 @@
+import type { IActionHttp, IActorHttpOutput, IActorHttpArgs, MediatorHttp } from '@comunica/bus-http';
+import { ActorHttp } from '@comunica/bus-http';
+import type { ActorHttpInvalidateListenable, IActionHttpInvalidate } from '@comunica/bus-http-invalidate';
+import type { TestResult } from '@comunica/core';
+import { ActionContextKey, failTest, passTest } from '@comunica/core';
+import type { IMediatorTypeTime } from '@comunica/mediatortype-time';
+
+export class ActorHttpLimitRate extends ActorHttp {
+  private readonly hostData: Map<string, IHostData>;
+  private readonly correctionMultiplier: number;
+  private readonly failureMultiplier: number;
+  private readonly limitByDefault: boolean;
+  private readonly allowOverlap: boolean;
+  private readonly httpInvalidator: ActorHttpInvalidateListenable;
+  private readonly mediatorHttp: MediatorHttp;
+
+  // Context key to indicate that the actor has already wrapped the given request
+  private static readonly keyWrapped = new ActionContextKey<boolean>('urn:comunica:actor-http-limit-rate#wrapped');
+
+  public constructor(args: IActorHttpLimitRateArgs) {
+    super(args);
+    this.mediatorHttp = args.mediatorHttp;
+    this.httpInvalidator = args.httpInvalidator;
+    this.httpInvalidator.addInvalidateListener(action => this.handleHttpInvalidateEvent(action));
+    this.correctionMultiplier = args.correctionMultiplier;
+    this.failureMultiplier = args.failureMultiplier;
+    this.limitByDefault = args.limitByDefault;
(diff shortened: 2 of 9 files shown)


packages/actor-http-limit-rate/README.md
```diff
@@ -0,0 +1,47 @@
+# Comunica HTTP Rate Limit Actor
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-http-limit-rate.svg)](https://www.npmjs.com/package/@comunica/actor-http-limit-rate)
+
+An [HTTP](https://github.com/comunica/comunica/tree/master/packages/bus-http) actor that performs simple rate limiting.
```
rubensworks, line 5:
```suggestion
An [HTTP](https://github.com/comunica/comunica/tree/master/packages/bus-http) actor that performs rate limiting.
```

packages/actor-http-limit-rate/lib/ActorHttpLimitRate.ts
```diff
@@ -0,0 +1,212 @@
...
+   */
+  public async waitForSlot(host: string): Promise<void> {
+    return new Promise<void>((resolve, reject) => {
+      const send = (): void => {
+        // Immediately reserve the free slot by incrementing the 'active requests' counter.
+        this.hostData[host].openRequests++;
+        resolve();
+      };
+
+      if (typeof this.hostData[host] === 'undefined') {
```
rubensworks, line 109:
```suggestion
      if (!this.hostData[host]) {
```

packages/actor-http-limit-rate/lib/ActorHttpLimitRate.ts
```diff
@@ -0,0 +1,212 @@
...
+    await this.waitForSlot(host);
+
+    // Delay the request when relevant
+    await this.delayRequest(host);
+
+    const timeStart = Date.now();
+
+    this.hostData[host].latestRequestTimestamp = timeStart;
+
+    const response = await this.mediatorHttp.mediate({
```
rubensworks, line 48: The fetch operation could also be rejected.
I guess this would need to be try-catched here to correctly handle the hostData?

packages/actor-http-limit-rate/README.md
```diff
@@ -0,0 +1,47 @@
...
+    }
+  ]
+}
+```
+
+### Config Parameters
+
+* `mediatorHttp`: A mediator over the [HTTP bus](https://github.com/comunica/comunica/tree/master/packages/bus-http).
+* `httpInvalidator`: A mediator over the [HTTP invalidate bus](https://github.com/comunica/comunica/tree/master/packages/bus-http-invalidate).
+* `historyLength`: The number of past requests to consider for the delay average.
```
rubensworks, line 46: Can you mention the default value for this and the next param?

rubensworks (changes requested):
Just to make sure I understand correctly, this limits concurrent requests to 10, and delays requests to avoid spikes?

I wonder if we really want this. Why wouldn't we want to send requests as soon as possible?
I suspect this actor will simply slow things without a clear benefit.

It's also unfortunate that the server-provided headers (such as the one from Wikidata) are not being taken into account anymore, because I thought this was the main motivation for this work?

surilindur:
> Just to make sure I understand correctly, this limits concurrent requests to 10, and delays requests to avoid spikes?

The history length is 10 by default, but the actor does not limit the number of concurrent requests. If all requests consistent succeed, it will scale the maximum concurrency up until infinity (which actually needs to be handled, so I need to fix that). It will also scale the interval between requests down until 0 if all requests would consistently complete immediately upon being sent. It would be easier to draw it somehow, but basically, if the server takes on average 1 second to reply to requests, the actor will scale the requests so that it sends 1 request/second to the host.

> Why wouldn't we want to send requests as soon as possible?

Because Comunica would then send up to 100+ requests/second regardless of whether the requests succeed or not, and the server would eventually stop replying entirely because of this. The HTTP retry actor handled status code 429, but ideally, the 429 should be avoided completely. Without rate limiting, the request process would be a repetition of spamming as many requests as possible, hitting 429 and waiting a while, then doing the same all over again. But if that is okay, then this actor is not needed, which is correct.

> It's also unfortunate that the server-provided headers (such as the one from Wikidata) are not being taken into account anymore, because I thought this was the main motivation for this work?

I need to check which headers those were. I have a feeling that I forgot something. :grimacing:

rubensworks:
> The history length is 10 by default, but the actor does not limit the number of concurrent requests. If all requests consistent succeed, it will scale the maximum concurrency up until infinity (which actually needs to be handled, so I need to fix that). It will also scale the interval between requests down until 0 if all requests would consistently complete immediately upon being sent. It would be easier to draw it somehow, but basically, if the server takes on average 1 second to reply to requests, the actor will scale the requests so that it sends 1 request/second to the host.

Aha ok, then I misunderstood.

The Wikidata comment may be something to look into first though.

surilindur:
This should now be ready for review! :partying_face: 

As for the Wikidata header, the HTTP status code 429 (too many requests) is already handled by the retry actor, so plugging this one between the retry actor and the proxy one should be sufficient, as this one does not do any retries on its own, and the retry actor handles the postponing of further requests based on that header (because otherwise, without a retry, it would be a failed request).

Should this also be added but not enabled by default, as with the VoID extractor?

rubensworks:
> Should this also be added but not enabled by default, as with the VoID extractor?

Yes.
For this PR at least.
But I will try out a way to include this in a backwards-compat manner into the config soon.

surilindur:
This is now ready for review @rubensworks. All the alternative functionality (overlapping requests, rate limit by default) are behind configuration options, to allow for different types of functionality.

The default configuration does not do any rate limiting before a request fails, and after the first failed request it will start spacing out requests in a way that matches the server response times, with higher response times reported internally for failed requests, by a configurable multiplier, using a configurable multiplier to smooth the tracking of response times.

The DBPedia SPARQL endpoint stopped responding to me this afternoon entirely, but the functionality is finished and it seems to behave as expected for TPF/QPF servers and other SPARQL endpoints.

surilindur:
The BSBM benchmark was having some networking issues coming from Java, but the WatDiv benchmark worked. I did a run with both the rate limiter and the VoID extraction enabled in their default configs, and the performance seems mostly identical. The queries take on average 0.9% longer, but the variance is already way more than that. Below is the LaTeX plot from the WatDiv TPF benchmark.
image

The rate limiter should go between the retry actor and the proxy one in the config.

rubensworks:
> performance seems mostly identical

Nice!

Let's put this one on hold until #1506 is merged, so we can evaluate the performance impact according to the CI first.

surilindur:
This is now rebased on master again and ready for review! :slightly_smiling_face: It does not do anything before the first failed request to a host, so the performance impact is non-existent for most cases.

rubensworks:
Looks good!

Could you add it in the default config (in a backwards-compat manner)? Like here: https://github.com/comunica/comunica/commit/f9646063f58d7e917776cfc6ddcf10f1695bdef9

surilindur:
This should now be ready for review @rubensworks, I added the actor by default via a new file using the other commit as a reference.

Some of the CI tests for mediators failed due to timeouts on Mac, and I am not sure why, since they seem local (so it should not be network-related) and only two of them had issues. I do not think it is related to the changes I made.

rubensworks (changes requested):
The code looks good to me!

Could you double-check coverage?

And if I remember correctly, you manually ran the TPF perf tests for this one as well, right?

surilindur:
> Could you double-check coverage?

The missing coverage in the CI comes from `packages/actor-function-factory-term-addition/lib/index.ts` according to Coveralls report, which does not seem to be related to any of my changes. I cloned the branch locally and ran the unit tests, but it reported 100% coverage. Would running the CI again fix it?

> And if I remember correctly, you manually ran the TPF perf tests for this one as well, right?

Yes, and the performance impact was non-existing, because none of the requests failed.

By default, for the rate limiter to do something, a request has to fail. And if a request fails by default in Comunica, the query also fails, I think. So to actually use the rate limit, a retry count of 1 or more should be set, or the lenient flag.

rubensworks:
> Would running the CI again fix it?

Possibly.
Could you amend and force-push your last commit to retrigger the CI?

Issue #1254, opened 2019-05-20
Support flags for REGEX / REPLACE


wschella:
The flags for regex/replace in SPARQL don't match up with those in JS, so passing them straight does not work.

PR #1464, opened 2024-11-28
Add Regex spec tests

jitsedesmet:
A [new PR on the rdf-tests repo](https://github.com/w3c/rdf-tests/pull/155) aims to add tests for the regex function.
This PR aims to add those tests to the Comunica suite. I expect some tests will need to be skipped, since our implementation of the regex function just calls the JavaScript regex function. Seen here: https://github.com/comunica/comunica/blob/master/packages/actor-function-factory-term-regex/lib/TermFunctionRegex.ts#L31-L32

Status:
* Most tests pass,
* Tests with the q flag fail since JS regex does not know that flag.

## Summary by CodeRabbit

- **New Features**
  - Introduced multiple test files to validate various regex functionalities, including case insensitivity, character class expressions, and quantifiers.
  - Added tests for regex operations with whitespace handling, multiline contexts, and specific quantifier behaviors.
  - Expanded dataset for testing regex quantifiers with new string properties.

- **Bug Fixes**
  - Implemented comprehensive testing for regex patterns to ensure expected behavior across different scenarios.

- **Documentation**
  - Each test file includes manifest entries detailing the queries, expected results, and data used for validation.
  - Updated README files for both the regex and replace packages to clarify regex evaluation capabilities and handling of specific flags.

### DIFF
--- a/engines/query-sparql-rdfjs-lite/webpack.config.ts
+++ b/engines/query-sparql-rdfjs-lite/webpack.config.ts
@@ -3,8 +3,8 @@ import { createConfig } from '../../webpack.config';
 const liteConfig = createConfig(__dirname);
 
 if (typeof liteConfig.performance === 'object') {
-  liteConfig.performance.maxAssetSize = 900_000;
-  liteConfig.performance.maxEntrypointSize = 900_000;
+  liteConfig.performance.maxAssetSize = 910_000;
+  liteConfig.performance.maxEntrypointSize = 910_000;
 }
 
 export default liteConfig;
--- a/packages/actor-function-factory-term-regex/README.md
+++ b/packages/actor-function-factory-term-regex/README.md
@@ -11,7 +11,6 @@ The `x` and `q` flag not present in the JS engine are implemented by preprocessi
 As a result of using the JS regex engine, the bundle size of this package is small, but some non-spec compliant edge cases are present.
 Examples are the skipped tests in [op.regex-test.ts](https://github.com/comunica/comunica/blob/master/packages/actor-function-factory-term-regex/test/op.regex-test.ts).
 
-
 This module is part of the [Comunica framework](https://github.com/comunica/comunica),
 and should only be used by [developers that want to build their own query engine](https://comunica.dev/docs/modify/).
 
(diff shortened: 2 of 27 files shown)


rubensworks:
Nice work @jitsedesmet!

Before I merge this, I wonder how much work it would be to add support for those unsupported flags?
Because ideally we should support all flags in the SPARQL spec at some point.

Maybe we could perform custom pre/post-processing steps on the regex and terms when those flags are detected?

jitsedesmet:
Quickly skimming over the [docs of the flags of the regex function](https://www.w3.org/TR/xpath-functions/#flags) tells me this shouldn't be too hard using some pre-processing. Regex engines are a little scary though as they all have weird edge cases.

When Comunica sees a REGEX SPARQL function, it will just pass that function to the JS-regex-engine ([implementation](https://github.com/comunica/comunica/blob/master/packages/actor-function-factory-term-regex/lib/TermFunctionRegex.ts#L30-L35)). A first thing to note here is that we do not add the `u` flag, which is best practice because that flag not _just_ allows the parsing of Unicode (see [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicode)):
> There are other changes to the parsing behavior that prevent possible syntax mistakes (which are analogous to [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode) for regex syntax)

All this is to say, an implementation for the flags shouldn't be too hard. An implementation that is 100% spec compliant probably is hard.

rubensworks:
Are you up for trying it? And if so, would you prefer merging this PR first, or implementing it as part of this PR?

jitsedesmet:
I will look at the support of those flags :) I think we can still add that i this PR.

rubensworks (approved):
Very nice work @jitsedesmet!

I also very much appreciate the comments in the code.
They definitely help improving code understandability!

rubensworks:
Before I merge, does this close #1254?
If not, what other flags would still need to be added in the future?
This PR here focuses only on q and x, right?

PR #1465, opened 2024-11-28
fix floating point error for very low cardinalities

maartyman:
I ran into an issue where comunica stopped producing results, I figured out that when joining low cardinalities the cardinality can keep decreasing until there is a floating point error and the cardinality is set to 0. This causes comunica to choose the join-multi-none actor and close the bindings streams. In reality, the cardinality was still one, but comunica didn't produce any results.

This PR is a possible fix, and I added a test for it as well.

## Summary by CodeRabbit

- **New Features**
	- Enhanced cardinality handling to prevent zero values in results, improving accuracy.
  
- **Bug Fixes**
	- Addressed potential float overflow issues related to cardinality calculations.

- **Tests**
	- Added new test cases to validate cardinality behavior under low value conditions, ensuring robustness.

### DIFF
--- a/packages/bus-rdf-join/lib/ActorRdfJoin.ts
+++ b/packages/bus-rdf-join/lib/ActorRdfJoin.ts
@@ -243,18 +243,24 @@ TS
     if (partialMetadata.cardinality) {
       cardinalityJoined = partialMetadata.cardinality;
     } else {
+      let hasZeroCardinality = false;
       cardinalityJoined = metadatas
         .reduce((acc: RDF.QueryResultCardinality, metadata) => {
           const cardinalityThis = ActorRdfJoin.getCardinality(metadata);
+          if (cardinalityThis.value === 0) {
+            hasZeroCardinality = true;
+          }
           return {
             type: cardinalityThis.type === 'estimate' ? 'estimate' : acc.type,
             value: acc.value * (optional ? Math.max(1, cardinalityThis.value) : cardinalityThis.value),
           };
         }, { type: 'exact', value: 1 });
-      if (cardinalityJoined.value > 1e-9) {
+      // The cardinality should only be zero if one of the entries has zero cardinality, not due to float overflow
+      if (!hasZeroCardinality) {
         cardinalityJoined.value *= (await this.mediatorJoinSelectivity.mediate({ entries, context })).selectivity;
-      } else {
-        cardinalityJoined.value = 1e-9;
+        if (cardinalityJoined.value === 0) {
+          cardinalityJoined.value = Number.MIN_VALUE;
+        }
       }
     }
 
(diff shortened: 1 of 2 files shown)


packages/bus-rdf-join/lib/ActorRdfJoin.ts
```diff
@@ -251,7 +251,11 @@ TS
             value: acc.value * (optional ? Math.max(1, cardinalityThis.value) : cardinalityThis.value),
           };
         }, { type: 'exact', value: 1 });
-      cardinalityJoined.value *= (await this.mediatorJoinSelectivity.mediate({ entries, context })).selectivity;
+      if (cardinalityJoined.value > 1e-9) {
```
rubensworks, line 254: Could you add a comment here describing why this is needed?
The reason for this may get lost in the future.

rubensworks (changes requested):
Aha, good catch!

Do you know why exactly these low cardinalities get produced?
Because that (also) seems like something we need to fix, or at least know about.

maartyman:
So I was running the interactive-short-1-nocity query from SolidBench on the profile/card of 77 people (the friends of a random person in the benchmark). In the standard configuration, the bind join is chosen with multi-smallest and a bunch of nested loop joins. This means that after the materialization in the bind join, the cardinality of all other triple patterns in the materialized query is 1 for the 77 sources (because it's a star-shaped query). I think this causes the selectivity to be incredibly low, decreasing the cardinality after each nested loop join by a substantial amount. After a few joins, I think it becomes so low that the exponent in the float overflows and the js runtime sets it to a 0 integer (during debugging, I think I saw a cardinality of around 10^(-190), but I'm not sure anymore). During debugging, the issue started occurring with 20 sources. When disabling the bind join and only using the hash joins, the issue still occurred, but with more than around 50 sources.

maartyman:
Before merging, I wanted to ask if you prefer another solution: check if the cardinality is 0 after applying the selectivity, if so set the cardinality to a very low number.

And also `1e-9` isn't small enough, I don't really know how small we need to go here? But I guess with small cardinalities, the join plan doesn't really mater...

maartyman:
I forgot to mention this before as well, but this issue was introduced in V3. Or at least before V3, the engine was less prone to this issue...

maartyman:
I think this is a better solution, and a better test as well. Also, if one of the entries has a cardinality of 0, the resulting cardinality should be 0. Otherwise, the join-multi-empty will never be used.

rubensworks:
This approach looks cleaner indeed!

However, I just realized that `constructResultMetadata` is also used for non-inner join types, such as optional and union. (e.g. see the `optional` argument to this method)
So some changes may be needed to cope with this.

rubensworks:
@maartyman This is just a ping in case you missed my last comment.

rubensworks:
Great work, thanks @maartyman!

Issue #1433, opened 2024-09-29
BIND-related regression issue


bergos:
#### Issue type:

- :bug: Bug

____
#### Description:

It looks like there is a regression issue. BIND-related tests in shacl-engine that passed on August 5th are now failing without any code changes to the shacl-engine codebase. Here are the queries and the links to the Turtle files used as sources. Let me know if a smaller, stripped-down example is required.

```sparql
SELECT $this WHERE {
  BIND ($this AS ?that) .
  FILTER (?that = ex:InvalidResource) .
}
```
[Turtle](https://github.com/rdf-ext/shacl-engine/blob/8142bda7e6e84c512911690fe9c1860251dfda02/test/assets/data-shapes/sparql/pre-binding/pre-binding-004.ttl#L26-L30)

```sparql
SELECT $this WHERE {
  BIND (true AS $this) .
}
```
[Turtle](https://github.com/rdf-ext/shacl-engine/blob/8142bda7e6e84c512911690fe9c1860251dfda02/test/assets/data-shapes/sparql/pre-binding/unsupported-sparql-005.ttl#L29-L32)

____

rubensworks:
@simonvbrae, could you have a look at this one?

simonvbrae:
Hi @bergos :wave:
Could you let me know how I can run an example without using `yarn run test`? I found it hard to understand `test.js` and `util.js` so I'd like to avoid debugging through them. Thanks!

bergos:
@simonvbrae The test code is a bit complicated. I've added simplified examples to my test repository:

https://github.com/bergos/comunica-tests/blob/master/pre-binding-004.js
https://github.com/bergos/comunica-tests/blob/master/unsupported-sparql-005.js

simonvbrae:
Thank you for adding the test examples.
I've ran them with versions of Comunica from before the 5th of August but none result in a triple being returned from the test files.
Are you certain that the test files are correct?

bergos:
[...]

Issue #1298, opened 2023-12-17
Problems related to initialBindings


bergos:
#### Issue type:

- :bug: Bug

____
#### Description:

I plan to use Comunica for SPARQL support in [shacl-engine](https://github.com/rdf-ext/shacl-engine). While running tests from the official test suite, I came across a few issues related to `initialBindings`. I created a repository with examples to reproduce the problem. There is a comment on top of each file with more details.

[Repository with examples](https://github.com/bergos/comunica-tests)

rubensworks:
Thanks for reporting these findings @bergos!

@joachimvh I suspect this is similar to recent issues you fixed in the algebra (#1223 and #1294). Could you check if this is something that needs changes in the algebra? If not, I can have a closer look to see if changes are needed on the Comunica-side.

joachimvh:
> @joachimvh I suspect this is similar to recent issues you fixed in the algebra (#1223 and #1294). Could you check if this is something that needs changes in the algebra?

Note that for both those issues I didn't take the initial bindings into account. There already was an issue with the original query without adding bindings when converting to algebra and back to SPARQL. So I had a look here to see if there also was a similar issue. 

In this case the generated SPARQL query was the same as the original query (again without applying bindings), except for the one in `pre-binding-002`. There the output query is 

```sparql
SELECT ?this WHERE {
  FILTER("false"^^<http://www.w3.org/2001/XMLSchema#boolean> 
         && (?this = <http://datashapes.org/sh/tests/sparql/pre-binding/pre-binding-002.test#InvalidResource>))
}
```

So probably for that one there is a bug in the algebra where the filter scope gets pulled out of their group graph patterns and applied to a higher level. Tbh, reading the spec on this again, this is probably a misinterpretation I made when originally implementing this. I opened https://github.com/joachimvh/SPARQLAlgebra.js/issues/105 for this.

But so for all the other ones I wouldn't know what the cause is. I didn't look into the effect of applying the bindings as I'm not sure how this works in Comunica. Is the query itself changed or is this applied afterwards?

rubensworks:
> So probably for that one there is a bug in the algebra where the filter scope gets pulled out of their group graph patterns and applied to a higher level. Tbh, reading the spec on this again, this is probably a misinterpretation I made when originally implementing this. I opened https://github.com/joachimvh/SPARQLAlgebra.js/issues/105 for this.

Thanks! :-)

[...]

PR #1467, opened 2024-11-29
Use VALUES to implement Initialbindings within a BIND operation

simonvbrae:
In response to #1433 this PR uses a `VALUES` clause to represent variables in `BIND` operations that also appear in initialbindings. It replaces the input of the `BIND` operation with a join of the input and a values clause containing the values from initialbindings. The principle is the same as in the original PR #1385.

This PR currently only takes into account the case where the BIND expression is a variable.

I've added full engine tests from the issue #1433 and from the original issue #1298.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new test suite for validating initial bindings in SPARQL queries.
	- Enhanced handling of various operations (extend, group, project, values) in the materialization process.

- **Bug Fixes**
	- Improved logic for handling variable bindings and transformations, ensuring correct behavior in strict target variable scenarios.

- **Tests**
	- Expanded test coverage for initial bindings and various operations to ensure robustness and accuracy.
	- Updated tests to reflect new logic for extend operations and variable matching.

### DIFF
--- a/engines/query-sparql/test/QuerySparql-test.ts
+++ b/engines/query-sparql/test/QuerySparql-test.ts
@@ -894,6 +894,107 @@ SELECT ?obsId {
         expect(bindings2).toMatchObject(expectedResult);
       });
     });
+
+    describe('initialbindings', () => {
+      it('should handle bindings used in BIND (=extend) operator correctly', async() => {
+        const bindingsFactory = new BindingsFactory();
+        const initialBindings = bindingsFactory.bindings([
+          [DF.variable('this'), DF.namedNode('http://datashapes.org/sh/tests/sparql/pre-binding/pre-binding-004.test#InvalidResource')]
+        ]);
+        
+        
+        const context: QueryStringContext = {
+          sources: [
+            {
+              type: 'serialized',
+              value: `
+              @prefix dash: <http://datashapes.org/dash#> .
+              @prefix ex: <http://datashapes.org/sh/tests/sparql/pre-binding/pre-binding-004.test#> .
+              @prefix mf: <http://www.w3.org/2001/sw/DataAccess/tests/test-manifest#> .
+              @prefix owl: <http://www.w3.org/2002/07/owl#> .
+              @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
+              @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+              @prefix sh: <http://www.w3.org/ns/shacl#> .
+              @prefix sht: <http://www.w3.org/ns/shacl-test#> .
+              @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+              
+              ex:
+              \tsh:declare [
+              \t\tsh:prefix "ex" ;
+              \t\tsh:namespace "http://datashapes.org/sh/tests/sparql/pre-binding/pre-binding-004.test#"^^xsd:anyURI ;
+              \t] .
+              
+              ex:TestShape
+                rdf:type sh:NodeShape ;
+                rdfs:label "Test shape" ;
+                sh:sparql ex:TestShape-sparql ;
+                sh:targetNode ex:InvalidResource ;
+              .
+              ex:TestShape-sparql
+                sh:prefixes ex: ;
+                sh:select """
+                \tSELECT $this
+              \tWHERE {
+              \t\tBIND ($this AS ?that) .
+              \t\tFILTER (?that = ex:InvalidResource) .
+              \t}""" ;
+              .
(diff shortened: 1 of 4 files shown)


engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -894,6 +894,175 @@ SELECT ?obsId {
...
+        SELECT $a WHERE {
+          {
+            FILTER (bound($a))
+          }
+          $a ex:property "testProperty" .
+          FILTER (bound($a)) .
+        }
+          `, context))).map(binding => [ ...binding ].sort(([ var1, _c1 ], [ var2, _c2 ]) => var1.value.localeCompare(var2.value)));
+
+        expect(bindings).toMatchObject(expectedResult);
```
rubensworks, line 944: Instead of sorting, you can use `toEqualBindingsStream` like in the other tests.

packages/utils-query-operation/lib/MaterializeBindings.ts
```diff
@@ -106,9 +106,29 @@ export function materializeOperation(
...
+        bindings,
+        algebraFactory,
+        bindingsFactory,
+        options,
+      );
+
+      // Join op.input with a values clause containing the variables that are in InitialBindings and
+      // are used in the current extend operation.
+      const inScopeVariables = [ op.variable ];
+      if (op.expression.expressionType === Algebra.expressionTypes.TERM && op.expression.term.termType === 'Variable') {
```
rubensworks, line 120: This may not work correctly if you have more complicated BIND clauses, such as `BIND (($this + $this) AS $this)`.
Something like `Util.inScopeVariables` (or a variant) from sparqlalgebrajs may help.

packages/utils-query-operation/lib/MaterializeBindings.ts
```diff
@@ -106,9 +106,29 @@ export function materializeOperation(
...
+        inScopeVariables.push(op.expression.term);
+      }
+      const values: Algebra.Operation[] =
+        createValuesFromBindings(factory, <Bindings> options.originalBindings, inScopeVariables);
```
rubensworks, line 124: The cast to `Bindings` may break if `options.originalBindings` is undefined.

packages/utils-query-operation/lib/MaterializeBindings.ts
```diff
@@ -106,9 +106,29 @@ export function materializeOperation(
...
+      if (values.length > 0) {
+        recursionResult = factory.createJoin([ ...values, recursionResult ]);
+      }
       return {
-        recurse: true,
-        result: op,
+        recurse: false,
+        result: factory.createExtend(recursionResult, op.variable, <Algebra.Expression>
+          materializeOperation(op.expression, bindings, algebraFactory, bindingsFactory, options)),
       };
```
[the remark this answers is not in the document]
rubensworks: This seems like an interesting thing to look into indeed @simonvbrae, thanks bot!

packages/utils-query-operation/lib/MaterializeBindings.ts
rubensworks: Would be good to also have a test case for this.

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -894,6 +894,175 @@ SELECT ?obsId {
         expect(bindings2).toMatchObject(expectedResult);
       });
     });
+
+    describe('initialbindings', () => {
```
rubensworks, line 898:
```suggestion
    describe('initialBindings', () => {
```

rubensworks (changes requested):
Thanks @simonvbrae!

packages/utils-query-operation/lib/MaterializeBindings.ts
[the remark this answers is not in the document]
simonvbrae: I've implemented this.

packages/utils-query-operation/lib/MaterializeBindings.ts
simonvbrae: I will look into it!

packages/utils-query-operation/lib/MaterializeBindings.ts
simonvbrae: I thought it would never be undefined because options.originalBindings is assigned like this
`originalBindings: 'originalBindings' in options ? options.originalBindings : bindings`.
Should I just assert it by adding a null check or what is the best way?

rubensworks: Ah yes, you're right, it should be fine!

PR #1483, opened 2025-01-15
Replace sparqljs by traqula

jitsedesmet:


packages/actor-query-parse-sparql/lib/ActorQueryParseSparql.ts
```diff
@@ -30,14 +32,13 @@ export class ActorQueryParseSparql extends ActorQueryParse {
...
-      factory: dataFactory,
+      dataFactory: <DataFactory<RDF.BaseQuad>> dataFactory,
     });
     const parsedSyntax = parser.parse(action.query);
-    const baseIRI = parsedSyntax.type === 'query' ? parsedSyntax.base : undefined;
+    const baseIRI = ('type' in parsedSyntax && parsedSyntax.type === 'query') ? parsedSyntax.base : undefined;
     return {
       baseIRI,
-      operation: translate(parsedSyntax, {
+      operation: translate(<any> parsedSyntax, {
```
rubensworks, line 41: Can we add a `// TODO: ...` here explaining that this is temporary, until sparqlalgebrajs has been ported to the traqula monorepo?

packages/actor-query-parse-sparql/lib/ActorQueryParseSparql.ts
```diff
@@ -4,8 +4,10 @@ import { KeysInitQuery } from '@comunica/context-entries';
 import type { IActorTest, TestResult } from '@comunica/core';
 import { failTest, passTestVoid } from '@comunica/core';
 import type { ComunicaDataFactory } from '@comunica/types';
+import type * as RDF from '@rdfjs/types';
+import { Sparql12Parser as SparqlParser } from '@traqula/engine-sparql-1-2';
```
rubensworks, line 8: For easier pluggability, I wonder if it wouldn't be easier if all `@traqula/engine-` packages just expose a `Parser` class (so it's always the same name).
This is also the approach we follow with comunica engines.

rubensworks: (this comment is not urgent btw)

packages/actor-query-parse-sparql/lib/ActorQueryParseSparql.ts
```diff
@@ -30,14 +32,13 @@ export class ActorQueryParseSparql extends ActorQueryParse {
     const parser = new SparqlParser({
       prefixes: this.prefixes,
       baseIRI: action.baseIRI,
-      sparqlStar: true,
-      factory: dataFactory,
+      dataFactory: <DataFactory<RDF.BaseQuad>> dataFactory,
     });
     const parsedSyntax = parser.parse(action.query);
-    const baseIRI = parsedSyntax.type === 'query' ? parsedSyntax.base : undefined;
+    const baseIRI = ('type' in parsedSyntax && parsedSyntax.type === 'query') ? parsedSyntax.base : undefined;
```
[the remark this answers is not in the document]
jitsedesmet: The check is needed because the result of `parsedSyntax` does not always have a key `type`. (The types of sparqlJS are wrong). For TS to be happy, we need to test whether it is present.
(It is not present in case the query is just a listing of prefixes)

rubensworks:
Really nice work @jitsedesmet!

PR #1492, opened 2025-01-20
Fix Query Termination for Triple Patterns with Zero Estimated Cardinality

RubenEschauzier:
This fix ensures that the query does not terminate when a triple pattern is estimated to have zero cardinality. This issue occurs when the estimation is incorrect or during link traversal, where triple patterns often have an estimated cardinality of zero due to the lack of prior knowledge.

Queries will still terminate as expected when the cardinality estimation is accurate.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved join operation logic to handle edge cases with zero cardinality entries more precisely
	- Enhanced test coverage for join operations with estimated zero cardinality

- **Tests**
	- Added new test case to validate join operation behavior with zero-cardinality entries


rubensworks:
Thanks!

Issue #915, opened 2022-01-25
Cleanup actor-init-query unit tests


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

The unit tests in `actor-init-query` are a bit of a mess, so we should clean them up.
The main problem with the current tests is that the mocking is way too complex, and for many tests we just check if things don't crash, while we should check for the expected output instead.

For all tests related to `ActorInitQuery`, `QueryEngineBase`, and `HttpServiceSparqlEndpoint`, we should start from scratch, and create new tests where we test each feature separately.

This depends on refactoring of `HttpServiceSparqlEndpoint`, for which a separate issue exists #764.

Issue #764, opened 2020-12-01
Refactor HttpServiceSparqlEndpoint


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

Currently, `HttpServiceSparqlEndpoint`(which exposes Comunica as a SPARQL endpoint), is a bit of an entangled mess.
One the one hand, this makes it very hard to maintain, but it also makes testing very difficult.

We should decouple main parts such as CLI invocation, HTTP serving, query handling, service description creation, ...

PR #1496, opened 2025-01-27
Update SPARQL endpoint code to support default-graph-uri

surilindur:
This is a collection of the work to support `default-graph-uri` with `HttpServiceSparqlEndpoint`, along with various adjustments to the way requests are handled to make it easier to implement the different cases. The SPARQL endpoint actor parses the `default-graph-uri` into the context, and `ActorQueryParseSparql` handles the wrapping of the query into a corresponding `Algebra.FROM` (or overrides the graph URIs in an existing one).

The motivation for this work is the need to be able to use Comunica as a SPARQL endpoint over multiple named graphs.

Any feedback would be welcome. I need to see if the CI passes, but the code itself works for the purpose it is meant for.

## Summary by CodeRabbit

- **New Features**
  - Enhanced SPARQL endpoint functionality with new graph URI handling.
  - Added support for more flexible context configuration in query processing.

- **Dependency Changes**
  - Replaced `rdf-quad` with `rdf-data-factory` in the actor-init-query package.
  - Added `@rdfjs/types` dependency in the actor-query-parse-sparql package.

- **Improvements**
  - Refined HTTP service request handling.
  - Improved error management in SPARQL query parsing.
  - Extended context key management for graph URIs.

- **Testing**
  - Expanded test coverage for SPARQL endpoint and query parsing functionality.
  - Established comprehensive testing framework for CLI argument handling.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -41,6 +41,7 @@ module.exports = config([
       '**/__mocks__/*.js',
       'packages/actor-dereference-file/**/*.ts',
       'packages/actor-http-native/**/*.ts',
+      'packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts',
       'packages/logger-bunyan/**/*.ts',
       'packages/packager/**/*.ts',
     ],
--- a/packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
+++ b/packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
@@ -1,73 +1,344 @@
-/* eslint-disable import/no-nodejs-modules,ts/no-require-imports,ts/no-var-requires */
-import type { Cluster } from 'node:cluster';
-import type { EventEmitter } from 'node:events';
-import * as http from 'node:http';
+import type { Worker, Cluster } from 'node:cluster';
+import { createServer } from 'node:http';
 import type { IncomingMessage, ServerResponse } from 'node:http';
-import * as querystring from 'node:querystring';
 import type { Writable } from 'node:stream';
-import * as url from 'node:url';
+import { inspect } from 'node:util';
 import { KeysQueryOperation } from '@comunica/context-entries';
-import { ActionContext } from '@comunica/core';
-import type { ICliArgsHandler, QueryQuads, QueryType } from '@comunica/types';
-import type * as RDF from '@rdfjs/types';
+import type { ICliArgsHandler, QueryType, QueryStringContext, IQueryQuadsEnhanced } from '@comunica/types';
 import { ArrayIterator } from 'asynciterator';
-
+import { DataFactory } from 'rdf-data-factory';
 import yargs from 'yargs';
-
-import type { IDynamicQueryEngineOptions } from '..';
-
-import { QueryEngineBase, QueryEngineFactoryBase } from '..';
-
+import type { ActorInitQuery } from './ActorInitQuery';
 import { CliArgsHandlerBase } from './cli/CliArgsHandlerBase';
 import { CliArgsHandlerHttp } from './cli/CliArgsHandlerHttp';
+import { QueryEngineBase } from './QueryEngineBase';
+import { QueryEngineFactoryBase } from './QueryEngineFactoryBase';
+import type { IDynamicQueryEngineOptions } from './QueryEngineFactoryBase';
+
+// The cluster module seemingly breaks when imported in CommonJS
+const cluster: Cluster = require('node:cluster');
+
+// The HTTP content negotiation library is missing all types, hence this workaround
+const negotiate: {
+  choose: (variants: { type: string; q: number }[], request: IncomingMessage) => { type: string; qts: number }[];
+} = require('negotiate');
 
 // Use require instead of import for default exports, to be compatible with variants of esModuleInterop in tsconfig.
-const clusterUntyped = require('node:cluster');
 const process: NodeJS.Process = require('process/');
-const quad = require('rdf-quad');
 
-// Force type on Cluster, because there are issues with the Node.js typings since v18
-const cluster: Cluster = clusterUntyped;
+const DF = new DataFactory();
 
 /**
  * An HTTP service that exposes a Comunica engine as a SPARQL endpoint.
  */
 export class HttpServiceSparqlEndpoint {
-  public static readonly MIME_PLAIN = 'text/plain';
-  public static readonly MIME_JSON = 'application/json';
-
-  public readonly engine: Promise<QueryEngineBase>;
-
-  public readonly context: any;
-  public readonly timeout: number;
-  public readonly port: number;
-  public readonly workers: number;
-
-  public readonly freshWorkerPerQuery: boolean;
-  public readonly contextOverride: boolean;
-
-  public lastQueryId = 0;
+  protected readonly port: number;
+  protected readonly timeout: number;
+  protected readonly workers: number;
+  protected readonly context: QueryStringContext;
+  protected readonly invalidateCacheBeforeQuery: boolean;
+  protected readonly freshWorkerPerQuery: boolean;
+  protected readonly allowContextOverride: boolean;
+  protected readonly endpointPath: string = '/sparql';
+  protected readonly engineFactory: QueryEngineFactoryBase<QueryEngineBase>;
+  protected readonly engineWrapper: (actorInitQuery: ActorInitQuery) => QueryEngineBase;
 
   public constructor(args: IHttpServiceSparqlEndpointArgs) {
     this.context = args.context || {};
     this.timeout = args.timeout ?? 60_000;
     this.port = args.port ?? 3_000;
     this.workers = args.workers ?? 1;
+    this.invalidateCacheBeforeQuery = Boolean(args.invalidateCacheBeforeQuery);
     this.freshWorkerPerQuery = Boolean(args.freshWorkerPerQuery);
-    this.contextOverride = Boolean(args.contextOverride);
-
-    this.engine = new QueryEngineFactoryBase(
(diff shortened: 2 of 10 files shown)


surilindur:
There is still something weird going on with the fresh worker flag, I need to investigate. It ends up killing the primary thread.

eslint.config.js
```diff
@@ -41,6 +41,7 @@ module.exports = config([
       '**/__mocks__/*.js',
       'packages/actor-dereference-file/**/*.ts',
       'packages/actor-http-native/**/*.ts',
+      'packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts',
```
rubensworks, line 44: This entry can be removed?

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -1,73 +1,344 @@
...
-import { QueryEngineBase, QueryEngineFactoryBase } from '..';
-
+import type { ActorInitQuery } from './ActorInitQuery';
 import { CliArgsHandlerBase } from './cli/CliArgsHandlerBase';
 import { CliArgsHandlerHttp } from './cli/CliArgsHandlerHttp';
+import { QueryEngineBase } from './QueryEngineBase';
+import { QueryEngineFactoryBase } from './QueryEngineFactoryBase';
+import type { IDynamicQueryEngineOptions } from './QueryEngineFactoryBase';
+
+// The cluster module seemingly breaks when imported in CommonJS
```
rubensworks, line 18: It did not break before though.
So this is probably not necessary.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -1,73 +1,344 @@
...
+    stdout: Writable,
+    stderr: Writable,
+    request: IncomingMessage,
+    response: ServerResponse,
+    engine: QueryEngineBase,
+    mediaTypeFormats: Record<string, string>,
+    mediaTypeWeights: Record<string, number>,
+  ): Promise<void> {
+    // Attempt to reconstruct the original full request URL with protocol and host
+    const requestProtocol = <string> request.headers['x-forwarded-proto'] ?? 'http';
```
rubensworks, line 81: This `x-forwarded-proto` stuff seems new and unrelated to this PR.
Is it really needed?
If so, could this be a separate PR?

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -1,73 +1,344 @@
...
+      // Requests should only be accepted at the specificed endpoint path
+      if (requestUrl.pathname !== this.endpointPath) {
+        throw new HTTPError(404, 'Not Found');
+      }
+
+      // Attempt to parse the request, and throw an error in case of failure
+      const operation = await this.parseOperation(requestUrl, request);
+
+      // Cache only needs to be invalidated when the worker is not fresh
+      if (this.invalidateCacheBeforeQuery && !this.freshWorkerPerQuery) {
```
rubensworks, line 98: This is not necessary. It's already done by QueryEngine.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -1,73 +1,344 @@
...
+        await engine.invalidateHttpCache();
+      }
+
+      // Execute the query, or generate the service description
+      const result: QueryType = operation.type === 'sd' ?
+        this.getServiceDescription(requestUrl, mediaTypeFormats) :
+        await engine.query(operation.queryString, operation.context);
+
+      // TODO: Fix media type negotiation to avoid serialization errors here
```
rubensworks, line 107: TODO?

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -1,73 +1,344 @@
...
+    mediaTypeWeights: Record<string, number>,
+  ): string {
+    // Convert the media type weights into format expected by the library
+    const variants = Object.entries(mediaTypeWeights).map(([ type, q ]) => ({ type, q })).sort(t => t.q);
+
+    // Attempt initial negotiation using the full pool of media types
+    const negotiatedVariant: { type: string; qts: number } | undefined = negotiate.choose(variants, request)
+      .sort((first: any, second: any) => second.qts - first.qts).at(0);
+
+    // TODO: Split the media types into pools based on response.resultType and do content negotiation over the
```
rubensworks, line 164: Unnecessary TODO?

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -1,73 +1,344 @@
...
+          };
+        }
+        if (requestBody.contentType.includes('application/x-www-form-urlencoded')) {
+          const requestBodyParams = new URLSearchParams(requestBody.content);
+          let requestBodyContext: QueryStringContext | undefined;
+          if (requestBodyParams.has('context')) {
+            try {
+              requestBodyContext = JSON.parse(requestBodyParams.get('context')!);
+            } catch {
+              break;
```
rubensworks, line 236: JSON parse error message is getting lost.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -1,73 +1,344 @@
...
+            return {
+              type: 'update',
+              queryString: requestBodyParams.get('update')!,
+              context: this.parseOperationParams(url.searchParams, requestBodyContext),
+            };
+          }
+        }
+        break;
+      default:
+        throw new HTTPError(405, 'Method Not Allowed');
```
rubensworks, line 256: Error message is very imprecise. The old error messages were much clearer, e.g. `Invalid POST body with context received...`.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -1,73 +1,344 @@
...
+
+  /**
+   * Reads the incoming HTTP request body into a string, using the Content-Encoding header.
+   * @param {IncomingMessage} request The incoming client request.
+   * @returns {IParsedRequestBody} The request body.
+   */
+  public async readRequestBody(request: IncomingMessage): Promise<IParsedRequestBody> {
+    return new Promise((resolve, reject) => {
+      if (!request.headers['content-type']) {
+        throw new HTTPError(400, 'Bad Request');
```
rubensworks, line 272: Also very imprecise error message here.

rubensworks (changes requested):
I appreciate the work you did, but it looks like you did _much_ more than just adding `default-graph-uri` support.
You basically rewrote the whole handling of SPARQL endpoint requests (which is needed at some point in any case https://github.com/comunica/comunica/issues/764 and will require part of https://github.com/comunica/comunica/issues/915).
As such, it's difficult for me to review this due to the invasiveness of the changes.

From the diffs I could pinpoint, it however looks like you very liberally modified things. 
As such, I don't have a high level of trust that everything still works in the same way as before, and nothing broke. This includes accumulating errors into generic error messages, whereas these were very low-level and precise before.

Furthermore, since this class did not have good unit tests before, it is impossible to know if these changes actually broke something.

In conclusion, I see two options of going forward:

1. We go forward with the refactoring of `HttpServiceSparqlEndpoint`. But we must ensure that all behaviour is identical as before. This also includes error/info logs. Besides detailed unit tests, this will also require extensive manual checks, such as re-running the SolidBench experiments and our performance benchmarks and ensuring the results (and log files) remain the same.
2. We revert these changes, and instead only implement `default-graph-uri` support.

The choice is up to you, but option 1 will require _a lot_ more work.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
surilindur: It could probably be a separate PR, yes. The goal was to allow using Comunica behind a proxy (such as nginx, with TLS handled by the proxy) where the endpoint host or protocol does not match the URI that Comunica is listening at, in which case the SPARQL SD triples would need to be based on the URI that the client application is talking to, rather than what Comunica is listening at.

rubensworks: Very useful indeed!

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
surilindur: Not necessarily unnecessary, since the content negotiation step currently does not take into consideration what type of output can be serialized in which format, so to be able to properly handle content negotiation, the available media types would need to be split into pools based on the outgoing response type.

Currently, a client can request `bindings` serialized as `application/n-quads`, and since Comunica has all media types in the same pool, it will resolve to use that media type and then fail at the serialization step. This failure is then shown to the client as HTTP 500, even though it should have been caught at content negotiation step and reported as 406 or something.

I do not know how to fix that because the types come from the engine that collects them from the mediator without doing any splitting there, so that is why I left the to-do note in there. :sweat:

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
surilindur: True, it could be reported to the client as invalid request!

surilindur:
The branch was not actually deleted, just renamed. When I find the time to address all the feedback, I will make new PRs for the endpoint.

PR #1497, opened 2025-01-28
Fix sd:defaultGraph parsing with QPF endpoints

surilindur:
This should fix the parsing of `sd:defaultGraph` on QPF endpoints, without causing the parser to register unrelated triples with that predicate. The fix relies on a whitelist of subjects that can be used to extract the predicate values from, which is not perfect, but is better than ignoring the subject entirely.

The parsing of the default graph with QPF broke due to the changes in 2c0af9eb76b7763afe39d07900f3320a061d94a4.

There is an added unit test for the QPF defaultGraph case, as well as the existing test for ignoring unrelated `sd:defaultGraph` entries. I removed the test for returning empty results without an `sd:endpoint` since that behaviour is not intended with QPF.

Manually testing against the linkeddatafragments.org TPF and QPF servers worked, with a simple spo query.

Any feedback would be welcome! :slightly_smiling_face: 

## Summary by CodeRabbit

- **Bug Fixes**
	- Refined metadata extraction logic for SPARQL services
	- Improved subject URI filtering mechanism

- **Tests**
	- Updated test cases to use localhost URLs
	- Added new test for parsing default graph metadata
	- Improved test coverage for endpoint and dataset parsing

### DIFF
--- a/packages/actor-rdf-metadata-extract-sparql-service/test/ActorRdfMetadataExtractSparqlService-test.ts
+++ b/packages/actor-rdf-metadata-extract-sparql-service/test/ActorRdfMetadataExtractSparqlService-test.ts
@@ -25,15 +25,18 @@ describe('ActorRdfMetadataExtractSparqlService', () => {
     });
 
     it('should test successfully', async() => {
-      await expect(actor.test({ url: 'http://example.org/', metadata: <any> undefined, requestTime, context })).resolves.toBeTruthy();
+      await expect(actor.test({ url: 'http://localhost/', metadata: <any> undefined, requestTime, context })).resolves.toBeTruthy();
     });
   });
 
   describe('run', () => {
-    const sparqlIri = DF.namedNode('http://example.org/sparql');
-    const sparqlService = DF.namedNode('http://example.org/endpoint');
-    const sparqlDefaultDataset = DF.namedNode('http://example.org/dataset');
-    const sparqlDefaultGraph = DF.namedNode('http://example.org/graph');
+    const voidSubset = DF.namedNode('http://rdfs.org/ns/void#subset');
+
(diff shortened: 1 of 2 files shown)


packages/actor-rdf-metadata-extract-sparql-service/test/ActorRdfMetadataExtractSparqlService-test.ts
```diff
@@ -53,70 +56,78 @@ describe('ActorRdfMetadataExtractSparqlService', () => {
...
-          defaultDataset: sparqlDefaultDataset.value,
-          defaultGraph: sparqlDefaultGraph.value,
+          sparqlService: endpointService.value,
+          defaultDataset: endpointDefaultDataset.value,
+          defaultGraph: endpointDefaultGraph.value,
         },
       });
     });
-
-    it('should return empty result without endpoint defined', async() => {
```
rubensworks, line 115: I guess we want to keep this one, but perhaps needs modifications?

surilindur: I am not sure what type of modifications would work. The original idea behind the test/logic was that if there is no SPARQL endpoint, then the other metadata from the Service Description does not make any sense, because it does not apply to anything. Now, it turned out parts of the metadata can apply to things outside a SPARQL endpoint (in the case of QPF). In that sense, I thought it makes sense to remove this test and modify the code so that it returns any metadata it can find, as some of it might be useful somewhere.

rubensworks: Ah, I misunderstood the test then. All good!

rubensworks:
Thanks for the quick for @surilindur!

PR #1502, opened 2025-02-19
Add workaround for DBPedia SPARQL HTTP status 405 with Retry-After

surilindur:
This is a workaround for DBPedia SPARQL endpoint sending a 405 instead of 429 for rate limiting, which should be considered 429 by the client for retry purposes. Assuming the same software could be used in other SPARQL endpoints, as well, there is no URI-specific handling, but instead just a check for 405 with `Retry-After`, which should be sufficient.

This also fixes a small bug with the timeout calculation for clearing up old delays.

### DIFF
--- a/packages/actor-http-retry/lib/ActorHttpRetry.ts
+++ b/packages/actor-http-retry/lib/ActorHttpRetry.ts
@@ -103,8 +103,12 @@ export class ActorHttpRetry extends ActorHttp {
         continue;
       }
 
-      if (response.status === 429 || response.status === 503) {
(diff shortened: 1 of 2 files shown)


surilindur:
This has a small bug somewhere, still. I need to find it...

surilindur:
Ah no, it was not a bug, I just forgot to enable retries during my testing. :joy:

This is now ready for review when you find the time @rubensworks

rubensworks:
Great, thanks @surilindur!

PR #1506, opened 2025-02-26
Adjust VoID extraction to lower the performance impact

surilindur:
Here is a set of changes to adjust the extraction of VoID descriptions to lower the overhead associated with it. Two redundant properties have also been removed from the dataset interface, because those can be used internally in the `getCardinality` function and therefore do not need to be exposed.

The overhead with TPF seems to be negligible (~7ms per page), but I will run the performance benchmarks to check, as with the rate limiter.


rubensworks:
> Does it make sense to keep the Wikidata and Rhea VoID descriptions in the tests? I thought it would be useful to test actual VoID/SPARQL SD, but they are kind of big.

The Rhea one seems to be reasonable in size.
The Wikidata is quite big though (17.5 MB), and will have a significant impact on cloning time, so I'd remove that one (or take a subset/alternative).

Everything else looks good to me!

surilindur:
I removed both of the actual descriptions, since they were indeed a bit big, and after removing the Wikidata one I had to expand the unit tests to account for some logic anyway.

This should now be ready for review! I kept the changes after the last review to a separate commit, resisting the urge to squash and force push. :sweat_smile:

PR #1507, opened 2025-03-04
Method to determine if the aggregatedStore has ended added

constraintAutomaton:


rubensworks:
As discussed offline, listening to `'data'` will force the stream in flow mode, which is not good for performance. So we'll need a different approach.

constraintAutomaton:
Yes, but I am working on this PR to make it work.

PR #1509, opened 2025-03-06
Avoid printing all metadata in join logging

surilindur:
This should fix the issue with extremely verbose debug logging after VoID descriptions are discovered.


rubensworks:
Thanks!

PR #1512, opened 2025-03-13
Bump lru-cache version

surilindur:
This bumps the `lru-cache` version to 11, and also adds the `--ignore-engines` flags to the CI install steps, because the new version does not officially claim to spport Node 18.

The Node [release schedule](https://nodejs.org/en/about/previous-releases#release-schedule) makes it look like Node 18 will also reach EOL at the end of April, so maybe dropping it would make sense soon.

Any feedback would be welcome, though.


surilindur:
Ah okay. I will just close this PR, and it can be done as part of that major version bump.

PR #1514, opened 2025-03-16
Fixes to dereference error handling

surilindur:
This is a small PR to address three issues I discovered while testing out the master branch with link traversal:
1. The URL of a failed fetch request does not get reported, but the actor does. This makes the warning messages useless, so I added the URL as extra data to the error message.
2. The media type cascade logic seemed off, so I changed it to prefer the variables, and only fall back to a function call as last resort. This seems to have changed nothing in my tests, but should avoid a function call if one of the variables does have a non-empty value.
3. The entire RDF parse operation was being attempted on sources that failed to be dereferenced, so I added that part of the code behind an if statement to avoid the redundant parse error messages when dereferencing fails. The error message about a failed fetch is more accurate, and the parse error was just a byproduct of a missing check.

Any feedback would be welcome. :slightly_smiling_face: None of these is a breaking change in my opinion.

### DIFF
--- a/packages/bus-dereference/lib/ActorDereferenceParse.ts
+++ b/packages/bus-dereference/lib/ActorDereferenceParse.ts
@@ -103,9 +103,8 @@ export abstract class ActorDereferenceParse<
           context,
           handle: { context, ...dereference, metadata: await this.getMetadata(dereference) },
           // eslint-disable-next-line ts/prefer-nullish-coalescing
-          handleMediaType: (dereference.mediaType ||
-            getMediaTypeFromExtension(dereference.url, this.mediaMappings)) ||
-            action.mediaType,
+          handleMediaType: dereference.mediaType || action.mediaType ||
+            getMediaTypeFromExtension(dereference.url, this.mediaMappings),
         })).handle;
         result.data = this.handleDereferenceStreamErrors(action, result.data);
       } catch (error: unknown) {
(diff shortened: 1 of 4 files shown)


packages/bus-dereference/lib/ActorDereferenceParse.ts
```diff
@@ -5,21 +5,22 @@ import { passTestVoid } from '@comunica/core';
...
  * Get the media type based on the extension of the given path,
  * which can be an URL or file path.
  * @param {string} path A path.
  * @param {Record<string, string>} mediaMappings A collection of mappings,
  * mapping file extensions to their corresponding media type.
- * @return {string} A media type or the empty string.
+ * @return {string} A media type or undefined.
  */
-export function getMediaTypeFromExtension(path: string, mediaMappings?: Record<string, string>): string {
+export function getMediaTypeFromExtension(path: string, mediaMappings?: Record<string, string>): string | undefined {
```
rubensworks, line 18: Is this change really needed? The current code seems to work fine.

rubensworks (changes requested):
Just one question, everything else looks good!

packages/bus-dereference/lib/ActorDereferenceParse.ts
surilindur: It is not needed, but it makes it easier to distinguish between an actual value and not-a-value. I think this function is only used internally here, since nothing else broke when I changed it. :sweat_smile:

surilindur: I think it is used in the unit tests only, or something else, same with a lot of `public` attributes on classes in Comunica. But I will revert it!

rubensworks:
Great, thanks!

Issue #1519, opened 2025-03-26
Federated SPARQL Endpoint Quering Error: "requires an operation with source annotation."


ecrum19:
#### Issue type:

- :bug: Bug 

____
#### Description:

When executing the federated query provided below, a bug related to source annotation is encountered (see Error). @surilindur is currently looking into the problem.

Sources: https://sparql.rhea-db.org/sparql https://sparql.uniprot.org/sparql

Query:
```
PREFIX ec: <http://purl.uniprot.org/enzyme/>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
PREFIX rh: <http://rdf.rhea-db.org/>
PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
# Query 11
# Retrieve the count of reactions mapped to each level (main class) of the enzyme classification
#
# This query mimics the Filter section of the Rhea website (Browse all reactions)
# https://www.rhea-db.org/rhea?query=
SELECT ?ecClass (STR(?ecName) AS ?ecClassName) (COUNT(?rhea) AS ?rheaCount)
WHERE {
  VALUES (?ecClass) { (ec:1.-.-.-)(ec:2.-.-.-)(ec:3.-.-.-)(ec:4.-.-.-)(ec:5.-.-.-) (ec:6.-.-.-) (ec:7.-.-.-)}
  ?ecNumber rdfs:subClassOf ?ecClass .
  ?ecClass skos:prefLabel ?ecName .
  ?rhea rdfs:subClassOf rh:Reaction .
  ?rhea rh:ec ?ecNumber .
} GROUP BY ?ecClass ?ecName
```

Error:
```
Query operation processing failed: none of the configured actors were able to handle the operation type pattern
    Error messages of failing actors:
        Actor urn:comunica:default:query-operation/actors#source requires an operation with source annotation.
```

____
#### Environment:

```
| software         | version
| ---------------- | -------
| Comunica Engine  | 4.1.0 
| node             | v20.11.1
| npm              | 10.2.4
| yarn             | 1.22.22
| Operating System | darwin (Darwin 24.3.0)
```

surilindur:
To add a note to self: it looks like an issue where a join over some triple patterns has been assigned a specific scoped source (according to the logical query plan), but instead of being sent to that source as a whole, that big join is being split into smaller ones that do not get the source assigned to them properly.

rubensworks:
@surilindur Maybe you can check the test implementations.

[...]

PR #1522, opened 2025-03-26
Avoid flattening joins with input-specific metadata

surilindur:
After a lot of debugging, it turned out the source assignment error in #1519 is most likely caused by the flattening of joins, when using the `createJoin` function from `sparqlalgebrajs`, as it seems to strip out the metadata from join entries when it performs the flattening step.

Disabling flattening seems to fix it, so I changed the parts of the code where I ran into this issue to not do flattening when any of the join entries have metadata associated with them.

After this change, the query in #1519, as well as another consistently reproducible query (`020_ns` from @ecrum19's example collection) no longer crash due to source assignment issues. Neither of them seem to complete within a reasonable amount of time, either, but at least the error is gone.

Any feedback would be welcome, though, because I am not sure if this is the correct way to fix this.

### DIFF
--- a/packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
+++ b/packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
@@ -147,7 +147,8 @@ export class ActorRdfJoinMultiBind extends ActorRdfJoin<IActorRdfJoinMultiBindTe
         // Send the materialized patterns to the mediator for recursive join evaluation.
         const operation = operations.length === 1 ?
           operations[0] :
-          algebraFactory.createJoin(operations);
+          // Flattening should only take place if none of the input operations have associated metadata
+          algebraFactory.createJoin(operations, operations.every(op => !op.metadata));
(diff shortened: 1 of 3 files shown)


packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
```diff
@@ -147,7 +147,8 @@ export class ActorRdfJoinMultiBind extends ActorRdfJoin<IActorRdfJoinMultiBindTe
         // Send the materialized patterns to the mediator for recursive join evaluation.
         const operation = operations.length === 1 ?
           operations[0] :
-          algebraFactory.createJoin(operations);
+          // Flattening should only take place if none of the input operations have associated metadata
+          algebraFactory.createJoin(operations, operations.every(op => !op.metadata));
```
rubensworks, line 151: @jitsedesmet Do take note of this when integrating sparqlalgebrajs.
Factory functions like these should probably default to *not* flatten, and only flatten if explicitly requested.
This is to avoid unexpected issues like this.

rubensworks:
Nice, good catch @surilindur!

PR #1528, opened 2025-04-01
Use VoID metadata for `ASK` queries in `QuerySourceSparql`

surilindur:
This is a change to use the VoID cardinality metadata to answer ASK queries in the SPARQL query source, when available.

When `useAskIfSupported` is enabled in `ActorOptimizeQueryOperationPruneEmptySourceOperations`, this will help avoid unnecessary ASK queries to the endpoints if VoID metadata is available, and will thus keep the number of HTTP requests the same as if `useAskIfSupported` were false.


surilindur:
I will close this since rebasing it would be quite a bit of work, and using estimates for ask seems like a bit of a stretch, as discussed offline.

PR #1534, opened 2025-04-09
aggregate store ending reporting mechanism

constraintAutomaton:
related to #1507

### DIFF
--- a/packages/actor-query-source-identify-hypermedia/test/QuerySourceHypermedia-test.ts
+++ b/packages/actor-query-source-identify-hypermedia/test/QuerySourceHypermedia-test.ts
@@ -741,56 +741,38 @@ describe('QuerySourceHypermedia', () => {
         ]);
         expect(mediatorsThis.mediatorQuerySourceIdentifyHypermedia.mediate).toHaveBeenCalledTimes(3);
 
-        const callsIt1Meta = it1Meta.mock.calls;
-        let lastCardinalityIt1 = 0;
-        for (const call of callsIt1Meta) {
-          const metadata = call[0];
-          expect(metadata).toStrictEqual({
-            a: 1,
-            state: expect.any(MetadataValidationState),
-            firstMeta: true,
-            cardinality: { type: 'exact', value: expect.any(Number) },
-          });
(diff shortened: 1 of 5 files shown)


packages/actor-query-source-identify-hypermedia/test/QuerySourceHypermedia-test.ts
```diff
@@ -741,101 +741,31 @@ describe('QuerySourceHypermedia', () => {
         ]);
         expect(mediatorsThis.mediatorQuerySourceIdentifyHypermedia.mediate).toHaveBeenCalledTimes(3);
 
-        expect(it1Meta).toHaveBeenCalledTimes(4);
```
rubensworks, line 744: Can you motivate these changes?
Is it because the number of calls can differ across runs?
If it not, I would suggest keeping the separate checks per call (and nr of times).

constraintAutomaton: During my testing, I **think** that noticed that for iterator 2, the number of calls varied. Also, it seemed like what really mattered was the final cardinality estimate, rather than how many times the metadata changed. I will  spend more time digging into whether those specific counts actually make sense. If I remember correctly, `it2` was called between 8 and 10 times, which I was not able to understand why.

constraintAutomaton: For the intermediary cardinality, I was also not really able to understand why at some very specif point it was evaluated (look iterator 2) ad how it was determined that it should be those specific values.

constraintAutomaton: I will follow up later for my first point. But my second point stand without further investigation (I mean it can be wrong, but I do not think I need to investigate more to be sure of my comment)

constraintAutomaton: As discussed in person the number of metadata update can vary with this implementation.

rubensworks:
Thanks!

PR #1535, opened 2025-04-10
Adjust HTTP request counting and related observer/listener creation

### LINKED ISSUE #1526: Query result metadata field `httpRequests` for `application/sparql-results+json` is inaccurate
#### Issue type:

- :bug: Bug 

____
#### Description:

The JSON serializer for SPARQL query results uses [an HTTP bus action observer](https://github.com/comunica/comunica/blob/0227f236be392992c77e872201623257e6080f60/packages/actor-query-result-serialize-sparql-json/lib/ActionObserverHttp.ts#L31) to record the number of HTTP requests performed. This observer records every individual run event on the bus. With only the fetch actor, this would function properly. However, with the rate limit actor, this count is multiplied by two (1 fetch + 1 rate limit wrap). With the retry actor enabled (so the `httpRetryCount` value assigned to greater than 0), it gets multiplied by three and has retries added to it, etc. which will become difficult to keep track of.

The solution here would probably be to adjust the approach to HTTP request counting to not use the HTTP bus run count, but rather something else that accurately depicts the actual number of requests sent out from the engine.

There is also something weird going on with the reporting of requests for planning in general, I assume a race condition, because the first *execution* request (with the bug described above) gets included in the planning phase count, which is wrong.

____
#### Environment:

The issue is environment-independent, tested on master branch at 0227f236be392992c77e872201623257e6080f60.

Older versions with only one HTTP bus actor running are not affected.

#### Crash log:

Not needed

surilindur:
This is a set of adjustments to how the HTTP request counts are tracked, and how the related listeners/observers are created. Basically, the goal is to split off the HTTP request count logic into a fully unit tested package, and then make the creation of the listener *and* the creation of the invalidate listener (since it followed the same logic) explicit via the config files, so that it is clear to a user what is being created and where. This also allows the Components.js builder to detect missing listeners at build time and not runtime during instantiation.

This fixes #1526 to some extent. The first execution request is still counted as a planning one, I think, but I chose not to add a weird hack for that here, because this PR is about fixes not workarounds.


packages/actor-http-retry/lib/ActorHttpRetry.ts
```diff
@@ -222,16 +222,13 @@ export class ActorHttpRetry extends ActorHttp {
   }
 }
 
-export interface IActorHttpQueueArgs extends IActorHttpArgs {
+export interface IActorHttpRetryArgs extends IActorHttpArgs {
```
surilindur, line 225: The name of this interface was supposed to mirror the actor name, but I forgot to change it apparently. So I included this small change here because I only noticed it now. The actor was called 'queue' during development.

packages/actor-query-result-serialize-sparql-json/lib/ActorQueryResultSerializeSparqlJson.ts
```diff
@@ -114,7 +112,12 @@ export class ActorQueryResultSerializeSparqlJson extends ActorQueryResultSeriali
           .map(([ key, value ]) => [ key.value, ActorQueryResultSerializeSparqlJson.bindingToJsonBindings(value) ])))}`;
           first = false;
           return res;
-        }).append(wrap(end(() => `\n]}${this.emitMetadata ? `,\n"metadata": { "httpRequests": ${this.httpObserver.requests} }` : ''}}\n`))),
```
surilindur, line 117: This is split to multiple lines to make it more readable for a developer, now with added condition checks.

packages/actor-query-result-serialize-stats/README.md
```diff
@@ -19,14 +19,12 @@ $ yarn add @comunica/actor-query-result-serialize-stats
 ## Configure
 
 After installing, this package can be added to your engine's configuration as follows:
-```text
+```json
 {
   "@context": [
-    ...
```
rubensworks, line 25: For consistency with all other READMEs, let's keep this as-is.

rubensworks, line 25: For consistency with all other READMEs, let's keep this as-is.

engines/config-query-sparql/config/config-rdfjs.json
```diff
@@ -30,39 +29,30 @@
...
     "ccqs:config/query-process/actors.json",
     "ccqs:config/query-process/mediators.json",
-    "ccqs:config/query-result-serialize/actors.json",
+    "ccqs:config/query-result-serialize/actors-rdfjs.json",
     "ccqs:config/query-result-serialize/mediators.json",
     "ccqs:config/query-source-identify/actors-rdfjs.json",
     "ccqs:config/query-source-identify/mediators.json",
     "ccqs:config/query-source-identify-hypermedia/actors-sparql.json",
     "ccqs:config/query-source-identify-hypermedia/mediators.json",
-
```
rubensworks, line 39: Please keep the empty lines.
They are here to indicate where the differences are to the default config.

packages/actor-context-preprocess-query-source-identify/lib/ActorContextPreprocessQuerySourceIdentify.ts
```diff
@@ -125,13 +125,10 @@ export interface IActorContextPreprocessQuerySourceIdentifyArgs extends IActorCo
    * @default {100}
    */
   cacheSize: number;
-  /* eslint-disable max-len */
   /**
    * An actor that listens to HTTP invalidation events
-   * @default {<default_invalidator> a <npmd:@comunica/bus-http-invalidate/^4.0.0/components/ActorHttpInvalidateListenable.jsonld#ActorHttpInvalidateListenable>}
```
rubensworks, line 131: I don't see the need to modify this to solve #1526.

If you want to change this, let's discuss this first separately.
But it seems unrelated to this PR.

packages/actor-http-limit-rate/lib/ActorHttpLimitRate.ts
```diff
@@ -139,13 +139,10 @@ export interface IActorHttpLimitRateArgs extends IActorHttpArgs {
    * The HTTP mediator.
    */
   mediatorHttp: MediatorHttp;
-  /* eslint-disable max-len */
   /**
    * An actor that listens to HTTP invalidation events
-   * @default {<default_invalidator> a <npmd:@comunica/bus-http-invalidate/^4.0.0/components/ActorHttpInvalidateListenable.jsonld#ActorHttpInvalidateListenable>}
```
rubensworks, line 145: Same here.

packages/actor-http-retry/lib/ActorHttpRetry.ts
```diff
@@ -222,16 +222,13 @@ export class ActorHttpRetry extends ActorHttp {
...
   /**
    * The HTTP mediator.
    */
   mediatorHttp: MediatorHttp;
-  /* eslint-disable max-len */
   /**
    * An actor that listens to HTTP invalidation events
-   * @default {<default_invalidator> a <npmd:@comunica/bus-http-invalidate/^4.0.0/components/ActorHttpInvalidateListenable.jsonld#ActorHttpInvalidateListenable>}
```
rubensworks, line 233: Same here.

packages/actor-query-result-serialize-sparql-json/lib/ActorQueryResultSerializeSparqlJson.ts
```diff
@@ -32,13 +31,12 @@ export class ActorQueryResultSerializeSparqlJson extends ActorQueryResultSeriali
    *   \ @defaultNested {{
    *       "application/sparql-results+json": "http://www.w3.org/ns/formats/SPARQL_Results_JSON"
    *     }} mediaTypeFormats
-   *   \ @defaultNested {true} emitMetadata
-   *   \ @defaultNested {<default_observer> a <caqrssj:components/ActionObserverHttp.jsonld#ActionObserverHttp>} httpObserver
    */
   public constructor(args: IActorQueryResultSerializeSparqlJsonArgs) {
     super(args);
+    this.emitMetadata = args.emitMetadata;
+    this.httpRequestCountObserver = args.httpRequestCountObserver;
```
rubensworks, line 38: These lines are not necessary. See base class of `Actor`.

rubensworks (changes requested):
In general, I'm also a bit worried about the breaking changes this will require.

I think some of the changes can be less invasive, to maintain backwards-compat.
You can always add inline TODOs to mark places that need to be changed once we work on the next major version.

surilindur:
The fix should now be implemented in a way that causes as few changes as possible. The actors to be tracked are defined using their URIs as strings, to avoid breakage when the fetch actor does not exist, which is the case with some of the configs.

rubensworks (changes requested):
Very nice, and fully backwards-compatible!

Could you also apply this change for the stats writer?
(this it the downside of this approach, but maybe we can refactor it for next major to be more generic)

surilindur:
It is already applied to the stats writer, as well.

rubensworks:
Oh, missed that, my bad. Looks good then!

PR #1536, opened 2025-04-14
Rework SPARQL source cardinality estimation to cache pattern counts

surilindur:
This is the set of changes that are needed to use cached `COUNT` query results for SPARQL source cardinality estimation, even without VoID metadata available. The main change here is splitting the cardinality estimation from VoID metadata into two parts:
* Triple pattern cardinality estimates, provided by VoID, based on the formulae from Hagedorn et al. with some vocabulary-based heuristics applied on top.
* Operation cardinality estimates, moved to `utils-query-operation` that break down larger operations down to triple pattern level, and then use pattern cardinalities to construct the higher-level cardinality estimate. This involves some purpose-built heuristics at the moment, but ideally this logic would later be moved into a separate cardinality estimation bus or something for more flexibility and easier implementation.

After these changes, the `QuerySourceSparql` actor uses cached triple pattern cardinalities from `COUNT` queries (done during planning phase) to reconstruct higher-level operation cardinalities when possible, and avoids a lot of larger `COUNT` queries during execution as a result.

With VoID metadata available, the reduction in `COUNT` queries is even bigger. :slightly_smiling_face:

### DIFF
--- a/packages/actor-query-source-identify-hypermedia-sparql/README.md
+++ b/packages/actor-query-source-identify-hypermedia-sparql/README.md
@@ -46,4 +46,4 @@ After installing, this package can be added to your engine's configuration as fo
 * `bindMethod`: The query operation for communicating bindings, defaults to `'values'`, alt: `'union'` or `'filter'`.
 * `countTimeout`: Timeout in ms of how long count queries are allowed to take. If the timeout is reached, an infinity cardinality is returned. Defaults to `3000`.
 * `cardinalityCountQueries`: If count queries should be sent to obtain the cardinality of (sub)queries. If set to false, resulting cardinalities will always be considered infinity. Defaults to `true`
-* `cardinalityEstimateConstruction`: If cardinality estimates for larger queries should be constructed locally from (sub)query cardinalities when possible. Defaults to `false`. If set to false, count queries will be sent for every operation.
+* `cardinalityEstimateConstruction`: If cardinality estimates for larger queries should be constructed locally from (sub)query cardinalities when possible. Defaults to `false`. If set to false, count queries will be sent for every operation at all levels.
--- a/packages/actor-query-source-identify-hypermedia-sparql/README.md
+++ b/packages/actor-query-source-identify-hypermedia-sparql/README.md
@@ -46,3 +46,4 @@ After installing, this package can be added to your engine's configuration as fo
 * `bindMethod`: The query operation for communicating bindings, defaults to `'values'`, alt: `'union'` or `'filter'`.
 * `countTimeout`: Timeout in ms of how long count queries are allowed to take. If the timeout is reached, an infinity cardinality is returned. Defaults to `3000`.
 * `cardinalityCountQueries`: If count queries should be sent to obtain the cardinality of (sub)queries. If set to false, resulting cardinalities will always be considered infinity. Defaults to `true`
+* `cardinalityEstimateConstruction`: If cardinality estimates for larger queries should be constructed locally from (sub)query cardinalities when possible. Defaults to `false`. If set to false, count queries will be sent for every operation.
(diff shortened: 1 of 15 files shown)


packages/actor-query-source-identify-hypermedia-sparql/lib/QuerySourceSparql.ts
```diff
@@ -252,32 +246,62 @@ export class QuerySourceSparql implements IQuerySource {
       }));
   }
 
+  /**
+   * Create a semi-normalized count query, by replacing the variables in subject, predicate and object
+   * positions for triple pattern operations. Any operations that are not simple patterns are unmodified.
```
rubensworks, line 251: I recommend mentioning here that you do this to increase cache hit rate.

packages/actor-rdf-metadata-extract-void/README.md
```diff
@@ -20,7 +20,7 @@ $ yarn add @comunica/actor-rdf-metadata-extract-void
 
 This actor adds the following entries to the metadata object.
 
-* `datasets`: An array containing the extracted VoID dataset metadata.
+* `patternCardinalityEstimator`: Function to provide cardinality estimates for triple patterns.
```
rubensworks, line 23: Why was this changed?
This `datasets` object looked like a nice design to me.
`patternCardinalityEstimator` seems like it can lead to more unintended conflicts, as it won't allow different actors to produce information for different datasets.

packages/types/lib/IDataset.ts
```diff
@@ -1,23 +0,0 @@
-import type { Algebra } from 'sparqlalgebrajs';
```
rubensworks, line 1: 😢

rubensworks (changes requested):
The general idea looks good, but I don't get why the metadata changes are needed.

Also, there are some merge conflicts because of some commits I just pushed 🙈

packages/actor-query-source-identify-hypermedia-sparql/lib/QuerySourceSparql.ts
```diff
@@ -196,11 +198,14 @@ export class QuerySourceSparql implements IQuerySource {
           return resolve(cachedCardinality);
         }
 
-        // Attempt to estimate locally prior to sending a COUNT request, as this should be much faster
-        const localEstimate = await this.estimateCardinality(operation);
-        if (localEstimate && Number.isFinite(localEstimate.value)) {
-          this.cache?.set(countQuery, localEstimate);
-          return resolve(localEstimate);
+        // Attempt to estimate locally prior to sending a COUNT request, as this should be much faster.
+        // The estimates may be off by varyind amounts, so this is set behind a configuration flag.
```
rubensworks, line 202:
```suggestion
        // The estimates may be off by varying amounts, so this is set behind a configuration flag.
```

packages/actor-query-source-identify-hypermedia-sparql/README.md
```diff
@@ -46,3 +46,4 @@ After installing, this package can be added to your engine's configuration as fo
 * `bindMethod`: The query operation for communicating bindings, defaults to `'values'`, alt: `'union'` or `'filter'`.
 * `countTimeout`: Timeout in ms of how long count queries are allowed to take. If the timeout is reached, an infinity cardinality is returned. Defaults to `3000`.
 * `cardinalityCountQueries`: If count queries should be sent to obtain the cardinality of (sub)queries. If set to false, resulting cardinalities will always be considered infinity. Defaults to `true`
+* `cardinalityEstimateConstruction`: If cardinality estimates for larger queries should be constructed locally from (sub)query cardinalities when possible. Defaults to `false`. If set to false, count queries will be sent for every operation.
```
rubensworks, line 49:
```suggestion
* `cardinalityEstimateConstruction`: If cardinality estimates for larger queries should be constructed locally from (sub)query cardinalities when possible. Defaults to `false`. If set to false, count queries will be sent for every operation at all levels.
```

rubensworks (approved):

rubensworks:
@ecrum19 FYI, `cardinalityEstimateConstruction` can be used now (see merged commit message).

rubensworks:
@ecrum19 Once you have some findings on the performance of enabling vs disabling `cardinalityEstimateConstruction`, please let me know. Because this option is disabled by default for now, but I highly suspect that we will be able to enable it by default.

rubensworks:
@ecrum19 Docker images won't be pushed because the CI is failing due to our server issues. Once you're going to need these images (and the servers are back up), ping me, and I'll restart the CI job.

PR #1549, opened 2025-04-25
Fix variable names in empty operations pruning actor

surilindur:
This is a fix for what appears to be a typo, where the variable names have an additional `?` in front of them, when there should be none. This issue was discovered by @ecrum19 during their experiments.


rubensworks:
Thanks @surilindur!

Could you create a separate PR with a small unit test?
Just so we avoid regressions on this in the future.

Issue #52, opened 2017-12-15
Add identity-based DISTINCT actor


rubensworks:
The current DISTINCT implementation works based on hashes. ~~While the chance of clashes is quite small~~, we should add an identity-based implementation as well.
**EDIT:** The chance of clashes became higher (#1510), and this is causing issues now.

This could work by for example just stringifying each object (e.g. using rdf-string), instead of hashing it.

We should add this in a new `actor-query-operation-distinct-identity` actor, as alternative to `actor-query-operation-distinct-hash`.

-----

## Bounty

A [bounty](https://comunica.dev/association/bounties/) has been placed on this issue by:

| |
|-|
| [<img alt="Comunica Association" src="https://comunica.dev/img/comunica_red.svg" height="50">](https://comunica.dev/association/) |
| €544 |

[Click here to learn more](https://comunica.dev/association/bounty_process/) if you're interested in claiming this bounty by resolving this issue.

rubensworks:
A bounty has been placed on this issue via the Comunica Association (see original post).

milafrerichs:
> As discussed with @rubensworks, I will work on this issue.

rubensworks:
^ FYI @karelklima.

ddeboer:
With this change, I’m getting duplicate quads again. Just so I understand correctly, does this require using a `DISTINCT` clause in my query?

ddeboer:
You’re right, @karelklima, adding `distinctConstruct` (`--context '{"distinctConstruct": true}'` in the CLI) solves it:

```
comunica-sparql-file --context '{"distinctConstruct": true}' registry.ttl \
                                                                                                        'PREFIX dcat: <http://www.w3.org/ns/dcat#> construct  where {?s a dcat:Dataset ; dcat:distribution ?d . ?d dcat:accessURL ?a }'
```

rubensworks:
This change had nothing to do with `distinctConstruct` AFAIK. So distinct being applied to construct queries without `distinctConstruct` being set to true in the past was definitely unexpected behaviour. The current behaviour of requiring `distinctConstruct` is correct.
This means #1590 can be closed (I hadn't read the issue that well yet).

PR #1550, opened 2025-05-06
feat: Add Identity based distinct actor

milafrerichs:
## Description

instead of using a hash function just stringifying the variables and checking those against the bindings

The library `rdf-string` can handle null/undefined values:
```js
expect(TermUtil.termToString(null)).toBeFalsy();
expect(TermUtil.termToString(undefined)).toBeFalsy();
```

### related issue: 
#52


rubensworks (changes requested):
Looks very nice!

Could you also include this actor in the default config (and remove the old distinct actor from the config)?
We'll need to modify the config in [a backwards-compat manner](https://github.com/comunica/comunica/tree/master/engines/config-query-sparql/config#config-versioning), so we'll need a new `query-operation/actors-v4-2-0.json` file (with propagation to parents).

packages/actor-query-operation-distinct-identity/lib/ActorQueryOperationDistinctIdentity.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * Create a new distinct filter function to Identity quads.
+   * This will maintain an internal Identity datastructure so that every quad object only returns true once.
+   * @return {(quad: RDF.Quad) => boolean} A distinct filter for quads.
+   */
+  public async newIdentityFilterQuads(): Promise<(quad: RDF.Quad) => boolean> {
+    const identities: Record<string, boolean> = {};
+    return (quad: RDF.Quad) => {
+      const identity = `${quad.subject.value}-${quad.predicate.value}-${quad.object.value}-${quad.graph.value}`;
```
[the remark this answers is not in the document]
rubensworks: Something like `Object.values(quadToStringQuad(quad)).join(' ');` should be sufficient I guess.

milafrerichs:
is that the correct way to add the configs? @rubensworks

milafrerichs:
> is that the correct way to add the configs? @rubensworks

apparently not. since `https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-query-operation-distinct-identity/^4.2.0/components/context.jsonld` does not exist yet. so this should be added in a different PR?

engines/config-query-sparql/config/config-default.json
```diff
@@ -24,7 +24,7 @@
     "ccqs:config/merge-bindings-context/mediators.json",
     "ccqs:config/optimize-query-operation/actors.json",
     "ccqs:config/optimize-query-operation/mediators.json",
-    "ccqs:config/query-operation/actors.json",
+    "ccqs:config/query-operation/actors-4-2-0.json",
```
rubensworks, line 27: This change will have to be made in config-default-v4-2-0.json (as a copy of config-default-v4-1-0.json).

And the configs in all query-sparql* packages will have to point to this new config-default-v4-2-0.json file.

All query-sparql* packages will also require swapping the old distinct actor for the new one in their package.json dependencies.

engines/config-query-sparql/config/query-operation/actors/query/distinct-v4-2-0.json
```diff
@@ -0,0 +1,16 @@
+{
+  "@context": [
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/runner/^4.0.0/components/context.jsonld",
+
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-query-operation-distinct-identity/^4.2.0/components/context.jsonld"
```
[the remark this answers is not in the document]
rubensworks: Must always refer to only the major version.

engines/config-query-sparql/config/config-default-v4-2-0.json
```diff
@@ -0,0 +1,71 @@
+{
```
rubensworks, line 1: I guess you copied this from `config-default.json`.
Could you start from `config-default-v4-1-0.json` instead?

engines/config-query-sparql/config/query-operation/actors/query/distinct-v4-2-0.json
```diff
@@ -0,0 +1,16 @@
...
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/runner/^4.0.0/components/context.jsonld",
+
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-query-operation-distinct-identity/^4.0.0/components/context.jsonld"
+  ],
+  "@id": "urn:comunica:default:Runner",
+  "@type": "Runner",
+  "actors": [
+    {
+      "@id": "urn:comunica:default:query-operation/actors#distinct",
+      "@type": "ActorQueryOperationDistinctHash",
```
rubensworks, line 12: CI is probably failing due to this.
Should be DistinctIdentity.

milafrerichs:
@rubensworks tests will fail because of these two tests. Should we delete them?

https://github.com/comunica/comunica/blob/94e1eacab069551590cc250074b36bce08720c4c/engines/query-sparql/test/Mediators-test.ts#L148C7-L149C7

```js
addTest<IActionHashBindings, IActorHashBindingsOutput>(
    'hash-bindings',
    ':query-operation/actors#distinct',
    'mediatorHashBindings',
    { allowHashCollisions: false },
    `Failed to obtaining hash functions for bindings`,
  );
  addTest<IActionHashQuads, IActorHashQuadsOutput>(
    'hash-quads',
    ':query-operation/actors#distinct',
    'mediatorHashQuads',
    { allowHashCollisions: false },
    `Failed to obtaining hash functions for quads`,
  );
  ```

or can we adjust them for the distinctIdentity part?

**Comment by rubensworks** [comment]
> @rubensworks tests will fail because of these two tests. Should we delete them?

Aha, those are mediator validation tests that depend on using one of the actors in the applicable bus.

The first one can probably be replaced with `:rdf-join/actors#inner-hash-def`.

The second one can probably be removed indeed, as there are no other usages of the hash-quads bus.

File: engines/query-sparql/config/config-default.json [hunk 6]
```diff
@@ -3,6 +3,6 @@
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/config-query-sparql/^4.0.0/components/context.jsonld"
   ],
   "import": [
-    "ccqs:config/config-default-v4-1-0.json"
+    "ccqs:config/config-default-v4-2-0.json"
```
rubensworks, line 6: Similar to this file, the other 3 engines will also require a reference to their `-v4-2-0` variant.

engines/config-query-sparql/config/config-file.json
```diff
@@ -3,7 +3,7 @@
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/config-query-sparql/^4.0.0/components/context.jsonld"
```
rubensworks, line 3: This `config-file.json` (and the next 2 files) can not be changed to ensure backwards-compat. Instead, a copy needs to be made with the `-v4-2-0.json` suffix, and that one needs to be referenced in the engines.

milafrerichs: jus to confirm:
`engines/config-query-sparql/config/config-file-v4-2-0.json`
`engines/config-query-sparql/config/config-rdfjs-lite-v4-2-0.json`
`engines/config-query-sparql/config/config-rdfjs-v4-2-0.json`

and how do we fix the dependency error? I thought to switch the `actor-query-operation-distinct-identity` for the `actor-query-operation-distinct-hash` but this seems not to be correct.

rubensworks: Should be correct, yes.

>  I thought to switch the actor-query-operation-distinct-identity for the actor-query-operation-distinct-hash but this seems not to be correct.

That is needed indeed.

> and how do we fix the dependency error?

The reported dependencies will have to be added. (some other deps are reported due to other actors being necessary for `v4-1-0`, but those weren't added in the last update yet, but will have to be included now)

milafrerichs: I updated the package.json to fix the depcheck error. so I removed the distinct-identity from sparql-rdf and sparql-rdf-lite

milafrerichs:
we should squash the commits in the end. :) or I rebase.

milafrerichs:
@rubensworks what should we do about this bundle size failure?

rubensworks:
You can bump it slightly here: https://github.com/comunica/comunica/blob/master/engines/query-sparql-rdfjs-lite/webpack.config.ts

engines/query-sparql-rdfjs-lite/config/config-default.json
```diff
@@ -3,6 +3,6 @@
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/config-query-sparql/^4.0.0/components/context.jsonld"
```
rubensworks, line 3: We forgot to also change `query-sparql-rdfjs`.

milafrerichs: what do you mean? it is changed there, right?

https://github.com/comunica/comunica/pull/1550/files/f4b80039964e321f9078d61f560ef38e92867d04#diff-9fc8647b28af4ac5396ff275b1ef5920c742a739004bb3970b0a24cadb5ae9daR6

rubensworks: Huh, I was looking for it before, but it's there indeed, nvm then :-)

rubensworks:
Nice work, thanks @milafrerichs!

PR #1559, opened 2025-05-29
Move primary Webpack config back to actor-init-query

surilindur:
This is a quick fix for the issue that was pointed out, where moving the Webpack config to the monorepo root in 99997787ae6e034354ff83cec9e9e7b654e7c609 broke use-cases where it was inherited from the actor-init-query package.

### DIFF
--- a/packages/actor-init-query/webpack.config.ts
+++ b/packages/actor-init-query/webpack.config.ts
@@ -1,5 +1,4 @@
 import { resolve } from 'node:path';
-import { ProgressPlugin } from 'webpack';
 import type { Configuration } from 'webpack';
 
 function createConfig(packagePath: string): Configuration {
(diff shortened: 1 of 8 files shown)


packages/actor-init-query/webpack.config.ts
```diff
@@ -22,9 +21,6 @@ function createConfig(packagePath: string): Configuration {
       libraryTarget: 'var',
       library: 'Comunica',
     },
-    plugins: [
-      new ProgressPlugin(),
```
rubensworks, line 26: Could we keep this one?

surilindur: Yes, if I figure out why the import keeps failing after moving the config back to the subpackage. It even failed when I added the webpack dependency into the subpackage.

surilindur: Adjusting the import fixed it, so it was probably just some ESM vs CJS thing. This is restored now.

rubensworks:
@surilindur CI is failing because if the linting issue mentioned above.

rubensworks:
Great, thanks for the quick fix @surilindur!

rubensworks:
@surilindur It looks like the browser builds now fail because of the `import.meta` usage: https://github.com/comunica/comunica/actions/runs/15326453917/job/43122998608

PR #1560, opened 2025-05-29
Add Node 24 to CI

surilindur:
This is just a small change to add Node 24 to the CI, and to run everything with it by default. This is both in preparation for Node 24 becoming LTS *and* to see if it fixes the browser build issue, because I have Node 23 locally and the CI was failing on Node 22. :thinking:

### DIFF
--- a/packages/actor-init-query/webpack.config.js
+++ b/packages/actor-init-query/webpack.config.js
@@ -1,10 +1,10 @@
-import { resolve } from 'node:path';
-import webpack from 'webpack';
+const path = require('node:path');
+const webpack = require('webpack');
 
-function createConfig(packagePath) {
+export function createConfig(packagePath) {
   return {
     devtool: 'source-map',
-    entry: resolve(packagePath, 'lib', 'index-browser.ts'),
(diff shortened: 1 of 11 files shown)


surilindur:
I changed the CI to use Node 18 by default, and fixed the `__dirname` workaround so it does not (hopefully) break on older versions, while still working on the new ones. :slightly_smiling_face:

surilindur:
The whole import thing is a mess now, I will just close this PR as not-a-solution and open an issue, maybe someone will look into it.

packages/actor-init-query/webpack.config.js
```diff
@@ -1,11 +1,10 @@
-import { resolve } from 'node:path';
-import webpack from 'webpack';
-import type { Configuration } from 'webpack';
+const path = require('node:path');
+const webpack = require('webpack');
 
-function createConfig(packagePath: string): Configuration {
+export function createConfig(packagePath) {
```
rubensworks, line 4: This is CJS, so `module.exports` must be used.

PR #1567, opened 2025-06-03
Fix browser builds on older Node versions

surilindur:
Changing the Webpack configs to CJS should hopefully fix the browser builds. They still work on Node 23, will see what the CI says about older ones.


rubensworks:
Looks great, thanks for the fix @surilindur! :-)

PR #1568, opened 2025-06-03
Fixes to CI Node versions

surilindur:
This contains the following changes:
* Adding Node 24 to CI
* Running everything on Node 18 by default, unless in a test matrix
* Fixing reference to a non-existing `matrix.node-version`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,8 @@ concurrency:
   cancel-in-progress: true
 
 env:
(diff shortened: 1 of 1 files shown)


rubensworks:
@surilindur Could you rebase to master, so we can confirm the CI still passes?

rubensworks:
Instead of one default node version, we probably need a `NODE_VERSION_LOWEST` and `NODE_VERSION_HIGHEST`.
All of the tooling checks etc can be run on the lowest, but things like performance testing can be run on the highest.

PR #1570, opened 2025-06-07
Fixes to readme badges

surilindur:
This is a small set of changes to fix some tiny annoyances with the CI badges that I have:
* Fix the indentation, because the badges are not indented, but everything else around them is.
* Remove the plastic style definitions from Gitter badges, because they look nothing like the other badges now.
* Lock the CI badge to master branch, I have no idea how or why, but this seems to fix the status showing up as failed, even though the latest run has been a success.

Any feedback would be welcome, though. I know this is a really random thing, but especially the CI badge has been bothering me for ages. :grimacing:

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,11 +11,11 @@
 </p>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


rubensworks:
Thanks! Much better indeed.

PR #1575, opened 2025-07-02
Deprioritize cartesian products in actor-rdf-join-inner-multi-smallest

RubenEschauzier:
In the multi-smallest actor, the two lowest cardinality triple patterns are chosen. This does not consider whether these triple patterns share variables. As a result, when the multi-smallest actor is used on query template C1 from the watdiv benchmark, it never finishes.

This fix addresses the problem and allows the C1 query to execute quickly, still slower than the bind-join, but relatively close.

### DIFF
--- a/packages/actor-rdf-join-inner-multi-smallest/lib/ActorRdfJoinMultiSmallest.ts
+++ b/packages/actor-rdf-join-inner-multi-smallest/lib/ActorRdfJoinMultiSmallest.ts
@@ -131,10 +131,6 @@ export class ActorRdfJoinMultiSmallest extends ActorRdfJoin<IActorRdfJoinMultiSm
           metadata.cardinality.value * requestItemTimes[i], 0),
     }, { ...sideData, sortedEntries });
   }
-
-  private sliceIndexes<T>(arr: T[], idxs: number[]): T[] {
-    return arr.filter((_, idx) => !idxs.includes(idx));
-  }
 }
 
 export interface IActorRdfJoinMultiSmallestArgs extends IActorRdfJoinArgs<IActorRdfJoinMultiSmallestTestSideData> {


packages/actor-rdf-join-inner-multi-smallest/lib/ActorRdfJoinMultiSmallest.ts
```diff
@@ -39,6 +39,30 @@ export class ActorRdfJoinMultiSmallest extends ActorRdfJoin<IActorRdfJoinMultiSm
     });
   }
 
+  /**
+   * Finds join indexes of lowest cardinality result sets, with priority on result sets that have common variables
+   * @param entries A sorted array of entries, sorted on cardinality
+   */
+  public getJoinIndexes(entries: IJoinEntryWithMetadata[]): [number, number] {
```
rubensworks, line 46: Could you also invoke this method in `getJoinCoefficients`?
As everything is still hardcoded to elements 0 and 1 there.

RubenEschauzier:
Will do, do you mean with the default Comunica configuration or Comunica with only multi-smallest enabled?

packages/actor-rdf-join-inner-multi-smallest/lib/ActorRdfJoinMultiSmallest.ts
RubenEschauzier: Done, though it makes no difference, as the calculations are done in the same manner, no matter what index is selected. For example, for the iterations attribute, we can calculate as follows:

```
 iterations:  metadatas.reduce((acc, metadata) => acc * metadata.cardinality.value, 1),
```
Is there any reason we don't do that?

rubensworks: Heh, you're right 😅 
Probably due to copy-pasting and simplifications based on the other actors.
Feel free to simplify.

rubensworks:
Thanks @RubenEschauzier!

Issue #1501, opened 2025-02-14
Support lesser-than operator for IRIs and bnodes


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

While not strictly required by the SPARQL spec, we should allow `<` to be applied on bnodes are IRIs.
`<=`, `>`, and `>=` support will follow automatically, since they are built using `<`.
This should be implemented here: https://github.com/comunica/comunica/blob/master/packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts

PR #1577, opened 2025-07-04
#1501: Support lesser-than operator for IRIs and bnodes

noahvsb:
https://github.com/comunica/comunica/issues/1501


packages/actor-term-comparator-factory-expression-evaluator/test/TermComparator-test.ts
```diff
@@ -133,8 +133,9 @@ describe('terms order', () => {
     await orderTestIsLower(dateTime('2000-01-01T00:00:00Z'), dateTime('2001-01-01T00:00:00Z'));
   });
   it('langString type comparison', async() => {
-    await orderTestIsEqual(DF.literal('a', 'de'), DF.literal('a', 'en'));
```
noahvsb, line 136: The [specs on operator mapping](https://www.w3.org/TR/sparql11-query/#OperatorMapping) led me to [RDFterm-equal(A, B)](https://www.w3.org/TR/sparql11-query/#func-RDFterm-equal). Which in turn led me to [6.5.1 Literal Equality](https://www.w3.org/TR/rdf-concepts/#section-Literal-Equality), which doesn't exist. 
But in the same file I found the [following](https://www.w3.org/TR/rdf-concepts/#section-Graph-Literal): "Literal term equality: Two literals are term-equal (the same RDF literal) if and only if the two [lexical forms](https://www.w3.org/TR/rdf-concepts/#dfn-lexical-form), the two [datatype IRIs](https://www.w3.org/TR/rdf-concepts/#dfn-datatype-iri), and the two [language tags](https://www.w3.org/TR/rdf-concepts/#dfn-language-tag) (if any) compare equal, character by character."

So this test should result to not equal.

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
```diff
@@ -31,6 +33,15 @@ export class TermFunctionLesserThan extends TermFunctionBase {
...
         .stringTest(() => (left, right) => left.localeCompare(right) === -1)
+        .set(
+          [ TypeURL.RDF_LANG_STRING, TypeURL.RDF_LANG_STRING ],
+          () => ([ left, right ]: LangStringLiteral[]) => {
+            if (left.str() !== right.str()) {
+              return bool(left.str() < right.str());
+            }
+            return bool(left.language < right.language);
+          },
+        )
```
noahvsb, line 54: The [specs on operator mapping](https://www.w3.org/TR/sparql11-query/#OperatorMapping) don't specify anything about the lesser than operator with language strings.
Also the [specs on orderby](https://www.w3.org/TR/sparql11-query/#modOrderBy) define that SPARQL doesn't define an ordering for language strings.

But in order to not have `"a"@en` and `"a"@de` be calculated as equal, I added this comparison. 

It got calculated as equal because of the following events that happened in [TermComparatorExpressionEvaluator](packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts).orderLiteralTypes():
It would first check if it's equal, then it would return false.
So then it checks if it's lesser than. But since that function didn't have support for language strings, it threw an error.
Because of that it fell back on simple string comparison, which compared only the `"a"` part and thus returned that it was equal.

The test started failing after supporting the language strings in the lesser than function, so that's how I discovered this.

packages/actor-term-comparator-factory-expression-evaluator/test/TermComparator-test.ts
jitsedesmet: I think it should since the [operator mapping](https://www.w3.org/TR/sparql11-query/#OperatorMapping) declares it. The term-equal is a dedicated [other function](https://www.w3.org/TR/sparql11-query/#func-RDFterm-equal).

noahvsb:
The specs that are failing are because of 2 SPARQL-star tests: [sparql-star-op-3](https://w3c.github.io/rdf-star/tests/sparql/eval/#sparql-star-op-3) and [sparql-star-op-4](https://w3c.github.io/rdf-star/tests/sparql/eval/#sparql-star-op-4).

Namely with the following triples:
```turtle
:x4 :left  << :a :b 123 >> .
:x4 :right << :c :d 123 >> .
```
op-3 does `FILTER ( ?left < ?right )` and op-4 does `FILTER ( ?left <= ?right )`. 

According to the [specs on operator mapping](https://www.w3.org/TR/sparql11-query/#OperatorMapping) A < B is undefined between named nodes, so comparing `:a` with `:c` results in a silent error and the triples aren't included in the result. But I defined it based on the [ORDER BY](https://www.w3.org/TR/sparql11-query/#modOrderBy) ordering for named nodes. Which is to order them by comparing them as simple literals and since `"a" < "c"` is true, the triples are included in the result.

packages/actor-term-comparator-factory-expression-evaluator/test/TermComparator-test.ts
jitsedesmet: My previous comment was wrong because `xsd:langString` is not a restriction on `xsd:string`.

packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts
```diff
@@ -27,88 +27,40 @@ export class TermComparatorExpressionEvaluator implements ITermComparator {
...
         .typedValue) {
         return 0;
       }
-      if ((<Eval.BooleanLiteral> this.lessThanFunction.applyOnTerms([ myLitA, myLitB ], this.internalEvaluator))
+      if ((<Eval.BooleanLiteral> this.lessThanFunction.applyOnTerms([ myTermA, myTermB ], this.internalEvaluator))
         .typedValue) {
         return -1;
       }
       return 1;
     } catch {
```
jitsedesmet, line 42: Is there still a catch needed? Since they now support 'term'?

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
```diff
@@ -81,6 +92,14 @@ export class TermFunctionLesserThan extends TermFunctionBase {
             return bool(this.quadComponentTest(left.graph, right.graph, exprEval) ?? false);
           },
           false,
+        ).set(
+          [ 'term', 'term' ],
+          exprEval => ([ left, right ]: [Term, Term]): BooleanLiteral => {
+            const termA = left.toRDF(exprEval.context.getSafe(KeysInitQuery.dataFactory));
+            const termB = right.toRDF(exprEval.context.getSafe(KeysInitQuery.dataFactory));
```
jitsedesmet: Do we really need this conversion? Typically, the Literals here are more expressive than what RDF.Term does, so it would be nice if we can rewrite that function to keep using the current model for literals.
(NOTE: in the PR now the orderBy does a translation from RDF.Term to our model to then translate back to RDF.term)

packages/actor-function-factory-term-greater-than-equal/test/op.greaterThenEqual-test.ts
```diff
@@ -194,11 +196,46 @@ describe('evaluation of \'>=\'', () => {
         [ '<<( <ex:a> <ex:b> 9 )>>', '<<( <ex:a> <ex:b> 123 )>>', 'false' ],
       ],
     });
+  });
+
+  describe('with named nodes operands like', () => {
     runFuncTestTable({
       ...config,
+<<<<<<< HEAD
```
jitsedesmet: Remainer of merge?

packages/actor-function-factory-term-lesser-than/test/op.lesserThan-test.ts
```diff
@@ -153,19 +181,52 @@ describe('evaluation of \'<\'', () => {
...
-        [ '<<( <ex:a> <ex:b> 123 )>>', '<<( <ex:a> <ex:b> 123 )>>', 'false' ],
-        [ '<<( <ex:a> <ex:b> 123 )>>', '<<( <ex:a> <ex:b> 123 )>>', 'false' ],
-        [ '<<( <ex:a> <ex:b> 123e0 )>>', '<<( <ex:a> <ex:b> 123 )>>', 'false' ],
-        [ '<<( <ex:a> <ex:b> 9 )>>', '<<( <ex:a> <ex:b> 123 )>>', 'true' ],
-        [ '<<( <ex:a> <ex:b> 123 )>>', '<<( <ex:a> <ex:b> 9 )>>', 'false' ],
+        [ '<< <ex:a> <ex:b> 123 >>', '<< <ex:a> <ex:b> 123.0 >>', 'false' ],
+        [ '<< <ex:a> <ex:b> 123 >>', '<< <ex:a> <ex:b> 123 >>', 'false' ],
+        [ '<< << <ex:a> <ex:b> 123 >> <ex:q> 999 >>', '<< << <ex:a> <ex:b> 123.0 >> <ex:q> 999 >>', 'false' ],
+        [ '<< << <ex:a> <ex:b> 9 >> <ex:q> 999 >>', '<< << <ex:a> <ex:b> 123.0 >> <ex:q> 999 >>', 'true' ],
+        // [ '<< <ex:q> << <ex:a> <ex:b> 9 >> 999 >>', '<< <ex:q> << <ex:a> <ex:b> 123.0 >> 999 >>', 'true' ],
```
jitsedesmet: Commented?

packages/actor-function-factory-term-lesser-than-equal/test/op.lesserThanEqual-test.ts
```diff
@@ -194,11 +196,46 @@ describe('evaluation of \'<=\'', () => {
         [ '<<( <ex:a> <ex:b> 123 )>>', '<<( <ex:a> <ex:b> 9 )>>', 'false' ],
       ],
     });
+  });
+
+  describe('with named nodes operands like', () => {
     runFuncTestTable({
       ...config,
+<<<<<<< HEAD
```
jitsedesmet: merge?

packages/actor-function-factory-term-greater-than/test/op.greaterThen-test.ts
```diff
@@ -161,11 +160,46 @@ describe('evaluation of \'>\'', () => {
         [ '<<( <ex:a> <ex:b> 9 )>>', '<<( <ex:a> <ex:b> 123 )>>', 'false' ],
       ],
     });
+  });
+
+  describe('with named nodes operands like', () => {
     runFuncTestTable({
       ...config,
+<<<<<<< HEAD
```
jitsedesmet: merge?

packages/actor-function-factory-term-greater-than/test/op.greaterThen-test.ts
```diff
@@ -64,9 +66,6 @@ describe('evaluation of \'>\'', () => {
         NaN 3f  = false
         3f  NaN = false
       `,
-      errorTable: `
-        "2"^^example:int "0"^^example:int = 'Argument types not valid for operator'
-      `,
```
jitsedesmet, line 69: Can you reinstate this test? Iit likely does not error anymore. You can just add a comment that states why. This is part of the dataType extensions but now it should mention as a comment what behaviour applies when you do not provide the type overloading.

jitsedesmet (changes requested):
Looks really good already. Good job, especially for a first PR ;)

What this PR does:
1. Adds support for term on the `<` operator. 
2. OrderBy now states: 'a'@en < 'a'@nl and uses the `<` and `=` operators directly (for any Term)

packages/actor-function-factory-term-greater-than/test/op.greaterThen-test.ts
noahvsb: It's an RDF term, namely a literal. Since the lesser than operator is now supported for RDF terms, the greater than is as well and it doesn't throw an error anymore. I defined it as comparing the string values, which are `"2"` and `"0"`.

I reinstated it and added a few more tests with RDF literals. The equality operator does throw some error on some of these, but it's pretty inconsistent.

The inconsistency of the equality operator also makes some tests fail in [TermComparator-test.ts](https://github.com/comunica/comunica/blob/9252e6d10e40f88e06032410ce2b845ccb9e8044/packages/actor-term-comparator-factory-expression-evaluator/test/TermComparator-test.ts).

noahvsb:
The `TermFunctionEquality` class throws an error if RDF literals aren't equal. Which results in some inconsistency in the tests for `<=` and `>=`.

It also makes it so that I can't safely remove the catch (which I did for this commit, thus resulting in tests failing).

packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts
```diff
@@ -27,88 +27,36 @@ export class TermComparatorExpressionEvaluator implements ITermComparator {
...
+    const myTermB: Eval.Term = this.internalEvaluator.transformer.transformRDFTermUnsafe(termB);
 
     try {
-      if ((<Eval.BooleanLiteral> this.equalityFunction.applyOnTerms([ myLitA, myLitB ], this.internalEvaluator))
+      if ((<Eval.BooleanLiteral> this.equalityFunction.applyOnTerms([ myTermA, myTermB ], this.internalEvaluator))
         .typedValue) {
         return 0;
       }
-      if ((<Eval.BooleanLiteral> this.lessThanFunction.applyOnTerms([ myLitA, myLitB ], this.internalEvaluator))
+      if ((<Eval.BooleanLiteral> this.lessThanFunction.applyOnTerms([ myTermA, myTermB ], this.internalEvaluator))
```
jitsedesmet: Could we rewrite this in function of only `<`?
So we would check `X < Y ` followed by `Y < X` and if  both are false they are equal? In that case we do not need the catch anymore either?

jitsedesmet: And we are already doing similar things for the lte operator right? So that would be nice :D

packages/actor-function-factory-term-lesser-than/test/op.lesserThan-test.ts
```diff
@@ -148,24 +176,65 @@ describe('evaluation of \'<\'', () => {
     });
   });
 
+  describe('with RDF literal operands like', () => {
```
jitsedesmet:
```suggestion
  describe('with literals of unknown types like', () => {
```

(Also check whether these tests are positioned close to the tests for [subtype substitution](https://www.w3.org/TR/xpath20/#dt-subtype-substitution))

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
```diff
@@ -102,4 +117,32 @@ export class TermFunctionLesserThan extends TermFunctionBase {
...
+    if (term.termType === 'blankNode') {
+      if (typeof (<BlankNode> term).value === 'string') {
+        return <string> (<BlankNode> term).value;
+      }
+      return (<RDF.BlankNode> (<BlankNode> term).value).value;
+    }
+    return term.str();
+  }
+
+  // SPARQL specifies that blankNode < namedNode < literal. Sparql star expands with < quads and we say < defaultGraph
```
jitsedesmet:
```suggestion
  // SPARQL specifies that blankNode < namedNode < literal. Sparql star expands with < quads and we say < defaultGraph: https://www.w3.org/TR/sparql11-query/#modOrderBy
```

packages/actor-function-factory-term-lesser-than-equal/test/op.lesserThanEqual-test.ts
```diff
@@ -181,6 +183,19 @@ describe('evaluation of \'<=\'', () => {
     });
   });
 
+  describe('with RDF literal operands like', () => {
```
jitsedesmet:
```suggestion
  describe('with literals of unknown types like', () => {
```

packages/actor-function-factory-term-greater-than/test/op.greaterThen-test.ts
```diff
@@ -148,6 +147,19 @@ describe('evaluation of \'>\'', () => {
     });
   });
 
+  describe('with RDF literal operands like', () => {
```
jitsedesmet:
```suggestion
  describe('with literals of unknown types like', () => {
```

packages/actor-function-factory-term-greater-than-equal/test/op.greaterThenEqual-test.ts
```diff
@@ -181,6 +183,19 @@ describe('evaluation of \'>=\'', () => {
     });
   });
 
+  describe('with RDF literal operands like', () => {
```
jitsedesmet:
```suggestion
  describe('with literals of unknown types like', () => {
```

packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts
jitsedesmet: We just need to be careful with the behaviour of NonLexicals probably?

jitsedesmet (changes requested):
We are really getting there! I like the way this is going. I have but a minor suggestion that we could explore.

packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts
noahvsb: > Could we rewrite this in function of only `<`? So we would check `X < Y ` followed by `Y < X` and if both are false they are equal?

Yes we could.

> In that case we do not need the catch anymore either?

This however won't be possible, as I explained before: https://github.com/comunica/comunica/pull/1577#issuecomment-3051855525.

packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts
```diff
@@ -27,88 +26,26 @@ export class TermComparatorExpressionEvaluator implements ITermComparator {
...
-      return 1;
+      return 0;
     } catch {
       // Fallback to string-based comparison
-      const compareType = this.comparePrimitives(myLitA.dataType, myLitB.dataType);
-      if (compareType !== 0) {
-        return compareType;
-      }
-      return this.comparePrimitives(myLitA.str(), myLitB.str());
+      return this.comparePrimitives(myTermA.str(), myTermB.str());
```
jitsedesmet: Can you add a comment why this is still needed? In the source code.

packages/actor-function-factory-term-lesser-than/test/op.lesserThan-test.ts
```diff
@@ -148,24 +178,80 @@ describe('evaluation of \'<\'', () => {
...
+  describe('with numeric and type discovery like', () => {
+    runFuncTestTable({
+      ...config,
+      config: new ActionContext().set(KeysExpressionEvaluator.superTypeProvider, {
+        cache: new LRUCache<string, any>({ max: 1_000 }),
+        discoverer: () => Eval.TypeURL.XSD_INTEGER,
+      }),
+      testTable: `
+         "2"^^example:int "2"^^example:int = false
+         "2"^^example:int "3"^^example:int = true
```
jitsedesmet: Add a test that shows the difference with lexicographic order

Like 02 < 5 or 25 < 100

The same test could be added in the unknown types underneath this.

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
```diff
@@ -102,4 +118,46 @@ export class TermFunctionLesserThan extends TermFunctionBase {
...
+    return this.comparePrimitives(this.getValue(termA), this.getValue(termB)) === -1;
+  }
+
+  private getValue(term: Term): string {
+    if (term.termType === 'blankNode') {
+      if (typeof (<BlankNode> term).value === 'string') {
+        return <string> (<BlankNode> term).value;
+      }
+      return (<RDF.BlankNode> (<BlankNode> term).value).value;
+    }
```
jitsedesmet: Are the explicit casts needed? can we not let TS derive this by itself?

jitsedesmet (changes requested):
Some minor nit picks still

packages/context-entries/lib/Keys.ts
```diff
@@ -350,3 +359,7 @@ export const KeysStatistics = {
     '@comunica/bus-context-preprocess:dereferencedLinks',
   ),
 };
+
+export enum FunctionLesserThanNonLexicalBehaviour {
+  throwsError, lexicalCompare,
+}
```
jitsedesmet: Could you use [string enums](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) instead? Imo it plays nicer since the string values provide some meaning.

jitsedesmet: (I think/ hope other enums here do something similar? We also dont necessarily need to create an enum, a type union might suffice.)

packages/context-entries/lib/Keys.ts
```diff
@@ -215,6 +215,15 @@ export const KeysInitQuery = {
    * A boolean value denoting whether results should be deduplicated or not.
    */
   distinctConstruct: new ActionContextKey<boolean>('@comunica/actor-init-query:distinctConstruct'),
+  /**
+   * An enum denoting the behaviour of the lesser than function when used with non lexical operands.
+   *
+   * throwsError: throws an error (default).
+   * lexicalCompare: treats them as lexicals and compares their string values.
+   */
+  functionLesserThenNonLexicalBehaviour: new ActionContextKey<FunctionLesserThanNonLexicalBehaviour>(
```
jitsedesmet:
```suggestion
  functionLesserThenNonLexicalBehaviour: new ActionContextKey<'throwsError' | 'lexicalCOmpare'>(
```

jitsedesmet: maybe 'throwsTypeError' is clearer? Since the spec talks about type errors?

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
```diff
@@ -31,11 +38,43 @@ export class TermFunctionLesserThan extends TermFunctionBase {
...
+            const nonLexical = <Term>left instanceof NonLexicalLiteral ?
+              left :
+                (<Term>right instanceof NonLexicalLiteral ? right : undefined);
+            if (nonLexical) {
+              return this.handleNonLexical(left, right, nonLexical, exprEval);
+            }
+            return bool(
+              toUTCDate(left.typedValue, exprEval.context.getSafe(KeysExpressionEvaluator.defaultTimeZone)).getTime() <
+              toUTCDate(right.typedValue, exprEval.context.getSafe(KeysExpressionEvaluator.defaultTimeZone)).getTime(),
+            );
```
jitsedesmet: Could we make this implementation maybe more similar to what happens in the helper class? SO this class would have a wrapper function that handles InvalidLiterals. And then calls the actual implementation if it does not. That wrapper function could then decide what to do based on the context entry.

jitsedesmet (changes requested):
Nice work, I think we can do a little better still :)

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
```diff
@@ -1,69 +1,115 @@
...
       arity: 2,
       operator: SparqlOperator.LT,
       overloads: declare(SparqlOperator.LT)
-        .numberTest(() => (left, right) => left < right)
+        .set(
+          [ C.TypeAlias.SPARQL_NUMERIC, C.TypeAlias.SPARQL_NUMERIC ],
+          exprEval => this.handleLiterals(exprEval),
+          false,
+        )
         .stringTest(() => (left, right) => left.localeCompare(right) === -1)
```
jitsedesmet, line 45: does this not need the custom non-lexical handling?
(Don't think it does since strings don't have invalid lexicals, but we should add a line of comment that highlights that.)

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
```diff
@@ -102,4 +188,55 @@ export class TermFunctionLesserThan extends TermFunctionBase {
...
+      const nonLexical = [ litA, litB ].find(arg => arg instanceof NonLexicalLiteral);
+      if (nonLexical) {
+        return this.handleNonLexicals(litA, litB, nonLexical, exprEval);
+      }
+
+      const compareType =
+        this.comparePrimitives(litA.dataType, litB.dataType);
+      if (compareType !== 0) {
+        return compareType === -1;
+      }
```
jitsedesmet, line 212: Could this be rewritten in function of the handleLiterals? (just wondering since this block is doing just that... If you say it's better like this, then let's keep it :) )

jitsedesmet (commented):
Very nice! I think after addressing these 2 things, it should be done. good work

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
noahvsb: I'm not sure, it definitely won't be easy since handleLiterals needs a function to be passed, but there's no real function to pass from this point. As it's in the middle of a different functionality and needs to continue like normal if there's no non lexicals.

noahvsb: I have some ideas on how I could rewrite it, but none result in better code in my opinion. For example when no function is passed on, I could return `undefined` in `handleLiterals` when there's no non lexicals. But then the necessity to check if there's a return value in this code block and not having a default function makes it uglier.

I could however make a function that does these 3 operations and call it both here and in `handleLiterals`, but then this code block wouldn't use `handleLiterals` like you asked.

noahvsb: Wait nvm, that last one will also need a check if there's a return value, which will uglify the code more than it currently is. So I think it's best to leave it in it's current state.

jitsedesmet (approved):
Very nice work! Good perseverance, the code is better for it!

packages/context-entries/lib/Keys.ts
```diff
@@ -215,6 +215,15 @@ export const KeysInitQuery = {
    * A boolean value denoting whether results should be deduplicated or not.
    */
   distinctConstruct: new ActionContextKey<boolean>('@comunica/actor-init-query:distinctConstruct'),
+  /**
+   * An enum denoting the behaviour of the lesser than function when used with non lexical operands.
+   *
+   * throwsError: throws an error (default).
+   * lexicalCompare: treats them as lexicals and compares their string values.
+   */
+  functionLesserThanNonLexicalBehaviour: new ActionContextKey<'throwsTypeError' | 'lexicalCompare'>(
```
rubensworks: I would make this a key under `KeysExpressionEvaluator`.
Do you agree @jitsedesmet?

Also, I'd simplify the name to `nonLiteralExpressionComparison`, and just make it a boolean (that defaults to `false`).
(If you have any other naming suggestions @jitsedesmet, feel free to chime in)

I would also add a CLI option for it, and a context entry shortcut. (see docs in this file on how to do this)
For the CLI option, I'd just add `--nonLiteralExpressionComparison` without shorthand.

rubensworks (changes requested):
If @jitsedesmet is happy, I'm happy as well :-)
Great work @noahvsb!

@jitsedesmet Could you double-check that all tests (unit, system, spec) are still passing? If I remember correctly, running the 1.2 spec tests require locally linking this N3.js version (to comunica and rdf-test-suite): https://github.com/rdfjs/N3.js/pull/486

rubensworks (approved):
Awesome work @noahvsb!
Will merge as soon as @jitsedesmet has confirmed the above.

jitsedesmet:
Tested and passed:
- Unit tests - no linking - pass
- spec - no linking - pass
- integration - no linking - failing  `TPF Datasets via QPF`

What I did: build https://github.com/rubensworks/N3.js/tree/feature/rdf-1-2 and `yarn link`
Then: on rdf-test-suite (main) link n3 and build and `yarn link`
Last: link n3 and rdf-test-suite to comunica and build

Test still fails:


```
      ✖ SELECT - TPF Datasets via QPF
        TPF datasets containing Link in the title through a QPF interface
        Error: Invalid query evaluation
      
              Query: PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
      SELECT *
      WHERE {
        ?dataset rdfs:label ?label
        FILTER(CONTAINS(?label, "Link"))
      }
      
              Data: [{"value":"https://data.linkeddatafragments.org/","type":""}]
      
              Result Source: https://comunica.github.io/manifest-ldf-tests/sparql/tpfdatasets/result.srj
      
              Expected: 
       [QueryResultBindings:
          Variables: [
        "?dataset",
        "?label"
      ]
          Bindings:  [
        {
          "?dataset": {
            "termType": "NamedNode",
            "value": "https://data.linkeddatafragments.org/linkedgeodata#dataset"
          },
          "?label": {
            "termType": "Literal",
            "value": "LinkedGeoData",
            "language": "",
            "datatype": {
              "termType": "NamedNode",
              "value": "http://www.w3.org/2001/XMLSchema#string"
            }
          }
        },
        {
          "?dataset": {
            "termType": "NamedNode",
            "value": "https://data.linkeddatafragments.org/linkedpolitics#dataset"
          },
          "?label": {
            "termType": "Literal",
            "value": "Linked Politics",
            "language": "",
            "datatype": {
              "termType": "NamedNode",
              "value": "http://www.w3.org/2001/XMLSchema#string"
            }
          }
        }
      ]
      ]
      
              Got: 
       [QueryResultBindings:
          Variables: [
        "?dataset",
        "?label"
      ]
          Bindings:  []
      ]
              
          at LdfTestCaseEvaluation.<anonymous> (/home/jitsedesmet/Documents/werken/ugent/comunica/node_modules/rdf-test-suite-ldf/lib/testcase/ldf/LdfTestCaseEvaluationHandler.js:94:32)
          at Generator.next (<anonymous>)
          at fulfilled (/home/jitsedesmet/Documents/werken/ugent/comunica/node_modules/rdf-test-suite-ldf/lib/testcase/ldf/LdfTestCaseEvaluationHandler.js:5:58)
          at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
        More info: https://comunica.github.io/manifest-ldf-tests/sparql/sparql-manifest.ttl#tpfdatasets
```

Issue #1307, opened 2024-03-05
Support RDF/JS Dataset as source


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

Besides the RDF/JS Source interface, we may also want to support [Datasets](https://rdf.js.org/dataset-spec/).
This can be implemented into the current rdfjs actor, where we should prefer the Source interface if the given object implements both.

PR #1578, opened 2025-07-07
#1307: added a wrapper to use RDF/JS Dataset as source

noahvsb:


packages/actor-query-source-identify-rdfjs/lib/ActorQuerySourceIdentifyRdfJs.ts
```diff
@@ -36,17 +37,32 @@ export class ActorQuerySourceIdentifyRdfJs extends ActorQuerySourceIdentify {
...
           dataFactory,
           await BindingsFactory.create(this.mediatorMergeBindingsContext, action.context, dataFactory),
         ),
         context: action.querySourceUnidentified.context ?? new ActionContext(),
       },
     };
   }
+
+  private isASource(obj: RDF.Source | RDF.Store | RDF.DatasetCore): obj is RDF.Source | RDF.Store {
+    const result = obj.match();
```
rubensworks: This call could be very expensive, as it essentially fetches all triples from the store/dataset.

I would recommend looking for the presence of certain members instead, such as `has`, `delete`, and `add`. https://rdf.js.org/dataset-spec/#datasetcore-interface

packages/actor-query-source-identify-rdfjs/lib/DatasetSourceWrapper.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  public match(
+    subject?: RDF.Term | null,
+    predicate?: RDF.Term | null,
+    object?: RDF.Term | null,
+    graph?: RDF.Term | null,
+  ): RDF.Stream<RDF.Quad> {
+    const datasetMatch: RDF.DatasetCore = this.dataset.match(subject, predicate, object, graph);
+
+    const quads = [ ...datasetMatch ];
```
rubensworks: Based on how `ArrayIterator` is implemented, it _may_ be possible to skip this step, and just pass `datasetMatch` directly.
It will require some `any` casting, but you could consider trying to see if/how it works.

packages/actor-query-source-identify-rdfjs/lib/ActorQuerySourceIdentifyRdfJs.ts
noahvsb: I did a rework, now there's no unnecessary `match()` calls anymore.

packages/actor-query-source-identify-rdfjs/lib/DatasetSourceWrapper.ts
noahvsb: I was unable to do it with `any` casting, it made the tests result in a timeout.

packages/actor-query-source-identify-rdfjs/test/ActorQuerySourceIdentifyRdfJs-test.ts
```diff
@@ -41,10 +41,31 @@ describe('ActorQuerySourceIdentifyRdfJs', () => {
   describe('An ActorQuerySourceIdentifyRdfJs instance', () => {
     let actor: ActorQuerySourceIdentifyRdfJs;
     let source: RDF.Source;
+    let dataset: RDF.DatasetCore;
 
     beforeEach(() => {
       actor = new ActorQuerySourceIdentifyRdfJs({ name: 'actor', bus, mediatorMergeBindingsContext });
       source = { match: () => <any> null };
+
+      // Empty dataset creation
```
jitsedesmet:
```suggestion
      // Mock of empty dataset
```

packages/actor-query-source-identify-rdfjs/lib/QuerySourceRdfJs.ts
```diff
@@ -187,15 +192,21 @@ export class QuerySourceRdfJs implements IQuerySource {
...
           QuerySourceRdfJs.nullifyVariables(operation.graph, quotedTripleFiltering),
         );
-        matches.on('error', reject);
-        matches.on('end', () => resolve(i));
-        matches.on('data', () => i++);
+
+        // If it's not a stream, turn it into one
+        if (typeof (<any>matches)[Symbol.asyncIterator] !== 'function') {
+          matches = <RDF.Stream>(new ArrayIterator(<RDF.DatasetCore> matches));
+        }
```
jitsedesmet: Would it make sense to do this in the constructor? Then everything is uniform more quickly?

packages/actor-query-source-identify-rdfjs/lib/ActorQuerySourceIdentifyRdfJs.ts
```diff
@@ -39,7 +38,7 @@ export class ActorQuerySourceIdentifyRdfJs extends ActorQuerySourceIdentify {
     return {
       querySource: {
         source: new QuerySourceRdfJs(
-          <RDF.Source> action.querySourceUnidentified.value,
+          <any> action.querySourceUnidentified.value,
```
jitsedesmet: any casting is typically considered bad practice. Maybe you can try using union types?

jitsedesmet: aha, you have that `QuerySourceReference` type?

engines/query-sparql/test/networkCache/34b8374ccc7d795a7c7401cfc509afe2
```diff
@@ -0,0 +1 @@
... \nundefined\n\rEncountered \" \"undef\" \"undef \"\" at line 1, column 1.\nWas expecting one of:\n    \"base\" ...\n    \"prefix\" ...\n    \"select\" ...\n    \"describe\" ...\n    \"construct\" ...\n    \"ask\" ...\n    \n\n\nFuseki - version 1.1.1 (Build date: 2014-10-02T16:36:17+0100)\n","headers":[["access-control-allow-headers","Origin, X-Requested-With, Content-Type, Accept"],["access-control-allow-methods","GET"],["access-control-allow-origin","*"],["cache-control","max-age=60, must-revalidate"],["connection","Keep-Alive"],["content-encoding","gzip"],["content-type","text/plain;charset=UTF-8"],["date","Mon, 07 Jul 2025 14:33:53 GMT"],["fuseki-request-id","242"],["keep-alive","timeout=5, max=100"],["referrer-policy","strict-origin"],["server","Fuseki (1.1.1)"],["set-cookie","connect.sid=s%3AueR6oRORT4lvCGDq9buW2LHJ.PW0VvKIEf6yLbCwW%2BQ%2F901czI1%2BFMkEB%2FtJC45kXTDM; Path=/; HttpOnly"],["strict-transport-security","max-age=31536000; includeSubdomains; preload"],["transfer-encoding","chunked"],["vary","Accept,Accept-Encoding,Accept-Charset"],["x-content-type-options","nosniff"],["x-frame-options","SAMEORIGIN"],["x-powered-by","Express"],["x-xss-protection","1; mode=block"]]}
```
jitsedesmet: woeps

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -529,14 +529,21 @@ describe('System test: QuerySparql', () => {
...
-        const store = new Store([
-          DF.quad(DF.namedNode('s'), DF.namedNode('p'), DF.namedNode('s')),
-          DF.quad(DF.namedNode('l'), DF.namedNode('m'), DF.namedNode('n')),
-        ]);
-        const result = <QueryBindings> await engine.query(`SELECT * WHERE {
+        // Store
+        const store = RdfStore.createDefault();
+        store.addQuad(DF.quad(DF.namedNode('s'), DF.namedNode('p'), DF.namedNode('s')));
+        store.addQuad(DF.quad(DF.namedNode('l'), DF.namedNode('m'), DF.namedNode('n')));
+        const storeResult = <QueryBindings> await engine.query(`SELECT * WHERE {
```
jitsedesmet: Is it the intention to overwrite this test? Could we make a new one?

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -529,14 +529,21 @@ describe('System test: QuerySparql', () => {
...
     }`, { sources: [ store ]});
-        await expect((arrayifyStream(await result.execute()))).resolves.toHaveLength(1);
+        await expect((arrayifyStream(await storeResult.execute()))).resolves.toHaveLength(1);
+
+        // Dataset
+        const dataset = store.asDataset();
+        const datasetResult = <QueryBindings> await engine.query(`SELECT * WHERE {
+      ?s ?p ?s.
+    }`, { sources: [ dataset ]});
+        await expect((arrayifyStream(await datasetResult.execute()))).resolves.toHaveLength(1);
```
jitsedesmet: for completeness we could check whether the results between both executions are equal?

jitsedesmet (commented):
Nice!
I just have a few questions.

engines/query-sparql/test/QuerySparql-test.ts
noahvsb: In practice, the test is still the same, but with less code duplication. Since I don't need to add the quads for the dataset.

packages/actor-query-source-identify-rdfjs/lib/QuerySourceRdfJs.ts
noahvsb: The problem is that the only reliable way to distinct an `RDF/JS Source` from an `RDF/JS DatasetCore` is to check the return type of the `match`-function. Since it's the only way to know for sure that we're talking about an `RDF/JS Source`

I have to check for `Source` distinctions from `DatasetCore` and not the other way aroung. Because that would make `DatasetCore` be preferred over `Source` if both are implemented by the same object.

So in order to not have an unnecessary call to `match` to check the return type, I waited pretty long to do the distinction until an actual `match`-call occurred.

rubensworks: +1 on @noahvsb's current approach.

jitsedesmet: I see, sounds good :D

engines/query-sparql/test/QuerySparql-test.ts
jitsedesmet: Yes, now that you do a comparison between the results, I like it more, looks good

jitsedesmet (approved):
To me, this looks good. @rubensworks will be responsible for a final decision.

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -529,14 +529,26 @@ describe('System test: QuerySparql', () => {
       });
 
       it('RDFJS Source', async() => {
-        const store = new Store([
```
rubensworks: I don't see the need for changing this test, can we keep it as-is?

rubensworks: If needed, a separate dedicated test can be added, which aims to test exactly the case of ensuring equal results when querying dataset vs store.

packages/actor-query-source-identify-rdfjs/lib/IRdfJsExtended.ts
```diff
@@ -48,3 +48,7 @@ export interface IRdfJsSourceExtended extends RDF.Source {
     graph: RDF.Term,
   ) => EventEmitter;
 }
+
+export interface IRdfJsSourceExtended extends RDF.Source, IRdfJsCommon {}
+
+export interface IRdfJsDatasetExtended extends RDF.DatasetCore, IRdfJsCommon {}
```
rubensworks: I see what you're doing here, but I don't think these current extensions make a lot of sense in their current form for datasets, since they are async.
For now, I don't think there's a big need for `IRdfJsDatasetExtended`, so I would recommend just keeping `IRdfJsSourceExtended` as-is.

packages/actor-query-source-identify-rdfjs/lib/QuerySourceRdfJs.ts
```diff
@@ -187,15 +192,21 @@ export class QuerySourceRdfJs implements IQuerySource {
...
           QuerySourceRdfJs.nullifyVariables(operation.predicate, quotedTripleFiltering),
           QuerySourceRdfJs.nullifyVariables(operation.object, quotedTripleFiltering),
           QuerySourceRdfJs.nullifyVariables(operation.graph, quotedTripleFiltering),
         );
-        matches.on('error', reject);
-        matches.on('end', () => resolve(i));
-        matches.on('data', () => i++);
+
+        // If it's not a stream, turn it into one
+        if (typeof (<any>matches)[Symbol.asyncIterator] !== 'function') {
```
rubensworks: Not all streams will have a `Symbol.asyncIterator` AFAIK. Might be better to just check for `typeof matches.on !== 'function'`.

rubensworks (changes requested):
Nice, almost there!

packages/actor-query-source-identify-rdfjs/lib/QuerySourceRdfJs.ts
```diff
@@ -78,8 +83,8 @@ export class QuerySourceRdfJs implements IQuerySource {
 
     // Get bindings directly if the source allows it
     // This will be more efficient, as it avoids the intermediary quads translation and representation.
-    if (this.source.matchBindings) {
-      const rawStream = this.source.matchBindings(
+    if ((<any> this.source).matchBindings) {
```
rubensworks: I suspect this any cast (and all of the other ones below) may not be necessary when replacing this check with `'matchBindings' in this.source`. Can you try?

rubensworks (changes requested):

PR #1579, opened 2025-07-08
Add federated explain actor

surilindur:
This is an implementation of an explain actor that produces the federated SPARQL serialization of the query as its output, after parsing and optimisation, with scoped source assignments converted into service clauses.

This will allow comparisons between manually federated queries, and the automated federation of Comunica, by comparing the explain output against the manually crafted query. This will also allow for the use of Comunica to split queries between sources, while using another engine to execute them.

Any feedback would be welcome. :slightly_smiling_face: Currently, it will only do the service clause conversion on triple patterns and BGPs. I need to do some more testing to see what other operations are likely to have sources assigned to them, in a way that makes sense to wrap them in service clauses.

### DIFF
--- a/packages/actor-query-process-explain-federated/README.md
+++ b/packages/actor-query-process-explain-federated/README.md
@@ -1,0 +1,41 @@
+# Comunica Explain Federated Query Process Actor
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-query-process-explain-federated.svg)](https://www.npmjs.com/package/@comunica/actor-query-process-explain-federated)
+
+A [Query Process](https://github.com/comunica/comunica/tree/master/packages/bus-query-process) actor
+that produces the federated version of a query after parsing and optimization,
+by converting algebra operation source assignments into `SERVICE` clauses,
+allowing for the inspection and comparison of automated query federation by the engine,
+as well as for the use of Comunica to handle the federation of a query but another engine to execute it.
+
+This module is part of the [Comunica framework](https://github.com/comunica/comunica),
+and should only be used by [developers that want to build their own query engine](https://comunica.dev/docs/modify/).
+
+[Click here if you just want to query with Comunica](https://comunica.dev/docs/query/).
+
+## Install
+
+```bash
+$ yarn add @comunica/actor-query-process-explain-federated
+```
+
+## Configure
+
+After installing, this package can be added to your engine's configuration as follows:
+```json
+{
+  "@context": [
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-query-process-explain-federated/^4.0.0/components/context.jsonld"
+  ],
+  "actors": [
+    {
+      "@id": "urn:comunica:default:query-process/actors#explain-federated",
+      "@type": "ActorQueryProcessExplainFederated"
+    }
+  ]
+}
+```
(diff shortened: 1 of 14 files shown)


packages/actor-query-process-explain-federated/README.md
```diff
@@ -0,0 +1,41 @@
+# Comunica Explain Federated Query Process Actor
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-query-process-explain-federated.svg)](https://www.npmjs.com/package/@comunica/actor-query-process-explain-federated)
+
+A [Query Process](https://github.com/comunica/comunica/tree/master/packages/bus-query-process) actor
+that produces the federated version of a query after parsing and optimization,
+by converting algebra operation source assignments into `SERVICE` clauses,
+allowing for the inspection and comparison of automated query federation by the engine,
```
rubensworks, line 8:
```suggestion
allowing for the inspection and comparison of source assignment and optimization by the engine,
```

packages/actor-query-process-explain-federated/lib/ActorQueryProcessExplainFederated.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  public async run(action: IActionQueryProcess): Promise<IActorQueryProcessOutput> {
+    // Parse and optimize the query
+    let { operation, context } = await this.queryProcessor.parse(action.query, action.context);
+    ({ operation, context } = await this.queryProcessor.optimize(operation, context));
+
+    const callbacks = Object.fromEntries([
+      Algebra.types.PATTERN,
+      Algebra.types.BGP,
```
rubensworks, line 43: We'll need something more generic here, as pretty much all operator types could have a source assigned to it.

rubensworks: Property paths are a special case here that may require some unwrapping (have a look at how comunica does source assignment for these).

packages/actor-query-process-explain-federated/lib/ActorQueryProcessExplainFederated.ts
```diff
@@ -0,0 +1,69 @@
...
+    ].map(at => [ at, (op: Algebra.Operation, factory: Factory) => {
+      const innerSource = <IQuerySource | undefined>(<any>op.metadata?.scopedSource)?.source?.innerSource;
+
+      if (innerSource && typeof innerSource.referenceValue === 'string') {
```
rubensworks, line 47: If it's not a string, we may want to either throw an error, or insert some dummy value.

packages/context-entries/lib/Keys.ts
```diff
@@ -186,7 +186,7 @@ export const KeysInitQuery = {
    */
   cliArgsHandlers: new ActionContextKey<ICliArgsHandler[]>('@comunica/actor-init-query:cliArgsHandlers'),
   /**
-   * Explain mode of the query. Can be 'parsed', 'logical', 'physical', or 'physical-json'.
+   * Explain mode of the query. Can be 'parsed', 'federated', 'logical', 'physical', or 'physical-json'.
```
rubensworks, line 189: I'd mention this one after 'logical'.

Issue #1515, opened 2025-03-17
Use service description sd:extensionFunction for SPARQL extension functions


jitsedesmet:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

Comunica currently supports [extension functions](https://www.w3.org/TR/sparql11-query/#extensionFunctions) by passing callbacks to the [context](https://comunica.dev/docs/modify/advanced/expression-evaluator/#config). This feature currently does not integrate well with the SPARQL service description, specifically the [sd:extensionFunction](https://www.w3.org/TR/2013/REC-sparql11-service-description-20130321/#sd-extensionFunction).

Two things are expected:
1. When comunica queries a SPARQL endpoint, it is to read the sd:extensionFunction and should not forward an extension function for which the Comunica context has an entry, but the endpoint does not explicitly list support. (meaning that when neither explicitly support the function, we still send it)
2. When Comunica is used as an endpoint, its service description should list all configured extension functions. 

As a result of **1**, a query like:
```sparql
SELECT * WHERE {
  ?person dbpo:birthDate ?bDate.
   BIND (time:periodMinInclusive(?bDate) AS ?bDateTz) .
}
```
will compute the function inside the engine in case the engine explicitly supports it, and the endpoint does not explicitly support it.
In all other cases, the engine does delegate the function execution to the endpoint.

I do wonder what should happen in case of federation?

PR #1581, opened 2025-07-15
#1515: list sd:extensionFunction in comunica's endpoint's service description and metadata extraction of other endpoint's service descriptions for sd:extensionFunction

noahvsb:


noahvsb:
@rubensworks, de test die faalt zal wel slagen als de CI nog een keer worden gerund. Maar diezelfde test faalt wel nog redelijk veel. Misschien best eens hiernaar kijken.

FAIL packages/actor-query-source-identify-hypermedia-sparql/test/QuerySourceSparql-test.ts

[](https://github.com/comunica/comunica/actions/runs/16288092802/job/45991277375?pr=1581#step:6:1433)● QuerySourceSparql › queryBindings › should not cache different queries

rubensworks:
Thanks!

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -553,12 +553,20 @@ export class HttpServiceSparqlEndpoint {
...
+
       // Append result formats
-      const formats = await engine.getResultMediaTypeFormats(new ActionContext(this.context));
+      const formats = await engine.getResultMediaTypeFormats(actionContext);
       for (const format in formats) {
         quads.push(quad(s, `${sd}resultFormat`, formats[format]));
       }
 
+      // Append extension functions
+      const functions = actionContext.get(KeysInitQuery.extensionFunctions);
```
jitsedesmet, line 565: I'm not sure, but I thought there are 2 ways to provide extension functions. But I think the other one is a callback and one is this map. Given the callback it will be hard to discover what extensions are available of course. We should make sure we properly document this behaviour on the website in in this README :)

packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -1371,6 +1373,20 @@ describe('HttpServiceSparqlEndpoint', () => {
       });
 
       it('should write the service description when no query was defined', async() => {
+        // Override instance with a context that includes some extensionFunctions
+        instance = new HttpServiceSparqlEndpoint({
```
jitsedesmet: Just to make sure, you can probably just resolve:
Do you need to overwrite the instance? It seems like it would create stateful tests, and that's not great. They should be independent if possible.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
jitsedesmet: To go forward, @rubensworks would it make sense if we would have the `pushDownNever` approach taken if extension functions are provided using the callback?
[docs](https://comunica.dev/docs/modify/advanced/expression-evaluator/#config)

jitsedesmet (commented):
Nice work! Already looks good to me. Just a single question and a note for a PR on the website :)

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
rubensworks: > To go forward, @rubensworks would it make sense if we would have the pushDownNever approach taken if extension functions are provided using the callback?

But this is unrelated to this PR, right?
I think even if we are using the callback, we should still be able to push down those expressions not handled by the callback.

packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
noahvsb: The overwriting isn't necessary, I can fix that. But it's still independent, because the instance gets changed with a `beforeEach`.

noahvsb: Besides my overwrite, there are 2 more on [line 1742](https://github.com/comunica/comunica/blob/50cdd91ad356327c46616d6fa12fffd01d8617b7/packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts#L1742) and [line 1807](https://github.com/comunica/comunica/blob/50cdd91ad356327c46616d6fa12fffd01d8617b7/packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts#L1807).

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
jitsedesmet: yes, it is unrelated. Ah, yes, at that point we can query the callback and check for a whether we have that. 
Sorry, for the confusion

packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
jitsedesmet: Ah, then it's ok!

jitsedesmet (approved):

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -553,12 +553,20 @@ export class HttpServiceSparqlEndpoint {
 
     let eventEmitter: EventEmitter;
     try {
+      const actionContext = ActionContext.ensureActionContext(this.context);
+
       // Append result formats
-      const formats = await engine.getResultMediaTypeFormats(new ActionContext(this.context));
+      const formats = await engine.getResultMediaTypeFormats(actionContext);
```
rubensworks, line 559: I think we can keep this line as-is.
Any reason why it was changed to `ActionContext.ensureActionContext`?

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -553,12 +553,20 @@ export class HttpServiceSparqlEndpoint {
...
       for (const format in formats) {
         quads.push(quad(s, `${sd}resultFormat`, formats[format]));
       }
 
+      // Append extension functions
+      const functions = actionContext.get(KeysInitQuery.extensionFunctions);
+      for (const value in functions) {
+        quads.push(quad(s, `${sd}extensionFunction`, value));
```
rubensworks, line 567: A question for @jitsedesmet: did we already have support for extension aggregates? (see `sd:extensionAggregate`)

packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
rubensworks: We've had issues with doing things like this in the past.
Let's make a local instance, such as `const localInstance` and use that one instead in the scope of this unit test.

packages/actor-rdf-metadata-extract-sparql-service/README.md
```diff
@@ -28,6 +28,7 @@ This actor adds the following entries to the metadata object.
 * `inputFormats`: List of `http://www.w3.org/ns/sparql-service-description#inputFormat` when available.
 * `resultFormats`: List of `http://www.w3.org/ns/sparql-service-description#resultFormat` when available.
 * `supportedLanguages`: List of `http://www.w3.org/ns/sparql-service-description#supportedLanguage` when available.
+* `extensionFunctions`: List of `https://www.w3.org/TR/2013/REC-sparql11-service-description-20130321/#sd-extensionFunction` when available.
```
rubensworks: The URL here seems wrong. (see other URLs above)

packages/actor-rdf-metadata-extract-sparql-service/lib/ActorRdfMetadataExtractSparqlService.ts
```diff
@@ -48,8 +49,10 @@ export class ActorRdfMetadataExtractSparqlService extends ActorRdfMetadataExtrac
           quad.subject.termType === 'BlankNode' ||
           acceptSubjectUris.has(quad.subject.value)
         ) {
+          const sd = 'http://www.w3.org/ns/sparql-service-description#';
```
rubensworks: Good idea! Maybe we can even make this a class-level static const.

rubensworks (changes requested):
Looks great @noahvsb! Just some minor comments.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
noahvsb: I just did this for less code duplication, because I also needed the `actionContext` for the extension functions.
In `engine.getResultMediaTypeFormats()` it does the `Action.ensureActionContext` and I think calling that twice wouldn't hurt. Maybe we could even remove the `Action.ensureActionContext` from `engine.getResultMediaTypeFormats()`, but I haven't looked into that.

rubensworks: > Maybe we could even remove the Action.ensureActionContext from engine.getResultMediaTypeFormats(), but I haven't looked into that.

Could you try? Because that would be even better indeed!

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
jitsedesmet: No, but it would not be very hard to implement, I would think.

packages/actor-init-query/lib/QueryEngineBase.ts
```diff
@@ -154,7 +154,6 @@ implements IQueryEngine<QueryStringContextInner, QueryAlgebraContextInner> {
    * @return {Promise<{[p: string]: number}>} All available SPARQL result media type formats.
    */
   public async getResultMediaTypeFormats(context?: any): Promise<Record<string, string>> {
-    context = ActionContext.ensureActionContext(context);
```
rubensworks: We definitely need to keep this one here!
I meant that the place where we call this method in our sparql endpoint probably doesn't require the `ActionContext.ensureActionContext` call.

packages/actor-rdf-metadata-extract-sparql-service/lib/ActorRdfMetadataExtractSparqlService.ts
```diff
@@ -1,3 +1,5 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { SD } from '@comunica/actor-rdf-metadata-extract-void/lib/Definitions';
```
rubensworks: This adds a strong dependency to `@comunica/actor-rdf-metadata-extract-void`, while we aim to make all packages in comunica as loosely coupled as possible.
We can just redefine `SD` as constant static field in the `ActorRdfMetadataExtractSparqlService` class.

rubensworks (changes requested):

packages/actor-init-query/lib/QueryEngineBase.ts
noahvsb: I was not able to do it without `ActionContext.ensureActionContext` in the endpoint. The extension functions aren't included in the service description without it.

rubensworks:
And there goes another one, nice work @noahvsb 🚀

PR #1587, opened 2025-07-19
Use ASK to verify cardinality estimate-based source assignments

surilindur:
This is a logic change to ensure Comunica follows the SPLENDID source assignment approach more carefully, by using ASK queries to avoid false positives caused by local cardinality estimation, that is used for pruning empty sources.

Previously, Comunica performed source assignment in a single pass, regardless of whether the estimate was produced locally using VoID descriptions or acquired from a remote source via COUNT query. This allows false positive source assignments for triple patterns with subject or object values that would restrict them to specific sources, because the VoID description does not encode subject or object information.

After this change, when the cardinality estimate from metadata is an estimate, and the source supports ASK queries, Comunica will issue an ASK query to verify positive source assignment results, to avoid false positives. When the cardinality is the result of a COUNT query, and is marked as accurate, it will be used as-is, and no further ASK queries are issues.

### DIFF
--- a/packages/actor-optimize-query-operation-prune-empty-source-operations/lib/ActorOptimizeQueryOperationPruneEmptySourceOperations.ts
+++ b/packages/actor-optimize-query-operation-prune-empty-source-operations/lib/ActorOptimizeQueryOperationPruneEmptySourceOperations.ts
@@ -7,7 +7,13 @@ import { ActorOptimizeQueryOperation } from '@comunica/bus-optimize-query-operat
 import { KeysInitQuery, KeysQuerySourceIdentify } from '@comunica/context-entries';
 import type { IActorTest, TestResult } from '@comunica/core';
 import { failTest, passTestVoid } from '@comunica/core';
-import type { ComunicaDataFactory, IActionContext, IQuerySourceWrapper, MetadataBindings } from '@comunica/types';
+import type {
+  ComunicaDataFactory,
+  IActionContext,
+  IQuerySourceWrapper,
+  MetadataBindings,
+  QueryResultCardinality,
+} from '@comunica/types';
 import { doesShapeAcceptOperation, getOperationSource } from '@comunica/utils-query-operation';
 import { Algebra, Factory, Util } from 'sparqlalgebrajs';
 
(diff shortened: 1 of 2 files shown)


packages/actor-optimize-query-operation-prune-empty-source-operations/lib/ActorOptimizeQueryOperationPruneEmptySourceOperations.ts
```diff
@@ -200,23 +206,31 @@ export class ActorOptimizeQueryOperationPruneEmptySourceOperations extends Actor
...
     }
 
-    // Send an ASK query
-    if (this.useAskIfSupported) {
-      const askOperation = algebraFactory.createAsk(input);
-      if (doesShapeAcceptOperation(await source.source.getSelectorShape(context), askOperation)) {
-        return source.source.queryBoolean(askOperation, context);
-      }
+    const askOperation = algebraFactory.createAsk(input);
+    const askSupported = doesShapeAcceptOperation(await source.source.getSelectorShape(context), askOperation);
```
rubensworks: Could we perform this check lazily?
Because sometimes it won't be needed, while it can be expensive to calculate.

packages/actor-optimize-query-operation-prune-empty-source-operations/lib/ActorOptimizeQueryOperationPruneEmptySourceOperations.ts
```diff
@@ -200,23 +206,31 @@ export class ActorOptimizeQueryOperationPruneEmptySourceOperations extends Actor
...
       bindingsStream.getProperty('metadata', (metadata: MetadataBindings) => {
         bindingsStream.destroy();
-        resolve(metadata.cardinality.value > 0);
+        resolve(metadata.cardinality);
       });
     });
+
+    // If the cardinality is an estimate, such as from a VoID description,
+    // verify it using ASK if the source supports it.
+    if (cardinality.type === 'estimate' && askSupported) {
```
rubensworks: Can VoID produce false negatives?
If not, there's no need to send ASK queries when cardinality is (estimated as) zero.

surilindur: That is true, I can add an exception for when the estimate is zero.

packages/actor-optimize-query-operation-prune-empty-source-operations/lib/ActorOptimizeQueryOperationPruneEmptySourceOperations.ts
surilindur: Would moving it behind the `useAskIfSupported` flag count as lazy?

rubensworks: As long as it's only calculated when it's actually required to be known, it's fine by me.

PR #1592, opened 2025-08-04
#1515: should not pushdown when comunica supports extension function, but endpoint doesn't

noahvsb:
#1515


packages/utils-query-operation/lib/Expressions.ts
```diff
@@ -2,6 +2,11 @@ import type * as RDF from '@rdfjs/types';
 import { uniqTerms } from 'rdf-terms';
 import { Algebra, Util } from 'sparqlalgebrajs';
 
+/**
+ * Get all variables inside the given expression.
+ * @param expression An expression.
+ * @return An array of variables, or undefined if the expression is unsupported for pushdown.
```
jitsedesmet: Can it return undefined? or just an []?

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
```diff
@@ -161,6 +165,18 @@ export class ActorOptimizeQueryOperationFilterPushdown extends ActorOptimizeQuer
       return true;
     }
 
+    // Don't push down extension functions comunica support, but no endpoint does
+    if (expression.expressionType === Algebra.expressionTypes.NAMED && context.has(KeysInitQuery.extensionFunctions)) {
+      const comunicaFunctions = context.getSafe(KeysInitQuery.extensionFunctions);
+      const functionName = expression.name.value;
+      if (functionName in comunicaFunctions &&
+        // Checks if there's no source that supports the extension function
+        !(sources.some(source => 'source' in source && source.source instanceof QuerySourceSparql &&
```
jitsedesmet:
```suggestion
        !(sources.some((source: {source: unknown}) => source.source instanceof QuerySourceSparql &&
```

You can skip the name check by adding a type like this. Maybe the type of that key is known, and then you can change that to the concrete type (instead of `unknown`)

packages/actor-optimize-query-operation-filter-pushdown/test/ActorOptimizeQueryOperationFilterPushdown-test.ts
```diff
@@ -269,7 +271,77 @@ describe('ActorOptimizeQueryOperationFilterPushdown', () => {
...
+          ]),
+        );
+        const src = <any> {};
+        const context = new ActionContext().set(KeysInitQuery.extensionFunctions, {
+          'https://example.com/functions#mock': async args => args[0],
+        });
+        expect(actor.shouldAttemptPushDown(op, [ src ], new Map(), context)).toBeFalsy();
+      });
+
+      it('returns true if both comunica and all sources support the extensionFunction', async() => {
```
jitsedesmet, line 298: Can you also add a test for when one endpoint supports the extension function, but the other does not?

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
```diff
@@ -161,6 +165,18 @@ export class ActorOptimizeQueryOperationFilterPushdown extends ActorOptimizeQuer
       return true;
     }
 
+    // Don't push down extension functions comunica support, but no endpoint does
```
jitsedesmet:
```suggestion
    // Don't push down extension functions comunica support, an endpoint doesn't
```

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
```diff
@@ -161,6 +165,18 @@ export class ActorOptimizeQueryOperationFilterPushdown extends ActorOptimizeQuer
...
+    if (expression.expressionType === Algebra.expressionTypes.NAMED && context.has(KeysInitQuery.extensionFunctions)) {
+      const comunicaFunctions = context.getSafe(KeysInitQuery.extensionFunctions);
+      const functionName = expression.name.value;
+      if (functionName in comunicaFunctions &&
+        // Checks if there's no source that supports the extension function
```
jitsedesmet:
```suggestion
        // Checks if there's a source that does not supports the extension function
```

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
```diff
@@ -161,6 +165,18 @@ export class ActorOptimizeQueryOperationFilterPushdown extends ActorOptimizeQuer
...
+        !(sources.some(source => 'source' in source && source.source instanceof QuerySourceSparql &&
+          source.source.extensionFunctions && source.source.extensionFunctions.includes(functionName)))) {
```
jitsedesmet:
```suggestion
        (sources.some(source => 'source' in source && source.source instanceof QuerySourceSparql &&
          source.source.extensionFunctions && !source.source.extensionFunctions.includes(functionName)))) {
```

I wonder if my suggestions are correct or there is something wrong in the documentation of your code? I feel like the different docs do not align?

packages/actor-optimize-query-operation-filter-pushdown/test/ActorOptimizeQueryOperationFilterPushdown-test.ts
```diff
@@ -269,7 +271,77 @@ describe('ActorOptimizeQueryOperationFilterPushdown', () => {
...
+            true,
+            0,
+            undefined,
+            undefined,
+            undefined,
+            [ 'https://example.com/functions#mock' ],
+          ),
+        };
+        const shapes = new Map();
+        shapes.set(src, {
```
jitsedesmet: Does the value object here satisfy some interface/ type? In that case you could annotate it as such: `{ ... } satisfies MyType`

jitsedesmet (commented):
Nice work already.  Could you also look into adding some kind of integration test for this feature?

packages/utils-query-operation/lib/Expressions.ts
noahvsb: I don't think so, since all `Algebra.expressionTypes` are covered in this switch. But I'm not sure if you can get around this in typescript or not.

packages/actor-optimize-query-operation-filter-pushdown/test/ActorOptimizeQueryOperationFilterPushdown-test.ts
noahvsb: I can, but I'm not sure what the outcome should be. Should it be passed down to all, to none or just to the one that supports it (if that last one is even possible)?

noahvsb: Judging by the following suggestions, I think it's safe to say that it should be passed down to none.

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
noahvsb: Changing `some` to `every` does the job.

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -421,6 +421,80 @@ describe('System test: QuerySparql', () => {
             .toThrow('Illegal simultaneous usage of extensionFunctionCreator and extensionFunctions in context');
         });
 
+        describe('filter pushdown behaviour with extension functions', () => {
+          let containsFilter: boolean;
```
jitsedesmet, line 430: Could you add a small word on what this test does?

Just the general how of how you pass a fetch and check whether a filter is contained from within that fetch.

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
```diff
@@ -170,6 +196,16 @@ export class ActorOptimizeQueryOperationFilterPushdown extends ActorOptimizeQuer
...
 
+  private sourceSupportsExtensionFunction(source: IQuerySource, functionName: string): boolean | undefined {
+    if (source instanceof QuerySourceSkolemized) {
+      return this.sourceSupportsExtensionFunction(source.innerSource, functionName);
+    }
+    if (source instanceof QuerySourceSparql || source instanceof QuerySourceHypermedia) {
+      return source.extensionFunctions.includes(functionName);
+    }
+    return false;
+  }
```
jitsedesmet: Call for @rubensworks to see whether this is okay. There was an issue that we have these wrapper sources that need to be unpacked until the 'actual source'. Since this function relies heavily on the existence of specific wrapper Sources and not any other (When creating a new wrapper source, this function would need to be patched.) 
-> I'm thus not sure whether this is the best approach.

jitsedesmet (commented):
Good work. If Ruben gives his blessing on the current approach for the wrapper sources, so do I.

I do wonder whether an additional test is required for the case where Comunica does not support an extension function, and one of 2 endpoints does support it.
-> The result would be push down since Comunica cannot perform it (unless through the extensionFunctionCreator - **which this PR does not cover)** (We should make this very clear in the docs.

I was also wondering whether we still want to implement a context entry that allows you to choose whether we perform pushdown?
Like `filterPushdownExtensionFunctionBehavior: 'never' | 'ifExplicitlySupported' | 'always'`.

engines/query-sparql-rdfjs-lite/webpack.config.js
```diff
@@ -3,8 +3,8 @@ const createConfig = require('@comunica/actor-init-query/webpack.config.js');
 const liteConfig = createConfig(__dirname);
 
 if (typeof liteConfig.performance === 'object') {
-  liteConfig.performance.maxAssetSize = 920_000;
-  liteConfig.performance.maxEntrypointSize = 920_000;
+  liteConfig.performance.maxAssetSize = 1_200_000;
+  liteConfig.performance.maxEntrypointSize = 1_200_000;
```
jitsedesmet: @rubensworks I assume this bundle size increases so much because of the new dependencies added? These dependencies are added because of the source wrapper issue already discussed above

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
rubensworks: This approach is not that flexible indeed, as it adds a strong dependency on existing source types and their relationship to them.

I would recommend including this in the `FragmentSelectorShape` logic. (see `IQuerySource#getSelectorShape` and `FragmentSelectorShapes.ts`)

engines/query-sparql-rdfjs-lite/webpack.config.js
rubensworks: Guessing this can be undone with the removed deps.

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
```diff
@@ -135,15 +145,18 @@ export class ActorOptimizeQueryOperationFilterPushdown extends ActorOptimizeQuer
...
    * @param operation The filter operation
    * @param sources The query sources in the operation
    * @param sourceShapes A mapping of sources to selector shapes.
+   * @param context The action context.
    */
   public shouldAttemptPushDown(
     operation: Algebra.Filter,
     sources: IQuerySourceWrapper[],
     sourceShapes: Map<IQuerySourceWrapper, FragmentSelectorShape>,
+    context: IActionContext,
```
rubensworks: Since we don't really need the rest of the context, I would just pass along the actual extensions functions object here.

packages/types/lib/IQuerySource.ts
```diff
@@ -194,6 +194,10 @@ export type FragmentSelectorShape = {
    * If bindings can be passed into the source as a filter.
    */
   filterBindings?: true;
+  /**
+   * The extension functions this source supports.
+   */
+  extensionFunctions?: string[];
```
rubensworks: I think we can make this fit a bit better in the current interface.

Could you try adding a variant of [this](https://github.com/comunica/comunica/blob/9eabbe6c6dfe52ac1b1fe5abb7a725eef523fdc4/packages/types/lib/IQuerySource.ts#L161-L164) that uses the same `operationType`, but fixed to `type` expression, and with an optional `extensionFunctions` field?

noahvsb: I have no idea what to do for the pattern in this instance. However I saw that you can add a named expression pattern, so it I had the idea to do a shape for each expression function, but that might overcomplicate things.

rubensworks: I mean adding something like
```
 | {
    operationType: 'type';
    type: Algebra.types.expression;
    extensionFunctions?: string[];
  } | 
```

noahvsb: Oops, I did `pattern: Algebra.Expression` because I made a variation of the wrong one, that's why I was having trouble

packages/utils-query-operation/lib/FragmentSelectorShapes.ts
```diff
@@ -76,11 +76,43 @@ function doesShapeAcceptOperationRecurseOperation(
...
+  return !(operation.patterns && !operation.patterns
+    .every((input: Algebra.Pattern) => doesShapeAcceptOperationRecurseShape(shapeTop, shapeTop, input, options)));
+}
+
+/**
+ * Check if the given shape accepts the given extension Function.
+ * @param shape A shape to test the query operation against.
+ * @param extensionFunction An extension function to test.
+ */
+export function doesShapeAcceptExtensionFunction(
```
rubensworks: I don't think we need these new functions.
I think we can integrate this logic into `doesShapeAcceptOperation`, by just passing the extension function as an expression object.

engines/query-sparql/test/networkCache/34b8374ccc7d795a7c7401cfc509afe2
```diff
@@ -0,0 +1 @@
... \nundefined\n\rEncountered \" \"undef\" \"undef \"\" at line 1, column 1.\nWas expecting one of:\n    \"base\" ...\n    \"prefix\" ...\n    \"select\" ...\n    \"describe\" ...\n    \"construct\" ...\n    \"ask\" ...\n    \n\n\nFuseki - version 1.1.1 (Build date: 2014-10-02T16:36:17+0100)\n","headers":[["access-control-allow-headers","Origin, X-Requested-With, Content-Type, Accept"],["access-control-allow-methods","GET"],["access-control-allow-origin","*"],["cache-control","max-age=60, must-revalidate"],["connection","Keep-Alive"],["content-encoding","gzip"],["content-type","text/plain;charset=UTF-8"],["date","Fri, 29 Aug 2025 09:13:26 GMT"],["fuseki-request-id","150"],["keep-alive","timeout=5, max=100"],["referrer-policy","strict-origin"],["server","Fuseki (1.1.1)"],["set-cookie","connect.sid=s%3APMgnzW5a2kwfHkgO9OhMxHS4.T2C24NiDDD%2BnRK9hr3e0%2BPmzsiqqwogijsqEL2JP%2F%2Fk; Path=/; HttpOnly"],["strict-transport-security","max-age=31536000; includeSubdomains; preload"],["transfer-encoding","chunked"],["vary","Accept,Accept-Encoding,Accept-Charset"],["x-content-type-options","nosniff"],["x-frame-options","SAMEORIGIN"],["x-powered-by","Express"],["x-xss-protection","1; mode=block"]]}
```
rubensworks: Where does this new cached test file come from? Is it needed? Are we doing any additional requests in tests?

noahvsb: It's probably because I got a timeout error, because my pc is slow. I should've deleted the file.

noahvsb:
I have no idea what happened, suddenly the direct POST isn't working anymore in the jest tests

packages/utils-query-operation/lib/FragmentSelectorShapes.ts
noahvsb: I don't think it'll work, because it'll always accept the extension function, because `QuerySourceSparql` has a disjunction with a wildcard as one of its children.
If we make that a conjunction, then it won't accept the operation (for the rule: Push down if federated and at least one accepts the filter) if comunica doesn't support the extension function.

rubensworks: I think we can keep the disjunction, but we could modify the meaning of the wildcard.
If we make it so that wildcard means "all operations excluding non-standard extension functions" (which we'll have to document in the jsdoc).
When checking for wildcard in the logic, we could check if we're making use of a non-standard extension function, and reject in that case.

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
```diff
@@ -161,6 +174,25 @@ export class ActorOptimizeQueryOperationFilterPushdown extends ActorOptimizeQuer
       return true;
     }
 
+    // Don't push down extension functions comunica support, but an endpoint doesn't
+    if (comunicaFunctions && expression.expressionType === Algebra.expressionTypes.NAMED) {
+      const extensionFunction = expression.name.value;
+      const extensionFunctionExpression: Algebra.NamedExpression = {
```
rubensworks: Why do we need to create a new named expression? Can't we just pass the current `expression` directly?

noahvsb: whoops

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
```diff
@@ -75,8 +84,9 @@ export class ActorOptimizeQueryOperationFilterPushdown extends ActorOptimizeQuer
       repeat = false;
       operation = Util.mapOperation(operation, {
         filter(op: Algebra.Filter, factory: Factory) {
+          const comunicaFunctions = action.context.get(KeysInitQuery.extensionFunctions);
```
rubensworks: Instead of `comunicaFunctions`, I'd just use something like `extensionFunctions` everywhere.

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
```diff
@@ -161,6 +174,25 @@ export class ActorOptimizeQueryOperationFilterPushdown extends ActorOptimizeQuer
       return true;
     }
 
+    // Don't push down extension functions comunica support, but an endpoint doesn't
```
rubensworks: I think there are some issues with this logic here:

1. If there's just a single source that supports the extension function, we should still push down, but then only call the extension functions client-side for those endpoints having no support.
2. In order to do 1., we may need some changes in `filterPushdown` (but not 100% certain). I'd start with adding some tests for expected behaviour, and implementing that in a test-driven approach.

packages/types/lib/IQuerySource.ts
```diff
@@ -164,9 +164,16 @@ export type FragmentSelectorShape = {
...
+    operationType: 'type';
+    type: Algebra.types.EXPRESSION;
+    /**
+     * The extension functions this source supports.
+     */
+    extensionFunctions?: string[];
   } | {
     /**
-     * All possible operations are accepted by this shape.
+     * All possible operations, except NamedExpressions (extension functions), are accepted by this shape.
```
rubensworks:
```suggestion
     * All possible operations are accepted by this shape.
     * As exception, extension functions are not accepted through wildcards, and must be explicitly listed via `extensionFunctions`.
```

packages/utils-query-operation/lib/FragmentSelectorShapes.ts
```diff
@@ -58,7 +64,7 @@ function doesShapeAcceptOperationRecurseShape(
       return shapeOperation.pattern.type === operation.type;
     }
     case 'wildcard': {
-      return true;
+      return operation.type !== Algebra.types.EXPRESSION || operation.expressionType !== Algebra.expressionTypes.NAMED;
```
rubensworks: This seems a bit too strict. Any kind of expression will be not allowed through this.

packages/utils-query-operation/lib/FragmentSelectorShapes.ts
```diff
@@ -76,11 +82,18 @@ function doesShapeAcceptOperationRecurseOperation(
...
-    .every((input: Algebra.Pattern) => doesShapeAcceptOperationRecurseShape(shapeTop, shapeTop, input, options))) {
-    return false;
-  }
-  return true;
+  return !(operation.patterns && !operation.patterns
+    .every((input: Algebra.Pattern) => doesShapeAcceptOperationRecurseShape(shapeTop, shapeTop, input, options)));
+}
+
+function isBuiltInFunctionIRI(iri: string): boolean {
+  return iri.includes('w3.org/2005/xpath-functions#') ||
```
rubensworks: @jitsedesmet Is this a safe check?

packages/utils-query-operation/test/FragmentSelectorShapes-test.ts
```diff
@@ -338,5 +337,105 @@ describe('FragmentSelectorShapes', () => {
...
+      beforeAll(() => {
+        extensionFunctionExpression = {
+          expressionType: Algebra.expressionTypes.NAMED,
+          name: <RDF.NamedNode> { value: 'mock1' },
+          args: [],
+          type: Algebra.types.EXPRESSION,
+        };
+      });
+
+      it('operation type (1)', () => {
```
rubensworks: Could you clarify the title tests?
It's not clear what is tested when just reading the title.

packages/utils-query-operation/test/FragmentSelectorShapes-test.ts
```diff
@@ -338,5 +337,105 @@ describe('FragmentSelectorShapes', () => {
...
+        expect(doesShapeAcceptOperation({
+          type: 'operation',
+          operation: {
+            operationType: 'type',
+            type: Algebra.types.EXPRESSION,
+            extensionFunctions: [ 'mock1' ],
```
rubensworks, line 359: Could you also add a variant for mock2? (so when the extension function is not supported).

packages/utils-query-operation/test/FragmentSelectorShapes-test.ts
```diff
@@ -338,5 +337,105 @@ describe('FragmentSelectorShapes', () => {
         undefined!,
       ))).toBeTruthy();
     });
+
+    describe('with extension functions', () => {
+      let extensionFunctionExpression: Algebra.NamedExpression;
+
+      beforeAll(() => {
+        extensionFunctionExpression = {
```
rubensworks, line 345: Could you also add one (or more) test(s) where the expression is contained within some other operation?

packages/types/lib/IQuerySource.ts
```diff
@@ -164,9 +164,17 @@ export type FragmentSelectorShape = {
...
+    operationType: 'type';
+    type: Algebra.types.EXPRESSION;
+    /**
+     * The extension functions this source supports.
+     */
+    extensionFunctions?: string[];
   } | {
     /**
      * All possible operations are accepted by this shape.
+     * As exception, extension functions are not accepted through wildcards, and must be.
```
rubensworks: The `and must be` here sounds weird.

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
```diff
@@ -135,15 +145,18 @@ export class ActorOptimizeQueryOperationFilterPushdown extends ActorOptimizeQuer
    * Check if the given filter operation must be attempted to push down, based on the following criteria:
    * - Always push down if aggressive mode is enabled
    * - Push down if the filter is extremely selective
+   * - Don't push down extension functions comunica support, but an endpoint doesn't
```
rubensworks:
```suggestion
   * - Don't push down extension functions comunica support, but a source does not
```

packages/actor-optimize-query-operation-group-sources/lib/ActorOptimizeQueryOperationGroupSources.ts
```diff
@@ -179,11 +175,23 @@ export class ActorOptimizeQueryOperationGroupSources extends ActorOptimizeQueryO
...
   ): Promise<O> {
-    if (source && doesShapeAcceptOperation(await source.source.getSelectorShape(context), operation)) {
-      this.logDebug(context, `Hoist ${inputs.length} source-specific operations into a single ${operation.type} operation for ${source.source.toString()}`);
-      operation = assignOperationSource(operation, source);
-      for (const input of inputs) {
-        removeOperationSource(input);
+    if (source) {
+      const shape: FragmentSelectorShape = await source.source.getSelectorShape(context);
+      if (doesShapeAcceptOperation(shape, operation)) {
+        // Checks for extension functions:
```
rubensworks: Since these 3 checks contain quite some logic, let's move them to a separate helper function, and move the docs to method-level jsdoc.

packages/actor-optimize-query-operation-group-sources/lib/ActorOptimizeQueryOperationGroupSources.ts
```diff
@@ -179,11 +175,23 @@ export class ActorOptimizeQueryOperationGroupSources extends ActorOptimizeQueryO
...
+        // The first 3 operands in the OR are checks to see if it's an extension function comunica supports
+        // If not, then we're good to go
+        // If it does, then only move the source annotation upwards if the source also supports the extension function
+        const extensionFunctions = context.get(KeysInitQuery.extensionFunctions);
+        const expression: Algebra.Operation | undefined = operation.expression;
```
rubensworks: We would not need the cast if we would first check the operation type, to ensure it's an expression.

packages/utils-query-operation/lib/FragmentSelectorShapes.ts
jitsedesmet: I will need to look in greater detail, but I don't think so. These named functions are supported by Comunica. [The spec](https://www.w3.org/TR/sparql11-query/#FunctionMapping) asks for these constructor functions to be implemented. But Many named functions can be suppported. I have not looked but maybe the better approach is just to ask the bus in Comunica for a function evaluator. - If the bus accepts, we have a function, otherwise not. Anyways, I will look closer later.

packages/utils-query-operation/lib/FragmentSelectorShapes.ts
```diff
@@ -58,7 +64,7 @@ function doesShapeAcceptOperationRecurseShape(
       return shapeOperation.pattern.type === operation.type;
     }
     case 'wildcard': {
-      return true;
+      return !isExtensionFunction(operation);
```
jitsedesmet, line 69: Can you explain why this is needed?

packages/utils-query-operation/lib/FragmentSelectorShapes.ts
jitsedesmet: Yeah, I don't really see it. @noahvsb what is the intention of this function? I doubt you care whether it is an extension function? (SO maybe the name is wrong?)
Does the function:
1. Check whether Comunica has support for the function?
2. Take a guess whether the endpoint supports the function
3. Something else?

packages/utils-query-operation/lib/FragmentSelectorShapes.ts
rubensworks: Jitse raises a good point: this should be briefly explained as comment in the code. Other code readers in the future will also wonder about it.

packages/utils-query-operation/lib/FragmentSelectorShapes.ts
noahvsb: Named expressions can either be built in functions or extension functions, so this check is used in my other function that checks if an expression is an extension function expression.

jitsedesmet: I would not call the named expressions that are required by the spec 'built-in functions'. (Because the spec uses builtin function as a term for functions that are not named, like `=`). That's why I suggest taking a more functional approach. So I feel like you are trying to do 2: you want to check if a minimal SPARQL-compliant engine can evaluate a given named function?

packages/utils-query-operation/test/FragmentSelectorShapes-test.ts
noahvsb: I did this, but this test fails, it returns true:

```ts
      it('should not accept when the expression is contained within a different expression and it\'s not compatible', () => {
        expect(doesShapeAcceptOperation({
          type: 'disjunction',
          children: [
            {
              type: 'operation',
              operation: {
                operationType: 'type',
                type: Algebra.types.EXPRESSION,
                extensionFunctions: [ 'mock2' ],
              },
              joinBindings: true,
            },
            {
              type: 'operation',
              operation: {
                operationType: 'type',
                type: Algebra.types.FILTER,
              },
            },
          ],
        }, <Algebra.Filter> {
          expression: extensionFunctionExpression,
          type: Algebra.types.FILTER,
        })).toBeFalsy();
      });
```

I was wondering if this might be the expected behaviour, before I put any time into fixing this.

rubensworks: Looks like a problem indeed. FILTER probably behaves as a wildcard on all expressions here as well, while it should only be a wildcard for all standard expressions.

packages/utils-query-operation/lib/FragmentSelectorShapes.ts
```diff
@@ -76,11 +84,21 @@ function doesShapeAcceptOperationRecurseOperation(
...
     return false;
   }
-  return true;
+  return !(operation.patterns && !operation.patterns
+    .every((input: Algebra.Pattern) => doesShapeAcceptOperationRecurseShape(shapeTop, shapeTop, input, options)));
+}
+
+function functionIsSparql11Required(iri: string): boolean {
+  return /^https?:\/\/www\.w3\.org\//u.test(iri);
+}
```
jitsedesmet: @rubensworks this test might be good enough (note: that is check the whole namespace though)
Interestingly enough, this works for SPARQL11 and SPARQL12 (since sparql1.2 does not introduce new required named functions). I do wonder whether newer versions of SPARQL might impose different requirements.
(Maybe that's a bridge we can cross when we get there)

rubensworks: Sure, that's fine by me!

Not a big fan of the name though, as we'd have to change it for every new SPARQL version.
What about something like `isStandardSparqlFunction`?

jitsedesmet: Yes, but than you would get into the question 'what is standard Sparql?' right?

rubensworks: Let's assume that's the latest supported SPARQL version.

jitsedesmet: sure :)

rubensworks:
> I did rebase, but did some things wrong, so had to merge the origin feature/extension-functions2 with my local one

Can you try making a different branch?
Because we won't be able to merge in this stage.

noahvsb:
> Can you try making a different branch? Because we won't be able to merge in this stage.

Oh ok, saw this too late. Ignore my above message, I'll make a different branch.

Issue #1545, opened 2025-04-22
Include VoID metadata in SPARQL endpoint


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

While we can now make use of [VoID](https://www.w3.org/TR/void/) descriptions to optimize federated queries over SPARQL endpoints, we don't expose it ourselves yet.

Let's add a flag when starting our SPARQL endpoint to optionally include VoID descriptions.
Depending on how fast it is to generate, we may consider persisting it on disk.

PR #1593, opened 2025-08-08
#1545: VoID description in SPARQL endpoint

noahvsb:
#1545


packages/actor-init-query/lib/cli/CliArgsHandlerHttp.ts
```diff
@@ -54,12 +54,19 @@ export class CliArgsHandlerHttp implements ICliArgsHandler {
...
+          type: 'boolean',
+          describe: 'Include a VoID description, which can be accessed at /void',
+        },
       })
       .check((args) => {
         if (args.version) {
           return true;
         }
-        if (args.context ? args.sources.length > 0 : args.sources.length === 0) {
+        if (args.context && args.sources.length > 0) {
```
jitsedesmet, line 66:
```suggestion
        if (args.context && args.sources.length === 0) {
```
?

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -426,6 +432,10 @@ export class HttpServiceSparqlEndpoint {
     readOnly: boolean,
     queryId: number,
   ): Promise<void> {
+    if (this.includeVoID && request.url?.split('/').at(-1)?.startsWith('void')) {
+      return this.writeVoIDDescription(engine, stdout, stderr, request, response, mediaType, headOnly);
```
surilindur: The `writeServiceDescription` method also produces a set of quads as the output, so I was wondering if it would be feasible to merge the VoID description output into the service description? 🤔 

Quite a few endpoints, for example Wikidata and UniProt, that expose a VoID description, do so within the service description, so there is no need to follow additional links or try additional URIs. The [VoID specification](https://www.w3.org/TR/void/#sparql-sd) also has it as an example publishing format. It is also the cleanest option of the ones offered by the spec (section 6.2).

surilindur: For reference, you can get the Wikidata and UniProt SD + VoID with curl:
```bash
curl -o ./wikidata-sd.nt -H "accept: application/n-triples" https://query.wikidata.org/sparql
curl -o ./uniprot-sd.nt -H "accept: application/n-triples" https://sparql.uniprot.org/sparql
```
They take slightly different approaches, but if I remember correctly, they both have a similar link path from the service to the (named or union) default graph, and/or the default dataset, which has the VoID description associated with it.

surilindur: The Comunica code to extract VoID description data is [here](https://github.com/comunica/comunica/blob/e943589fbe08c99c270019d5b2772288d17fb48b/packages/actor-rdf-metadata-extract-void/lib/ActorRdfMetadataExtractVoid.ts#L52) in `ActorRdfMetadataExtractVoid`.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -568,17 +578,112 @@ export class HttpServiceSparqlEndpoint {
...
+
+    let eventEmitter: EventEmitter;
+    try {
+      // Dublin Core Metadata Terms
+      if (this.context.dcterms) {
+        quads.push(quad(s, vocabulary, dcterms));
+        for (const key in this.context.dcterms) {
+          quads.push(quad(s, `${dcterms}${key}`, this.context.dcterms[key]));
+        }
+      }
```
jitsedesmet: The feature of adding dcterms to `/void` seems pretty undiscoverable right now. I have noticed that you did provide sufficient documentation on [your website PR](https://github.com/comunica/website/pull/63) (very nice!). @rubensworks I wonder whether docs on the website is sufficient? It seems like the 'context' here is not pretty well described in Comunica cli itself.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -568,17 +578,112 @@ export class HttpServiceSparqlEndpoint {
...
+              this.cachedStatistics.push(quad(s, `${vd}triples`, xsdInteger(triples)));
+              this.cachedStatistics.push(quad(s, `${vd}properties`, xsdInteger(triples)));
+              this.cachedStatistics.push(quad(s, `${vd}distinctSubjects`, xsdInteger(binding.get('distinctSubjects').value)));
+              this.cachedStatistics.push(quad(s, `${vd}distinctObjects`, xsdInteger(binding.get('distinctObjects').value)));
+            });
+
+            bindings.on('error', reject);
+
+            bindings.on('end', resolve);
+          });
```
jitsedesmet: I think there is a cleaner API described on the website with `for async`.

jitsedesmet: https://comunica.dev/docs/query/getting_started/query_app/#3-2-consuming-binding-results-as-an-async-iterable

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
jitsedesmet: Additionally, I would open up the functionality to support additional static properties to the `/void` endpoint. Something like `voidEndpointPropertyList`?

packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -272,6 +278,24 @@ describe('HttpServiceSparqlEndpoint', () => {
       expect(err).toContain('exposes a SPARQL endpoint');
       expect(err).toContain('At least one source must be provided');
     });
+
+    it('handles a source being passed outside the context when a context is passed', async() => {
```
jitsedesmet, line 283: 👍 For initiative

packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -1522,6 +1555,312 @@ describe('HttpServiceSparqlEndpoint', () => {
...
+          quad(s, `${dcterms}created`, `"2025/08/07"^^${xsd}date`),
+          quad(s, `${vd}triples`, `"2"^^${xsd}integer`),
+          quad(s, `${vd}properties`, `"2"^^${xsd}integer`),
+          quad(s, `${vd}distinctSubjects`, `"1"^^${xsd}integer`),
+          quad(s, `${vd}distinctObjects`, `"2"^^${xsd}integer`),
+        ]));
+      });
+
+      // eslint-disable-next-line max-len
+      it('should write the VoID description when includeVoid is true and the request url starts with \'void\' for HEAD', async() => {
```
jitsedesmet: it should not write? or it should?

jitsedesmet (commented):
Very nice work already. I have a few remarks

packages/actor-init-query/lib/cli/CliArgsHandlerHttp.ts
noahvsb: When a context is provided, there can't be any sources provided outside of the context for some reason. Which is why it uses greater than zero.

noahvsb: I remember trying to make it work, but I stumbled into some problems. I forgot which ones, but I can try to find them again.

packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
noahvsb: It should write the response header I guess. I copied the title from the service description test, where it states: "should write the service description when no query was defined for HEAD". So then I should probably change that title too.

packages/actor-init-query/lib/cli/CliArgsHandlerHttp.ts
jitsedesmet: ah right, I already got confused, yes, you are right :)

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -338,12 +344,12 @@ export class HttpServiceSparqlEndpoint {
...
     const requestUrl = url.parse(request.url ?? '', true);
     if (requestUrl.pathname === '/' || request.url === '/') {
-      stdout.write('[301] Permanently moved. Redirected to /sparql.');
+      stdout.write('[301] Permanently moved. Redirected to /sparql.\n');
       response.writeHead(301, { 'content-type': HttpServiceSparqlEndpoint.MIME_JSON, 'Access-Control-Allow-Origin': '*', Location: `http://localhost:${this.port}/sparql${requestUrl.search ?? ''}` });
       response.end(JSON.stringify({ message: 'Queries are accepted on /sparql. Redirected.' }));
       return;
     }
-    if (requestUrl.pathname !== '/sparql') {
+    if (!(requestUrl.pathname === '/sparql')) {
```
surilindur: This can probably be reverted, now that the other condition is gone.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -591,6 +606,78 @@ export class HttpServiceSparqlEndpoint {
...
 
+  public async getVoIDQuads(
+    engine: QueryEngineBase,
+    stdout: Writable,
+    response: http.ServerResponse,
+  ): Promise<RDF.Quad[]> {
+    const dataset = '_:defaultDataset';
+    const vd = 'http://rdfs.org/ns/void#';
+    const dcterms = 'http://purl.org/dc/terms/';
+    const feature = `${vd}feature`;
```
surilindur: It should be okay to have the features on the service itself, and the statistics in an `sd:defaultGraph` on an `sd:defaultDataset`, like in the Wikidata and UniProt endpoints. Something like that, as an output for the combined service description + VoID:
```turtle
@prefix sd: <http://www.w3.org/ns/sparql-service-description#> .
@prefix void: <http://rdfs.org/ns/void#> .

<https://comunica.dev/sparql> a sd:Service ;
    # Ensure the endpoint URI is present, if someone uses it
    sd:endpoint <https://comunica.dev/sparql> ;
    # Comunica supports SERVICE clauses, so could add this one
    sd:feature sd:BasicFederatedQuery ;
    # Could declare union default graph when configured for it
    sd:feature sd:UnionDefaultGraph ;
    # Listing of supported language versions,
    # not sure how to include because they depend on the parsers,
    # and the support for updates depends on the allow updates CLI flag, as well
    sd:supportedLanguage sd:SPARQL10Query ;
    sd:supportedLanguage sd:SPARQL11Query ;
    sd:supportedLanguage sd:SPARQL11Update ;
    # Supported result formats listed with sd:resultFormat
    sd:resultFormat <https://www.w3.org/ns/formats/SPARQL_Results_XML> ;
    # ... all the other formats ...
    # Pointer to default graph, only if VoID description is enabled,
    # since everything before this one applies without VoID, as well
    sd:defaultDataset _:defaultDataset .

# The things below are VoID-specific, too, since they should not be needed without it:

# The default dataset of this endpoint, which currently only has one graph,
# but later support could be added to enumerate all the named graphs being served
_:defaultDataset a sd:Dataset ;
    sd:defaultGraph _:defaultGraph .

# The default graph, which is either the union default graph or an implicit default,
# if Comunica supports implicit default graphs somehow (never checked it).
# This is the one where the results go into from the COUNT queries.
_:defaultGraph a sd:Graph ;
    void:triples 1234 ;
    void:distinctSubjects 12 ;
    void:distinctObjects 34 .
```
Sorry about being vague, I hope that clarifies it. 😓

Basically, the existing SPARQL SD implementation is (mostly) okay, and for VoID, just the steps from `sd:Service` (existing) --`sd:defaultDataset`--> `sd:Dataset` --`sd:defaultGraph`--> `sd:Graph` are needed + the VoID statistics on that default graph.

Since UniProt and Wikidata do it that way, it should be sufficient for Comunica, as well, I think.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -591,6 +603,82 @@ export class HttpServiceSparqlEndpoint {
...
+        this.context,
+      );
+
+      try {
+        for await (const bindings of bindingsStream) {
+          const xsdInteger = (n: string): string =>
+            `"${n}"^^http://www.w3.org/2001/XMLSchema#integer`;
+          const triples = bindings.get('triples')!.value;
+          this.cachedStatistics.push(quad(graph, `${vd}triples`, xsdInteger(triples)));
+          this.cachedStatistics.push(quad(graph, `${vd}properties`, xsdInteger(triples)));
```
surilindur: The actual property count could be acquired from `COUNT(DISTINCT ?p)` I think, because not every triple might have a unique property on it.

surilindur (commented):
The VoID description produced by the engine looks okay to me now, I ran it locally and it seems to make sense. The structure follows the example of Wikidata and UniProt, so it should be okay. Awesome work! 🥳 Just the code to get unique predicates needs a small fix.

The only missing thing that I could think of would be named graph handling. Comunica supports named graphs, and does not query all of them by default unless union default graph is defined. Therefore, it would make sense to treat these separately, for example by:

1. Always producing a VoID description for each named graph in the input data. This could look like the following, based off Wikidata and UniProt:

```turtle
</sparql> a sd:Service ;
    # this part is all the same, because Comunica cannot distinguish datasets
    # ... other stuff ...
    sd:defaultDataset _:defaultDataset .

_:defaultDataset a sd:Dataset ;
    # here, all the named graphs should be enumerated
    sd:namedGraph <urn:some:named:graph> ;
    sd:namedGraph <urn:other:graph> .
    # NOTE: there is no default graph, so one cannot be defined

<urn:some:named:graph> a sd:Graph ;
    # ... other metrics, too
    void:triples 123 .

<urn:other:graph> a sd:Graph ;
    # ... other metrics, too
    void:triples 123 .
```

Maybe a query like that would work:

```sparql
SELECT
    (?g AS ?graph)
    (COUNT(?s) AS ?triples)
    (COUNT(DISTINCT ?s) AS ?distinctSubjects)
    (COUNT(DISTINCT ?p) AS ?properties)
    (COUNT(DISTINCT ?o) AS ?distinctObjects)
WHERE {
    GRAPH ?g {
        ?s ?p ?o
    }
} GROUP BY ?g
```

2. Adding the union default graph bits when when the union default flag is set for Comunica itself:

```turtle
</sparql> a sd:Service ;
    # this part is all the same, because Comunica cannot distinguish datasets
    # ... other stuff ...
    # NOTE: we can now declare the union default graph behaviour,
    # because that is how Comunica behaves with the flag set
    sd:feature sd:UnionDefaultGraph ;
    # the dataset remains the same
    sd:defaultDataset _:defaultDataset .

_:defaultDataset a sd:Dataset ;
    # here, all the named graphs should be enumerated (again)
    sd:namedGraph <urn:some:named:graph> ;
    sd:namedGraph <urn:other:graph> ;
    # this time, the default can be added, too
    sd:defaultGraph _:defaultGraph .

<urn:some:named:graph> a sd:Graph ;
    # ... other metrics, too
    void:triples 123 .

<urn:other:graph> a sd:Graph ;
    # ... other metrics, too
    void:triples 123 .

_:defaultGraph a sd:Graph ;
    # ... again, all the metrics, but now for the union default graph
    void:triples 123 .
```

There is/was a separate issue for handling the union default graphs in Comunica vs source, though, so this might end up being related to that.

I have no clue if the named graph vs union should be in the scope of this work, but it does affect the VoID description produced. Maybe @rubensworks knows.

Also, if Comunica every supports querying a default named graph, the `sd:defaultGraph` would need to point at that existing default named graph, instead of generating a new default with blank node URI. But that is probably an issue for when something like it is supported.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
noahvsb: Oops, I thought it wasn't distinct properties since it's not called `void:distinctProperties`. I found it weird that it wasn't distinct, I should've known haha.

surilindur: It makes more sense (when reading) when it is listed together with  `void:classes`, but less on its own. :sweat_smile: 

There is a handy table in the spec for those: https://www.w3.org/TR/void/#statistics

noahvsb:
This is how the service + void description currently looks with the following example context:
```json
{
    "sources": [
        "https://www.rubensworks.net/",
        "https://ruben.verborgh.org/profile/"
    ],
    "dcterms": {
        "title": "comunica SPARQL endpoint",
        "creator": "me"
    }
}
```

```turtle
</sparql> a <http://www.w3.org/ns/sparql-service-description#Service>;
    <http://www.w3.org/ns/sparql-service-description#endpoint> </sparql>;
    <http://www.w3.org/ns/sparql-service-description#url> </sparql>;
    <http://www.w3.org/ns/sparql-service-description#feature> <http://www.w3.org/ns/sparql-service-description#BasicFederatedQuery>;
    <http://www.w3.org/ns/sparql-service-description#supportedLanguage> <http://www.w3.org/ns/sparql-service-description#SPARQL10Query>, <http://www.w3.org/ns/sparql-service-description#SPARQL11Query>;
    <http://www.w3.org/ns/sparql-service-description#resultFormat> <https://comunica.linkeddatafragments.org/#results_stats>, <http://www.w3.org/ns/formats/SPARQL_Results_JSON>, <http://www.w3.org/ns/formats/Shaclc>, <http://www.w3.org/ns/formats/ShaclcExtended>, <http://www.w3.org/ns/formats/JSON-LD>, <http://www.w3.org/ns/formats/N-Quads>, <http://www.w3.org/ns/formats/N-Triples>, <http://www.w3.org/ns/formats/TriG>, <http://www.w3.org/ns/formats/N3>, <http://www.w3.org/ns/formats/Turtle>, <https://comunica.linkeddatafragments.org/#results_tree>, <https://comunica.linkeddatafragments.org/#results_table>, <http://www.w3.org/ns/formats/SPARQL_Results_XML>, <http://www.w3.org/ns/formats/SPARQL_Results_TSV>, <http://www.w3.org/ns/formats/SPARQL_Results_CSV>, <https://comunica.linkeddatafragments.org/#results_simple>, <https://comunica.linkeddatafragments.org/#results_JSON>;
    <http://www.w3.org/ns/sparql-service-description#defaultDataset> _:defaultDataset.
_:defaultDataset a <http://www.w3.org/ns/sparql-service-description#Dataset>, <http://rdfs.org/ns/void#Dataset>;
    <http://rdfs.org/ns/void#sparqlEndpoint> </sparql>;
    <http://rdfs.org/ns/void#vocabulary> <http://purl.org/dc/terms/>;
    <http://purl.org/dc/terms/title> <comunica SPARQL endpoint>;
    <http://purl.org/dc/terms/creator> <me>;
    <http://www.w3.org/ns/sparql-service-description#defaultGraph> _:defaultGraph.
_:defaultGraph a <http://www.w3.org/ns/sparql-service-description#Graph>;
    <http://rdfs.org/ns/void#triples> 11550;
    <http://rdfs.org/ns/void#distinctSubjects> 3715;
    <http://rdfs.org/ns/void#properties> 79;
    <http://rdfs.org/ns/void#distinctObjects> 5648.```

**Comment by noahvsb** [comment]
I want to give an updated version of the output, but I forgot to test the actual query and it throws errors, so I still need to fix that.

**Comment by noahvsb** [comment]
I just realized that default graph only contains the classes statistic, odd.

**Comment by rubensworks** [comment]
Very cool, could you have a look at the output @surilindur? To see if it contains everything we would need for VoID extraction and query planning?

@noahvsb Could you report on execution times for generating this?
Both on the sources in your example, but perhaps also for https://fragments.dbpedia.org/2016-04/en, and for https://dbpedia.org/sparql
(Without having looked at the code) do you do any caching for subsequent requests to the service description?

**Comment by noahvsb** [comment]
I started counting from the call of the statistics fetch until the last VoID quad was added to the list of quads of the service description.

- https://www.rubensworks.net/, https://ruben.verborgh.org/profile/: 2387 ms
- https://fragments.dbpedia.org/2016-04/en: ERROR, see below (438 ms)
- https://dbpedia.org/sparql: 31422 ms

To prove that caching works, here are the execution times the 2nd time for each of the sources.

- https://www.rubensworks.net/, https://ruben.verborgh.org/profile/: 0 ms
- https://fragments.dbpedia.org/2016-04/en: ERROR, see below (8 ms)
- https://dbpedia.org/sparql: 8 ms

The error I got with https://fragments.dbpedia.org/2016-04/en was preceded by this warning:

```
[2025-09-08T09:00:38.631Z]  WARN: Metadata extraction for https://fragments.dbpedia.org/2016-04/en: failed: Could not retrieve https://fragments.dbpedia.org/2016-04/en: (HTTP status 404):
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>.
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
@prefix owl: <http://www.w3.org/2002/07/owl#>.
@prefix skos: <http://www.w3.org/2004/02/skos/core#>.
@prefix xsd: <http://www.w3.org/2001/XMLSchema#>.
@prefix dc: <http://purl.org/dc/terms/>.
@prefix dcterms: <http://purl.org/dc/terms/>.
@prefix dc11: <http://purl.org/dc/elements/1.1/>.
@prefix foaf: <http://xmlns.com/foaf/0.1/>.
@prefix geo: <http://www.w3.org/2003/01/geo/wgs84_pos#>.
@prefix dbpedia: <http://dbpedia.org/resource/>.
@prefix dbpedia-owl: <http://dbpedia.org/ontology/>.
@prefix dbpprop: <http://dbpedia.org/property/>.
@prefix hydra: <http://www.w3.org/ns/hydra/core#>.
@prefix void: <http://rdfs.org/ns/void#>.

<#metadata> {
<#metadata> foaf:primaryTopic <>.
<https://fragments.dbpedia.org/2016-04/en#dataset> a void:Dataset, hydra:Collection;
    dcterms:title "DBpedia 2016-04".
<https://fragments.dbpedia.org/2015-10/en#dataset> a void:Dataset, hydra:Collection;
    dcterms:title "DBpedia 2015-10".
<https://fragments.dbpedia.org/2015/en#dataset> a void:Dataset, hydra:Collection;
    dcterms:title "DBpedia 2015-04".
<https://fragments.dbpedia.org/2014/en#dataset> a void:Dataset, hydra:Collection;
    dcterms:title "DBpedia 2014".
<https://fragments.dbpedia.org/#dataset> a void:Dataset, hydra:Collection;
    dcterms:title "dataset index"
}
 { actor: 'urn:comunica:default:query-source-identify/actors#hypermedia' }
```

This is the error (the same as the warning, but for completion I'm adding it here):

```
[500] Server error in results: Could not retrieve https://fragments.dbpedia.org/2016-04/en: (HTTP status 404):
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>.
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#>.
@prefix owl: <http://www.w3.org/2002/07/owl#>.
@prefix skos: <http://www.w3.org/2004/02/skos/core#>.
@prefix xsd: <http://www.w3.org/2001/XMLSchema#>.
@prefix dc: <http://purl.org/dc/terms/>.
@prefix dcterms: <http://purl.org/dc/terms/>.
@prefix dc11: <http://purl.org/dc/elements/1.1/>.
@prefix foaf: <http://xmlns.com/foaf/0.1/>.
@prefix geo: <http://www.w3.org/2003/01/geo/wgs84_pos#>.
@prefix dbpedia: <http://dbpedia.org/resource/>.
@prefix dbpedia-owl: <http://dbpedia.org/ontology/>.
@prefix dbpprop: <http://dbpedia.org/property/>.
@prefix hydra: <http://www.w3.org/ns/hydra/core#>.
@prefix void: <http://rdfs.org/ns/void#>.

<#metadata> {
<#metadata> foaf:primaryTopic <>.
<https://fragments.dbpedia.org/2016-04/en#dataset> a void:Dataset, hydra:Collection;
    dcterms:title "DBpedia 2016-04".
<https://fragments.dbpedia.org/2015-10/en#dataset> a void:Dataset, hydra:Collection;
    dcterms:title "DBpedia 2015-10".
<https://fragments.dbpedia.org/2015/en#dataset> a void:Dataset, hydra:Collection;
    dcterms:title "DBpedia 2015-04".
<https://fragments.dbpedia.org/2014/en#dataset> a void:Dataset, hydra:Collection;
    dcterms:title "DBpedia 2014".
<https://fragments.dbpedia.org/#dataset> a void:Dataset, hydra:Collection;
    dcterms:title "dataset index"
}
```

File: packages/actor-init-query/lib/cli/CliArgsHandlerHttp.ts [hunk 10]
```diff
@@ -54,12 +54,19 @@ export class CliArgsHandlerHttp implements ICliArgsHandler {
           type: 'boolean',
           describe: 'If the query context can be overridden through POST requests',
         },
+        includeVoID: {
```
rubensworks:
```suggestion
        emitVoid: {
```
Also changed capitalization, which I know is correct, but might be complicated for users to get right otherwise.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -45,16 +45,20 @@ export class HttpServiceSparqlEndpoint {
 
   public readonly freshWorkerPerQuery: boolean;
   public readonly contextOverride: boolean;
+  public readonly includeVoID: boolean;
```
rubensworks: I would use `includeVoid` anywhere, to simplify capitalization.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -465,6 +471,13 @@ export class HttpServiceSparqlEndpoint {
       return;
     }
 
+    // If the query was an update query, update the cachedStatistics
+    if (this.includeVoID && queryBody.value.includes('INSERT')) {
+      this.fetchVoIDStatistics(engine).catch(() => {
```
rubensworks: Why is the `fetchVoIDStatistics` needed?
I guess we can just reset `this.cachedStatistics`, which will trigger recalc later when needed?

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -591,6 +610,168 @@ export class HttpServiceSparqlEndpoint {
...
   }
 
+  /**
+   * Returns a list of all necessary VoID quads.
+   * @param {QueryEngineBase} engine A SPARQL engine.
+   * @param {module:stream.internal.Writable} stdout
+   * @param {module:http.IncomingMessage} request
+   * @param {module:http.ServerResponse} response
+   */
+  public async getVoIDQuads(
```
rubensworks: There's quite some logic here. Could we move this to a dedicated class (e.g. `VoidMetadataEmitter`), and construct and invoke that one from somewhere here?

rubensworks: Most unit tests could then also be moved to a new test file.

rubensworks (changes requested):
Nice work @noahvsb.

The perf show that we should document this as an experimental feature, but it's still useful to include at this stage already.
In the future, we should think about timeout strategies for slow queries, and partial VoID descriptions in case some of those queries timed out.

Below some feedback on this PR.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
noahvsb: I thought it would be more optimized by doing it in the background immediately after the update, instead of waiting for the next service description GET.

rubensworks: It may never even be needed again. E.g. when multiple insert queries are done in sequence.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -567,11 +580,17 @@ export class HttpServiceSparqlEndpoint {
         quads.push(quad(s, `${sd}extensionFunction`, value));
       }
 
+      if (!this.emitVoid) {
```
rubensworks: This seems incorrect. The negation should be removed I guess? (also above?)

noahvsb: If you don't emit, then that means the same as include right?

noahvsb: I think I got confused with omit

packages/actor-init-query/lib/cli/CliArgsHandlerHttp.ts
```diff
@@ -54,12 +54,19 @@ export class CliArgsHandlerHttp implements ICliArgsHandler {
           type: 'boolean',
           describe: 'If the query context can be overridden through POST requests',
         },
+        emitVoid: {
+          type: 'boolean',
+          describe: 'Emit the VoID description, which can normally be found in the service description',
```
rubensworks:
```suggestion
          describe: 'Generate VoID descriptions and include them in the service description',
```

Issue #1253, opened 2018-11-18
Implement automatic type casting for xs:untypedAtomic 


wschella:
See [https://www.w3.org/TR/xpath-31/#dt-function-conversion](https://www.w3.org/TR/xpath-31/#dt-function-conversion). Relevant for all functions, somewhat important for unary plus.

PR #1594, opened 2025-08-13
#1253: automatic type casting for xs:untypedAtomic

noahvsb:
According to [the specs](https://www.w3.org/TR/xpath-functions-31/#casting-from-strings): 
"If the value is xs:untypedAtomic, it is treated in exactly the same way as a string containing the same sequence of characters."
So in `TypeHandling.ts` I added `[TypeURL.XSD_UNTYPED_ATOMIC]: TypeURL.XSD_STRING,` to the `extensionTableInput` and it worked. xsd:UntypedAtomic isn't really a subtype of xsd:sting by definition, but this way was pretty easy. It's well documented inside the source code as well.

The following query produces results now:
```SPARQL
PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
PREFIX ex: <http://example.org/>

SELECT ?person ?age
WHERE {
  ?person ex:age ?age .
  FILTER(xsd:decimal(?age) > 30)
  FILTER(DATATYPE(?age) = xsd:untypedAtomic)
}
```

I also tested the same query in oxigraph and there it produced no results. You first have to cast to str (`FILTER(xsd:decimal(str(?age)) > 30)`), which is not correct according to the specs. 
I could test it with jena as well, but I remember having a lot of trouble with it last time. So unless it's necessary I'd rather not do it.


packages/utils-expression-evaluator/lib/util/TypeHandling.ts
```diff
@@ -37,6 +37,11 @@ export const extensionTableInput: Record<KnownLiteralTypes, OverrideType> = {
   [TypeURL.XSD_ENTITY]: TypeURL.XSD_NC_NAME,
   [TypeURL.XSD_ID]: TypeURL.XSD_NC_NAME,
   [TypeURL.XSD_ID_REF]: TypeURL.XSD_NC_NAME,
+  // https://www.w3.org/TR/xpath-functions-31/#casting-from-strings
```
jitsedesmet, line 44: Can you explain starting from: https://www.w3.org/TR/xpath-31/#dt-function-conversion
(just like you did in the conversation on this PR :) )

So the reasoning is clear that https://www.w3.org/TR/xpath-31/#dt-function-conversion says do cast and that cast says that untyped Is like string?

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -1488,6 +1488,77 @@ SELECT ?option WHERE {
       });
     });
 
+    describe('xsd:untypedAtomic', () => {
```
jitsedesmet: I don't think these tests need to operate on this level, so it might be better to find a place to write this as a unit test?
(Just to not make this test file to big)

jitsedesmet (changes requested):
Looks really good already. Nice solution using the typeHandling system and documenting why it's a special case.
I would just suggest moving the test elsewhere as to not overload that singe integration-test-file

engines/query-sparql/test/QuerySparql-test.ts
noahvsb: Do you have a suggestion for the place?

rubensworks:
Rerunning.

jitsedesmet (approved):
Looks good to me. @rubensworks if you are happy with the approach taken, this can be accepted.

Issue #1540, opened 2025-04-21
Specify base IRI of parsed documents


constraintAutomaton:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

When using the binary version of `comunica-sparql-file`, the `-b` or `--baseIRI` option does not work. The IRI produced still uses the file path of the current folder instead of the IRI specified via the flag.

### How to reproduce

```bash
comunica-sparql-file {my-rdf-file} -b "http://example.org/" -f {my-query-file}
```

____
#### Environment:

```
| software         | version
| ---------------- | -------
| Comunica Engine  | 4.1.0 
| node             | v23.9.0
| npm              | 11.3.0
| yarn             | 1.22.19
| Operating System | linux (Linux 6.14.2-arch1-1)
```

rubensworks:
The `-b` flag is used to specify the base IRI of the query, not the RDF documents, for which the base IRI is not overridable at the moment.
But this could be a useful feature indeed, but then perhaps behind a separate flag, as both may not (always) be the same.

PR #1597, opened 2025-08-19
#1540: Specify base IRI of parsed documents

noahvsb:
#1540


engines/query-sparql-file/test/QuerySparqlFile-test.ts
```diff
@@ -15,15 +15,33 @@ describe('System test: QuerySparqlFile', () => {
...
+
+SELECT ?person ?name
+WHERE {
+  <Alice> foaf:knows ?person .
+  ?person a <Person> ;
+          foaf:name ?name .
+}
+        `;
+        const baseIRI = 'http://example.org/';
+        const context: QueryStringContext = { sources: [{ value: p }], baseIRI, fileBaseIRI: baseIRI };
```
rubensworks: To improve test robustness, I'd use a different value for baseIRI and fileBaseIRI.

packages/bus-dereference/lib/ActorDereference.ts
```diff
@@ -82,6 +82,10 @@ export interface IActorDereferenceOutput extends IActorOutput {
    * as this may have changed due to redirects.
    */
   url: string;
+  /**
+   * BaseIRI.
```
rubensworks: Make sure to explain the intended behaviour here. If defined, `baseIRI` will be used over `url`.

rubensworks (changes requested):

engines/query-sparql-file/test/QuerySparqlFile-test.ts
```diff
@@ -15,15 +15,35 @@ describe('System test: QuerySparqlFile', () => {
...
+          foaf:name ?name .
+}
+        `;
+        const baseIRI = 'http://example.org/';
+        const fileBaseIRI = 'http://different.org/';
+        const context: QueryStringContext = { sources: [{ value: p }], baseIRI, fileBaseIRI };
+
+        const result = await arrayifyStream(await engine.queryBindings(query, context));
+        // Result should be length 1 if you use the same baseIRI, which shows that it actually makes a difference
+        expect(result).toHaveLength(0);
```
rubensworks: I would recommend actually inspecting the results to ensure the right `fileBaseIRI` was used.
You may need to tweak the query slightly to actually get some results.
Because 0 results could also be caused by some other bugs, so this is not a strong guarantee for correctness.

rubensworks (changes requested):

engines/query-sparql-file/test/QuerySparqlFile-test.ts
```diff
@@ -15,15 +18,41 @@ describe('System test: QuerySparqlFile', () => {
...
+
+        const expectedResult = [
+          [
+            [ DF.variable('name'), DF.literal('Bob', DF.namedNode('http://www.w3.org/2001/XMLSchema#string')) ],
+            [ DF.variable('person'), DF.namedNode(`${fileBaseIRI}Bob`) ],
+          ],
+        ];
+
+        const result = await arrayifyStream(await engine.queryBindings(query, context));
+        expect(result.map(binding => [ ...binding ])).toMatchObject(expectedResult);
```
rubensworks: Best to use `toEqualBindingsArray` instead.
See other tests that use `toEqualBindingsArray` as example.

rubensworks (approved):
Excellent work @noahvsb 🚀

PR #1598, opened 2025-08-22
Support for Accept-Post header

noahvsb:
As described [here](https://github.com/comunica/comunica/discussions/1482#discussioncomment-12136129)


engines/config-query-sparql/config/rdf-metadata-extract/actors-v4-1-0.json
```diff
@@ -7,6 +7,7 @@
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-rdf-metadata-extract-hydra-pagesize/^4.0.0/components/context.jsonld",
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-rdf-metadata-extract-request-time/^4.0.0/components/context.jsonld",
     "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-rdf-metadata-extract-allow-http-methods/^4.0.0/components/context.jsonld",
+    "https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-rdf-metadata-extract-post-accepted/^1.0.0/components/context.jsonld",
```
rubensworks: Adding actors is a breaking change, so we'll need to do config versioning. See https://github.com/comunica/comunica/tree/master/engines/config-query-sparql/config#config-versioning

packages/actor-query-source-identify-hypermedia-sparql/test/QuerySourceSparql-test.ts
```diff
@@ -852,6 +852,80 @@ describe('QuerySourceSparql', () => {
...
+      it('should perform a direct HTTP POST request when acceptPost doesn\'t include the url-encoded type', async() => {
+        source = getSource([ 'application/sparql-query' ]);
+
+        await expect(source.queryBindings(operationIn, ctx)).toEqualBindingsStream(expectedResult);
+
+        expect(mediatorHttp.mediate).toHaveBeenCalledTimes(2);
+        expect(mediatorHttp.mediate).toHaveBeenCalledWith({
+          context: ctx,
+          init: {
+            body: 'SELECT (COUNT(*) AS ?count) WHERE { undefined:s ?p undefined:o. }',
```
jitsedesmet, line 920: When you check for the body: `expect.any(URLSearchParams)`, this test would also pass right? In that case, I would try to be more specific in what kind of body you expect in the previous test?

jitsedesmet (commented):
Looks nice, thank you Noah!
I have a single comment to increase the test quality.

packages/actor-query-source-identify-hypermedia-sparql/test/QuerySourceSparql-test.ts
noahvsb: It wouldn't pass, since the body is just a string. But it also wouldn't hurt to be more specific in my previous test.

jitsedesmet (approved):
Nice work :partying_face:

packages/actor-query-source-identify-hypermedia-sparql/test/QuerySourceSparql-test.ts
```diff
@@ -852,6 +852,80 @@ describe('QuerySourceSparql', () => {
       });
     });
 
+    describe('acceptPost', () => {
```
rubensworks:
```suggestion
    describe('when acceptPost is defined', () => {
```

packages/actor-rdf-metadata-extract-post-accepted/README.md
```diff
@@ -0,0 +1,39 @@
+# Comunica Post Accepted RDF Metadata Extract Actor
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-rdf-metadata-extract-post-accepted.svg)](https://www.npmjs.com/package/@comunica/actor-rdf-metadata-extract-post-accepted)
+
+A comunica Post Accepted RDF Metadata Extract Actor.
```
rubensworks: Could you make this description more concrete, similar to https://github.com/comunica/comunica/tree/master/packages/actor-rdf-metadata-extract-patch-sparql-update ?

Possibly with a link to https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Post

packages/actor-rdf-metadata-extract-post-accepted/README.md
```diff
@@ -0,0 +1,39 @@
...
+      "@id": "urn:comunica:default:rdf-metadata-extract/actors#post-accepted",
+      "@type": "ActorRdfMetadataExtractPostAccepted"
+    }
+  ]
+}
+```
+
+### Config Parameters
+
+* `postAccepted`: `[ 'application/x-www-form-urlencoded', 'application/sparql-query', ... ]`.
```
rubensworks: These are not config parameters (there are none for this actor, so this section can go), but a metadata entry that is added.
See https://github.com/comunica/comunica/tree/master/packages/actor-rdf-metadata-extract-patch-sparql-update on how this could be described.

packages/actor-rdf-metadata-extract-post-accepted/lib/ActorRdfMetadataExtractPostAccepted.ts
```diff
@@ -0,0 +1,29 @@
...
+    super(args);
+  }
+
+  public async test(_action: IActionRdfMetadataExtract): Promise<TestResult<IActorTest>> {
+    return passTestVoid();
+  }
+
+  public async run(action: IActionRdfMetadataExtract): Promise<IActorRdfMetadataExtractOutput> {
+    const metadata: IActorRdfMetadataExtractOutput['metadata'] = {};
+    if (action.headers?.get('accept-post')) {
```
rubensworks: Let's store the return value of `get()` in a variable.
We don't know how `Headers` is implemented, so if it's slow, we should minimize calls to it.

yarn.lock
```diff
@@ -5985,6 +5985,24 @@ fetch-sparql-endpoint@^6.1.0:
     stream-to-string "^1.0.0"
     yargs "^17.0.0"
 
+fetch-sparql-endpoint@^6.2.0:
```
rubensworks, line 5970: Can you check to ensure all comunica packages with dependencies on `fetch-sparql-endpoint` have been updated to this version?
Because ideally, the existing `fetch-sparql-endpoint` should have been updated in the lockfile.

rubensworks (changes requested):

packages/actor-rdf-metadata-extract-post-accepted/lib/ActorRdfMetadataExtractPostAccepted.ts
noahvsb: I copied this part from the accept-put extract actor, so I'll change it there too.

PR #1614, opened 2025-09-11
Integrate Traqula and Traqula Algebra

jitsedesmet:
This PR aims to integrate the new version of Traqula in Comunica. The new version has a different AST as sparqlJS, allowing for round tripping generation. Additionally, we in lined and modified SPARQLAlgebra.js to take this new AST and generate algebra. This algebra is marginally different from the algebra generated by SPARQLAlgebra.JS (variables in a VALUES class are no longer prefixed using a '?'.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,6 @@
     "typedoc": "^0.27.0",
     "typescript": "^5.3.3",
     "webpack": "^5.0.0",
-    "webpack-bundle-analyzer": "^4.10.2",
     "webpack-cli": "^6.0.0"
   },
   "resolutions": {
--- a/packages/actor-rdf-parse-n3/package.json
+++ b/packages/actor-rdf-parse-n3/package.json
@@ -38,6 +38,6 @@
     "@comunica/bus-rdf-parse": "^4.0.2",
     "@comunica/context-entries": "^4.0.2",
     "@comunica/types": "^4.0.2",
-    "n3": "*"
+    "n3": "^1.17.0"
   }
 }
--- a/packages/actor-rdf-serialize-n3/package.json
+++ b/packages/actor-rdf-serialize-n3/package.json
@@ -38,6 +38,6 @@
   },
   "dependencies": {
     "@comunica/bus-rdf-serialize": "^4.0.2",
-    "n3": "*"
+    "n3": "^1.17.0"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -670,11 +670,6 @@
     enabled "2.0.x"
     kuler "^2.0.0"
 
-"@discoveryjs/json-ext@0.5.7":
-  version "0.5.7"
-  resolved "https://registry.yarnpkg.com/@discoveryjs/json-ext/-/json-ext-0.5.7.tgz#1d572bfbbe14b7704e0ba0f39b74815b84870d70"
-  integrity sha512-dBVuXR082gk3jsFp7Rd/JI4kytwGHecnCoTtXFb7DB6CNHp4rg5k1bhg0nWdLGLnOV71lmDzGQaLMy8iPLY0pw==
-
 "@discoveryjs/json-ext@^0.6.1":
   version "0.6.3"
   resolved "https://registry.yarnpkg.com/@discoveryjs/json-ext/-/json-ext-0.6.3.tgz#f13c7c205915eb91ae54c557f5e92bddd8be0e83"
@@ -1655,11 +1650,6 @@
   integrity sha512-jTXRsoSPONAs8Za9QEQdyjFn+0ZQFjCiIztAIF6bi1HqhBzG9Ma7g1WotyiGqFSBRZjIEqMdT8RUlbk1QVhzCQ==
   dependencies:
     playwright "1.47.2"
-
-"@polka/url@^1.0.0-next.24":
-  version "1.0.0-next.29"
-  resolved "https://registry.yarnpkg.com/@polka/url/-/url-1.0.0-next.29.tgz#5a40109a1ab5f84d6fd8fc928b19f367cbe7e7b1"
-  integrity sha512-wwQAWhWSuHaag8c4q/KN/vCoeOJYshAIvMQwD4GpSb3OiZklFfvAgmj0VCBBImRpuF/aFgIRzllXlVX93Jevww==
 
 "@pollyjs/adapter-node-http@^6.0.5":
   version "6.0.6"
@@ -2989,17 +2979,12 @@ acorn-jsx@^5.3.2:
   resolved "https://registry.npmjs.org/acorn-jsx/-/acorn-jsx-5.3.2.tgz"
   integrity sha512-rq9s+JNhf0IChjtDXxllJ7g41oZk5SlXtp0LHwyA5cejwn7vKmKp4pPri6YEePv2PU65sAsegbXtIinmDFDXgQ==
 
-acorn-walk@^8.0.0, acorn-walk@^8.1.1:
+acorn-walk@^8.1.1:
   version "8.3.4"
   resolved "https://registry.yarnpkg.com/acorn-walk/-/acorn-walk-8.3.4.tgz#794dd169c3977edf4ba4ea47583587c5866236b7"
   integrity sha512-ueEepnujpqee2o5aIYnvHU6C0A42MNdsIDeqy5BydrkuC5R1ZuUFnm27EeFJGoEHJQgn3uleRvmTXaJgfXbt4g==
   dependencies:
     acorn "^8.11.0"
-
-acorn@^8.0.4:
-  version "8.15.0"
-  resolved "https://registry.yarnpkg.com/acorn/-/acorn-8.15.0.tgz#a360898bc415edaac46c8241f6383975b930b816"
-  integrity sha512-NZyJarBfL7nWwIq+FDL6Zp/yHEhePMNnnJ0y3qfieCrmNvYct8uvtiV41UvlSe6apAfk0fY1FbWx+NwfmpvtTg==
 
 acorn@^8.11.0, acorn@^8.4.1:
   version "8.12.1"
@@ -4258,11 +4243,6 @@ commander@^2.20.0:
   resolved "https://registry.npmjs.org/commander/-/commander-2.20.3.tgz"
   integrity sha512-GpVkmM8vF2vQUkj2LvZmD35JxeJOLCwJ9cUkugyk2nuhbv3+mJvpLYYt+0+USMxE+oj+ey/lJEnhZw75x/OMcQ==
 
-commander@^7.2.0:
-  version "7.2.0"
-  resolved "https://registry.yarnpkg.com/commander/-/commander-7.2.0.tgz#a36cb57d0b501ce108e4d20559a150a391d97ab7"
-  integrity sha512-QrWXB+ZQSVPmIWIhtEO9H+gwHaMGYiF5ChvoJ+K9ZGHG/sVsa6yiesAD1GC/x46sET00Xlwo1u49RVVVzvcSkw==
-
 comment-parser@1.4.1:
   version "1.4.1"
   resolved "https://registry.npmjs.org/comment-parser/-/comment-parser-1.4.1.tgz"
@@ -4737,11 +4717,6 @@ dateformat@^3.0.3:
   resolved "https://registry.npmjs.org/dateformat/-/dateformat-3.0.3.tgz"
   integrity sha512-jyCETtSl3VMZMWeRo7iY1FL19ges1t55hMo5yaam4Jrsm5EPL89UQkoQRyiI+Yf4k8r2ZpdngkV8hr1lIdjb3Q==
 
-debounce@^1.2.1:
-  version "1.2.1"
-  resolved "https://registry.yarnpkg.com/debounce/-/debounce-1.2.1.tgz#38881d8f4166a5c5848020c11827b834bcb3e0a5"
-  integrity sha512-XRRe6Glud4rd/ZGQfiV1ruXSfbvfJedlV9Y6zOlP+2K04vBYiJEte6stfFkCP03aMnY5tsipamumUjL14fofug==
-
 debug@2.6.9, debug@^2.2.0, debug@^2.3.3:
   version "2.6.9"
   resolved "https://registry.npmjs.org/debug/-/debug-2.6.9.tgz"
@@ -5103,7 +5078,7 @@ dtrace-provider@~0.8:
   dependencies:
     nan "^2.14.0"
 
-duplexer@^0.1.1, duplexer@^0.1.2:
+duplexer@^0.1.1:
   version "0.1.2"
   resolved "https://registry.npmjs.org/duplexer/-/duplexer-0.1.2.tgz"
   integrity sha512-jtD6YG370ZCIi/9GTaJKQxWTZD045+4R4hTk/x1UyoqadyJ9x9CgSi1RlVDQF8U2sxLLSnFkCaMihqljHIWgMg==
@@ -6841,13 +6816,6 @@ graphql@^15.5.2:
   resolved "https://registry.npmjs.org/graphql/-/graphql-15.8.0.tgz"
   integrity sha512-5gghUc24tP9HRznNpV2+FIoq3xKkj5dTQqf4v0CpdPbFVwFkWoxOM+o+2OC9ZSvjEMTjfmG9QT+gcvggTwW1zw==
 
-gzip-size@^6.0.0:
-  version "6.0.0"
-  resolved "https://registry.yarnpkg.com/gzip-size/-/gzip-size-6.0.0.tgz#065367fd50c239c0671cbcbad5be3e2eeb10e462"
-  integrity sha512-ax7ZYomf6jqPTQ4+XCpUGyXKHk5WweS+e05MBO4/y3WJ5RkmPXNKvX+bx1behVILVwr6JSQvZAku021CHPXG3Q==
-  dependencies:
-    duplexer "^0.1.2"
-
 handlebars@^4.7.7:
   version "4.7.8"
   resolved "https://registry.npmjs.org/handlebars/-/handlebars-4.7.8.tgz"
@@ -7034,7 +7002,7 @@ hosted-git-info@^7.0.0:
   dependencies:
     lru-cache "^10.0.1"
 
-html-escaper@^2.0.0, html-escaper@^2.0.2:
+html-escaper@^2.0.0:
   version "2.0.2"
   resolved "https://registry.npmjs.org/html-escaper/-/html-escaper-2.0.2.tgz"
   integrity sha512-H2iMtd0I4Mt5eYiapRdIDjp+XzelXQ0tFE4JS7YFwFevXXMmOp9myNrUvCg0D6ws8iqkRPBfKHgbwig1SmlLfg==
@@ -9570,11 +9538,6 @@ morgan@^1.10.0:
     on-finished "~2.3.0"
     on-headers "~1.0.2"
 
-mrmime@^2.0.0:
-  version "2.0.1"
-  resolved "https://registry.yarnpkg.com/mrmime/-/mrmime-2.0.1.tgz#bc3e87f7987853a54c9850eeb1f1078cd44adddc"
-  integrity sha512-Y3wQdFg2Va6etvQ5I82yUhGdsKrcYox6p7FfL1LbK2J4V01F9TGlepTIhnK24t7koZibmg82KGglhA1XK5IsLQ==
-
 ms@2.0.0:
   version "2.0.0"
   resolved "https://registry.npmjs.org/ms/-/ms-2.0.0.tgz"
@@ -9630,7 +9593,7 @@ mv@~2:
     ncp "~2.0.0"
     rimraf "~2.4.0"
 
-n3@*, n3@2.0.0-beta.1, n3@^1.0.0, n3@^1.11.1, n3@^1.16.3, n3@^1.17.0, n3@^1.22.3, n3@^1.6.3:
+n3@2.0.0-beta.1, n3@^1.0.0, n3@^1.11.1, n3@^1.16.3, n3@^1.17.0, n3@^1.22.3, n3@^1.6.3:
   version "2.0.0-beta.1"
   resolved "https://registry.yarnpkg.com/n3/-/n3-2.0.0-beta.1.tgz#e966ccd02548cdd58a429c9c8b1e3dbae5982613"
   integrity sha512-mOAlmgDP6XFKSxIL9OEX/wrmoiMyawl40PgXWYtjH6zUtnUKyppqen3JkEZBsmUV68CZp5OaLN+ii5t6A6Zrew==
@@ -10265,11 +10228,6 @@ open@^8.4.0:
     define-lazy-prop "^2.0.0"
     is-docker "^2.1.1"
     is-wsl "^2.2.0"
-
-opener@^1.5.2:
-  version "1.5.2"
-  resolved "https://registry.yarnpkg.com/opener/-/opener-1.5.2.tgz#5d37e1f35077b9dcac4301372271afdeb2a13598"
-  integrity sha512-ur5UIdyw5Y7yEj9wLzhqXiy6GZ3Mwx0yGI+5sMn2r0N0v3cKJvUmFH5yPP+WXh9e0xfyzyJX95D8l088DNFj7A==
 
 optionator@^0.9.3:
   version "0.9.3"
@@ -12118,15 +12076,6 @@ simple-update-notifier@^2.0.0:
   dependencies:
     semver "^7.5.3"
 
-sirv@^2.0.3:
-  version "2.0.4"
-  resolved "https://registry.yarnpkg.com/sirv/-/sirv-2.0.4.tgz#5dd9a725c578e34e449f332703eb2a74e46a29b0"
-  integrity sha512-94Bdh3cC2PKrbgSOUqTiGPWVZeSiXfKOVZNJniWoqrWrRkB1CJzBU3NEbiTsPcYy1lDsANA/THzS+9WBiy5nfQ==
-  dependencies:
-    "@polka/url" "^1.0.0-next.24"
-    mrmime "^2.0.0"
-    totalist "^3.0.0"
-
 sisteransi@^1.0.5:
   version "1.0.5"
   resolved "https://registry.npmjs.org/sisteransi/-/sisteransi-1.0.5.tgz"
@@ -12981,11 +12930,6 @@ toml-eslint-parser@^0.9.0, toml-eslint-parser@^0.9.3:
   dependencies:
     eslint-visitor-keys "^3.0.0"
 
-totalist@^3.0.0:
-  version "3.0.1"
-  resolved "https://registry.yarnpkg.com/totalist/-/totalist-3.0.1.tgz#ba3a3d600c915b1a97872348f79c127475f6acf8"
-  integrity sha512-sf4i37nQ2LBx4m3wB74y+ubopq6W/dIzXg0FDGjsYnZHVa1Da8FH853wlL2gtUhg+xJXjfk3kUZS3BRoQeoQBQ==
-
 touch@^3.1.0:
   version "3.1.0"
   resolved "https://registry.npmjs.org/touch/-/touch-3.1.0.tgz"
@@ -13601,24 +13545,6 @@ webidl-conversions@^3.0.0:
   resolved "https://registry.npmjs.org/webidl-conversions/-/webidl-conversions-3.0.1.tgz"
   integrity sha512-2JAn3z8AR6rjK8Sm8orRC0h/bcl/DqL7tRPdGZ4I1CjdF+EaMLmYxBHyXuKL849eucPFhvBoxMsflfOb8kxaeQ==
 
-webpack-bundle-analyzer@^4.10.2:
-  version "4.10.2"
-  resolved "https://registry.yarnpkg.com/webpack-bundle-analyzer/-/webpack-bundle-analyzer-4.10.2.tgz#633af2862c213730be3dbdf40456db171b60d5bd"
-  integrity sha512-vJptkMm9pk5si4Bv922ZbKLV8UTT4zib4FPgXMhgzUny0bfDDkLXAVQs3ly3fS4/TN9ROFtb0NFrm04UXFE/Vw==
-  dependencies:
-    "@discoveryjs/json-ext" "0.5.7"
-    acorn "^8.0.4"
-    acorn-walk "^8.0.0"
-    commander "^7.2.0"
-    debounce "^1.2.1"
-    escape-string-regexp "^4.0.0"
-    gzip-size "^6.0.0"
-    html-escaper "^2.0.2"
-    opener "^1.5.2"
-    picocolors "^1.0.0"
-    sirv "^2.0.3"
-    ws "^7.3.1"
-
 webpack-cli@^6.0.0:
   version "6.0.1"
   resolved "https://registry.yarnpkg.com/webpack-cli/-/webpack-cli-6.0.1.tgz#a1ce25da5ba077151afd73adfa12e208e5089207"
@@ -13928,11 +13854,6 @@ write-pkg@^3.1.0:
   dependencies:
     sort-keys "^2.0.0"
     write-json-file "^2.2.0"
-
-ws@^7.3.1:
-  version "7.5.10"
-  resolved "https://registry.yarnpkg.com/ws/-/ws-7.5.10.tgz#58b5c20dc281633f6c19113f39b349bd8bd558d9"
-  integrity sha512-+dbF1tHwZpXcbOJdVOkzLDxZP1ailvSxM6ZweXTegylPny803bFhA+vqBYw4s31NSAk4S2Qz+AKXK9a4wkdjcQ==
 
 ws@~8.11.0:
   version "8.11.0"


packages/actor-expression-evaluator-factory-default/lib/AlgebraTransformer.ts
```diff
@@ -13,10 +13,32 @@ export class AlgebraTransformer extends ExprEval.TermTransformer {
...
+              type: Types.EXPRESSION,
+              expressionType: ExpressionTypes.TERM,
+              term: expr.term.predicate,
+            }, {
+              type: Types.EXPRESSION,
+              expressionType: ExpressionTypes.TERM,
+              term: expr.term.object,
+            }],
+          });
+        }
```
jitsedesmet: @rubensworks Does this kind of rewrite make sense to you?
So this issue is that <<( )>> in a filter can contain variables, as such, you can not just act like it is a regular term (since the bindings need to be assigned). I would say that the algebra construction of a Quad within a filter is actually more like an operator (so `triple()`). This transformation here will basically rewrite any occurrence of <<( )>> in a filter to `triple`. I only wonder whether this rewrite should be here and not in the algebra itself? (I think here is fine?)

If you wonder why this is needed, look at the test in QuerySparql-test.ts (note that the test is skipped but does actually work).

rubensworks: Can't this be done directly within Traqula?
I would also suspect this transformation may have to be applied recursively.

jitsedesmet: Recursion is a good question, but I think it should be handled here. I will alter the current test to a recursion test (since the current one is part of the spec).

As we talked about offline, I think Traqula should not do this actually since it would make Traqula too opinionated towards Comunica.

jitsedesmet: Yes, no question, the recursion works: https://github.com/jitsedesmet/comunica/blob/feature/sparql12/packages/actor-expression-evaluator-factory-default/lib/AlgebraTransformer.ts#L67

jitsedesmet: Maybe related is: https://github.com/w3c/sparql-query/issues/284

packages/actor-abstract-path/lib/ActorAbstractPath.ts
```diff
@@ -16,6 +16,8 @@ import type {
 import type { BindingsFactory } from '@comunica/utils-bindings-factory';
 import { assignOperationSource, getOperationSource, getSafeBindings } from '@comunica/utils-query-operation';
 import type * as RDF from '@rdfjs/types';
+import type { Factory } from '@traqula/algebra-transformations-1-2';
+import { Algebra } from '@traqula/algebra-transformations-1-2';
```
rubensworks: Given these version-specific imports, I wonder how compatible everything will be across SPARQL versions.
Let's say Comunica is at version 1.2, and someone makes his own SPARQL extension with Traqula with corresponding algebra package, does he have to replace _all_ of these imports across the Comunica codebase, or will this be compatible?

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
```diff
@@ -10,8 +10,8 @@ import { passTestVoid } from '@comunica/core';
 import type { ComunicaDataFactory, FragmentSelectorShape, IActionContext, IQuerySourceWrapper } from '@comunica/types';
 import { doesShapeAcceptOperation, getOperationSource } from '@comunica/utils-query-operation';
 import type * as RDF from '@rdfjs/types';
+import { Factory, Algebra, utils } from '@traqula/algebra-transformations-1-2';
```
rubensworks: I always found the `Util` name in sparqlalgebrajs a bit annoying, since it's very generic.
Since we're still in the breaking change stage, could we give it a more precise name? E.g. `algebraUtils`?

packages/actor-query-parse-sparql/lib/ActorQueryParseSparql.ts
```diff
@@ -4,17 +4,16 @@ import { KeysInitQuery } from '@comunica/context-entries';
 import type { IActorTest, TestResult } from '@comunica/core';
 import { failTest, passTestVoid } from '@comunica/core';
 import type { ComunicaDataFactory } from '@comunica/types';
-import type * as RDF from '@rdfjs/types';
-import { Parser as SparqlParser } from '@traqula/engine-sparql-1-2';
-import type { DataFactory } from 'rdf-data-factory';
-import { translate } from 'sparqlalgebrajs';
+import { toAlgebra } from '@traqula/algebra-sparql-1-2';
+import { Parser as SparqlParser } from '@traqula/parser-sparql-1-2';
+import { Factory } from '@traqula/rules-sparql-1-2';
```
rubensworks: Similar to my comment above, could we name this `AlgebraFactory` instead?

packages/actor-query-parse-sparql/lib/ActorQueryParseSparql.ts
```diff
@@ -4,17 +4,16 @@ import { KeysInitQuery } from '@comunica/context-entries';
 import type { IActorTest, TestResult } from '@comunica/core';
 import { failTest, passTestVoid } from '@comunica/core';
 import type { ComunicaDataFactory } from '@comunica/types';
-import type * as RDF from '@rdfjs/types';
-import { Parser as SparqlParser } from '@traqula/engine-sparql-1-2';
-import type { DataFactory } from 'rdf-data-factory';
-import { translate } from 'sparqlalgebrajs';
+import { toAlgebra } from '@traqula/algebra-sparql-1-2';
+import { Parser as SparqlParser } from '@traqula/parser-sparql-1-2';
```
rubensworks, line 8: Would it make sense to export this as `SparqlParser` from the package directly?
Or even `QueryParser`?
(this one seems less important than my other renaming suggestions, given that this one probably won't cause name clashes)

packages/actor-query-parse-sparql/package.json
```diff
@@ -44,7 +44,11 @@
     "@comunica/bus-query-parse": "^4.0.2",
     "@comunica/context-entries": "^4.0.2",
     "@comunica/core": "^4.0.2",
-    "@traqula/engine-sparql-1-2": "^0.0.1-alpha.148",
-    "sparqlalgebrajs": "^5.0.1"
+    "@traqula/algebra-sparql-1-2": "^0.0.15",
+    "@traqula/parser-sparql-1-2": "^0.0.15",
+    "@traqula/rules-sparql-1-2": "^0.0.15"
+  },
+  "devDependencies": {
```
rubensworks, line 51: This is incorrect. Type packages must always be actual deps.

packages/actor-query-source-identify-hypermedia-sparql/lib/QuerySourceSparql.ts
```diff
@@ -15,18 +15,21 @@ import type {
 import type { BindingsFactory } from '@comunica/utils-bindings-factory';
 import { MetadataValidationState } from '@comunica/utils-metadata';
 import type * as RDF from '@rdfjs/types';
+import { toAst } from '@traqula/algebra-sparql-1-2';
+import type { Factory, Algebra } from '@traqula/algebra-transformations-1-2';
+import { utils } from '@traqula/algebra-transformations-1-2';
+import { Generator } from '@traqula/generator-sparql-1-2';
```
rubensworks, line 23: `SparqlGenerator` or `QueryGenerator`?

packages/utils-expression-evaluator/lib/expressions/Term.ts
```diff
@@ -428,13 +422,9 @@ export class NonLexicalLiteral extends Literal<{ toString: () => 'undefined' }>
...
 
   public override coerceEBV(): boolean {
-    const isNumericOrBool =
-      isSubTypeOf(this.dataType, TypeURL.XSD_BOOLEAN, this.openWorldType) ||
-      isSubTypeOf(this.dataType, TypeAlias.SPARQL_NUMERIC, this.openWorldType);
-    if (isNumericOrBool) {
-      return false;
-    }
-    throw new Err.EBVCoercionError(this);
+    // Always throws in [SPARQL 1.2](https://www.w3.org/TR/sparql12-query/#ebv),
```
rubensworks, line 425: I'm guessing this can still change, based on the resolution of https://github.com/w3c/sparql-query/issues/286

rubensworks (changes requested):
What a massive PR 🤯
Great work 👍

My main comment relates to the version-specific package names, and how they will be compatible with other SPARQL versions. (which is definitely a situation we will encounter in the future)

packages/actor-abstract-path/lib/ActorAbstractPath.ts
jitsedesmet: The Algebra between 1.1 and 1.2 happen to be the same (I just reexport them in the 1.2 package).
In case the Algebra does not change between the SPARQL extension and our algebra, there will be minimal changes required in Comunica. (only the source would need to be edited)

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
jitsedesmet: During the creation of the API of Traqula I have wondered the many times. As you can see, it also exports just `Factory` (and not `AlgebraFactory`). I think I like this approach since you can always change the import name using `as`. Unless you have a good argument why not to do that.

packages/actor-query-parse-sparql/lib/ActorQueryParseSparql.ts
jitsedesmet: Yes, this is an AST factory actually. The reasoning behind it is that we have these different versions, and all of them have an ASTFactory, So actually we have an ASTSparql11Factory, and a ASTSparql12Factory...
This gets reaallly lengthy. And most users will actually not need to know the versions?

packages/utils-expression-evaluator/lib/expressions/Term.ts
jitsedesmet: Yes, it can, but if they decide to change this, the spec tests will also change. So I would NOT wait for this PR until that one is closed.

packages/actor-query-parse-sparql/package.json
jitsedesmet: Oh, why?

packages/actor-abstract-path/lib/ActorAbstractPath.ts
rubensworks: > there will be minimal changes required in Comunica

But how large will these changes be, exactly?

Could you maybe set up a small prototype experiment where you try adding some fictional operator to SPARQL, and then see what needs to be done exactly to add an actor for it in comunica? (without actually implementing the actor of course)
The goal here would be that we can just add this new actor, without having to change any other code in comunica (besides the parser actor of course).

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
rubensworks: My main concern here is that this approach is inconsistent with all other factories (e.g. DataFactory and BindingsFactory).

packages/utils-expression-evaluator/lib/expressions/Term.ts
rubensworks: I agree.

packages/actor-abstract-path/lib/ActorAbstractPath.ts
jitsedesmet: Adding an operator will not change anything. You alter the parser to accept the operator and you create a FunctionFactory Actor for it in Comunica, you are done.

rubensworks: Ok, great, that's what we want indeed.
Could you try it out in a quick prototype to confirm?
(because we can still do breaking changes now if needed, and will be more difficult later)

jitsedesmet: Sure: Do I just add support for the SPARQL 1.2 adjust buildtin for SPARQL1.2?
How did we say again that we would support a parser for Comunica? I was thinking just add a class in the Sparql actor of Comunica that draws from the Traqula 1.2 Parser?

rubensworks: I would add an actual operator (like union or filter), can be a non-sensical one, just for testing. Operators may have more influence on typings than functions.

We'll need to figure out a good pattern for swapping parsing rules indeed.

Maybe we should even write up a tutorial for the comunica website based corresponding to the steps you will be taking for this prototype.

packages/actor-query-parse-sparql/lib/ActorQueryParseSparql.ts
jitsedesmet: I don't think the renaming here is necessary. Just for the same reason as our Factory will just be called AstFactory and Algebra Factory. There is no mention of SPARQL, nor of 1.1 and 1.2 just because name collisions will occur less often

packages/actor-abstract-path/lib/ActorAbstractPath.ts
jitsedesmet: Discussion continues in #1633

packages/actor-abstract-path/lib/ActorAbstractPath.ts
```diff
@@ -383,30 +386,31 @@ export abstract class ActorAbstractPath extends ActorQueryOperationTypedMediated
...
+    if (Algebra.isKnownOperation(operation, Algebra.Types.LINK) ||
+      Algebra.isKnownOperation(operation, Algebra.Types.NPS)) {
+      const source = getOperationSource(operation);
+      if (!source) {
+        throw new Error(`Could not find a required source on a link path operation`);
       }
+      return [ source ];
     }
+    // TODO: @RT what does this do? Do we throw an error? Or is empty list good?
+    return [];
```
rubensworks: I think we should throw here.

packages/algebra-sparql-comunica/README.md
```diff
@@ -0,0 +1,9 @@
+# Algebra SPARQL Comunica
+
+A wrapper around and algebra exposed by Traqula.
+Comunica assumes a base Algebra and this package represents that base algebra.
+Using a comunica package allows the monorepo to have more independent evolution than Traqula without restricting itself too much.
+
+TODO: write more
```
rubensworks: TODO

packages/algebra-sparql-comunica/README.md
```diff
@@ -0,0 +1,9 @@
+# Algebra SPARQL Comunica
```
rubensworks: Also make sure to follow the same format as other READMEs in comunica.

packages/algebra-sparql-comunica/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@comunica/algebra-sparql-comunica",
+  "version": "4.0.2",
+  "description": "Self maintained Algebra that Comunica views as its base",
```
rubensworks:
```suggestion
  "description": "Query algebra definitions and utilities",
```

packages/algebra-sparql-comunica/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@comunica/algebra-sparql-comunica",
```
rubensworks: What do you think of this package name?
```suggestion
  "name": "@comunica/utils-algebra",
```

packages/utils-algebra/lib/Algebra.ts
```diff
@@ -0,0 +1,97 @@
...
+
+type withMeta = { metadata?: Record<string, unknown> };
+
+// Base types
+export type BaseOperation = Algebra.BaseOperation & withMeta;
+export type BaseExpression = Algebra.BaseExpression & withMeta;
+export type Operation = BaseOperation;
+export type Expression = BaseExpression;
+
+// https://www.typescriptlang.org/docs/handbook/2/conditional-types.html#distributive-conditional-types
```
rubensworks, line 19: Besides the link, could you also explain what happens, and why it's needed?

packages/actor-context-preprocess-query-source-skolemize/lib/utils.ts
```diff
@@ -213,32 +213,28 @@ export function deskolemizeOperation<O extends Algebra.Operation>(
...
-        return {
-          result: Object.assign(factory.createPattern(
-            deskolemizeTermNestedThrowing(dataFactory, op.subject, sourceId),
-            deskolemizeTermNestedThrowing(dataFactory, op.predicate, sourceId),
-            deskolemizeTermNestedThrowing(dataFactory, op.object, sourceId),
-            deskolemizeTermNestedThrowing(dataFactory, op.graph, sourceId),
-          ), { metadata: op.metadata }),
-          recurse: false,
-        };
+    return Algebra.mapOperation<'unsafe', O>(operation, {
```
rubensworks: Not sure what this `'unsafe'` does here.
Not saying you need to change it.
Could you explain it to me here?

packages/actor-expression-evaluator-factory-default/lib/AlgebraTransformer.ts
```diff
@@ -1,41 +1,55 @@
+import { Algebra, AlgebraFactory } from '@comunica/algebra-sparql-comunica';
 import type { MediatorFunctionFactory } from '@comunica/bus-function-factory';
 import { KeysExpressionEvaluator } from '@comunica/context-entries';
 import type { Expression, IActionContext, OperatorExpression } from '@comunica/types';
 import * as ExprEval from '@comunica/utils-expression-evaluator';
-import { Algebra as Alg } from 'sparqlalgebrajs';
 
 export class AlgebraTransformer extends ExprEval.TermTransformer {
+  private readonly AF = new AlgebraFactory();
```
rubensworks, line 8: Random thought: is it (still) cheap to make this factory?
If yes, great.
If not, we should consider moving it into the query context, and creating it only once.

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
```diff
@@ -271,226 +255,197 @@ export class ActorOptimizeQueryOperationFilterPushdown extends ActorOptimizeQuer
...
+      if (!this.variablesIntersect([ operation.variable ], expressionVariables)) {
+        return [ true, factory.createExtend(
+          this.filterPushdown(expression, expressionVariables, operation.input, factory, context)[1],
+          operation.variable,
+          operation.expression,
+        ) ];
+      }
+      return [ false, factory.createFilter(operation, expression) ];
+    }
+    if (Algebra.isKnownOperation(operation, Algebra.Types.FILTER)) {
```
rubensworks: I sometimes see `isKnownOperation`, `isKnownOperationSub`, or `isKnownSub`.
It may become confusing for users to know what to use when.
Do you think it would it be feasible (and does it make sense) to always use the same method name for all variants?

packages/actor-query-process-explain-physical/lib/MemoryPhysicalQueryPlanLogger.ts
```diff
@@ -165,14 +165,10 @@ export class MemoryPhysicalQueryPlanLogger implements IPhysicalQueryPlanLogger {
...
 
-      // eslint-disable-next-line ts/switch-exhaustiveness-check
-      switch (operation.type) {
-        case 'pattern':
-          data.pattern = this.quadToString(operation);
-          break;
-        case 'project':
-          data.variables = operation.variables.map(variable => variable.value);
-          break;
+      if (Algebra.isKnownOperation(operation, Algebra.Types.PATTERN)) {
```
rubensworks: How is `isKnownOperation` implemented exactly?
Just thinking in terms of performance if it is comparable to switch-case performance (which I guess can be possible).

packages/actor-query-source-identify-hypermedia-sparql/lib/QuerySourceSparql.ts
```diff
@@ -312,7 +315,13 @@ export class QuerySourceSparql implements IQuerySource {
...
    */
   public static operationToQuery(operation: Algebra.Operation): string {
-    return toSparql(operation, { sparqlStar: true });
+    // This query source only handles the Known Algebra from @comunica/algebra-sparql-comunica.
+    // It will likely throw when unknown algebra operations are being translated
+    // or the translation will not happen correctly.
+    const ast = toAst(Algebra.asKnown(operation));
+    // 'Minify' pretty printed string
+    return this.queryStringGenerator.generate(ast)
+      .replaceAll(/[\n \t]+/gu, ' ').trim();
```
rubensworks: Not sure if this replaceAll is correct, as it may cause unwanted changes in literals.
Maybe such minification should be done by the generator? As an optional flag?

packages/actor-query-source-identify-hypermedia-sparql/lib/QuerySourceSparql.ts
```diff
@@ -312,7 +315,13 @@ export class QuerySourceSparql implements IQuerySource {
    * @return {string} A query string.
    */
   public static operationToQuery(operation: Algebra.Operation): string {
-    return toSparql(operation, { sparqlStar: true });
+    // This query source only handles the Known Algebra from @comunica/algebra-sparql-comunica.
+    // It will likely throw when unknown algebra operations are being translated
+    // or the translation will not happen correctly.
```
rubensworks, line 328: You can add a TODO here, to delegate this to a bus.

packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
```diff
@@ -170,15 +170,15 @@ export class ActorRdfJoinMultiBind extends ActorRdfJoin<IActorRdfJoinMultiBindTe
...
     let valid = true;
-    Util.recurseOperation(operation, {
-      [Algebra.types.EXTEND](): boolean {
+    Algebra.visitOperation(operation, {
+      [Algebra.Types.EXTEND]: { preVisitor: () => {
         valid = false;
-        return false;
-      },
-      [Algebra.types.GROUP](): boolean {
+        return { shortcut: true };
```
rubensworks, line 176: What does the shortcut mean exactly? Does it mean, don't visit further children, or don't visit anything anymore (including neighbours and parents)?

packages/algebra-sparql-comunica/lib/algebra/helpers.ts
```diff
@@ -0,0 +1,42 @@
...
+):
+  val is Extract<KnownOperation, { type: Type; subType: SubType }> extends object ?
+    Extract<KnownOperation, { type: Type; subType: SubType }> : { type: Type; subType: SubType } {
+  return val.type === type && val.subType === subType;
+}
+
+// ----------------------- manipulators --------------------
+
+type _NeedRef = Algebra.Operation;
+const transformer = new TransformerSubType<KnownOperation>({ shallowKeys: [ 'metadata' ], ignoreKeys: [ 'metadata' ]});
```
rubensworks: What is the difference between `shallowKeys` and `ignoreKeys`?
In any case, I'd add a brief comment to explain what is going on here, and why.

packages/algebra-sparql-comunica/lib/algebra/helpers.ts
```diff
@@ -0,0 +1,42 @@
+import type { Algebra } from '@traqula/algebra-transformations-1-2';
+import { TransformerSubType } from '@traqula/core';
+import type { CloseSingle, OpenSingle, KnownOperation } from './remapping';
+
+export function asKnown<T extends object>(arg: T): CloseSingle<T> {
```
rubensworks: Since these helpers will be used often, can you add jsdoc for them?

tsconfig.json
```diff
@@ -18,8 +18,15 @@
     "inlineSources": true,
     "preserveConstEnums": true,
     "removeComments": false,
-    "sourceMap": true
+    "sourceMap": true,
+    "skipLibCheck": true
```
rubensworks: Can you remove this option again?

tsconfig.json
```diff
@@ -18,8 +18,15 @@
...
   },
+  "references": [
```
rubensworks: Also remove here.

tsconfig.json
```diff
@@ -29,6 +36,7 @@
     "performance/*/lib/**/*"
   ],
   "exclude": [
-    "**/node_modules"
+    "**/node_modules",
+    "../../../PhD/**"
```
rubensworks: Also here.

rubensworks (changes requested):
Really impressive work @jitsedesmet!
Several comments/questions, but nothing major (I suspect).

packages/algebra-sparql-comunica/package.json
jitsedesmet: Ja, way better!

packages/actor-expression-evaluator-factory-default/lib/AlgebraTransformer.ts
jitsedesmet: It is :) Its a normal factory ;)

packages/actor-query-source-identify-hypermedia-sparql/lib/QuerySourceSparql.ts
jitsedesmet: ah yes, good call!

packages/algebra-sparql-comunica/lib/algebra/helpers.ts
jitsedesmet: These keys are explained in traqula (which you do not see since it is local only (or in the pr). But the docs is there.

packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
jitsedesmet: These keys are explained in traqula (which you do not see since it is local only (or in the pr). But the docs is there.

(it does not visist children nor neighbours and parents)

packages/algebra-sparql-comunica/lib/algebra/helpers.ts
rubensworks: Ok, but could you explain why `metadata` are configured like this here? (in a jscomment)

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
jitsedesmet: Now that you have seen the implementation (a single if), do you still think it makes sense to merge them into one function?

packages/actor-context-preprocess-query-source-skolemize/lib/utils.ts
jitsedesmet: Yes, I am actually not 100% sure about this. Traqula will force you to write this most of the time. What the transformer does:
it takes a copy of the current object and then performs the mapping over all children of the copy. This means that the values of your keys have been mapped (to anything, since the transformer cannot know what you map it too (an object can suddenly be mapped to undefined, null, a string, some new object...).

The default transformation(safe) will thus let you map over an object of type `{ [key in keyof OrigObject]: unknow }` instead of mapping over the type of your object (since again, your childeren have been mapped).
The unsafe option will make it so you instead get your original type (at the expense that your current object does actually not match it's type)

packages/actor-optimize-query-operation-filter-pushdown/lib/ActorOptimizeQueryOperationFilterPushdown.ts
rubensworks: Seems fine indeed :)

packages/actor-context-preprocess-query-source-skolemize/lib/utils.ts
rubensworks: If it's going to be necessary in most cases, it might make sense to flip the default value of this, so that `safe` or `strict` can be passed instead.

jitsedesmet: Will change in Traqula

jitsedesmet: Actually... Thinking about this some more, I think changing this is actually not easy since you still want to have that second argument which decides what the output type is. Furthermore, changing the order of this generic types does not make since because the second argument can only be used when using the unsafe mode

packages/actor-query-source-identify-hypermedia-sparql/lib/QuerySourceSparql.ts
jitsedesmet: I will remove the regex and make a new PR when Traqula has a non-pretty print mode

jitsedesmet: (Comunica should still be able to work with the pretty print since it is still a correct query, it will just be a longer string)

packages/actor-context-preprocess-query-source-skolemize/lib/utils.ts
rubensworks: Would it work if we'd have a `mapOperationStrict` and `mapOperation` (where the latter is non-strict, corresponding to what Comunica would use)?

packages/actor-query-source-identify-hypermedia-sparql/lib/QuerySourceSparql.ts
rubensworks: Works for me 👍

packages/utils-algebra/README.md
```diff
@@ -0,0 +1,25 @@
+# Comunica Algebra
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Futils-algebra.svg)](https://www.npmjs.com/package/@comunica/utils-algebra)
+
+Exposes the algebra used by Comunica.
```
rubensworks, line 5: Could you follow the same format as other utils packages?

Being:
- A short explanation of that the package does.
- A copy of the paragraph "This module is part of.."
- A copy of the line "Click here if you just want to query with Comunica..."

After that, you can add additional sections with more details if needed.

packages/utils-algebra/lib/utils.ts
```diff
@@ -0,0 +1,84 @@
...
+export const inScopeVariables = algebraUtils.inScopeVariables;
+
+/**
+ * Type guard that checks if an operation is of a certain type known by Comunica.
+ * In case the type matches one known by Comunica,
+ * the type guard will conclude the operation contains all member Comunica expects from this operation-type
+ * @param val
+ * @param type
+ */
+export function isKnownOperation<Type extends KnownOperation['type']>(val: Operation, type: Type):
```
rubensworks: I had another look at this code, and I still think there are opportunities here for simplification.
Because these functions will be confusing for users to determine which one they need to use.
So this comment is mainly for the purpose of improving developer experience.

What is the difference between type and subType for example?
This should be clear here in the signature and/or docs.

Also, I think it's possible to merge `isKnownOperationSub ` into `isKnownOperation` using an optional param.

Also, are there cases outside of expressions where `isKnownSub` is used?
If not, could we rename it to something like `isKnownExpression`?
Or even better: is there a way to merge it into `isKnownOperation` by for example checking if the given operation is of type expression?

jitsedesmet: The subType is currently only used by expressions. Using an optional parameter might be a worthwhile investigation. And since 'isKnownSub' is indeed not used a lot we could drop it for that single isKnownOperation function.

I think isKnownSub used a few times in succession, but a lambda could be introduced there to fix the Algebra.Types.Expression

rubensworks: > I think isKnownSub used a few times in succession, but a lambda could be introduced there to fix the Algebra.Types.Expression

If no other abstraction is possible, I think a helper function here is still fine.
As long as it's clear for developers what the "main" and most commonly used one is.

package.json
```diff
@@ -92,6 +92,7 @@
     "typedoc": "^0.27.0",
     "typescript": "^5.3.3",
     "webpack": "^5.0.0",
+    "webpack-bundle-analyzer": "^4.10.2",
```
rubensworks: This one can be removed again.

packages/actor-rdf-parse-n3/package.json
```diff
@@ -38,6 +38,6 @@
     "@comunica/bus-rdf-parse": "^4.0.2",
     "@comunica/context-entries": "^4.0.2",
     "@comunica/types": "^4.0.2",
-    "n3": "^1.17.0"
+    "n3": "*"
```
rubensworks: Can you undo this change?
We should never have wildcard version ranges.

packages/actor-rdf-serialize-n3/package.json
```diff
@@ -38,6 +38,6 @@
   },
   "dependencies": {
     "@comunica/bus-rdf-serialize": "^4.0.2",
-    "n3": "^1.17.0"
+    "n3": "*"
```
rubensworks: Also here.

rubensworks (changes requested):
3 minor comments, then we should be good to go! :-)

rubensworks (approved):
Really excellent work @jitsedesmet!
This is a big one, and will have major impact!

Issue #1537, opened 2025-04-15
Rewrite SERVICE clauses as query operation with annotated source


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

Right now, we have a dedicated actor for handling SERVICE clauses, while we could live without it.
Instead, we could add an optimize actor that rewrites SERVICE operations into operations with that source as annotation.

This would allow more optimizations to take place if explicit SERVICE clauses are mentioned in the query, such as the Multi Bind Source join actor, as well as several optimize actors.

We should pay attention to nested service clauses, so that they still end up in their original form to the serviced source.

Related to this, we may also want to allow zero sources to be passed to comunica if all data-producing operators (triple patterns, property paths) are either annotated with a source, or part of an operation annotated with a source.

rubensworks:
This has been implemented for Comunica v5.

PR #1616, opened 2025-09-11
Cache identified sources in service operation actor

surilindur:
This is a small optimisation from the TRIPLE project use case, where Comunica was repeatedly requesting metadata for the same SPARQL endpoint, due to SERVICE clauses being evaluated in a bind join.

The actual solution, as discussed offline, is #1537, but this could be a solution until that one is implemented.

Any feedback would be welcome. :slightly_smiling_face:

### DIFF
--- a/packages/actor-query-operation-service/lib/ActorQueryOperationService.ts
+++ b/packages/actor-query-operation-service/lib/ActorQueryOperationService.ts
@@ -131,6 +131,7 @@ export interface IActorQueryOperationServiceArgs extends IActorQueryOperationTyp
   /**
    * An actor that listens to HTTP invalidation events
    * @default {<default_invalidator> a <npmd:@comunica/bus-http-invalidate/^4.0.0/components/ActorHttpInvalidateListenable.jsonld#ActorHttpInvalidateListenable>}
+   * TODO: make required in next-major.
    */
   httpInvalidator?: ActorHttpInvalidateListenable;
   /* eslint-enable max-len */


packages/actor-query-operation-service/lib/ActorQueryOperationService.ts
```diff
@@ -98,4 +127,16 @@ export interface IActorQueryOperationServiceArgs extends IActorQueryOperationTyp
    * The mediator for identifying query sources.
    */
   mediatorQuerySourceIdentify: MediatorQuerySourceIdentify;
+  /* eslint-disable max-len */
+  /**
+   * An actor that listens to HTTP invalidation events
+   * @default {<default_invalidator> a <npmd:@comunica/bus-http-invalidate/^4.0.0/components/ActorHttpInvalidateListenable.jsonld#ActorHttpInvalidateListenable>}
+   */
+  httpInvalidator: ActorHttpInvalidateListenable;
```
rubensworks: While this should normally be mandatory indeed, we should make it optional here for backwards-compat.

surilindur: Should it be optional even with the default value declared? There are no config changes required, because of the default value declaration. The only issues would arise if someone instantiates the actor manually in code, which probably never happens. :thinking:

rubensworks: It could cause issues when someone has a precompiled config (so a JS file), bumps comunica, and does not recompile the config.

packages/actor-query-operation-service/lib/ActorQueryOperationService.ts
```diff
@@ -98,4 +127,16 @@ export interface IActorQueryOperationServiceArgs extends IActorQueryOperationTyp
    * The mediator for identifying query sources.
    */
   mediatorQuerySourceIdentify: MediatorQuerySourceIdentify;
+  /* eslint-disable max-len */
+  /**
+   * An actor that listens to HTTP invalidation events
+   * @default {<default_invalidator> a <npmd:@comunica/bus-http-invalidate/^4.0.0/components/ActorHttpInvalidateListenable.jsonld#ActorHttpInvalidateListenable>}
+   */
```
rubensworks, line 135:
```suggestion
   * TODO: make required in next-major.
   */
```

PR #1617, opened 2025-09-11
Prevent `ActorIteratorTransformRecordIntermediateResults` from crashing when `StatisticIntermediateResults` is missing in context

constraintAutomaton:
When running link traversal actors (such as the type index), some internal queries are executed without including statistic tracking in their context. This makes sense, since we’re not interested in collecting intermediary results for these internal queries. However, the absence of `StatisticIntermediateResults` in the context currently causes the engine to crash.

This PR updates `ActorIteratorTransformRecordIntermediateResults` to gracefully handle the absence of `StatisticIntermediateResults` by simply passing through the stream unchanged.

Another option could also be to make the `test` fail, however I am not sure if it would be a good idea if it is the only actor in a bus.

@RubenEschauzier

### DIFF
--- a/packages/actor-iterator-transform-record-intermediate-results/lib/ActorIteratorTransformRecordIntermediateResults.ts
+++ b/packages/actor-iterator-transform-record-intermediate-results/lib/ActorIteratorTransformRecordIntermediateResults.ts
@@ -66,10 +66,7 @@ export class ActorIteratorTransformRecordIntermediateResults extends ActorIterat
   public async testIteratorTransform(
     action: ActionIteratorTransform,
   ): Promise<TestResult<IActorTest>> {
-    const statisticIntermediateResults = <StatisticIntermediateResults>(
-      action.context.get(KeysStatistics.intermediateResults)
-    );
-    if (!statisticIntermediateResults) {
+    if (!action.context.has(KeysStatistics.intermediateResults)) {
       return failTest(
         `Missing required context value: ${KeysStatistics.intermediateResults.name}. It must be defined before running ${this.name}.`,
       );


packages/actor-iterator-transform-record-intermediate-results/lib/ActorIteratorTransformRecordIntermediateResults.ts
```diff
@@ -23,48 +23,57 @@ export class ActorIteratorTransformRecordIntermediateResults extends ActorIterat
...
-    const output = action.stream.map((data) => {
-      statisticIntermediateResults.updateStatistic(
-        {
+  public async transformIteratorQuads(
+    action: IActionIteratorTransformQuads,
+  ): Promise<ITransformIteratorOutput<AsyncIterator<RDF.Quad>, MetadataQuads>> {
+    const statisticIntermediateResults = <StatisticIntermediateResults>(
+      action.context.get(KeysStatistics.intermediateResults)
+    );
+    if (statisticIntermediateResults !== undefined) {
```
RubenEschauzier:
```suggestion
    if (statisticIntermediateResults) {
```
Here also

packages/actor-iterator-transform-record-intermediate-results/lib/ActorIteratorTransformRecordIntermediateResults.ts
```diff
@@ -23,48 +23,57 @@ export class ActorIteratorTransformRecordIntermediateResults extends ActorIterat
...
-        {
+  public async transformIteratorBindings(
+    action: IActionIteratorTransformBindings,
+  ): Promise<
+    ITransformIteratorOutput<AsyncIterator<RDF.Bindings>, MetadataBindings>
+  > {
+    const statisticIntermediateResults = <StatisticIntermediateResults>(
+      action.context.get(KeysStatistics.intermediateResults)
+    );
+    if (statisticIntermediateResults !== undefined) {
```
RubenEschauzier:
```suggestion
    if (statisticIntermediateResults) {
```
This style seems more common in Comunica, and it should be safe here since statisticIntermediateResults is always either undefined or an object. I don't have a strong opinion though, so either way is fine for me.

rubensworks: +1

constraintAutomaton:
I've made the test method based implementation from my testing the engine does not crash in LTQP.

packages/actor-iterator-transform-record-intermediate-results/lib/ActorIteratorTransformRecordIntermediateResults.ts
```diff
@@ -63,8 +81,9 @@ export class ActorIteratorTransformRecordIntermediateResults extends ActorIterat
     return { stream: output, metadata: action.metadata };
   }
 
-  public async testIteratorTransform(_action: ActionIteratorTransform):
-  Promise<TestResult<IActorTest>> {
+  public async testIteratorTransform(
```
RubenEschauzier, line 66: The test function of the base class https://github.com/comunica/comunica/blob/5f9a196f83170081f51a01d329e3ee67e3a1698e/packages/bus-iterator-transform/lib/ActorIteratorTransform.ts#L61 returns the result of the `testIteratorTransform` function as the test result if the actor can wrap a certain operation. So, I think you could move the logic in the `test` function above to this `testIteratorTransform` function.

constraintAutomaton: I am confused why this method exist. Why not "just" overloading the test method (like I did)? But, if we keep this pattern, I understand what you mean.

RubenEschauzier: The idea is similar to the `rdf-join` bus: it separates common tests (e.g., verifying whether the actor is configured to wrap a given stream) from actor-specific tests. This way, you don’t need to call super.test() in overrides, and all custom test logic lives in the abstract function.

constraintAutomaton: I saw the idea, I am not sure if it is better, but I will conform to it.

packages/actor-iterator-transform-record-intermediate-results/lib/ActorIteratorTransformRecordIntermediateResults.ts
```diff
@@ -63,8 +63,17 @@ export class ActorIteratorTransformRecordIntermediateResults extends ActorIterat
...
+    action: ActionIteratorTransform,
+  ): Promise<TestResult<IActorTest>> {
+    const statisticIntermediateResults = <StatisticIntermediateResults>(
+      action.context.get(KeysStatistics.intermediateResults)
+    );
+    if (!statisticIntermediateResults) {
```
rubensworks: You can just do `context.has`

PR #1620, opened 2025-09-15
made the update query check sufficient for the sparql endpoint with void description

noahvsb:


rubensworks:
Nice, thanks @noahvsb!

PR #1621, opened 2025-09-15
fixed some small things in the void description output

noahvsb:


packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -550,13 +550,13 @@ export class HttpServiceSparqlEndpoint {
       return;
     }
 
-    const s = request.url;
+    const s = `http://localhost:${this.port}${request.url}`;
```
rubensworks: Comunica will not always necessarily run on localhost.
There should be another property in `request` you can use. I don't remember which one though.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -550,13 +550,13 @@ export class HttpServiceSparqlEndpoint {
       return;
     }
 
-    const s = request.url;
+    const s = `http://${request.headers.host}${request.url}`;
```
rubensworks: Can you detect http/https from the request as well?
If not, we should add this to the config file.

packages/actor-init-query/lib/VoidMetadataEmitter.ts
```diff
@@ -86,6 +86,32 @@ export class VoidMetadataEmitter {
     return quads;
   }
 
+  private convertValue(
+    key: string,
+    value: string,
+  ): string {
+    const xsd = 'http://www.w3.org/2001/XMLSchema#';
+    const stringLiterals = new Set([ 'alternative', 'description', 'title' ]);
+    const dateLiterals = new Set([
```
rubensworks: Let's make these static final in the class, to avoid re-creating them every call.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -550,13 +550,14 @@ export class HttpServiceSparqlEndpoint {
       return;
     }
 
-    const s = request.url;
+    const protocol = request.socket && 'encrypted' in request.socket && request.socket.encrypted ? 'https' : 'http';
```
rubensworks: I'm sorry I didn't notice this before, but it should actually be safe to just use relative URLs, and not include the base URL, as RDF parsers can determine these on their own.
But I don't remember how the rest of this logic here was implemented, so you definitely need to double-check this.
You can do this by setting up a comunica sparql endpoint locally, and then querying over that endpoint (either via the comunica cli, or using https://query.comunica.dev/).

noahvsb: I doubled checked and it failed with the message: 'Failed to fetch'. But doing it with the full URLs also fails.
I'm doing [this](https://query.comunica.dev/#datasources=http%3A%2F%2Flocalhost%3A3000%2Fsparql&query=SELECT%20*%0AWHERE%20%7B%20%3Fs%20%3Fp%20%3Fo%20%7D&executeOnLoad=true) currently.

noahvsb: When navigating to http://localhost:3000/sparql manually and doing the query manually with http://localhost:3000/sparql?query=SELECT%20*%0AWHERE%20%7B%20%3Fs%20%3Fp%20%3Fo%20%7D, everything works.

noahvsb: Oh wait, I'm stupid, ofcourse query.comunica.dev can't reach my localhost, I'll try using the CLI.

noahvsb: I've confirmed that with and without full URLs the output is the same.

rubensworks: Ok, thanks for checking. Then we can remove the absolute URL, and keep it relative to keep the code simpler.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -555,8 +555,8 @@ export class HttpServiceSparqlEndpoint {
     const quads: RDF.Quad[] = [
       // Basic metadata
       quad(s, 'http://www.w3.org/1999/02/22-rdf-syntax-ns#type', `${sd}Service`),
-      quad(s, `${sd}endpoint`, '/sparql'),
-      quad(s, `${sd}url`, '/sparql'),
+      quad(s, `${sd}endpoint`, s),
```
rubensworks: I believe these should remain `/sparql`. Or does something fail without it?

packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -1517,6 +1518,9 @@ describe('HttpServiceSparqlEndpoint', () => {
       });
 
       it('should handle errors in service description stringification', async() => {
+        // Make this a https request for coverage
+        request.socket = { encrypted: true };
```
rubensworks: Not needed anymore?

packages/actor-init-query/test/VoidMetadataEmitter-test.ts
```diff
@@ -167,6 +168,9 @@ describe('VoidMetadataEmitter', () => {
   });
 
   it('should only query statistics once when doing multiple VoID description requests', async() => {
+    // Make this a https request for coverage
+    request.socket = { encrypted: true };
```
rubensworks: Also not needed anymore?

PR #1622, opened 2025-09-16
#1515: should not pushdown when comunica supports extension function, but endpoint doesn't

noahvsb:
#1515


rubensworks (approved):
This is very useful work @noahvsb, thanks for this! 🚀

Issue #930, opened 2022-02-14
Expose ESM modules


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

Currently, Comunica only exposes CommonJS modules for Node.js.
In the future, we should expose both ESM and CJS modules, as this will make it easier to use Comunica in other environments, like the browser.

Some pointers:
- https://2ality.com/2021/06/typescript-esm-nodejs.html
- https://www.typescriptlang.org/docs/handbook/2/modules.html
- https://github.com/RubenVerborgh/AsyncIterator/blob/main/package.json

rubensworks:
Maybe dual packaging is not such a good idea after all.
Node documents [dual packaging hazard](https://nodejs.org/api/packages.html#packages_dual_package_hazard), so we may not want to do this.

Might be better to fully transition to ESM, or until we do, stick with pure CJS.

Another alternative could be to just publish ESM and CJS under different package names, but this would complicate publishing a lot.

jeswr:
As discussed with @rubensworks I will work on this issue.

jeswr:
I will now wait for #1377, #1398 and #1400 land on main; and then continue.

rubensworks:
Hi @jeswr 👋, just checking in to see what the status is on this bounty.

rubensworks:
Ok, no worries @jeswr, thanks for the update.

I'm going to unassigned you from this issue for now.
But in case you'd like to continue with it any time in the future, please let me know.
Until then, anyone else is free to work on this bounty.

rubensworks:
Hi @Pixel998, could you contact me at ruben.taelman@ugent.be?
Because this one will take a lot of work, and requires deep understanding of the codebase.

jacoscaz:
Having recently done some work on this topic and just in case it may help others, it is possible to use Webpack (and likely other bundlers) to produce an ESM bundle for any given configuration of Comunica: https://github.com/quadstorejs/quadstore-comunica/blob/a942cb89d05d7191f2edce6c83a130e32779f7a2/engine/webpack.config.js

jeswr:
I was just cleaning up my open PRs, I have the following still open in relation to this issue that I suspect you'll want to take over:
 - https://github.com/comunica/asyncjoin/pull/22
 - https://github.com/LinkedSoftwareDependencies/Components.js/pull/145

jeswr:
*haven't read through full message - just the bit I was pinged on*

[...]

PR #1624, opened 2025-09-18
Update rdfa-streaming-parser to v3.0.1

csarven:
https://github.com/rubensworks/rdfa-streaming-parser.js/issues/49#issuecomment-3305527990

### DIFF
--- a/packages/actor-rdf-parse-html-rdfa/package.json
+++ b/packages/actor-rdf-parse-html-rdfa/package.json
@@ -45,6 +45,6 @@
     "@comunica/context-entries": "^4.4.0",
     "@comunica/core": "^4.4.0",
     "@comunica/types": "^4.4.0",
-    "rdfa-streaming-parser": "^2.0.1"
+    "rdfa-streaming-parser": "^3.0.1"
   }
 }


rubensworks:
We expect to release Comunica v5 shortly after SPARQL 1.2 becomes a candidate recommendation. This will probably be somewhere in the next couple of months.

elf-pavlik:
Would it make sense to add an `*` in this table warning that until RDFa parser dependency gets updated one shouldn't use `rdf-dereference` if they want to support RDFa content? https://github.com/rubensworks/rdf-dereference.js?tab=readme-ov-file#rdf-dereference

elf-pavlik:
Whatever temporary solution could be made available I think it would still be helpful to add a

>[!WARNING]
> what to do to get RDFa working before Comunica v5 lands

in `rdf-dereference` README. And possibly other packages that developers are using.

Currently I'm already in the middle or major detour trying to improve ESM support across CSS, LDO, sai-js, hopefully #930 and eventually componentsjs. With that in mind I will be very happy to use provided solution but can't really allocate any time to help make it available to everyone using comunica and its dependents to parse RDFa.

rubensworks:
Comunica v5 has been released (new rdf-parse release is also available), which includes this update.

PR #1626, opened 2025-09-19
Add support for sd:propertyFeature in SPARQL endpoints

surilindur:
Here are the changes needed to support propertyFeatures from SPARQL SD, as they are implemented in the IDSM endpoint. This allows the first part of the TRIPLE project use-case (Solid pod + IDSM + ChEBI) to execute with all triples at the top level in the query.

The changes are:
1. Adding the parsing of `sd:propertyFeature` to the SPARQL service metadata extractor.
2. Forwarding the property features to the SPARQL source actor.
3. Changing the cardinality estimation functionality and the ASK query implementation in the SPARQL source to assume the source contains results when property features are present both in the query and the endpoint. Otherwise, the engine would prune the sources as empty during query planning.

I also restructured the SPARQL source unit tests, to make sure they test the functions independently, and only test the SPARQL actor itself, so that issues can be better isolated when tests fail. The system / integration tests will catch issues outside the scope of the unit tests.

Any feedback would be welcome. Thank you!

### DIFF
--- a/packages/actor-query-source-identify-hypermedia-sparql/test/QuerySourceSparql-test.ts
+++ b/packages/actor-query-source-identify-hypermedia-sparql/test/QuerySourceSparql-test.ts
@@ -1794,7 +1794,7 @@ WHERE { undefined:s ?p undefined:o. }` }),
         { propertyFeatures: [ propertyFeature.value ]},
       );
       const operation = AF.createPath(DF.variable('s'), AF.createLink(DF.namedNode('.ex:p')), DF.variable('params'));
-      expect(source.operationUsesPropertyFeatures(operation)).toBeTruthy();
+      expect(source.operationUsesPropertyFeatures(operation)).toBeFalsy();
     });
 
     it('should return true if source and query share property features in property path nps', () => {
@@ -1836,7 +1836,7 @@ WHERE { undefined:s ?p undefined:o. }` }),
         { propertyFeatures: [ propertyFeature.value ]},
       );
       const operation = AF.createPath(DF.variable('s'), AF.createNps([ DF.namedNode('ex:p') ]), DF.variable('params'));
-      expect(source.operationUsesPropertyFeatures(operation)).toBeTruthy();
+      expect(source.operationUsesPropertyFeatures(operation)).toBeFalsy();
     });
   });
 });


packages/actor-query-source-identify-hypermedia-sparql/lib/ActorQuerySourceIdentifyHypermediaSparql.ts
```diff
@@ -68,11 +68,7 @@ export class ActorQuerySourceIdentifyHypermediaSparql extends ActorQuerySourceId
       this.cardinalityCountQueries && !isSingularSource,
       this.cardinalityEstimateConstruction,
       this.forceGetIfUrlLengthBelow,
-      action.metadata.defaultGraph,
-      action.metadata.unionDefaultGraph,
-      action.metadata.datasets,
-      action.metadata.extensionFunctions,
-      action.metadata.postAccepted,
+      action.metadata,
```
rubensworks, line 71: Good one, this will be easier to maintain indeed :-)

packages/actor-query-source-identify-hypermedia-sparql/lib/QuerySourceSparql.ts
```diff
@@ -354,6 +367,37 @@ export class QuerySourceSparql implements IQuerySource {
...
+   * @returns {boolean} Whether the operation makes use of property features.
+   */
+  public operationUsesPropertyFeatures(query: Algebra.Operation, operation: Algebra.Operation): boolean {
+    if (this.propertyFeatures) {
+      // Collect all the variables associated with propertyFeatures from the full query
+      const propertyFeatureVariables = new Set<string>();
+      Util.recurseOperation(query, {
+        [Algebra.types.PATTERN]: (op) => {
+          if (op.predicate.termType === 'NamedNode' && this.propertyFeatures!.has(op.predicate.value)) {
+            if (op.subject.termType === 'Variable') {
```
rubensworks: If I understand correctly, this ensures that the method only returns true if the actually selected variables overlap with the variables included in triple patterns with property features.
I wonder if we really need to be so precise?
Can't we just check for the presence of property features as predicates? And ignore checking of variable overlap?

In either case, we are only checking for triple patterns here (PATTERN).
I think we will also need to check for property paths (NPS and LINK I suspect). (unless the SD spec explicitly excludes these from propertyFeatures of course)

surilindur:
I will revert the unit tests to the old format, but the issue with that is it tests a lot of stuff unrelated to the SPARQL source, such as the `fetch-sparql-endpoint` library and the SPARQL results parsers. So they are not really unit tests in that sense, in the way I always thought of unit tests, because they do not test an individual function of the SPARQL source, for example. They also rely on chaining multiple functions from the SPARQL source for the test to pass.

There were some issues with some 'expect toHaveBeenCalledTimes' not matching after my code additions, but I will look into figuring out why again. It is slightly frustrating to figure it out when the calls take place across multiple functions in a single unit test. :frowning_face:

packages/actor-query-source-identify-hypermedia-sparql/lib/QuerySourceSparql.ts
surilindur: Just checking the properties was also my first idea, but it did not work. The issue is that the property features are used like this (at least in IDSM in the TRIPLE use-case queries), where `?resultValue` is projected out, and `ex:propertyFeature` is the only declared property feature in the SPARQL SD:
```
[
    ex:resultKey ?resultValue
] ex:propertyFeature [
    ex:parameterA "value a" ;
    ex:parameterB ex:optionB
]
```
So to actually check if an individual pattern is involved with a property feature, it has to check the surroundings, too, from the original query, to get the necessary context/connections. Otherwise `?var0 ex:parameterA "value a"` on its own would appear to not use property features. At least that was the idea I was going for.

I will look into adding property paths, as well. The [spec](https://www.w3.org/TR/sparql11-service-description/#sd-propertyFeature) does not really say a lot about property features:

> Relates an instance of `sd:Service` to a resource representing an implemented feature that extends the SPARQL Query or Update language and that is accessed by using the named property.

The IDSM endpoint declares them as `sd:Feature`, which makes sense, looking at the list of options.

rubensworks: Ok, I understand the issue now.

But I wonder if the IDSM SD is correct in this case.
Shouldn't `ex:parameterA` and `ex:parameterB` also be declared as property feature?
Because they clearly also have special meaning.

In either case, we can go forward with your current implementation.
But if you also believe the IDSM SD is lacking, we should contact them, and ask for changes.
Also, we should then add a TODO in our code to remove the workaround once the IDSM SD is amended.

surilindur: I updated the implementation to remove this variable hop workaround. When the IDSM service description is updated, I will test it and then request a new review.

packages/actor-query-source-identify-hypermedia-sparql/lib/QuerySourceSparql.ts
```diff
@@ -354,6 +361,42 @@ export class QuerySourceSparql implements IQuerySource {
...
+              this.propertyFeatures!.has(op.predicate.iri.value)
+            ) ||
+            (
+              op.predicate.type === Algebra.types.NPS &&
+              op.predicate.iris.some(iri => this.propertyFeatures!.has(iri.value))
+            )
+          ) {
+            propertyFeaturesUsed = true;
+          }
+          return false;
```
rubensworks, line 390: I think you'll need to enable recursion calls here, because paths can be nested.
You can return early if `propertyFeaturesUsed` is already true of course.

surilindur: Oh, right. :sweat_smile: Thank you for catching that!

surilindur: This should be fixed now, the recurse function seems to go through all the paths automatically, so I just made it check for link and negated property set.

PR #1629, opened 2025-09-23
Add federated Wikidata queries to web benchmark

surilindur:
This is the version of the VALUES clause expansion actor that I used for the Wikidata experiments. It works around some source assignment issues by filling in the bindings into the operations directly, allowing for VALUES bindings without any matches at the source to be pruned during query optimisation, as well.

I will need to check the variable scoping, to make sure nothing gets substituted that is meant to be projected out of the query, now that I think of it... :sweat_smile: 

The idea is open for comments, though.


rubensworks:
I only quickly skimmed this PR.
Before I go through it in more detail, are you sure this kind of transformation is correct?
Because it looks like you are converting VALUES clauses (which can contain disjunctions of solution mappings) into a conjunction of _all_ these solution mappings.
So if I understand correctly, these transformed queries will not be isomorphic.

rubensworks:
Thanks!

PR #1634, opened 2025-09-26
Fix LIMIT not terminating the traversal part of query execution during LTQP

RubenEschauzier:
When performing link traversal with a LIMIT query, reaching the LIMIT caused the `MediatedLinkedRdfSourcesAsyncRdfIterator` to flag itself as `wasForcefullyClosed`; however, this was not sufficient to stop the actual traversal to new links. This PR changes the logic of when a new iterator (link) can be started, thus fixing this problem.


RubenEschauzier:
Right, forgot to look at the tests. I will make sure they're updated.

PR #1641, opened 2025-10-28
Docs: add more docs to utils algebra

jitsedesmet:


packages/utils-algebra/lib/utils.ts
```diff
@@ -78,18 +79,66 @@ const transformer = new TransformerSubTyped<KnownOperation>({
...
 });
 
+/**
+ * Transform a single operation.
+ * The transformation calls the preVisitor starting from the startObject.
+ * The preVisitor can dictate whether transformation should be stopped.
+ * Note that stopping the transformation also prevets further copying.
+ * The transformer itself transforms operations starting with the deepest one that can be visited.
+ * The transformer callback is performed on a copy of the original.
+ * @param startObject
```
rubensworks: `startObject` and `nodeCallBacks` lack a description, also below.

Also, does this method have a `void` result? Or does it return something else?
In the latter case, we need to document what exactly it returns.

packages/utils-algebra/lib/utils.ts
```diff
@@ -78,18 +79,66 @@ const transformer = new TransformerSubTyped<KnownOperation>({
   [Types.COPY]: { ignoreKeys: new Set([ 'source', 'destination', 'metadata' ]) },
 });
 
+/**
+ * Transform a single operation.
+ * The transformation calls the preVisitor starting from the startObject.
+ * The preVisitor can dictate whether transformation should be stopped.
+ * Note that stopping the transformation also prevets further copying.
```
rubensworks:
```suggestion
 * Note that stopping the transformation also prevents further copying.
```

packages/utils-algebra/lib/utils.ts
```diff
@@ -78,18 +79,66 @@ const transformer = new TransformerSubTyped<KnownOperation>({
   [Types.COPY]: { ignoreKeys: new Set([ 'source', 'destination', 'metadata' ]) },
 });
 
+/**
+ * Transform a single operation.
+ * The transformation calls the preVisitor starting from the startObject.
+ * The preVisitor can dictate whether transformation should be stopped.
```
rubensworks: How does one configure this?

packages/utils-algebra/lib/utils.ts
```diff
@@ -78,18 +79,66 @@ const transformer = new TransformerSubTyped<KnownOperation>({
...
+ * Note that stopping the transformation also prevets further copying.
+ * The transformer itself transforms operations starting with the deepest one that can be visited.
+ * The transformer callback is performed on a copy of the original.
+ * @param startObject
+ * @param nodeCallBacks
+ */
 export const mapOperation: (typeof transformer.transformNode<'unsafe', Operation>) = <any>
```
rubensworks: Given the importance of this function, I'd consider including a small example in the jsdocs.

packages/utils-algebra/lib/utils.ts
```diff
@@ -78,18 +79,66 @@ const transformer = new TransformerSubTyped<KnownOperation>({
...
   transformer.transformNodeSpecific.bind(transformer);
+/**
+ * Shares the functionality and first two arguments with {@link mapOperation}.
+ * The third argument allows you to also transform based on the subType of operations.
+ * Note that when a callback for the subtype is provided, the callback for the general type is NOT executed.
+ */
 export const mapOperationSubStrict = transformer.transformNodeSpecific.bind(transformer);
 
+/**
+ * Similar to {@link mapOperation}, but without copying the startObject.
```
rubensworks: While it's ok to say a function is similar, I'd not use it as a reason to lessen its documentation.
You can just copy-paste parts.

packages/utils-algebra/lib/utils.ts
```diff
@@ -78,18 +79,66 @@ const transformer = new TransformerSubTyped<KnownOperation>({
...
+ * The pre-visitor visits starting from the root, going deeper, while the actual visitor goes in reverse.
```
rubensworks, line 313: > while the actual visitor goes in reverse.

What is the reason for this?

packages/utils-algebra/lib/utils.ts
```diff
@@ -78,18 +79,66 @@ const transformer = new TransformerSubTyped<KnownOperation>({
...
+ * The third argument allows you to also transform based on the subType of operations.
+ * Note that when a callback for the subtype is provided, the callback for the general type is NOT executed.
+ */
 export const visitOperationSub = transformer.visitNodeSpecific.bind(transformer);
 
+/**
+ * Detects all in-scope variables.
+ * In practice this means iterating through the entire algebra tree, finding all variables,
+ * and stopping when a project function is found.
+ * @param {Operation} op - Input algebra tree.
```
rubensworks: These are the kinds of `@param` and `@returns` descriptions we also need above :-)

packages/utils-algebra/lib/utils.ts
```diff
@@ -78,18 +86,136 @@ const transformer = new TransformerSubTyped<KnownOperation>({
...
+ * @return the result of transforming the requested descendant operations (based on the preVisitor)
+ * using a transformer that works its way back up from the descendant to the startObject.
+ */
 export const mapOperationStrict = transformer.transformNode.bind(transformer);
 
-export const mapOperationSub: (typeof transformer.transformNodeSpecific<'unsafe', Operation>) = <any>
-  transformer.transformNodeSpecific.bind(transformer);
+/**
+ * Transform a single operation. e.g.:
+ * `mapOperation({type: 'distinct', input: {type: 'project' }}, { project: { transform: () => 5 } })`
```
rubensworks: Shall we use a more realistic example?
Something similar to https://github.com/comunica/comunica/blob/next/major/packages/actor-optimize-query-operation-assign-sources-exhaustive/lib/ActorOptimizeQueryOperationAssignSourcesExhaustive.ts#L75-L88 for example.
Having a previsitor in there could also be useful.

rubensworks: (also for the docs below)

packages/utils-algebra/lib/utils.ts
```diff
@@ -78,18 +86,136 @@ const transformer = new TransformerSubTyped<KnownOperation>({
...
+ * @param startObject the object from which we will start visiting,
+ *   potentially visiting its descendants along the way.
+ * @param nodeCallBacks a dictionary mapping the various operation types to objects optionally
+ *    containing preVisitor and visitor.
+ *    The preVisitor allows you to provide {@link VisitContext} for the current object,
+ *    altering how it will be visited.
+ *    The visitor allows you to visit the object from deepest to the outermost object.
+ *    This is useful if you for example want to manipulate the objects you visit during your visits,
+ *    similar to {@link mapOperation}.
+ * @param nodeSpecificCallBacks: Same as nodeCallBacks but using an additional level of indirection to
```
rubensworks:
```suggestion
 * @param nodeSpecificCallBacks Same as nodeCallBacks but using an additional level of indirection to
```

packages/utils-algebra/lib/utils.ts
```diff
@@ -78,18 +86,136 @@ const transformer = new TransformerSubTyped<KnownOperation>({
...
+ *     indicate the subType.
+ */
 export const visitOperationSub = transformer.visitNodeSpecific.bind(transformer);
 
+/**
+ * Detects all in-scope variables.
+ * In practice this means iterating through the entire algebra tree, finding all variables,
+ * and stopping when a project function is found.
+ * @param {Operation} op - Input algebra tree.
```
rubensworks:
```suggestion
 * @param {Operation} op Input algebra tree.
```

packages/utils-algebra/lib/utils.ts
```diff
@@ -78,18 +86,136 @@ const transformer = new TransformerSubTyped<KnownOperation>({
...
+ * @param visitor the visitor to be used to traverse the various nodes.
+ *        Allows you to provide a visitor with different default preVisitor cotexts.
+ * @returns {RDF.Variable[]} - List of unique in-scope variables.
```
rubensworks:
```suggestion
 * @returns {RDF.Variable[]} List of unique in-scope variables.
```

packages/utils-algebra/lib/utils.ts
```diff
@@ -78,18 +86,136 @@ const transformer = new TransformerSubTyped<KnownOperation>({
...
+ * returns `{type: distinct, input: [1, 5] }`
+ * @param startObject the object from which we will start the transformation,
+ *   potentially visiting and transforming its descendants along the way.
+ * @param nodeCallBacks a dictionary mapping the various operation types to objects optionally
+ *    containing preVisitor and transformer.
+ *    The preVisitor allows you to provide {@link TransformContext} for the current object,
+ *    altering how it will be transformed.
+ *    The transformer allows you to manipulate the copy of the current object,
+ *    and expects you to return the value that should take the current objects place.
+ * @param nodeSpecificCallBacks: Same as nodeCallBacks but using an additional level of indirection to
```
rubensworks:
```suggestion
 * @param nodeSpecificCallBacks Same as nodeCallBacks but using an additional level of indirection to
```

rubensworks: (also in some other places it seems)

packages/utils-algebra/lib/utils.ts
jitsedesmet: My problem with the more realistic examples is that they do not show the actual data. So my intention is to show both the data and the mapping. I just got the idea of a mapping: Wrap a distinct around the outermost project.

rubensworks: Within comunica, the example I linked above is representative for most usages of this method.
I suspect you also noticed this during your refactoring.

But showing with actual data does indeed make sense, so perhaps we can add both?

jitsedesmet: I personally think the current approach (new) suffices. What do you think?

rubensworks: Sure, that works!

Two minor nits:
- Make sure that the example code is syntactically valid TS (e.g. `returns =`), because some people (including myself) may copy-paste that code.
- Can you modify the code so that indentation and newline placement is similar to other places in Comunica?

packages/utils-algebra/lib/utils.ts
```diff
@@ -78,18 +86,179 @@ const transformer = new TransformerSubTyped<KnownOperation>({
...
+ * Transform a single operation.
+ * e.g. wrapping a distinct around the outermost project:
+ * ```ts
+ * mapOperation(
+ *   { type: 'slice', input: { type: 'project', input: { type: 'join', input: [{ type: 'project' }, { type: 'bgp' }]}}},
+ *   { project: { preVisitor: () => ({ continue: false }), transform: projection =>
+ *     algebraFactory.createDistinct(projection) }},
+ * );
+ * const returns = { type: 'slice', input: { type: 'distinct', input: { type: 'project', input: { type: 'join', input:
+ *           [{ type: 'project' }, { type: 'bgp' }]}}}};
```
rubensworks: What I had in mind was something more like this for all examples:
```
mapOperation(
  {
    type: 'slice',
    input: {
      type: 'project',
      input: {
        type: 'join',
        input: [{ type: 'project' }, { type: 'bgp' }],
      },
    },
  },
  {
    [Algebra.Types.PROJECT]: {
      preVisitor: () => ({ continue: false }),
      transform: projection => algebraFactory.createDistinct(projection).
    },
  },
);
...
```
IMO, this is easier to understand when quickly glancing at the example.

jitsedesmet: aha, okay, fully expanded. Sure

PR #1652, opened 2026-01-22
Group repetitive logger warnings to reduce console noise

### LINKED ISSUE #1375: Compact similar filter warnings in logs
#### Issue type:

- :bug: Bug 

____
#### Description:

Filter warnings become intrusive to seeing the result of queries, in particular when running `npx -p @comunica/query-sparql comunica-sparql http://www.w3.org/ns/odrl/2/ "CONSTRUCT { ?s ?p ?o } WHERE { ?s ?p ?o filter(contains(?o, \"Purpose\")) }"` the result is [this](https://gist.github.com/jeswr/8b6ca39b6fa2032b883c3be89496e79f) which makes it hard to see the single result of

```ttl
<http://www.w3.org/ns/odrl/2/purpose> <http://www.w3.org/2000/01/rdf-schema#label> "Purpose"@en.
```

It may be worth considering having a default behavior where there is 1 message at the end saying "Warnings were produced when running this query, use the `-w` flag to show details".

**Implementation suggestion for bounty:**

The `Logger` should have a utility method for tracking the last X warning messages, and group messages if many/multiple similar messages are detected, which can be used by `LoggerPretty`.
The first message should always be emitted immediately, but any similar later messages may be postponed.

-----

## Bounty

| |
|-|
| €1088 |

voidten:
#### Description

This PR reduces the intrusiveness of repetitive warnings (e.g., from filters) by grouping consecutive identical log messages.

Closes #1375

### DIFF
--- a/engines/query-sparql/test/QuerySparql-test.ts
+++ b/engines/query-sparql/test/QuerySparql-test.ts
@@ -1922,7 +1922,7 @@ WHERE {
       });
     });
 
-    describe('logger warning grouping', () => {
+    describe('logger warning grouping (no browser)', () => {
       class TestLogger extends Logger {
         public readonly warnings: string[] = [];
 


packages/logger-pretty/lib/LoggerPretty.ts
```diff
@@ -27,6 +27,7 @@ export class LoggerPretty extends Logger {
     this.level = args.level;
     this.levelOrdinal = Logger.getLevelOrdinal(this.level);
     this.actors = args.actors;
+    process.on('exit', () => this.flush());
```
rubensworks: Flushing on process exist won't always work, e.g. when the engine is invoked via JavaScript and long-running over potentially multiple queries. This also won't work in a browser environment such as https://query.comunica.dev/.

packages/logger-pretty/lib/LoggerPretty.ts
```diff
@@ -50,7 +51,11 @@ export class LoggerPretty extends Logger {
   }
 
   public warn(message: string, data?: any): void {
-    this.log('warn', LoggerPretty.COLOR_YELLOW, message, data);
+    this.logGrouped(message, (count) => {
+      for (let i = 0; i < count; i++) {
```
rubensworks: This doesn't seem that valuable. This just prints the original lines, without merging them into a single line.

packages/logger-pretty/lib/LoggerPretty.ts
voidten: I've replaced the exit listener with a debounced flush mechanism

packages/types/lib/Logger.ts
```diff
@@ -26,10 +26,74 @@ export abstract class Logger {
...
+        return;
+      }
+      delete this.activeLogGroups[key];
+      if (pending.count > 0) {
+        pending.callback(pending.count);
+      }
+    }
+
+    if (!this.debounceTimeout) {
+      this.debounceTimeout = setTimeout(() => {
```
rubensworks: The problem with this timeout approach, is that if a query takes less 100ms to execute, that the process will still run for at least 100ms, which is not what we want.
We'll need a better way of calling this flush method.
I don't have a good idea right now how to solve this, but I'll think about it a bit as well.

voidten: Another approach could be to explicitly call `logger.flush()` when the query execution finishes.

rubensworks: I like that idea, that could work indeed!

This may be a good place to call this:
https://github.com/comunica/comunica/blob/master/packages/actor-init-query/lib/QueryEngineBase.ts#L206-L239
For bindings and quads, flush would have to be called as an on-end listener on the bindings/quadstream.
For boolean and void, flush would have to be called after the execute promise resolved.

packages/actor-init-query/lib/QueryEngineBase.ts
```diff
@@ -204,11 +204,15 @@ implements IQueryEngine<QueryStringContextInner, QueryAlgebraContextInner> {
...
    */
   public static internalToFinalResult(internalResult: IQueryOperationResult): QueryType {
+    const logger = internalResult.context?.get(KeysCore.log);
     switch (internalResult.type) {
       case 'bindings':
         return {
           resultType: 'bindings',
-          execute: async() => internalResult.bindingsStream,
+          execute: async() => {
+            internalResult.bindingsStream.on?.('end', () => logger?.flush());
```
rubensworks: I'd move the `logger` defined check outside of the on-end listener, so that the end listener only gets registered if logger is defined. (also below)

PR #1674, opened 2026-02-26
Add `disableClearErrorMessages` flag to SPARQL parse actor (#1673)

### LINKED ISSUE #1673: Add flag: disable-clear-error-messages
#### Issue type:

____
#### Description:

Currently, the SPARQL parser does not provide clear error messages since we task the lexer to not track sourceLocations.
As a result, however, the parser error messages are way less clear:

```sparql
SELECT ?movie ?title ?name
WHERE {
  FILTER LANGMATCHES(LANG(?title), "EN" #        <- missing `)`
  FILTER LANGMATCHES(LANG(?name),  "EN")
}
```

Results in error: 
```
Parse error
Expecting --> ) <-- but found --> 'FILTER' <--
```
Since most users will care about clear error messages, I suggest we make clear messages the default. 
Since pipeline use cases might not care about clear error messages, I suggest we do add a flag in the context `disable-clear-error-messages`.
When enabled, actors can choose to optimize themselves at the cost of clear error messages.

Context keys reside in `packages/context-entries/lib/Keys.ts`.
Of course, the parser actor (`actor-query-parse-sparql`) should test its behaviour .

jitsedesmet:
Resolves #1673

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -387,6 +387,6 @@ jobs:
     steps:
       - name: Deploy to GitHub Pages
         id: deployment
-        uses: actions/deploy-pages@v4
+        uses: actions/deploy-pages@v5
         with:
           artifact_name: documentation


packages/actor-query-parse-sparql/lib/ActorQueryParseSparql.ts
```diff
@@ -81,4 +83,10 @@ export interface IActorQueryParseSparqlArgs extends IActorQueryParseArgs {
    * }}
    */
   prefixes?: Record<string, string>;
+  /**
+   * If true, the parser will use an optimized lexer config that disables clear error messages.
+   * Enabling this can improve performance at the cost of less informative parse error messages.
+   * @default {false}
+   */
+  disableClearErrorMessages?: boolean;
```
rubensworks: Instead of "clear" messages, I would talk about "verbose" vs "minimal" error messages.

```suggestion
  minimalErrorMessages?: boolean;
```

rubensworks (changes requested):

PR #1676, opened 2026-03-02
Bugfix/triple function error handling

### LINKED ISSUE #1675: Implementation of Triple should throw on invalid
#### Issue type:

- :bug: Bug 

____
#### Description:

The current implementation of the [SPARQL 1.2 function `TRIPLE`](https://www.w3.org/TR/sparql12-query/#func-triple) does not throw an error like the spec requests.

jitsedesmet:
Resolves #1675

### DIFF
--- a/packages/types/lib/ExpressionEvaluator.ts
+++ b/packages/types/lib/ExpressionEvaluator.ts
@@ -87,6 +87,8 @@ export enum ExpressionType {
   Term = 'term',
   Variable = 'variable',
 }
+
+// TODO: next/major - edit these to reflect the RDFJS capitalization (and general typings?)
 export type TermType = 'namedNode' | 'literal' | 'blankNode' | 'quad' | 'defaultGraph';
 
 export type TermExpression = IExpressionProps & {


packages/actor-function-factory-term-triple/lib/TermFunctionTriple.ts
```diff
@@ -17,12 +18,15 @@ export class TermFunctionTriple extends TermFunctionBase {
...
         .onTerm3(
-          _ => (...args) => new Quad(
-            args[0],
-            args[1],
-            args[2],
-            new DefaultGraph(),
-          ),
+          _ => (subject, predicate, object) => {
+            if (subject.termType !== 'namedNode' && subject.termType !== 'blankNode') {
+              throw new ExpressionError(`TRIPLE: invalid subject term type: ${subject.termType}`);
```
rubensworks: I wouldn't use the `TRIPLE: ` prefix. Other `ExpressionError`'s don't have it either, so it would be a bit inconsistent.

Also, no need to include `subject.termType`, as the error will be caught higher up, and the bindings will be printed there. It's better that the error message remains constant in these places, so it can be more easily grouped: https://github.com/comunica/comunica/commit/5dadcc8a7b6110c0190af675df9b642e818a55b8

```suggestion
              throw new ExpressionError(`Subjects in triple terms must either be named nodes or blank nodes`);
```

Similar to the one below.

packages/actor-function-factory-term-triple/lib/TermFunctionTriple.ts
```diff
@@ -17,12 +18,15 @@ export class TermFunctionTriple extends TermFunctionBase {
...
       overloads: declare(SparqlOperator.TRIPLE)
         .onTerm3(
-          _ => (...args) => new Quad(
-            args[0],
-            args[1],
-            args[2],
-            new DefaultGraph(),
-          ),
+          _ => (subject, predicate, object) => {
+            if (subject.termType !== 'namedNode' && subject.termType !== 'blankNode') {
```
rubensworks, line 22:
```suggestion
            if (subject.termType !== 'NamedNode' && subject.termType !== 'BlankNode') {
```

packages/actor-function-factory-term-triple/lib/TermFunctionTriple.ts
```diff
@@ -17,12 +18,15 @@ export class TermFunctionTriple extends TermFunctionBase {
...
+              throw new ExpressionError(`TRIPLE: invalid subject term type: ${subject.termType}`);
+            }
+            if (predicate.termType !== 'namedNode') {
```
rubensworks, line 25:
```suggestion
            if (predicate.termType !== 'NamedNode') {
```

packages/actor-query-parse-sparql/lib/ActorQueryParseSparql.ts
```diff
@@ -13,13 +13,15 @@ import { AstFactory } from '@traqula/rules-sparql-1-2';
  */
 export class ActorQueryParseSparql extends ActorQueryParse {
   public readonly prefixes: Record<string, string> | undefined;
+  public readonly disableClearErrorMessages: boolean;
   private readonly parser: SparqlParser;
 
   public constructor(args: IActorQueryParseSparqlArgs) {
     super(args);
     this.prefixes = Object.freeze(args.prefixes);
+    this.disableClearErrorMessages = args.disableClearErrorMessages ?? false;
```
rubensworks: Unrelated to this PR?

packages/actor-function-factory-term-triple/lib/TermFunctionTriple.ts
jitsedesmet: why? This is part of rdfjs/types right? It starts with a lower case letter?

rubensworks: No, it starts with a capital letter. I'm surprised the unit tests pass even right now...

jitsedesmet: Hmm, I opened the code in my ide to verify. Maybe it's comunica' internal representation? 
it is: https://github.com/comunica/comunica/blob/master/packages/types/lib/ExpressionEvaluator.ts#L90

jitsedesmet: that's, unfortunate, that they do not match

rubensworks: Ah, that's unfortunate indeed. Could you add a TODO comment to fix this in next/major?

PR #1677, opened 2026-03-09
Replace Karma browser tests with Playwright

### LINKED ISSUE #1524: Improve browser isomorphism
#### Issue type:

____
#### Description:

Even though we have checks in the CI to ensure Comunica runs correctly in browser environments,
it has some flaws that lead to some issues still popping up (#1523).

Concretely, we require 2 things:

1. Pure browser-only tests of Comunica in the CI. Our current tests make use of Karma, and require some polyfilling due to the use of Karma, which leads to these things also being polyfilled into Comunica, which means that we could be missing some things, since we don't want to require polyfilling. (@surilindur did some work in this direction with Playwright: https://github.com/comunica/comunica/issues/1523#issuecomment-2761233839) These pure browser-only tests should either replace our Karma tests, or we should introduce some (not too many) new pure browser-only tests (e.g. using Playwright).
2. Fixing all of the issues that pop up during these browser-only tests. AFAIK, this is currently only an issue with `process.env` being used within the graphql dependency. This may be fixed by updating to the latest graphql version, submitting a PR, or removing the graphql dependency.

voidten:
Karma-based browser tests were relying on a test harness that introduced polyfills into the browser environment, which weakens confidence that Comunica works in a real browser without Node-like shims.

This PR replaces that path with pure browser Playwright coverage and updates CI accordingly.

@rubensworks I investigated the issue’s second point about `process.env` being used and could not reproduce it anymore.

Fixes #1524

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,6 +141,8 @@ jobs:
           cache: yarn
       - name: Install dependencies and build
         run: yarn install --frozen-lockfile
+      - name: Build browser bundles
+        run: yarn run build:browser
       - name: Install Playwright browsers
         run: yarn playwright install --with-deps
       - name: Run browser tests


playwright.config.ts
```diff
@@ -0,0 +1,27 @@
+import { defineConfig, devices } from '@playwright/test';
+
+export default defineConfig({
+  forbidOnly: Boolean(process.env.CI),
+  fullyParallel: true,
+  reporter: 'html',
```
rubensworks, line 6: I suspect we may also want to change the reporter in the CI? https://playwright.dev/docs/test-reporters#reporters-on-ci

voidten: We use the HTML reporter, but in CI Playwright also adds `dot` for terminal output, which is what we’re seeing in the [Browser tests job](https://github.com/comunica/comunica/actions/runs/22838652194/job/66240266515?pr=1677).

rubensworks: Do you know what the output looks like if any of the tests fail?

voidten: s

rubensworks: Ok, great, that's perfect as-is then.

PR #1689, opened 2026-04-03
Add custom `SERVICE` executors

### LINKED ISSUE #1660: Custom `SERVICE` executors
### Discussed in https://github.com/comunica/comunica/discussions/1659

<div type='discussions-op-text'>

<sup>Originally posted by **namedgraph** February  2, 2026</sup>
Hi,

Does Comunica support something like Jena's [Custom Service Executors](https://jena.apache.org/documentation/query/custom_service_executors.html)?

What I want to do is to register my service implementation that would intercept the `SERVICE <customer:service>` patterns, process the incoming bindings, and return updated bindings. So that from the user's perspective it will look like another SPARQL service even though internally it might be something completely different.</div>

-----

We should allow SERVICE handler injection, similar to how we allow passing extension functions: https://comunica.dev/docs/query/advanced/extension_functions/
I suspect allowing some kind of callback should be possible, which takes service operation and bindings as input, and a bindingsstream as output.

-----

## Bounty

| |
|-|
| €1088 |

voidten:
This PR adds support for custom `SERVICE` executors through query context.

Users can now provide either:

- `serviceExecutors`: a map from service IRI to executor
- `serviceExecutorCreator`: a callback that resolves an executor for a service IRI

This makes `SERVICE` extensible in the same way Comunica already supports custom extension functions.

Fixes #1660


rubensworks:
@jitsedesmet Given the relatedness of this to extension functions, could you have a look at this one?

packages/actor-optimize-query-operation-query-source-identify/lib/ActorOptimizeQueryOperationQuerySourceIdentify.ts
```diff
@@ -86,25 +89,41 @@ export class ActorOptimizeQueryOperationQuerySourceIdentify extends ActorOptimiz
...
+
+      // Identify sources of SERVICE targets, unless the whole query is passed to the source (e.g. for SPARQL endpoints)
+      // and no custom executor needs to intercept a SERVICE clause locally.
+      if (Object.keys(serviceExecutors).length > 0 ||
+        !await passFullOperationToSource(action.operation, querySources ?? [], context)) {
+        const serviceSources: Record<string, IQuerySourceWrapper> = {};
+        for (const [ service, serviceNamedNode ] of services.entries()) {
+          const serviceExecutor = serviceExecutors[service];
+          serviceSources[service] = serviceExecutor ?
+              { source: new QuerySourceServiceExecutor(serviceNamedNode.value, serviceExecutor) } :
```
rubensworks, line 121: Doing this within this actor feels a bit weird. I recommend adding a new actor for it.

voidten: Done!

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -1214,6 +1215,61 @@ SELECT ?person ?name ?book ?title {
...
+              throw new Error('service failed');
+            },
+          },
+        });
+        await expect(bindingsStream).toEqualBindingsStream([
+          BF.bindings([
+            [ DF.variable('value'), DF.literal('outer') ],
+          ]),
+        ]);
+      });
```
jitsedesmet, line 1358: Nice tests already. Maybe this is a stretch, but I think a test that works using a variable SERVICE would also be valuable: https://www.w3.org/TR/sparql12-query/#rServiceGraphPattern

Something like:

```
SELECT ?value WHERE {
  VALUES ?service { <https://example.org/service> }
  SERVICE ?service {
    <urn:s> <urn:p> ?value .
  }
}
```

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -1214,6 +1215,61 @@ SELECT ?person ?name ?book ?title {
...
+          },
+        });
+        await expect(bindingsStream).toEqualBindingsStream([
+          BF.bindings([
+            [ DF.variable('value'), DF.literal('works') ],
+          ]),
+        ]);
+      });
+
+      it('with a silent custom SERVICE executor failure should preserve outer bindings', async() => {
```
jitsedesmet, line 1338: Could you also add a single test that uses the `serviceExecutorCreator`?
This could be in combination with the variable service test below, where you would, for example, create services on the fly.

I am thinking of a use case where you would parse a base IRI to mean some function should be called, but the IRI also has some query parameters to further specify the functionality. I think that would be cool.

So, like:

`<https://example.com/addXtoLiterals?x=${myNum}>`
`<https://example.com/addXtoLiterals?x=5>`, `<https://example.com/addXtoLiterals?x=10>`

packages/actor-optimize-query-operation-assign-sources-exhaustive/lib/ActorOptimizeQueryOperationAssignSourcesExhaustive.ts
```diff
@@ -32,10 +32,11 @@ export class ActorOptimizeQueryOperationAssignSourcesExhaustive extends ActorOpt
 
     const sources = action.context.get(KeysQueryOperation.querySources) ?? [];
     const serviceSources = action.context.get(KeysQueryOperation.serviceSources) ?? {};
-    if (sources.length === 0 && Object.keys(serviceSources).length === 0) {
+    const hasServiceSources = Object.keys(serviceSources).length > 0;
+    if (sources.length === 0 && !hasServiceSources) {
       return { operation: action.operation, context: action.context };
     }
-    if (await passFullOperationToSource(action.operation, sources, action.context)) {
+    if (!hasServiceSources && (await passFullOperationToSource(action.operation, sources, action.context))) {
```
jitsedesmet, line 41: I'm unsure what is going on here, I feel like it deserves some comments? What do you think, @rubensworks ? 

If I understand correctly:

`// Algebra patterns with a custom service implementation should not be pushed down to the sources since they will fail to recognise the special behavior` 


(I see this should be added to line 86 instead?)


I am also wondering. Do you correctly check whether the service is one registered or not? Meaning: do you resolve the registered services in time before this optimize actor is called?

packages/actor-optimize-query-operation-assign-sources-exhaustive/test/ActorOptimizeQueryOperationAssignSourcesExhaustive-test.ts
```diff
@@ -38,6 +38,18 @@ describe('ActorOptimizeQueryOperationAssignSourcesExhaustive', () => {
       }),
     },
   };
+  const sourceServiceExecutor: IQuerySourceWrapper = <any> {
+    source: {
+      referenceValue: 'source1',
+      supportsServiceOperationInjection: true,
```
jitsedesmet, line 44: Sources now have a boolean they can say whether they accept the named services or not? Would it be interesting to also make this a boolean callback (or a list)? Similar to how you are able to create service providers on the fly?

@rubensworks

packages/actor-optimize-query-operation-query-source-identify/lib/ActorOptimizeQueryOperationQuerySourceIdentify.ts
```diff
@@ -86,27 +86,34 @@ export class ActorOptimizeQueryOperationQuerySourceIdentify extends ActorOptimiz
...
-            }
-          },
+    const services: Set<string> = new Set();
+    algebraUtils.visitOperation(action.operation, {
+      [Algebra.Types.SERVICE]: {
+        preVisitor: () => ({ continue: false }),
+        visitor: (serviceOperation) => {
+          if (serviceOperation.name.termType === 'NamedNode') {
+            services.add(serviceOperation.name.value);
+          }
```
jitsedesmet: What happens if it is a variable instead? And will this code function correctly if it is?

packages/actor-optimize-query-operation-service-executor/lib/ActorOptimizeQueryOperationServiceExecutor.ts
```diff
@@ -0,0 +1,84 @@
...
+      }
+    }
+
+    return {
+      context: modified ? action.context.set(KeysQueryOperation.serviceSources, serviceSources) : action.context,
+      operation: action.operation,
+    };
+  }
+
+  public static getServices(operation: Algebra.Operation): Map<string, RDF.NamedNode> {
```
jitsedesmet:
```suggestion
  public static getNamedServices(operation: Algebra.Operation): Map<string, RDF.NamedNode> {
```

packages/actor-optimize-query-operation-service-executor/lib/ActorOptimizeQueryOperationServiceExecutor.ts
```diff
@@ -0,0 +1,84 @@
...
+    const services: Map<string, RDF.NamedNode> = new Map();
+    algebraUtils.visitOperation(operation, {
+      [Algebra.Types.SERVICE]: {
+        preVisitor: () => ({ continue: false }),
```
jitsedesmet, line 85: Why do SERVICES within SERVICES not need to be handled?

jitsedesmet (commented):
I think this looks pretty good.
I agree with the API, since I don't know much beyond that, I don't know whether my feedback will be of great help. Regardless, I looked at the code, and where I had questions, I asked them.

packages/actor-optimize-query-operation-query-source-identify/lib/ActorOptimizeQueryOperationQuerySourceIdentify.ts
```diff
@@ -86,27 +86,34 @@ export class ActorOptimizeQueryOperationQuerySourceIdentify extends ActorOptimiz
...
-      const services: Set<string> = new Set();
-      algebraUtils.visitOperation(action.operation, {
-        [Algebra.Types.SERVICE]: {
-          preVisitor: () => ({ continue: false }),
-          visitor: (serviceOperation) => {
-            if (serviceOperation.name.termType === 'NamedNode') {
-              services.add(serviceOperation.name.value);
-            }
-          },
+    const services: Set<string> = new Set();
```
rubensworks: We can probably skip this collection if we can't pass the full operation to the source and we don't have KeysQueryOperation.serviceSources in the context.

packages/actor-optimize-query-operation-service-executor/lib/QuerySourceServiceExecutor.ts
```diff
@@ -0,0 +1,202 @@
...
+    const lenient = Boolean(context.get(KeysInitQuery.lenient));
+    const silentFallback = lenient ? QuerySourceServiceExecutor.createSilentFallbackBinding(context) : undefined;
+    const serviceBindingsPromise = Promise.resolve()
+      .then(async() => this.serviceExecutor(
+        { ...serviceOperation, input: operation },
+        context.get(KeysQueryOperation.joinBindings),
+        context,
+        options,
+      ));
+    const bindings = new class extends BufferedIterator<Bindings> {
```
rubensworks: Let's make this a dedicated class within a separate file.

packages/types/lib/ServiceExecutor.ts
```diff
@@ -0,0 +1,15 @@
...
+import type { BindingsStream } from './Bindings';
+import type { IActionContext } from './IActionContext';
+import type { IQueryBindingsOptions } from './IQuerySource';
+
+export type AsyncServiceExecutor = (
+  serviceOperation: Algebra.Service,
+  binding: RDF.Bindings | undefined,
+  context: IActionContext,
+  options?: IQueryBindingsOptions,
+) => BindingsStream | Promise<BindingsStream>;
```
rubensworks: For simplicity, I would make it always return `Promise<BindingsStream>`.

packages/types/lib/ServiceExecutor.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export type AsyncServiceExecutorCreator = (serviceNamedNode: RDF.NamedNode) =>
+Promise<AsyncServiceExecutor | undefined> | AsyncServiceExecutor | undefined;
```
rubensworks: Same here, can also be always async.

packages/actor-optimize-query-operation-service-executor/lib/ActorOptimizeQueryOperationServiceExecutor.ts
voidten: I added a comment for this.

The reason is that once we hit a SERVICE clause, its input is delegated to that service/source. So nested SERVICE clauses should remain part of that delegated input instead of being collected/reassigned locally, otherwise we could change who is responsible for evaluating the nested SERVICE.

packages/actor-optimize-query-operation-assign-sources-exhaustive/lib/ActorOptimizeQueryOperationAssignSourcesExhaustive.ts
voidten: I added comments around this.

.github/workflows/ci.yml
```diff
@@ -201,10 +201,10 @@ jobs:
         run: git config --global core.autocrlf input
       - name: Check out repository
         uses: actions/checkout@v6
-      - name: Use Node.js ${{ env.NODE_VERSION_LOWEST }}
+      - name: Use Node.js 20.x
```
voidten: The Docker job passes again after pinning the Docker image build job back to Node 20.x.

This seems to confirm the failure was related to the newer Node/npm environment after `NODE_VERSION_LOWEST` was bumped to 22.x.

rubensworks: This should remain unchanged, since Node 20 is not in LTS anymore.

voidten: The issue appears to be that `lerna-docker` installs `verdaccio@5.29.2`, which is incompatible with Node 22 because Node 22 removes the `crypto.createCipher` API still used by that Verdaccio version. This causes Verdaccio to crash during authentication. I believe the fix is to upgrade the Verdaccio version pinned in `lerna-docker` to ≥ `5.31.0`.

See https://github.com/verdaccio/verdaccio/issues/4601

rubensworks: That's not what the CI logs are saying though.
Node 22 is working fine for lerna-docker on the master branch.

voidten: The 404 is just Verdaccio not having the package at install time. But earlier logs show Verdaccio crashing during auth (`_crypto.createCipher`), which causes `lerna publish` to fail. So the package never got published to the local registry in the first place.

rubensworks: I see. Surprising that we don't see it on master.
Could you make a PR to https://github.com/rubensworks/lerna-docker/tree/master to bump the version?

voidten: It actually does crash on master too, but it does not fail the build because nothing new is being published or installed.

In this PR, the added package makes the issue visible since `npm install` needs Verdaccio to contain it, which it does not.

Opened https://github.com/rubensworks/lerna-docker/pull/5

engines/config-query-sparql/config/optimize-query-operation/actors-v5-1-3.json
```diff
@@ -42,6 +43,11 @@
         { "@id": "urn:comunica:default:optimize-query-operation/actors#query-source-identify" }
       ]
     },
+    {
```
rubensworks, line 46: This would have to be moved to 5-2-3, since we're at that version now on master.

packages/types/lib/ServiceExecutor.ts
```diff
@@ -0,0 +1,15 @@
...
+import type { BindingsStream } from './Bindings';
+import type { IActionContext } from './IActionContext';
+import type { IQueryBindingsOptions } from './IQuerySource';
+
+export type AsyncServiceExecutor = (
+  serviceOperation: Algebra.Service,
+  binding: RDF.Bindings | undefined,
+  context: IActionContext,
+  options?: IQueryBindingsOptions,
+) => Promise<BindingsStream>;
```
rubensworks, line 12: Besides `BindingsStream`, I would also allow returning an `Iterable` of bindings, for convenience of implementors.
Since this iterable won't have any metadata, it will require slightly different processing here within comunica.

packages/actor-optimize-query-operation-assign-sources-exhaustive/lib/ActorOptimizeQueryOperationAssignSourcesExhaustive.ts
```diff
@@ -32,10 +32,13 @@ export class ActorOptimizeQueryOperationAssignSourcesExhaustive extends ActorOpt
 
     const sources = action.context.get(KeysQueryOperation.querySources) ?? [];
     const serviceSources = action.context.get(KeysQueryOperation.serviceSources) ?? {};
-    if (sources.length === 0 && Object.keys(serviceSources).length === 0) {
+    const hasServiceSources = Object.keys(serviceSources).length > 0;
```
rubensworks, line 35: To be honest, I don't fully understand why the changes within this file are needed, and what their impact is exactly.

packages/actor-optimize-query-operation-assign-sources-exhaustive/lib/ActorOptimizeQueryOperationAssignSourcesExhaustive.ts
```diff
@@ -137,4 +120,78 @@ export class ActorOptimizeQueryOperationAssignSourcesExhaustive extends ActorOpt
...
+      // Nested SERVICE clauses belong to this delegated SERVICE input and should not be transformed locally.
+      {},
+    );
+  }
+
+  public static supportsServiceOperationInjection(source: unknown): boolean {
+    if (source === null || typeof source !== 'object') {
+      return false;
+    }
+    const { supportsServiceOperationInjection, innerSource } = <{
```
rubensworks, line 190: Let's use fragment selector shapes instead for this.

packages/actor-optimize-query-operation-query-source-identify/lib/ActorOptimizeQueryOperationQuerySourceIdentify.ts
```diff
@@ -86,30 +86,71 @@ export class ActorOptimizeQueryOperationQuerySourceIdentify extends ActorOptimiz
         .set(KeysQueryOperation.querySources, querySources);
     }
 
-    // Identify sources of SERVICE targets, unless the whole query is passed to the source (e.g. for SPARQL endpoints)
-    if (!await passFullOperationToSource(action.operation, querySources ?? [], context)) {
-      const services: Set<string> = new Set();
-      algebraUtils.visitOperation(action.operation, {
+    // Identify sources of SERVICE targets, unless the whole query is passed to the source (e.g. for SPARQL endpoints).
```
rubensworks, line 89: Similar to before, it's unclear to me why the changes within this file are needed.

packages/actor-optimize-query-operation-service-executor/lib/ActorOptimizeQueryOperationServiceExecutor.ts
```diff
@@ -0,0 +1,163 @@
...
+import { passTestVoid } from '@comunica/core';
+import type { AsyncServiceExecutor, IActionContext, IQuerySourceWrapper } from '@comunica/types';
+import { Algebra, algebraUtils } from '@comunica/utils-algebra';
+import type * as RDF from '@rdfjs/types';
+import { QuerySourceServiceExecutor } from './QuerySourceServiceExecutor';
+
+/**
+ * A comunica SERVICE Executor Optimize Query Operation Actor.
+ */
+export class ActorOptimizeQueryOperationServiceExecutor extends ActorOptimizeQueryOperation {
```
rubensworks, line 18: Why is this implemented here, and not on the query-source-identify bus?

packages/actor-optimize-query-operation-service-executor/lib/QuerySourceServiceExecutor.ts
```diff
@@ -0,0 +1,153 @@
...
+    this.serviceExecutor = serviceExecutor;
+  }
+
+  public async getFilterFactor(_context: IActionContext): Promise<number> {
+    return 1;
+  }
+
+  public async getSelectorShape(_context: IActionContext): Promise<FragmentSelectorShape> {
+    return {
+      type: 'operation',
```
rubensworks, line 37: I suspect we should only be accepting SERVICE operation types here.

packages/actor-optimize-query-operation-service-executor/lib/QuerySourceServiceExecutorBindingsIterator.ts
```diff
@@ -0,0 +1,75 @@
+import type { Bindings, BindingsStream } from '@comunica/types';
+import { BufferedIterator } from 'asynciterator';
+
+export interface IQuerySourceServiceExecutorBindingsIteratorArgs {
```
rubensworks, line 4: Is it necessary to define a separate interface for this? If it's only use in the ctor, let's just inline it.

packages/actor-optimize-query-operation-service-executor/lib/QuerySourceServiceExecutorBindingsIterator.ts
```diff
@@ -0,0 +1,75 @@
...
+  public override _read(_count: number, done: () => void): void {
+    if (this.started) {
+      done();
+      return;
+    }
+    this.started = true;
+
+    this.serviceBindingsPromise
+      .then((serviceBindings) => {
+        serviceBindings.on('data', (binding: Bindings) => {
```
rubensworks, line 40: This sets the bindings stream into flow-mode, and ignores backpressure handling.
Could be solved using transform iterator.

rubensworks (changes requested):

voidten:
Sorry for the delay! This PR should be ready for review again now.

packages/actor-query-source-identify-service-executor/lib/ActorQuerySourceIdentifyServiceExecutor.ts
```diff
@@ -0,0 +1,68 @@
...
+import { ActionContext, failTest, passTestVoidWithSideData } from '@comunica/core';
+import type { ComunicaDataFactory, IActionContext, ServiceExecutor } from '@comunica/types';
+import { QuerySourceServiceExecutor } from './QuerySourceServiceExecutor';
+
+/**
+ * A comunica Service Executor Query Source Identify Actor.
+ * It identifies IRIs for which a custom SERVICE executor is registered in the query context
+ * as sources that evaluate SERVICE clauses through that executor.
+ */
+export class ActorQuerySourceIdentifyServiceExecutor extends ActorQuerySourceIdentify<ServiceExecutor> {
```
rubensworks, line 18: Could you also open a PR to https://github.com/comunica/website that documents how to use these service executors with a simple example?
Let's also include in the example how the bindings factory can be extracted from the context (I'm pretty sure it's in there, but I haven't double-checked, if not, it should be easily constructable from it).

packages/actor-optimize-query-operation-assign-sources-exhaustive/lib/ActorOptimizeQueryOperationAssignSourcesExhaustive.ts
```diff
@@ -143,4 +151,65 @@ export class ActorOptimizeQueryOperationAssignSourcesExhaustive extends ActorOpt
...
+   * Check if the given operation contains a SERVICE clause that must be evaluated as a whole by its source.
+   * @param operation An operation.
+   * @param serviceShapes Mapping of SERVICE names to the selector shapes of their sources.
+   */
+  public hasWholeServiceClauses(
+    operation: Algebra.Operation,
+    serviceShapes: Record<string, FragmentSelectorShape>,
+  ): boolean {
+    let found = false;
+    algebraUtils.visitOperation(operation, {
```
rubensworks, line 186: I think `visitOperation` now also has an async variant (right @jitsedesmet?).
So let's move the shape retrieval from the service operation (that is now done as preprocessing step in `getServiceShapes`) into this method directly.
This way, not all shapes will always be extracted.

packages/actor-optimize-query-operation-query-source-identify/lib/ActorOptimizeQueryOperationQuerySourceIdentify.ts
```diff
@@ -97,8 +97,17 @@ export class ActorOptimizeQueryOperationQuerySourceIdentify extends ActorOptimiz
         .set(KeysQueryOperation.querySources, querySources);
     }
 
-    // Identify sources of SERVICE targets, unless the whole query is passed to the source (e.g. for SPARQL endpoints)
-    if (!await passFullOperationToSource(action.operation, querySources ?? [], context)) {
+    // Identify sources of SERVICE targets, unless the whole query is passed to the source (e.g. for SPARQL endpoints).
```
rubensworks, line 100: My previous comment still holds.
Are you sure actors are really needed? I would suspect we can solve it using just fragment selector shapes (possibly with some changes, similar to how extension functions are handled).

rubensworks (changes requested):

packages/actor-optimize-query-operation-assign-sources-exhaustive/lib/ActorOptimizeQueryOperationAssignSourcesExhaustive.ts
jitsedesmet: yes https://github.com/comunica/comunica/pull/1796

PR #1694, opened 2026-04-10
Fix: argument compatible functions

jitsedesmet:
Traqula did not handle the language direction information in SPARQL1.2 well. This has been fixed in Traqula through a recent patch. That patch caused Comunica's spec tests to fail (they passed "accidentally" before). This PR ensures Comunica passes the spec test by fixing the string compatibility check. We moved the patch, that happened on each function separately to the Helper Expression Evaluator Builder.


jitsedesmet:
no additional tests need to be added (the issue was able to fly under the radar due to a bug in Traqula)

jitsedesmet:
@rubensworks that one failing test seems like something unrelated went wrong?

rubensworks (approved):
Failing test can be ignored indeed.

Could you add a PR description so I have something to reference when creating a squashed commit message?

rubensworks:
Thanks!

Issue #1643, opened 2025-11-03
Cases of incorrect GRAPH handling


rubensworks:
#### Issue type:

- :bug: Bug 

____
#### Description:

We currently push down all GRAPH operators into triple patterns as quad patterns.

In https://github.com/comunica/comunica/commit/7e215eea862e317bc8f56aa1cced97f9afd6e9a1#diff-a43c9682d61fb2c893590c0d57a66fbedad5f4d2d54f7a7a44b851f1aa14764d, we fixed an edge-case where this translation causes problems in combination with MINUS.
However, https://github.com/w3c/rdf-tests/pull/233 and https://github.com/w3c/rdf-tests/pull/234 show two additional edge-cases where we fail.
While we could probably add hacks similar to MINUS to work around these issues, we probably need a cleaner and more sustainable solution.

One solution I currently have in mind is to make this triple-to-quad translation not happen when the translation is known to be unsafe (a whitelist approach could also work).
(unsafe translations correspond to https://github.com/w3c/rdf-tests/pull/233, https://github.com/w3c/rdf-tests/pull/234, and https://github.com/w3c/rdf-tests/pull/228)
For those unsafe translations, the `graph` operator would be kept.
This means that in Comunica, we will require a new actor for handling `graph` operators.
Since this `graph` operator produces data from a source (just like triple patterns and property paths), this will require some changes in source assignment (e.g. the exhaustive source assignment actor, filter pushdown, ...)

Since this is a quite invasive change, it would be better to keep this for Comunica v5, and accept that these cases fail in v4.

@jitsedesmet do you think these changes make sense? If so, could you look into making the necessary changes in Traqula? I can then look into the corresponding changes on the Comunica side (`next/major`).
This is not super urgent, but would be good if we can include this in the v5 release.

When fixed, we need to remove the `--skip` flag in the query-sparql test suite run command.

____

jitsedesmet:
Sure, I can look at that. I actually already did in the context of the new test you mentioned, I wanted to always keep the graph operation, the problem with that is in construct templates and update queries.
In the template, and in the insert and delete fields, the values are just lists of `RDF.Quad`. So this did not seem trivial...
What would you suggest for these fields? To change the types so they can keep the graph information?
I ask because I think the cleanest way to do this is to keep the graph statements and never push down from within the algebra. We could then provide a util function that does the pushdown based on `mapOperation`, which Comunica's actor-graph-operator could use.

Additionally, what would the whitelist be?

[...]

PR #1707, opened 2026-05-08
Feat: own graph operator

jitsedesmet:
Draft pull request for #1643 .


jitsedesmet:
(Tests will fail until Traqula has been updated)

jitsedesmet:
I doubt this optimizer is of much help. Checking the restrictions comes at a cost, and since the optimization is so heavily restricted, with very few gains, I don't think we should keep the optimizer. What do you think @rubensworks ?

PR #1708, opened 2026-05-12
Feat: own graph operator bis

jitsedesmet:
We should still investigate what to do with: https://github.com/comunica/comunica/pull/1707#discussion_r3210336141


packages/actor-optimize-query-operation-quad-substitution/lib/ActorOptimizeQueryOperationQuadSubstitution.ts
```diff
@@ -0,0 +1,178 @@
...
+  /**
+   * Recursively walk the algebra tree and substitute safe GRAPH operations with quad patterns.
+   */
+  public static substituteQuads(
+    algebraFactory: AlgebraFactory,
+    operation: Algebra.Operation,
+  ): Algebra.Operation {
+    return algebraUtils.mapOperation(operation, {
+      [Algebra.Types.GRAPH]: {
+        preVisitor: () => ({ continue: false }),
```
jitsedesmet, line 53: This can be a continue true, and then during the substitution you should not traverse the graph operation. Similarly, when finding a GRAPH operation within the graph OPERATOR, you cannot substitute since, it would have been substituted in case that operation was safe.

packages/actor-optimize-query-operation-quad-substitution/lib/ActorOptimizeQueryOperationQuadSubstitution.ts
```diff
@@ -0,0 +1,178 @@
...
+      },
+    });
+  }
+
+  /**
+   * Check whether it is safe to substitute a GRAPH operation with quad patterns.
+   *
+   * Substitution is safe when:
+   * 1. The inner tree has at least one pattern-like operation (PATTERN, PATH, LINK, NPS)
+   * 2. If the graph name is a variable, the inner tree does NOT contain MINUS or GROUP
```
jitsedesmet, line 77: We should better define this in function of when it is NOT safe

packages/actor-optimize-query-operation-quad-substitution/lib/ActorOptimizeQueryOperationQuadSubstitution.ts
```diff
@@ -0,0 +1,178 @@
...
+   *
+   * For variable graphs, we do NOT recurse into PROJECT (subquery) nodes because:
+   * - Subqueries create a variable scope boundary
+   * - The graph variable might collide with a same-named variable inside the subquery
+   * - Patterns inside subqueries should not be counted as substitutable
```
jitsedesmet, line 82: I actually wonder about what the semantics of something like this would be...
I think we simply should not substitute?

jitsedesmet: I suspect in case the subsitution would be safe, we should simply add the graph var to the projected variables... but then we should still ensure no variables clash... So maybe we can see this as a possible optimization? @rubensworks

packages/actor-optimize-query-operation-quad-substitution/lib/ActorOptimizeQueryOperationQuadSubstitution.ts
```diff
@@ -0,0 +1,178 @@
...
+   */
+  public static isSafeToSubstitute(graphOp: Algebra.Graph): boolean {
+    let hasPatterns = false;
```
jitsedesmet, line 85: This is not needed if you assume you are safe from the start and only can be changed to non safe.... its not about containing a patterns, it's about not containing empty BGP or NOP.

packages/actor-optimize-query-operation-quad-substitution/lib/ActorOptimizeQueryOperationQuadSubstitution.ts
```diff
@@ -0,0 +1,178 @@
...
+   * For variable graphs, does not recurse into subqueries (PROJECT) to respect scope boundaries.
+   */
+  public static substituteGraphInOperation(
+    algebraFactory: AlgebraFactory,
+    operation: Algebra.Operation,
+    graphTerm: Algebra.Graph['name'],
+  ): Algebra.Operation {
+    const noRecurse = {
+      preVisitor: () => ({ continue: false }),
+      transform: (op: Algebra.Operation) => op,
```
jitsedesmet, line 145: no need to provide the identity.

packages/actor-query-operation-graph/lib/ActorQueryOperationGraph.ts
```diff
@@ -0,0 +1,416 @@
...
+
+    const result = getSafeBindings(
+      await this.mediatorQueryOperation.mediate({ operation: discoveryOp, context }),
+    );
+
+    const bindings = await result.bindingsStream.toArray();
+    return bindings
+      .map(b => b.get(graphVar))
+      .filter((v): v is RDF.NamedNode => v?.termType === 'NamedNode');
+  }
```
jitsedesmet, line 288: These things should be done in a streaming fashion since, in link traversal, data is not static!

packages/actor-query-operation-graph/lib/ActorQueryOperationGraph.ts
```diff
@@ -0,0 +1,416 @@
...
+          visitor: (path: Algebra.Path) => {
+            const source = getOperationSource(path);
+            if (source && !seen.has(source)) {
+              seen.add(source);
+              sources.push(source);
+            }
+          },
+        },
+      });
+    }
```
jitsedesmet, line 321: I assume that there will be a better way.

packages/actor-optimize-query-operation-quad-substitution/lib/ActorOptimizeQueryOperationQuadSubstitution.ts
rubensworks: If it's too complex, I would leave it out for now.

Issue #721, opened 2020-08-22
Run all SPARQL spec tests


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request

____
#### Description:

The package.json file of engines/query-sparql contains several `spec:*` scripts to invoke [SPARQL spec tests](https://github.com/w3c/rdf-tests/blob/gh-pages/sparql11/data-sparql11/manifest-all.ttl).
We should make sure that all of them are invoked in our GitHub CI workflow:

* [x] `spec:query`: http://www.w3.org/TR/sparql11-query/
* [x] `spec:update`: http://www.w3.org/TR/sparql11-update/ (depends on #435)
* [ ] `spec:csv-tsv`: http://www.w3.org/TR/sparql11-results-csv-tsv/ (requires CSV/TSV parsers)
* [ ] `spec:json`: http://www.w3.org/TR/sparql11-results-json/ (should be enableable directly)
* [ ] `spec:fed`: http://www.w3.org/TR/sparql11-federated-query/ (requires support in rdf-test-suite)
* [ ] `spec:sd`: http://www.w3.org/TR/sparql11-service-description/ (?)
* [ ] `spec:prot`: http://www.w3.org/TR/sparql11-protocol/ (depends on #435)
* [ ] `spec:graphstore`: http://www.w3.org/TR/sparql11-http-rdf-update/ (depends on #435)

Each one that succeeds should be added to the `spec` script.

Some of these may require changes to https://github.com/rubensworks/rdf-test-suite.js

Afterwards, we should share our earl report on https://github.com/w3c/rdf-tests/

It's possible to take up this bounty in smaller parts.

-----

## Bounty

A [bounty](https://comunica.dev/association/bounties/) has been placed on this issue by:

| |
|-|
| [<img alt="Comunica Association" src="https://comunica.dev/img/comunica_red.svg" height="50">](https://comunica.dev/association/) |
| *Variable budget* |

[Click here to learn more](https://comunica.dev/association/bounty_process/) if you're interested in claiming this bounty by resolving this issue.

noahvsb:
The SPARQL spec test link is outdated. I think this is the updated one: https://github.com/w3c/rdf-tests/blob/main/sparql/sparql11/manifest-all.ttl

kavixo:
Hi, I’m interested in taking this on.

yui-stingray:
Hi, I would be interested in taking a small, well-scoped slice of this bounty if the association/company is open to assigning partial work.

[...]

PR #1715, opened 2026-05-25
[codex] Run SPARQL JSON result spec tests

maiqiu-cat:
## Summary
- make `spec:json-1-1` run the SPARQL JSON result-format manifest directly, so the JSON result tests execute instead of returning 0/0 through the aggregate manifest selection
- include `spec:json-1-1` in the aggregate `spec` script as a focused smaller part of #721

## Verification
- `yarn --cwd engines/query-sparql run spec:json-1-1`
- commit hook: `eslint . --cache`, `yarn run build:ts && yarn run build:components`, `jest`
- `yarn --cwd engines/query-sparql run spec` reached and passed the new JSON step locally; the later existing `spec:sparql12-xsd-datetime-duration` step failed in my environment due a `raw.githubusercontent.com` connect timeout

Refs #721.


rubensworks:
This does not solve the issue.

PR #1720, opened 2026-06-15
Refactor/cleanup expr eval tests

jitsedesmet:
This PR cleans up some test code of the expression-evaluation packages that originally imported using file imports. They now import using utils-jest.

The PR also fixes some evaluation helper functions of Comunica that we wrong (as discoverd in our composability demo: https://github.com/jitsedesmet/demo-mixed-composability/blob/main/comunica/packages/actor-function-factory-term-adjust/lib/TermFunctionAdjust.ts )

PR #1721, opened 2026-06-19
Align CSV serialization with spec and enable tests

voidten:
This PR aligns the CSV query result serialization with the [SPARQL 1.1 Query Results CSV and TSV Formats spec](https://www.w3.org/TR/sparql11-results-csv-tsv/). Previously, IRIs were incorrectly wrapped in angle brackets (`< >`). However, section 3.2 explicitly states: *"The entry in each field is the string corresponding to the RDF term value... without syntax to denote what kind of term it is."* To resolve this, I removed the faulty fallback logic in the CSV serializer.

This fix is required for `spec:csv-tsv-1-1` to pass. Since `spec:json-1-1` is already passing, both commands have now been appended to the `spec` script.


packages/actor-init-query/spec/sparql-engine-base.js
```diff
@@ -12,6 +12,16 @@ module.exports = function(engine) {
     query(data, queryString, options) {
       return this.queryLdf([{ type: 'rdfjs', value: source(data) }], null, queryString, options);
     },
+    async queryResultFormat(data, queryString, mediaType, options) {
+      const result = await engine.query(queryString, {
+        baseIRI: options.baseIRI,
+        sources: [{ type: 'rdfjs', value: source(data) }],
+        httpRetryCount: 3,
+        httpRetryDelayFallback: 10,
+        httpRetryDelayLimit: 100,
```
rubensworks: These 3 lines can be removed, since we're just querying over an rdfjs source.

rubensworks (changes requested):
Looks great!

And nice that you also fixed the parser issues. This shows that running these spec tests already pays off :-)

rubensworks:
The traqula bump is not necessary. We’ll do that in another PR. So the traqula-specific spec tests are allowed to fail here.

rubensworks (approved):
Looks good! Will merge after #1720

rubensworks:
@voidten Could you resolve the yarn.lock merge conflict? I can merge after that.

rubensworks:
I suspect you reset the lockfile, which caused the latest traqula version to be selected (which has some issues). Best to start from the lockfile on master and install from that.

rubensworks:
Nvm my earlier comment. Didn’t see your new commit yet :-)

rubensworks:
Thanks!

PR #1724, opened 2026-07-03
Refactor to remove Node.js polyfills and implement patches for browser compatibility

smessie:
This pull request improves browser compatibility and build reliability by removing the need for Node.js polyfills in the browser bundle. Instead, it introduces targeted patches for dependencies that previously required Node globals, streamlining the build process and reducing bundle complexity. The most important changes are grouped below:

**Build Process and Dependency Management**

* The `postinstall` script in `package.json` now runs `patch-package` to apply dependency patches automatically after installation.
* Added `patch-package` and `postinstall-postinstall` as devDependencies in `package.json`.
* Removed the `vite-plugin-node-polyfills` dependency from `package.json` and from the Vite config, eliminating the need for global Node.js polyfills in browser builds. [[1]](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L85) [[2]](diffhunk://#diff-abaa5bd92bd9d997db426c85c58f90738fa160fd74c02a3c9b262c250a7defe7L3-R14) [[3]](diffhunk://#diff-abaa5bd92bd9d997db426c85c58f90738fa160fd74c02a3c9b262c250a7defe7L24-L39)

**Browser Compatibility Patches**

* Added a patch for `readable-stream` to use a bundled process shim instead of relying on global polyfills, resolving issues with `process.nextTick` in the browser.
* Added a patch for `graphql` to guard access to `process.env.NODE_ENV`, preventing runtime errors when `process` is undefined in the browser.
* Added a patch for `http-link-header` to check for the existence of `Buffer` before using it, improving compatibility in browser environments.

**Code Cleanup**

* Removed the manual `process` polyfill injection from `ActorInitQuery-browser.ts`, as it is no longer needed due to the new patching approach.
* Updated comments and removed related code in `vite.config.base.ts` to reflect that Node.js polyfills are no longer required, thanks to the new patches.

These changes collectively ensure that browser builds are more robust and maintainable by handling Node.js-specific dependencies through targeted patches rather than global polyfills.


rubensworks:
Great that these patches indicate where the changes need to be made, but I'm not a fan of patching code in external dependencies.
Shall we make a PR to readable-stream to upstream these changes?

PR #1725, opened 2026-07-05
Compute rdfjs source cardinality metadata lazily, with a pattern cache

jeswr:
### Summary

`QuerySourceRdfJs` eagerly counts the cardinality of every pattern it is asked to evaluate, to populate the stream's `metadata` property. Profiling CSS-shaped workloads (SPARQL over in-memory rdf-js stores) on WatDiv shows this counting — not actual matching — dominating join-heavy queries: under a bind-join, the engine

1. counts patterns whose metadata is **never consumed** (bound single-pattern sub-operations mediate straight to `ActorQueryOperationSource`, whose `metadata` thunk is never invoked by the bind-join), and
2. **re-counts identical patterns once per outer binding** during recursive re-planning of the remaining join entries — including fully-unbound patterns whose counts cannot change — at O(matching quads) per count in stores such as N3.js (`N3Store._countInIndex`).

This PR makes the metadata computation lazy and caches pattern cardinalities:

- **`deferMetadata` utility** (in `@comunica/utils-metadata`): defers a stream's metadata computation until the property is first requested via `getProperty`. Applied to both `queryBindings` flows of `QuerySourceRdfJs` (`match` and `matchBindings`).
- **Laziness propagation through skolemization**: `skolemizeQuadStream` / `skolemizeBindingsStream` previously registered their metadata-inheritance listener eagerly, which would have forced the source's computation; they now inherit lazily too.
- **Pattern cardinality LRU cache** in `QuerySourceRdfJs` (new `cacheSize` parameter on `ActorQuerySourceIdentifyRdfJs`, default 1024, `0` disables — mirroring the COUNT-query cache of `QuerySourceSparql`). The cache is invalidated whenever the source's reported `size` changes, and is only enabled for sources exposing a numeric `size` (e.g. `Store`/`DatasetCore` implementations).

### Profile evidence (before)

WatDiv scale-10 (1,079,876 quads) in an `n3.Store`, `@comunica/query-sparql` from this monorepo (`d55a644`), Node v22.23.1, `--cpu-prof`:

- WatDiv **C2**: 47.6 s wall, **76.1 % of wall time inside `store.countQuads`** (11,294 calls totalling 36.9 s; instrumented separately). The six fully-unbound predicate patterns of the query were each re-counted 1,240–1,481× (e.g. `?v4 wsdbm:makesPurchase ?v7`, cardinality 15,000: 1,372 counts = 18.1 s).
- Mixed workload (L2,L3,S1,S4,F1,F3,F5,C1,C3): `N3Store._countInIndex` alone = **30.8 % of total self time**, plus 19.7 % GC largely fed by the same churn.

### After

Same C2 query: countQuads drops from 11,294 calls / 36,916 ms to **2,295 calls / 102 ms (0.7 % of wall)**; `store.match` calls are unchanged (11,294), i.e. the actual query evaluation work and results are identical.

### Query-level benchmarks

3 alternating baseline/branch pairs (fresh process per leg, same machine/era; medians of reps; WatDiv scale-10 over `n3.Store`, first template instantiation). Machine is a shared 2-core box, so ratios are the claim, not absolutes:

| Query | base p1 | fix p1 | base p2 | fix p2 | base p3 | fix p3 | speedup |
|---|---:|---:|---:|---:|---:|---:|---:|
| F1 | 168 ms | 107 ms | 169 ms | 115 ms | 171 ms | 151 ms | 1.1–1.6× |
| F3 | 1,851 ms | 436 ms | 1,393 ms | 395 ms | 1,837 ms | 393 ms | **3.5–4.7×** |
| F5 | 1,153 ms | 332 ms | 1,118 ms | 382 ms | 1,488 ms | 355 ms | **2.9–4.2×** |
| C1 | 4,803 ms | 1,318 ms | 3,709 ms | 1,088 ms | 3,445 ms | 851 ms | **3.4–4.1×** |
| C2 | 50,596 ms | 13,907 ms | 47,465 ms | 13,110 ms | 47,737 ms | 13,442 ms | **3.6×** |
| C3 | 6,397 ms | 4,031 ms | 4,639 ms | 3,737 ms | 3,721 ms | 3,457 ms | 1.1–1.6× |

Result counts identical across all templates in all runs (e.g. C3: 48,802 bindings). L/S templates, property paths, trivial selects (~0.8 ms/op engine overhead) and `INSERT`/`DELETE` update workloads show **no measurable change** (within noise, verified with interleaved in-process A/B batches in both leg orders).

### Correctness considerations

- Metadata contents are unchanged; only *when* the counting happens changes. It was already computed asynchronously relative to stream consumption before this PR.
- The cache is keyed on the nullified pattern terms (`termToString`, JSON-encoded to avoid ambiguity) and cleared whenever `source.size` differs from the size at fill time. A mutation that leaves the store at exactly the same size between two queries could serve a stale cardinality until the size next changes; cardinality metadata is planning-advisory, and `cacheSize: 0` opts out. Sources without a synchronous `size` are never cached.
- `setMetadata` no longer mutates `operation.graph` when substituting the union-default-graph dummy (it previously left the dummy variable assigned to the caller's operation object in the `matchBindings` flow).

### Tests

- New unit tests: `deferMetadata` (utils-metadata), cardinality caching/invalidations/laziness/`cacheSize: 0`/fallback counting (actor-query-source-identify-rdfjs), lazy + invalidation-driven metadata inheritance (skolemize utils).
- Changed lib files at 100 % statement/branch/function coverage in per-package runs; affected consumer suites green: `bus-rdf-join`, all `actor-rdf-join-*`, `actor-abstract-path`, `actor-optimize-query-operation-prune-empty-source-operations`, `actor-query-source-identify-hypermedia-none`, `actor-query-source-identify-compositefile`, and the `engines/query-sparql` system suite (123 tests). Lint clean on changed files.

### Notes

Generated by an agent (Claude Fable 5) on @jeswr's behalf — draft for maintainer review. The residual C2 cost (≈13 s) is the bind-join re-*executing* the unbound star-shaped component once per outer binding (`store.match` on unbound patterns 1,372×); that is a join-planning matter, out of scope here.

### DIFF
--- a/packages/actor-optimize-query-operation-query-source-skolemize/lib/utils.ts
+++ b/packages/actor-optimize-query-operation-query-source-skolemize/lib/utils.ts
@@ -8,6 +8,7 @@ import type {
 } from '@comunica/types';
 import { Algebra, AlgebraFactory, algebraUtils } from '@comunica/utils-algebra';
 import { BlankNodeScoped } from '@comunica/utils-data-factory';
+import { deferMetadata } from '@comunica/utils-metadata';
 import type * as RDF from '@rdfjs/types';
 import type { AsyncIterator } from 'asynciterator';
 import { mapTermsNested } from 'rdf-terms';
@@ -102,7 +103,8 @@ export function skolemizeQuadStream(
       metadata.state.addInvalidateListener(inheritMetadata);
     });
   }
-  inheritMetadata();
+  // Inherit lazily upon the first request, so that sources can avoid computing unrequested metadata.
+  deferMetadata(ret, inheritMetadata);
   return ret;
 }
 
@@ -125,7 +127,8 @@ export function skolemizeBindingsStream(
       metadata.state.addInvalidateListener(inheritMetadata);
     });
   }
-  inheritMetadata();
+  // Inherit lazily upon the first request, so that sources can avoid computing unrequested metadata.
+  deferMetadata(ret, inheritMetadata);
   return ret;
 }
 
--- a/packages/actor-optimize-query-operation-query-source-skolemize/test/utils-test.ts
+++ b/packages/actor-optimize-query-operation-query-source-skolemize/test/utils-test.ts
@@ -1,0 +1,79 @@
+import { BindingsFactory } from '@comunica/utils-bindings-factory';
+import { MetadataValidationState } from '@comunica/utils-metadata';
+import type * as RDF from '@rdfjs/types';
+import { ArrayIterator } from 'asynciterator';
+import { DataFactory } from 'rdf-data-factory';
+import { skolemizeBindingsStream, skolemizeQuadStream } from '../lib';
+
+const DF = new DataFactory();
+const BF = new BindingsFactory(DF);
+
+describe('utils', () => {
+  describe('skolemizeQuadStream', () => {
+    let inner: ArrayIterator<RDF.Quad>;
+    beforeEach(() => {
+      inner = new ArrayIterator([
+        DF.quad(DF.namedNode('s'), DF.namedNode('p'), DF.blankNode('o')),
+      ], { autoStart: false });
+    });
+
+    it('should inherit metadata lazily upon the first request', async() => {
+      const state = new MetadataValidationState();
+      inner.setProperty('metadata', { state, cardinality: { type: 'exact', value: 1 }});
+      const out = skolemizeQuadStream(DF, inner, '0');
+      await expect(new Promise(resolve => out.getProperty('metadata', resolve))).resolves
+        .toEqual(expect.objectContaining({ cardinality: { type: 'exact', value: 1 }}));
+    });
+
+    it('should re-inherit metadata upon invalidation', async() => {
+      const state = new MetadataValidationState();
+      inner.setProperty('metadata', { state, cardinality: { type: 'exact', value: 1 }});
+      const out = skolemizeQuadStream(DF, inner, '0');
+      await expect(new Promise(resolve => out.getProperty('metadata', resolve))).resolves
+        .toEqual(expect.objectContaining({ cardinality: { type: 'exact', value: 1 }}));
+
+      inner.setProperty('metadata', {
+        state: new MetadataValidationState(),
+        cardinality: { type: 'exact', value: 2 },
+      });
+      state.invalidate();
+      await new Promise(resolve => setImmediate(resolve));
+      await expect(new Promise(resolve => out.getProperty('metadata', resolve))).resolves
+        .toEqual(expect.objectContaining({ cardinality: { type: 'exact', value: 2 }}));
+    });
+  });
+
+  describe('skolemizeBindingsStream', () => {
+    let inner: ArrayIterator<RDF.Bindings>;
+    beforeEach(() => {
+      inner = new ArrayIterator<RDF.Bindings>([
+        <RDF.Bindings> BF.fromRecord({ a: DF.blankNode('a0') }),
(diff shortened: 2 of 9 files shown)


jeswr:
**Follow-up: where the residual WatDiv C2 time goes (design notes, no code change proposed here).**

After this PR, C2 (scale-10, 1.08 M quads in an `n3.Store`) evaluates in ~13.4 s (was 47.6 s). We profiled and instrumented the residual before attempting a fix; the findings suggest any further win is a planning/architecture matter, so we're documenting them instead of forcing a patch. All figures from `--cpu-prof` + prototype-level instrumentation on the post-PR tree (`bdf161b`), shared 2-core box, ratios are the claim.

**Physical plan shape** (`explain physical`): multi-bind on `?v0 legalName ?v1` (12 retailers) → per retailer, multi-bind on `(R offers ?v2)` (~60–130 offers) → per offer, a recursive join of the 6–8 remaining (mostly unbound) entries. That inner join is instantiated **~1,372 times** (once per retailer×offer).

**Residual decomposition** (13.4 s wall, 25.8 s total self time incl. ~11 s dataset load):
- `countQuads` is gone: 11,294 calls / 36.9 s → 2,295 calls / 0.10 s (0.5 % of wall).
- **Data volume is NOT the cost**: the 11,285 per-occurrence entry streams pull only **136,994 quads total** (~16 per stream — first-buffer fills) before being discarded; `store.match` emissions to consumers total 1,589 quads.
- The cost is **per-occurrence constant overhead ≈ 10 ms × 1,372 occurrences**: stream-chain construction and teardown (`N3Store` stream machinery ≈ 23.7 % self), GC from the same churn (**31.5 % self**), mediation/context/metadata plumbing (long tail). Each occurrence constructs full bindings-stream chains for 6–8 entries, most of which the join actor immediately destroys (non-smallest entries; ~90 % of offer-level joins are killed by `ActorRdfJoinMultiEmpty` because `eligibleRegion Country5` has cardinality 0 for that offer).

**Why we didn't patch it:**
1. *Cost-model angle*: `ActorRdfJoinMultiBind`'s coefficients model per-binding cost as cardinality-proportional with `requestInitialTimes` ≈ 0 for local sources — the ~10 ms/occurrence constant (mediation + stream construction + GC share) is unmodeled, so bind wins over hash/smallest for shapes where it shouldn't. Fixing this means calibrating a constant per-sub-operation overhead term in the join cost model — config-sensitive and benchmark-wide in impact; it needs maintainer input on how (and whether) to parameterize it.
2. *Laziness angle*: constructing join-entry streams lazily (deferred `TransformIterator` sources) helps only entries that are never read; here most entries **are** touched (~16 quads) by the bind machinery's eager sub-stream buffering (`autoStart` on sub-bindings, `maxBufferSize: 128`), which exists deliberately ("Enable auto-start on sub-bindings during depth-first binding for best performance").
3. *Caching angle*: memoizing results of binding-invariant sub-operations (the unbound `?v4`-star is re-evaluated per offer) is the biggest possible win but introduces result-lifetime/mutation semantics that don't currently exist in the engine.

If maintainers see value in any of these directions (esp. 1), we're happy to build the benchmark evidence for a concrete proposal — the harness (WatDiv scale-10 over rdf-js stores + per-pattern instrumentation) is reproducible from the PR description.

*Generated by an agent (Claude Fable 5) on @jeswr's behalf.*

rubensworks:
We recently did a similar change wrt cardinality caching.
The metadata deferring is not done yet though; how significant is it in terms of performance?

PR #1726, opened 2026-07-05
Compute SPARQL source metadata lazily, skip count queries for bound operations

jeswr:
### Summary

Follow-up to #1725, applying the same lazy-metadata principle to `QuerySourceSparql` (**stacked on #1725** — it uses the `deferMetadata` utility introduced there; the last commit is this PR's own diff).

Two changes in `QuerySourceSparql`:

1. **Lazy metadata attachment**: `attachMetadata` (which issues a `SELECT (COUNT(*) ...)` request to the endpoint) now runs upon the first request of the stream's `metadata` property instead of on every `queryBindings`/`queryQuads` call. Count queries are avoided entirely when the metadata is never consumed.
2. **No count queries for bound operations**: sub-operations evaluated with `joinBindings` (the chunked `VALUES` pushdowns issued by `ActorRdfJoinMultiBindSource`) no longer send a COUNT query, and report `{ type: 'estimate', value: ∞ }` instead — the same convention already used when `cardinalityCountQueries` is disabled or the count times out. Rationale: the join was already planned on the *unbound* operation's cardinality; the per-chunk exact COUNT costs one extra endpoint request per bound chunk, and (verified by tracing) its only consumer in the default config is the hypermedia metadata accumulator, whose per-iterator metadata read is structural (link traversal) rather than cardinality-driven.

### Measured (end-to-end, default `@comunica/query-sparql` config)

Two local HTTP SPARQL endpoints (in-process, backed by an unmodified engine over rdf-js stores, constant across legs), federated join `?s <urn:bought> ?m . ?m <urn:item> ?o` across 300 links (~19 bound chunks of the default block size), fresh engine per leg, 2 alternating pairs:

| leg | data requests | COUNT requests | total | results |
|---|---:|---:|---:|---:|
| master (`d55a644`) | 20 | 23 | 43 | 300 |
| this PR | 20 | **4** | **24** | 300 |

**44 % fewer endpoint requests**; the 4 remaining COUNTs are genuinely consumed (per-pattern source selection by `ActorOptimizeQueryOperationPruneEmptySourceOperations` and join planning). Results identical in all runs. Wall-clock on the (noisy, shared 2-core) measurement box was not treated as evidence; the request counts are deterministic across runs.

Single-endpoint whole-query pushdown (scenario A) issues no COUNT on master either (the count path is not reached for non-pattern root operations), so no change there — noted for honesty.

### Why not defer in the hypermedia layer too

`LinkedRdfSourcesAsyncRdfIterator` requests every sub-iterator's metadata eagerly, but that read is load-bearing: metadata carries the next-page links and gates traversal termination (`startIteratorsForNextUrls`). Making that lazy would require a way for sources to declare "no next links" — a design-level change we did not attempt here. This PR removes the wasted *endpoint requests* behind that read instead.

### Tests

- All 98 tests in `actor-query-source-identify-hypermedia-sparql` green; `QuerySourceSparql.ts` coverage at parity with master (99.49 % stmts / 98.55 % branch; the only uncovered line pre-dates this PR).
- New/updated tests: laziness (no COUNT when metadata unread), estimate metadata for `joinBindings` operations, count-response error path, non-pattern and non-variable-term count normalization, local-cardinality-estimation path now asserts the estimated metadata.
- `engines/query-sparql` system suite (123 tests) green. Lint clean.

### Notes

Generated by an agent (Claude Fable 5) on @jeswr's behalf — draft for maintainer review. If maintainers prefer the bound-operation count skip behind a flag (e.g. alongside `cardinalityCountQueries`), that is a small follow-up — kept flag-free here to minimize config surface.

### DIFF
--- a/packages/actor-optimize-query-operation-query-source-skolemize/lib/utils.ts
+++ b/packages/actor-optimize-query-operation-query-source-skolemize/lib/utils.ts
@@ -8,6 +8,7 @@ import type {
 } from '@comunica/types';
 import { Algebra, AlgebraFactory, algebraUtils } from '@comunica/utils-algebra';
 import { BlankNodeScoped } from '@comunica/utils-data-factory';
+import { deferMetadata } from '@comunica/utils-metadata';
 import type * as RDF from '@rdfjs/types';
 import type { AsyncIterator } from 'asynciterator';
 import { mapTermsNested } from 'rdf-terms';
@@ -102,7 +103,8 @@ export function skolemizeQuadStream(
       metadata.state.addInvalidateListener(inheritMetadata);
     });
   }
-  inheritMetadata();
+  // Inherit lazily upon the first request, so that sources can avoid computing unrequested metadata.
+  deferMetadata(ret, inheritMetadata);
   return ret;
 }
 
@@ -125,7 +127,8 @@ export function skolemizeBindingsStream(
       metadata.state.addInvalidateListener(inheritMetadata);
     });
   }
-  inheritMetadata();
+  // Inherit lazily upon the first request, so that sources can avoid computing unrequested metadata.
+  deferMetadata(ret, inheritMetadata);
   return ret;
 }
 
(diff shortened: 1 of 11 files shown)


rubensworks:
@jeswr Is this one still planned? I see some merge conflicts.

PR #1729, opened 2026-07-06
Added new rdf parser and rdf writer actors

pietercolpaert:
Comunica today uses N3.js for parsing and serializing Turtle, TriG, N-Triples and N-Quads. N3.js scope however is bigger than just that. This pull request proposes to instead use a dedicated parser and writer that only does this. This ensures the bundle size is smaller. On performance tests, the RDF1.2 compliant parser outperforms N3.js between 1x and 2x.


pietercolpaert:
With the fix of Node 22, this should now be ready for review

PR #1732, opened 2026-07-22
Run Service Description Tests

voidten:
This PR adds `spec:sd-1-1`, which starts a local Comunica endpoint with a minimal RDF source, waits for it to become available, and runs the Service Description conformance suite against that endpoint to the `spec` command.


engines/query-sparql/spec/sparql-engine.js
```diff
@@ -1,2 +1,59 @@
+/* eslint-disable import/no-nodejs-modules */
+const http = require('node:http');
+const path = require('node:path');
+/* eslint-enable import/no-nodejs-modules */
+const { HttpServiceSparqlEndpoint } = require('@comunica/actor-init-query');
+const createTestEngine = require('@comunica/actor-init-query/spec/sparql-engine-base.js');
 const QueryEngine = require('@comunica/query-sparql').QueryEngine;
-module.exports = require('@comunica/actor-init-query/spec/sparql-engine-base.js')(new QueryEngine());
```
rubensworks, line 2: Instead of making the changes within this file here, could be do it within `actor-init-query/spec/sparql-engine-base.js`?

And that exported function could take an optional argument to indicate whether or not to expose this `startServiceDescriptionEndpoint` function.

rubensworks (changes requested):

engines/query-sparql/spec/sparql-engine.js
voidten: Done 👍

engines/query-sparql/package.json
```diff
@@ -71,7 +71,7 @@
     "spec:prot-1-1": "yarn run --silent spec:base-1-1 -s http://www.w3.org/TR/sparql11-protocol/",
     "spec:graphstore-1-1": "yarn run --silent spec:base-1-1 -s http://www.w3.org/TR/sparql11-http-rdf-update/",
     "spec:sparql12-xsd-datetime-duration": "node ../../node_modules/rdf-test-suite/bin/Runner.js spec/sparql-engine.js https://raw.githubusercontent.com/kasei/sparql-12/xsd_datetime_duration/tests/xsd_functions/manifest.ttl -c ../../.rdf-test-suite-cache/ --skip \"http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#adjust_.*\"",
-    "spec": "yarn run spec:query-1-0 && yarn run spec:query-1-1 && yarn run spec:update-1-1 && yarn run spec:csv-tsv-1-1 && yarn run spec:json-1-1 && yarn run spec:sparql12-xsd-datetime-duration && yarn run spec:query-1-2",
+    "spec": "yarn run spec:query-1-0 && yarn run spec:query-1-1 && yarn run spec:update-1-1 && yarn run spec:csv-tsv-1-1 && yarn run spec:json-1-1 && yarn run spec:sd-1-1 && yarn run spec:sparql12-xsd-datetime-duration && yarn run spec:query-1-2",
```
rubensworks, line 74: Could you also enable the corresponding earl script?

packages/actor-init-query/spec/sparql-engine-base.js
```diff
@@ -79,3 +89,51 @@ function source(data) {
...
+      });
+    });
+
+    const address = server.address();
+    if (!address || typeof address === 'string') {
+      await close();
+      throw new Error('Could not determine the service description endpoint address.');
+    }
+
+    const service = new HttpServiceSparqlEndpoint({ port: address.port });
```
rubensworks: One problem here: the `engine` above is not injected into `HttpServiceSparqlEndpoint`.
I think we should modify `HttpServiceSparqlEndpoint` to accept an optional `engine` in `IHttpServiceSparqlEndpointArgs`.

rubensworks (changes requested):

packages/actor-init-query/spec/sparql-engine-base.js
voidten: Done 👍

rubensworks (approved):
Looks great, thanks @voidten!

PR #1734, opened 2026-07-27
ActorRdfJoinOperationBind rejects on deep incompatible operations

jitsedesmet:
There is a discrepancy between the two bind Join actors on the conditions for their correct execution. I unified the condition through a public static function.

I can confirm that this branch fixes my local problem and added some integration tests showcasing the issue.


rubensworks:
Good catch!

PR #1619, opened 2025-09-12
#1501: Support lesser-than operator for IRIs and bnodes

jitsedesmet:
This PR will finish the work of @noahvsb performed on PR #1577

PR #1735, opened 2026-08-03
#1501: Support lesser-than operator for IRIs and bnodes

noahvsb:
Taking over #1619, which took over #1577.


packages/actor-function-factory-term-lesser-than-equal/lib/TermFunctionLesserThanEqual.ts
noahvsb, line 1: Uses the `lessThanFunction` to make the logic more consistent with `TermFunctionLesserThan.ts`, with the exception of a special case for the numbers, see comments in code.

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
```diff
@@ -1,71 +1,108 @@
...
-        .copy({
+        .set(
+          [ TypeURL.RDF_LANG_STRING, TypeURL.RDF_LANG_STRING ],
+          () => ([ left, right ]: LangStringLiteral[]) => {
+            if (left.str() !== right.str()) {
+              return bool(left.str() < right.str());
+            }
+            return bool(left.language < right.language);
+          },
+        )
```
noahvsb, line 61: The order of language strings is undefined: https://www.w3.org/TR/sparql11-query/#modOrderBy. But if this isn't defined, it will say `"a"@en` and `"a"@de` are equal, which is even more wrong. 

See https://github.com/comunica/comunica/pull/1577#discussion_r2185392293.

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
```diff
@@ -1,71 +1,108 @@
...
 export class TermFunctionLesserThan extends TermFunctionBase {
   public constructor(private readonly equalityFunction: ITermFunction) {
     super({
       arity: 2,
       operator: SparqlOperator.LT,
       overloads: declare(SparqlOperator.LT)
-        .numberTest(() => (left, right) => left < right)
+        .set(
+          [ C.TypeAlias.SPARQL_NUMERIC, C.TypeAlias.SPARQL_NUMERIC ],
+          exprEval => this.compareLiterals(exprEval),
```
noahvsb, line 47: This more complex approach with `compareLiterals` is used to handle non-lexical literals. You will see it appear a lot throughout this file.

packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts
noahvsb, line 1: Implemented this using only the `lesserThanFunction` for more consistent logic.

This in combination with setting `KeysExpressionEvaluator.nonLiteralExpressionComparison` to `true` in the actor, led to the fallback being unused and so it was removed.

noahvsb (commented):
# recap

#1501 was fixed, now the lesser-than operator has optional support for IRIs and bnodes. A new context option `nonLiteralExpressionComparison` will enable this: if true, it compares non literal expressions instead of throwing an error. This also includes non-lexical literals. The option defaults to false.

See comments per files for additional smaller changes.

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -1826,6 +1826,29 @@ SELECT ?option WHERE {
       });
     });
 
+    it('nonLiteralExpressionComparison set to true', async() => {
```
jitsedesmet:
```suggestion
    describe('nonLiteralExpressionComparison set to true', async() => {
       it('compares nonLexical literals', () => {
```

packages/utils-jest/lib/expressionEvaluator/generalEvaluation.ts
```diff
@@ -16,7 +16,7 @@ const BF = new BindingsFactory(DF);
 export interface IGeneralEvaluationArg {
   bindings?: RDF.Bindings;
   expression: string;
-  generalEvaluationConfig?: IActionContext;
+  evaluationActionContext?: IActionContext;
```
jitsedesmet: This rename is a breaking API change, so we cannot do it.

packages/utils-expression-evaluator/lib/index.ts
```diff
@@ -83,6 +83,7 @@ export {
   NonLexicalLiteral,
   StringLiteral,
   isNonLexicalLiteral,
+  ISerializable,
```
jitsedesmet, line 87: is the import needed? That would be weird since Nothing else changed?

packages/types/lib/IActionContext.ts
```diff
@@ -15,6 +15,11 @@
  * For instance, this context can contain a list of datasources over which operators should query.
  */
 export interface IActionContext {
+  /**
+   * @param key The key to set in the new copy
+   * @param value The value that should be assigned to the key in the new copy
+   * @return A copy of the original context with relating the key to the new value.
```
jitsedesmet:
```suggestion
   * @return A copy of the original context, but also mapping the key to the new value.
```
Something like this?

packages/actor-term-comparator-factory-expression-evaluator/test/TermComparator-test.ts
```diff
@@ -39,17 +43,17 @@ function dateTime(value: string): RDF.Literal {
...
-  const context = typeDiscoveryCallback ?
+  const context = (typeDiscoveryCallback ?
     getMockEEActionContext().set(KeysExpressionEvaluator.superTypeProvider, {
       discoverer: typeDiscoveryCallback,
       cache: new LRUCache<string, any>({ max: 1_000 }),
     }) :
-    getMockEEActionContext();
-  const equalityFunc = new TermFunctionEquality();
+    getMockEEActionContext())
+    .set(KeysExpressionEvaluator.nonLiteralExpressionComparison, true);
```
jitsedesmet: Can you split this up? It becomes a little hard to read.

packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts
jitsedesmet: Are you correctly changing the context such that the comparator called though this function is actually calles with the compare nonLiterals true? (I don't seem to find it?)

jitsedesmet: Nvm, found it. Maybe we should add some JSComment on the requested InternalEValuator in the constructor here saying that it expects this?

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
jitsedesmet: I don't see it. If we do not define it, we will throw right? When is this casted to false?

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
```diff
@@ -81,11 +118,52 @@ export class TermFunctionLesserThan extends TermFunctionBase {
...
+    return (args) => {
+      const nonLexical = args.find(arg => arg instanceof NonLexicalLiteral);
+      if (nonLexical) {
+        if (this.shouldThrowNonLexicalError(exprEval)) {
+          throw new Err.InvalidLexicalForm(
+            nonLexical.toRDF(exprEval.context.getSafe(KeysInitQuery.dataFactory)),
+          );
+        }
+        const [ left, right ] = args;
+        return this.comparePrimitives(left.str(), right.str()) === -1;
```
jitsedesmet: This would mean as soon as one is nonLexical, we compare by value.

I wonder if this is the kind of comparison we want. Maybe it would be nice if the non-lexicals grouped together?

So you would have, for example 
1. all IRIs
2. all bNodes
3. all Literals
4. all nonLexical literals.

It might be a style choice, but it's something we should think about an something we document clearly. @rubensworks

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
```diff
@@ -102,4 +180,61 @@ export class TermFunctionLesserThan extends TermFunctionBase {
...
+
+  // SPARQL specifies that blankNode < namedNode < literal. Sparql star expands with < quads and we say < defaultGraph:
+  // https://www.w3.org/TR/sparql11-query/#modOrderBy
+  private readonly _TERM_ORDERING_PRIORITY = {
+    blankNode: 0,
+    namedNode: 1,
+    literal: 2,
+    quad: 3,
+    defaultGraph: 4,
+  };
```
jitsedesmet, line 242: I wonder whether this is documented clearly? In jsdocs but maybe also on the website. Since this behavior can be enabled to run in the evaluator, the expr eval docs might be a good place?

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
```diff
@@ -102,4 +180,61 @@ export class TermFunctionLesserThan extends TermFunctionBase {
     );
     return (<BooleanLiteral>componentLess).typedValue;
   }
+
+  private lesserThanTerms(termA: Term, termB: Term, exprEval: IInternalEvaluator): boolean {
+    if (this.shouldThrowNonLexicalError(exprEval)) {
+      throw new InvalidArgumentTypes([ termA, termB ], SparqlOperator.LT, `
+To enable comparison, set the ${KeysExpressionEvaluator.nonLiteralExpressionComparison.name} flag to true.`);
+    }
```
jitsedesmet: I feel like the context option is doing two different thing that we might want to split:

1. `nonLexicalComaprison`: enables whether non lexical literals can be compared.
2. `fullTermComparison`: enables whether bnodes, IRIs and by extension Triple Terms can be compared.

To me they are distinct since one is a expression evaluator restriction based on dataType while the other is a restriction on term types.


@rubensworks would splitting make sense?

jitsedesmet (changes requested):
Nice work! I have some feedback and questions to @rubensworks :)

packages/utils-expression-evaluator/lib/index.ts
noahvsb: it's used in `TermFunctionLesserThan.ts`

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
noahvsb: If you would throw, then [this test](https://github.com/noahvsb/comunica/blob/88e4d7d13ea13739bc484fdd8d383cadb68db8f1/packages/actor-term-comparator-factory-expression-evaluator/test/TermComparator-test.ts#L143) would fail, because `"a"@de` would be seen as equal to `"a"@en`:

([TermComparatorExpressionEvaluator.ts](https://github.com/noahvsb/comunica/blob/88e4d7d13ea13739bc484fdd8d383cadb68db8f1/packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts))
ln 18: termA === termB would resolve to false, because they're not equal
ln 33 and 37: the literal values are compared, the `"a"`'s, and since these are equal both if's don't get taken
ln 41: 0 is returned

However I ran the spec test and they still succeed if you throw, so I think the test is wrong and should be changed, what are your thoughts on this?

packages/context-entries/lib/Keys.ts
```diff
@@ -273,6 +273,15 @@ export const KeysExpressionEvaluator = {
     '@comunica/utils-expression-evaluator:defaultTimeZone',
   ),
   actionContext: new ActionContextKey<IActionContext>('@comunica/utils-expression-evaluator:actionContext'),
+  /**
+   * A boolean denoting the behaviour of the lesser than function when used with non literals operands.
```
rubensworks:
```suggestion
   * A boolean denoting the behaviour of the lesser than function when used with non-literals operands.
   * Such non-literals are IRIs and blank nodes. Their comparison is not required by default in SPARQL.
```

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
rubensworks: For this, I would do whatever is easiest to implement, as long as it's consistent.
According to @jitsedesmet, this would probably be the interleaved approach.

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
rubensworks: Yes, let's split!

rubensworks (changes requested):

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
noahvsb: See https://github.com/comunica/website/pull/77

packages/actor-function-factory-term-lesser-than-equal/lib/TermFunctionLesserThanEqual.ts
```diff
@@ -2,52 +2,43 @@ import type { ITermFunction } from '@comunica/bus-function-factory';
...
   bool,
   declare,
   SparqlOperator,
 } from '@comunica/utils-expression-evaluator';
+import { nonLexicalHandler } from '@comunica/utils-expression-evaluator/lib/functions/Helpers';
+import * as C from '@comunica/utils-expression-evaluator/lib/util/Consts';
 
 export class TermFunctionLesserThanEqual extends TermFunctionBase {
   public constructor(
-    private readonly equalityFunction: ITermFunction,
```
rubensworks, line 14: This is strictly a breaking change.
Can we keep it, but with a TODO comment? (`// TODO: remove in next major, as it's unused`).

packages/actor-term-comparator-factory-expression-evaluator/lib/ActorTermComparatorFactoryExpressionEvaluator.ts
```diff
@@ -25,7 +25,9 @@ export class ActorTermComparatorFactoryExpressionEvaluator extends ActorTermComp
    * @param context.context IActionContext
    */
   public async run({ context }: IActionTermComparatorFactory): Promise<IActorTermComparatorFactoryOutput> {
-    context = Eval.prepareEvaluatorActionContext(context);
+    context = Eval.prepareEvaluatorActionContext(context)
+      .set(KeysExpressionEvaluator.nonLexicalComparison, true)
+      .set(KeysExpressionEvaluator.fullTermComparison, true);
```
rubensworks, line 30: Does this force these options to always be true?
I guess we don't want this?

packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts
```diff
@@ -6,8 +6,8 @@ import type * as RDF from '@rdfjs/types';
 
 export class TermComparatorExpressionEvaluator implements ITermComparator {
   public constructor(
+    // The internal evaluator is expected to have a context with nonLexicalComparison and fullTermComparison set to true
     private readonly internalEvaluator: InternalEvaluator,
-    private readonly equalityFunction: ITermFunction,
```
rubensworks, line 10: Same as above, keep to avoid a breaking change.

rubensworks (changes requested):

packages/actor-term-comparator-factory-expression-evaluator/lib/ActorTermComparatorFactoryExpressionEvaluator.ts
noahvsb: @jitsedesmet added this last year if I remember correctly, this way the fallback to string-based comparison can be left out.

rubensworks: I may not fully understand what's happening here, and why. Can you give some more details?

noahvsb: If you take a look at the previous implementation of `TermComparatorExpressionEvaluator.ts`, you see that a ton of extra work is needed besides the usage of the equality and lesser than functions, including the string comparison fallback. With these options set to true and using only the lesser than function, all of this extra work can be left out. Because the lesser than function now handles it.

rubensworks: Ok, I get it now, thanks for the clarification!

noahvsb:
The last commit's message explains only half of what I did, I also fixed the failing jest tests.

rubensworks (approved):
Looks good to me!

If @jitsedesmet also approves, I will merge.

packages/utils-expression-evaluator/lib/functions/Helpers.ts
```diff
@@ -200,6 +200,7 @@ addInvalidHandling = false,
     return this.set(
       [ 'term', 'term', 'term' ],
       expressionEvaluator => ([ t1, t2, t3 ]: [Term, Term, Term]) => op(expressionEvaluator)(t1, t2, t3),
+      false,
```
jitsedesmet: Why this change? it seems like the default is true? So the behavior between `onTerm3` would differ from that of `onTerm1` (it might even be breaking?)

packages/utils-expression-evaluator/lib/functions/Helpers.ts
```diff
@@ -550,3 +551,20 @@ export function expressionToVar(
 ): RDF.Variable {
   return dataFactory.variable(variableExpression.name.slice(1));
 }
+
+export function nonLexicalHandler(
```
jitsedesmet: This function is in the same kind of space as `wrapInvalidLexicalProtected` already in this file. I see how they differ, but we should add some JS docs on both that describes their link/ own goal. I also think because of that, the name of this function can bet better?

`nonLexicalComparatorWrapper`

or something like that?

jitsedesmet: Also be clear on the fact that undefined in this return means `do it yourself, it is safe`

packages/utils-expression-evaluator/lib/functions/Helpers.ts
```diff
@@ -550,3 +551,20 @@ export function expressionToVar(
...
+  right: Literal<ISerializable>,
+): undefined | -1 | 0 | 1 {
+  const nonLexical = [ left, right ].find(arg => arg instanceof NonLexicalLiteral);
+  if (nonLexical) {
+    if (!exprEval.context.get(KeysExpressionEvaluator.nonLexicalComparison)) {
+      throw new Err.InvalidLexicalForm(
+        nonLexical.toRDF(exprEval.context.getSafe(KeysInitQuery.dataFactory)),
+      );
+    }
+    return left.str() === right.str() ? 0 : (left.str() < right.str() ? -1 : 1);
```
jitsedesmet, line 590: I am not entirely sure, but I thought `.str()` does not cover the dataType itself? I this a decision we want? Or this what we used to have?

packages/types/lib/IActionContext.ts
```diff
@@ -15,6 +15,11 @@
  * For instance, this context can contain a list of datasources over which operators should query.
  */
 export interface IActionContext {
+  /**
+   * @param key The key to set in the new copy
+   * @param value The value that should be assigned to the key in the new copy
+   * @return A copy of the original context, but also mapping the key to the new value.
```
jitsedesmet:
```suggestion
   * @return A copy of the original context, additionally mapping the key to the new value.
```

packages/context-entries/lib/Keys.ts
```diff
@@ -273,6 +273,28 @@ export const KeysExpressionEvaluator = {
     '@comunica/utils-expression-evaluator:defaultTimeZone',
   ),
   actionContext: new ActionContextKey<IActionContext>('@comunica/utils-expression-evaluator:actionContext'),
+  /**
+   * A boolean denoting the behaviour of the lesser than function when used with non-lexical literal operands.
```
jitsedesmet:
```suggestion
   * A boolean denoting the behaviour of the lesser than function when used with non-lexical literal operands. Non-lexical literal are literals whose value does not belong in the space of it's datatype, e.g. "not-a-number"^^xsd:int
```

packages/context-entries/lib/Keys.ts
```diff
@@ -273,6 +273,28 @@ export const KeysExpressionEvaluator = {
...
+   *
+   * true: compares both operands:
+   *         if only 1 operand is non-lexical: the lexical is lower.
+   *         if both are non-lexical: their string values are compared.
```
jitsedesmet: This behavior does not match the one implemented in the Helpers.ts? It looks like there is only a case for 'at least one is non-lexical'?

packages/context-entries/lib/Keys.ts
jitsedesmet: I would also not only talk about `<` but rather about `comparators, e.g. <, >, <=, =` etc. (For an outsider it might not be clear `<` is used as the core operator to implement the others)

packages/context-entries/lib/Keys.ts
```diff
@@ -273,6 +273,28 @@ export const KeysExpressionEvaluator = {
...
+    '@comunica/utils-expression-evaluator:nonLexicalComparison',
+  ),
+  /**
+   * A boolean denoting the behaviour of the lesser than function when used with non-literal and mixed operands.
+   * Such non-literals are IRIs, blank nodes, languageStrings and triple terms.
+   * Their comparison is not required by default in SPARQL.
+   *
+   * true: treats them as literals and compares their string values.
+   * false: throws an error (default).
+   */
```
jitsedesmet, line 294: Language strings are literals.

I would not say we treat the non-literal-terms as non lieterals since we do follow the relative ordering described in https://www.w3.org/TR/sparql12-query/#modOrderBy ?

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -1910,6 +1826,83 @@ SELECT ?option WHERE {
...
+            [ DF.variable('l2'), DF.literal('true', bool) ],
+          ],
+        ];
+
+        const bindings = (await arrayifyStream(await engine.queryBindings(`
+PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
+SELECT 
+  (( "a"^^xsd:dateTime < "b"^^xsd:boolean ) AS ?l1)
+  (( "a"^^xsd:boolean < "a"^^xsd:dateType ) AS ?l2)
+WHERE { }
```
jitsedesmet, line 1899: aha, nvm my previous comment, it seems like we do look at the datatype itself, nice.
I do wonder whether we can add a test that shows the behaviour between our tho configurations for comparing:

`"a"^^xsd:number < <ex://a>`

packages/actor-term-comparator-factory-expression-evaluator/test/TermComparator-test.ts
```diff
@@ -39,17 +43,24 @@ function dateTime(value: string): RDF.Literal {
...
 
 function orderByFactory(typeDiscoveryCallback?: SuperTypeCallback): ITermComparator {
-  const context = typeDiscoveryCallback ?
-    getMockEEActionContext().set(KeysExpressionEvaluator.superTypeProvider, {
-      discoverer: typeDiscoveryCallback,
-      cache: new LRUCache<string, any>({ max: 1_000 }),
-    }) :
-    getMockEEActionContext();
-  const equalityFunc = new TermFunctionEquality();
+  const getContext = () => {
```
jitsedesmet: Why the callBack? we evaluate only once and it has no arguments? So it could just be a construction as part of the function?

packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts
```diff
@@ -178,6 +184,16 @@ export class CliArgsHandlerBase implements ICliArgsHandler {
...
         },
+        nonLexicalComparison: {
+          type: 'boolean',
+          describe: 'When true, compares non-lexical literals.' +
+            'Throws an error otherwise.',
+        },
+        fullTermComparison: {
+          type: 'boolean',
+          describe: 'When true, compares IRIs, blank nodes, languageStrings and triple terms.' +
+            'Throws an error otherwise.',
```
jitsedesmet:
```suggestion
        fullTermComparison: {
          type: 'boolean',
          describe: 'When true, compares IRIs, blank nodes, languageStrings and triple terms.' +
            'Throws an expression error otherwise, which is caught by FILTER and BIND.',
```

packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts
```diff
@@ -178,6 +184,16 @@ export class CliArgsHandlerBase implements ICliArgsHandler {
           type: 'boolean',
           describe: 'If the query engine should deduplicate resulting triples',
         },
+        nonLexicalComparison: {
+          type: 'boolean',
+          describe: 'When true, compares non-lexical literals.' +
+            'Throws an error otherwise.',
```
jitsedesmet: Same here - expression error, not system error.

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
```diff
@@ -81,25 +121,120 @@ export class TermFunctionLesserThan extends TermFunctionBase {
...
+      const blankNode = <BlankNode> term;
+      if (typeof blankNode.value === 'string') {
+        return blankNode.value;
+      }
+      return blankNode.value.value;
+    }
+    return term.str();
+  }
+
+  private comparePrimitives(valueA: any, valueB: any): -1 | 0 | 1 {
```
jitsedesmet: I know you moved this but I saw potential ;)

```suggestion
  private comparePrimitives<T>(valueA: T, valueB: T): -1 | 0 | 1 {
```

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
```diff
@@ -1,71 +1,111 @@
...
-          toUTCDate(right, exprEval.context.getSafe(KeysExpressionEvaluator.defaultTimeZone)).getTime())
-        .copy({
+        .set(
+          [ TypeURL.RDF_LANG_STRING, TypeURL.RDF_LANG_STRING ],
+          exprEval => ([ left, right ]: LangStringLiteral[]) => {
+            this.fullTermComparisonCheck(exprEval, left, right);
+            if (left.str() !== right.str()) {
+              return bool(left.str() < right.str());
+            }
+            return bool(left.language < right.language);
```
jitsedesmet: I thought this was a linter rule? We can avoid negation (making this more complicated) while achieving the same readbility:


```suggestion
            if (left.str() === right.str()) {
              return bool(left.language < right.language);
            }
            return bool(left.str() < right.str());
```

jitsedesmet (changes requested):
Very nice work! I only have some minor nit picks on documentation that will also need to be resolved in the docs.

packages/utils-expression-evaluator/lib/functions/Helpers.ts
noahvsb: I changed it to fix a bug where it would throw a non-lexical error when comparing triple terms even with `nonLexicalComparison` set to true. For `onTerm1` the default is false for some reason, so the behaviors are equal.

jitsedesmet: Oh, seems like that's true for onTerm1. Could we add an additional argument instead and default that to true? Within the coparators we can then put it to false ourself. 

If the bug is also present for other operations, we should fix it in a separate PR.

packages/utils-expression-evaluator/lib/functions/Helpers.ts
noahvsb: This function is exclusively used for literals of the same datatype (or different numeric types).

noahvsb: Comparison of literals with different types is handled in `lesserThanTerms`.

packages/context-entries/lib/Keys.ts
noahvsb: To be clear, it hasn't been implement for `=` yet. I was going to do that in a different PR. So I'll exclude that from the examples for now.

packages/utils-expression-evaluator/lib/functions/Helpers.ts
noahvsb: I can't really think of another operation that allows non-lexical literals apart from `ORDER BY`, but that uses the lesser than function, so it's already fixed.

packages/utils-expression-evaluator/lib/functions/Helpers.ts
```diff
@@ -184,9 +192,8 @@ addInvalidHandling = true,
   }
 
   public onTerm1<T extends Term>(
-    op: (expressionEvaluator: IInternalEvaluator) =>
-    (term: T) => Term,
-addInvalidHandling = false,
+    op: (expressionEvaluator: IInternalEvaluator) => (term: T) => Term,
+    addInvalidHandling = true,
```
jitsedesmet: The default here cannot change (breaking). Maybe we can add a `TODO: next major can change default to false matching ...`

packages/utils-expression-evaluator/lib/functions/Helpers.ts
```diff
@@ -550,3 +560,33 @@ export function expressionToVar(
 ): RDF.Variable {
   return dataFactory.variable(variableExpression.name.slice(1));
 }
+
+/**
+ * If any of the arguments is a non-lexical literal, an error is thrown or string comparison is performed
```
jitsedesmet:
```suggestion
 * If any of the arguments is a non-lexical literal, an error is thrown or string comparison of their value is performed
```

packages/context-entries/lib/Keys.ts
```diff
@@ -273,6 +273,28 @@ export const KeysExpressionEvaluator = {
...
+   * e.g. "not-a-number"^^xsd:int
+   *
+   * true: compares both operands by datatypes first and if they're equal, their string values are compared.
+   * false: throws an error (default).
+   */
+  nonLexicalComparison: new ActionContextKey<boolean>(
+    '@comunica/utils-expression-evaluator:nonLexicalComparison',
+  ),
+  /**
+   * A boolean denoting the behaviour of the lesser than function when used with non-literal and mixed operands.
```
jitsedesmet: same reference to comparators can be used here, not refering only to `<`.

packages/actor-function-factory-term-str/lib/TermFunctionStr.ts
```diff
@@ -15,7 +15,7 @@ export class TermFunctionStr extends TermFunctionBase {
       arity: 1,
       operator: SparqlOperator.STR,
       overloads: declare(SparqlOperator.STR)
-        .onTerm1(() => term => string(term.str()))
+        .onTerm1(() => term => string(term.str()), false)
```
jitsedesmet: The fact that these changes are required shows we did a breaking change. Sadly, we cannot ;)

engines/query-sparql/test/QuerySparql-test.ts
jitsedesmet: I meant for all configurations of the 2 context options, so when `nonLexical` is on but the other not, when both are on, when non are on etc.

jitsedesmet (changes requested):

noahvsb:
How is 100% coverage reached for only half of the spec tests.

engines/query-sparql/test/QuerySparql-test.ts
noahvsb: But that'll just throw an error, meaning we won't get the result of the first two.

packages/utils-expression-evaluator/lib/functions/Helpers.ts
noahvsb: Setting the default to true for `onTerm3` makes coverage fail, because it's only used once, meaning the default is never used. I'll undo it.

engines/query-sparql/test/QuerySparql-test.ts
jitsedesmet: That should not be the case. An error thrown within the expression evaluation of `EXTEND`/ `BIND` (what you do here) should result in the variable not being bound.

NOTE: This is precisely why you cannot view ExpressionErrors like 'regular errors,' since the scope in which they appear very much matters on the expected behavior.

packages/utils-expression-evaluator/lib/functions/Helpers.ts
jitsedesmet: I am confused? The default was true right? So it should stay true?

jitsedesmet (changes requested):
@noahvsb CI fails, and https://github.com/comunica/comunica/pull/1735/changes#r3802617604 was not resolved?

jitsedesmet (changes requested):
Still some not resolved

jitsedesmet (changes requested):
Still some unresolved

packages/actor-function-factory-term-triple/lib/TermFunctionTriple.ts
```diff
@@ -27,6 +27,7 @@ export class TermFunctionTriple extends TermFunctionBase {
             }
             return new Quad(subject, predicate, object, new DefaultGraph());
           },
+          false,
```
jitsedesmet, line 30: Do we still need to put this to false? Why?

packages/actor-function-factory-term-lesser-than-equal/test/op.lesserThanEqual-test.ts
```diff
@@ -183,24 +200,160 @@ describe('evaluation of \'<=\'', () => {
...
+        
+        "a"^^xsd:integer           "b"^^xsd:decimal           = 'Invalid lexical form'
+        "a"^^xsd:yearMonthDuration "b"^^xsd:yearMonthDuration = 'Invalid lexical form'
+        "a"^^xsd:dayTimeDuration   "b"^^xsd:dayTimeDuration   = 'Invalid lexical form'
+        "a"^^xsd:time              "b"^^xsd:time              = 'Invalid lexical form'
+      `,
+    });
+  });
+
+  describe('with non lexical operands and nonLiteralComparison like', () => {
```
jitsedesmet:
```suggestion
  describe('with non lexical operands and nonLexicalComparison like', () => {
```

packages/actor-function-factory-term-lesser-than-equal/test/op.lesserThanEqual-test.ts
```diff
@@ -183,24 +200,160 @@ describe('evaluation of \'<=\'', () => {
...
+        "a"^^xsd:time              "b"^^xsd:time              = true
+      `,
+      errorTable: `
+        "a"^^xsd:boolean     "a"^^xsd:dateTime   = 'Argument types not valid'
+        "true"^^xsd:boolean  "a"^^xsd:dateTime   = 'Argument types not valid'
+      `,
+    });
+  });
+
+  describe('with quoted triple operandsand fullTermComparison like', () => {
```
jitsedesmet:
```suggestion
  describe('with quoted triple operands fullTermComparison like', () => {
```

jitsedesmet (changes requested):
Last nit picks :)

packages/actor-function-factory-term-triple/lib/TermFunctionTriple.ts
noahvsb: To fix a bug where a triple term containing a non-lexical literal with another term would throw a non-lexical error even when `nonLexicalComparison` is true. Because the wrapper that gets added when this is set to true doesn't take the `nonLexicalComparison` option in account.

jitsedesmet (approved):

packages/actor-function-factory-term-lesser-than-equal/lib/TermFunctionLesserThanEqual.ts
```diff
@@ -2,52 +2,45 @@ import type { ITermFunction } from '@comunica/bus-function-factory';
...
   BooleanLiteral,
+  NumericLiteral,
 } from '@comunica/utils-expression-evaluator';
 import {
   bool,
   declare,
   SparqlOperator,
 } from '@comunica/utils-expression-evaluator';
+import { nonLexicalComparisonHandler } from '@comunica/utils-expression-evaluator/lib/functions/Helpers';
+import * as C from '@comunica/utils-expression-evaluator/lib/util/Consts';
```
rubensworks: It's better to import from `'@comunica/utils-expression-evaluator'`, to avoid breakages when that package re-organizes exports.

rubensworks (changes requested):
Just one minor comment. Should be good to go after this!

jitsedesmet:
Good stuff @noahvsb ! Took us a year, but Comunica will allow you to compare triple terms like no other :D Nice work

rubensworks:
@noahvsb There are still a few `/lib/functions/Helpers`... imports in the PR.

noahvsb:
I don't understand why 100% coverage isn't reached, all I changed were imports and exports.

rubensworks:
@noahvsb CI seems to fail due to coverage not being 100%.

Issue #1736, opened 2026-08-03
Run SPARQL 1.0 spec tests


jitsedesmet:
#### Issue type:

- :heavy_plus_sign: Feature request 

____
#### Description:

Description:
The package.json file of engines/query-sparql contains several spec:* scripts to invoke SPARQL spec tests.
We should also run the [SPARQL 1.0 spec tests](https://w3c.github.io/rdf-tests/sparql/sparql10/manifest.ttl) and make sure that all of them are invoked in our GitHub CI workflow.

It is possible that the addition of these specs tests requires changes to both the engine and potentially also to the SPARQL parser, maintained by as part of the Traqula repo.

Afterwards, we could maybe share our earl report on https://github.com/w3c/rdf-tests/

This issue may be split up in the future to track specific issues discovered during the process of adding spec test proven SPARQL 1.0 support.

noahvsb:
I've run the SPARQL 1.0 spec tests (see #1737), there are a total of 90 failing tests, here are the issues:

### Equality operator
- [x] ill-typed literals should be compared (`=`) instead of throwing an error and more `open-eq-*` related failures -- ill-typed literals are compared in #1751, which fixes `open-eq-07`. But the other failures require `mf:XsdDateOperations`, which we won't implement, so they're skipped
- [x] some clauses with xsd:date and xsd:dateTime comparison (`=`) are failing -- skipped, as we won't implement `mf:XsdDateOperations` (defined [here](https://www.w3.org/2001/sw/DataAccess/tests/README.html))
 
#### Sorting -- PR: #1739
- [x] does ascending sort on `ORDER BY DESC(?name)` with string literals (both untyped and datatyped)
- [x] sorts integers and decimals incorrectly: 23 < 9, 1.5 (dec) < 1 (int)
- [x] sorts `?emp` first and then `?name` with `ORDER BY ?name DESC(?emp)`
- [x] IRIs are lower than RDF literals (https://www.w3.org/TR/rdf-sparql-query/#modOrderBy), yet literals get sorted first in ascending sort
- [x] odd sort in the "Builtin sort"- and "Function sort"-test

### Graph handling issue:
- [ ] for variable graphs with empty BGP it should list all named graphs, but none are listed -- Known graph Issue: https://github.com/comunica/comunica/issues/1643
- [ ] for a non-existing graph, it should return nothing, but it returned an empty binding


### Implementation error in langmatches -- PR: #1741 
[...]

PR #1738, opened 2026-08-06
fixed the failing date-* SPARQL 1.0 spec tests (kinda)

noahvsb:
part of #1736

There were 2 issues:

*Literals with datatypes `xsd:date` and `xsd:dateTime` are expected to be compared instead of throwing an error.*

The specs for SPARQL 1.0 don't mention anything about this, but there's an [operator extensibility section](https://www.w3.org/TR/rdf-sparql-query/#operatorExtensibility) which it could fall under. I followed the advice from [this email](https://lists.w3.org/Archives/Public/public-sparql-dev/2018OctDec/0010.html) to go with the flow, so I implemented this. The [specs for SPARQL 1.1](https://www.w3.org/TR/sparql11-query/#operatorExtensibility) are the same in that section. The [specs for SPARQL 1.2](https://www.w3.org/TR/sparql12-query/#func-sameValue) say that if the SPARQL processor can determine this equality we don't throw an error, so we get a bit of freedom there, which means this new implementation is still correct under the SPARQL 1.1 and SPARQL 1.2 specs (the spec tests for these versions still succeed as well).

-------

*Comparison between `xsd:date`s with and without a timezone can be undetermined.*

- Comparison between `"2006-08-23Z"^^xsd:date` and `"2006-08-23"^^xsd:date` is undetermined (not equal, not inequal)
- Comparison between `"2001-01-01Z"^^xsd:date` and `"2006-08-23"^^xsd:date` is determined, they're inequal
- Comparison between `"2006-08-23T09:00:00+01:00"^^xsd:dateTime` and `"2006-08-23"^^xsd:date` is determined, they're inequal

The only implementation that I could see that makes this test pass is to compare it twice, once with the lower bound of timezones (-14h) attached to the "unzoned" date and once with the upper bound (+14h), if these comparisons conflict each other, it's undetermined, else it's determined. This works for the first two, but not for the latter, so we'd only do this implementation for the date to date comparison and not the dateTime to date comparison. But this not only is cursed, it's also against the specs: https://www.w3.org/TR/xpath-functions/#comp.datetime says "If either operand to a comparison function on date or time values does not have an (explicit) timezone then, for the purpose of the operation, an implicit timezone, provided by the dynamic context [Section C.2 Dynamic Context Components ](https://www.w3.org/TR/xpath-31/#id-xp-evaluation-context-components)XP31, is assumed to be present as part of the value." (this is for `op:dateTime-equal`, but `op:date-equal` follows the same rules (https://www.w3.org/TR/xpath-functions/#func-date-equal). So that's why I decided to not implement this and leave the test as skipped.

### DIFF
--- a/engines/query-sparql/spec/sparql-engine.js
+++ b/engines/query-sparql/spec/sparql-engine.js
@@ -1,4 +1,2 @@
-process.env.TZ = 'UTC';
-
 const QueryEngine = require('@comunica/query-sparql').QueryEngine;
 module.exports = require('@comunica/actor-init-query/spec/sparql-engine-base.js')(new QueryEngine());
--- a/engines/query-sparql/spec/sparql-engine.js
+++ b/engines/query-sparql/spec/sparql-engine.js
@@ -1,2 +1,4 @@
+process.env.TZ = 'UTC';
+
 const QueryEngine = require('@comunica/query-sparql').QueryEngine;
 module.exports = require('@comunica/actor-init-query/spec/sparql-engine-base.js')(new QueryEngine());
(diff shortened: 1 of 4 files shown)


rubensworks:
@jitsedesmet Can you review first and ping me once it's ok for you?

noahvsb:
The date-* tests will remain skipped, see https://comunica.dev/docs/query/advanced/specifications/#xsddateoperation.

PR #1744, opened 2026-08-15
Detect and correct TPF metadata with an invalid protocol

### LINKED ISSUE #1686: Add warning when TPF metadata is invalid.
#### Issue type:

____
#### Description:

Occasionally, it happens that TPF servers are hosted on the https protocol, while the base URL is set to http. This leads to things like the metadata graph using the http protocol, which makes Comunica (and other TPF clients) unable to understand the TPF controls.

Since this is a common case, we should detect these http/https mixups, handle them, and emit a warning.

rubensworks:
Closes #1686

TPF servers are occasionally hosted over https while their base URL is configured as http (or vice-versa). All URLs in their metadata are then exposed under the wrong protocol, which breaks two things:

1. **`ActorRdfMetadataPrimaryTopic`** matches `void:subset` objects against the page URL to locate the metadata graph. The comparison fails, so no metadata graph is detected, and all metadata triples leak into the data stream and surface as query results.
2. **`ActorRdfMetadataExtractHydraControls`** looks up Hydra links by page URL (`hydraProperties[link][pageUrl]`). Nothing matches, so `hydra:next` is lost and only the first page of every fragment is ever read. The search form template also stays on the wrong protocol.

## Changes

- **`ActorRdfMetadataExtractHydraControls`**: added `getProtocolMismatchOrigins()` and `correctProtocolMismatches()`. Before links and search forms are extracted, every Hydra subject/object exposed under the page's own host but the opposite protocol is rewritten to the page's protocol, and a warning is logged:

  > `Invalid metadata detected in https://…: controls are exposed under http://… instead of https://…. These have been corrected, but the server should be reconfigured with a valid base URL.`

  Rewriting is scoped to the exact same host: the invalid origin must be followed by `/`, `?`, `#`, or end-of-string, so URLs such as `http://example.org.other.com/` are left alone. Non-http(s) and unparseable page URLs are untouched, and when nothing matches, the original object is returned unchanged without a warning.
- **`ActorRdfMetadataPrimaryTopic`**: the `void:subset` comparison now ignores an http/https mismatch, so the metadata graph is still detected and the data stream stays clean. URLs on a different host still do not match.

## Performance

The detection runs once per dereferenced page. Measured on a realistic `fragments.dbpedia.org` fragment (Node 22, 200k iterations):

| Case | Cost |
| --- | --- |
| Document without Hydra controls (plain RDF) | ~0 (early return) |
| Valid metadata (common case) | ~1.0 µs |
| Invalid metadata (correction applied) | ~6.9 µs |

Valid metadata is detected without copying anything; only pages that actually need correcting pay for the rewrite. On the extractor's `run()` this is ~3.3 µs/page, i.e. ~1.5% of the RDF parsing of a single TPF page (210 µs) and a far smaller fraction once the HTTP request is counted.

## Testing

- **Unit tests**: 10 new cases in the Hydra controls test (origin detection including ports, non-http and invalid URLs; correction in both directions; empty/valid/other-host no-ops; and a full `run()` asserting corrected `first`/`next`, search form template, `getUri()` output, and exactly one warning) and 2 new cases in the primary-topic test (mismatched protocol detected, different host not). Both changed files remain at 100% coverage.
- **System test** (`engines/query-sparql/test/QuerySparql-test.ts`): a mocked two-page TPF interface served at `https://tpf.example.org/ds` that declares all of its controls under `http://…`, modelled on a real `fragments.dbpedia.org` response. It asserts that results from both pages are returned without any metadata leaking in, that the only requests made are the two https ones, and that the warning reaches the logger.

The system test was confirmed to be non-vacuous: with the two `lib` changes reverted, it fails with 21 results (metadata leaking, page 2 never fetched) instead of 2.

---
_Generated by [Claude Code](https://claude.ai/code/session_01N6KzbLao4C7rf598BUnJ9G)_

### DIFF
--- a/packages/actor-rdf-metadata-extract-hydra-controls/lib/ActorRdfMetadataExtractHydraControls.ts
+++ b/packages/actor-rdf-metadata-extract-hydra-controls/lib/ActorRdfMetadataExtractHydraControls.ts
@@ -93,23 +93,30 @@ export class ActorRdfMetadataExtractHydraControls extends ActorRdfMetadataExtrac
     hydraProperties: Record<string, Record<string, string[]>>,
     context: IActionContext,
   ): Record<string, Record<string, string[]>> {
+    // Documents without any Hydra controls, such as plain RDF documents, can never contain invalid controls.
+    if (Object.keys(hydraProperties).length === 0) {
+      return hydraProperties;
+    }
+
     const origins = this.getProtocolMismatchOrigins(pageUrl);
     if (!origins) {
       return hydraProperties;
     }
 
-    // Rewrite all URLs that are exposed under the invalid origin, and remember if any of them were found.
-    let mismatchDetected = false;
-    const correctUrl = (url: string): string => {
-      // Only consider URLs for which the invalid origin is followed by a path, query, fragment, or nothing at all,
-      // so that hosts that merely start with the same characters are not corrected.
-      if (url.startsWith(origins.invalid) &&
-        (url.length === origins.invalid.length || '/?#'.includes(url[origins.invalid.length]))) {
-        mismatchDetected = true;
-        return origins.valid + url.slice(origins.invalid.length);
-      }
-      return url;
-    };
+    // Only URLs for which the invalid origin is followed by a path, query, fragment, or nothing at all are
+    // considered, so that hosts that merely start with the same characters are not corrected.
+    const isInvalidUrl = (url: string): boolean => url.startsWith(origins.invalid) &&
+      (url.length === origins.invalid.length || '/?#'.includes(url[origins.invalid.length]));
+
+    // Detect invalid URLs before correcting them,
+    // so that nothing has to be copied for the common case of valid metadata.
+    if (!this.hasInvalidUrl(hydraProperties, isInvalidUrl)) {
+      return hydraProperties;
+    }
+
+    const correctUrl = (url: string): string => isInvalidUrl(url) ?
+      origins.valid + url.slice(origins.invalid.length) :
+      url;
     const correctedHydraProperties: Record<string, Record<string, string[]>> = {};
     for (const [ property, subjects ] of Object.entries(hydraProperties)) {
       const correctedSubjects: Record<string, string[]> = correctedHydraProperties[property] = {};
@@ -123,11 +130,28 @@ export class ActorRdfMetadataExtractHydraControls extends ActorRdfMetadataExtrac
       }
     }
 
-    if (!mismatchDetected) {
-      return hydraProperties;
-    }
     this.logWarn(context, `Invalid metadata detected in ${pageUrl}: controls are exposed under ${origins.invalid} instead of ${origins.valid}. These have been corrected, but the server should be reconfigured with a valid base URL.`);
     return correctedHydraProperties;
+  }
+
+  /**
+   * Check if any URL within the given Hydra properties is invalid.
+   * @param hydraProperties The collected Hydra properties.
+   * @param isInvalidUrl A predicate determining if a URL is invalid.
+   */
+  protected hasInvalidUrl(
+    hydraProperties: Record<string, Record<string, string[]>>,
+    isInvalidUrl: (url: string) => boolean,
+  ): boolean {
+    for (const property in hydraProperties) {
+      const subjects = hydraProperties[property];
+      for (const subject in subjects) {
+        if (isInvalidUrl(subject) || subjects[subject].some(object => isInvalidUrl(object))) {
+          return true;
+        }
+      }
+    }
+    return false;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


packages/actor-rdf-metadata-extract-hydra-controls/lib/ActorRdfMetadataExtractHydraControls.ts
```diff
@@ -6,10 +6,19 @@ import type {
...
 import { passTestVoid } from '@comunica/core';
+import type { IActionContext } from '@comunica/types';
 import type * as RDF from '@rdfjs/types';
 import type { UriTemplate } from 'uritemplate';
 import { parse as parseUriTemplate } from 'uritemplate';
 
+/**
+ * The protocols that servers commonly mix up, mapped to the protocol they may have been mixed up with.
+ */
+const PROTOCOL_ALTERNATIVES: Record<string, string> = {
```
rubensworks: This is overkill.
Only consider the case that is tested in QuerySparql-test.
This is the only case that actually occurs in practise.

Then, the URL parsing is not necessary, and can be replaced by a `.startsWith` or something similar, to avoid parsing overhead.
Other parts of the code may become simplifiable then as well.

packages/actor-rdf-metadata-primary-topic/lib/ActorRdfMetadataPrimaryTopic.ts
```diff
@@ -50,7 +50,7 @@ export class ActorRdfMetadataPrimaryTopic extends ActorRdfMetadata {
       const primaryTopics: Record<string, string> = {};
       action.quads.on('data', (quad) => {
         if (quad.predicate.value === 'http://rdfs.org/ns/void#subset' &&
-          quad.object.value === action.url) {
+          equalsIgnoringHttpProtocol(quad.object.value, action.url)) {
```
rubensworks: Are you sure this is necessary?
If the mismatched protocol was corrected in the metadata extractor actor, are you sure this special handling is needed here?

packages/actor-rdf-metadata-extract-hydra-controls/lib/ActorRdfMetadataExtractHydraControls.ts
rubensworks: Narrowed to the https-page/http-controls case in 0ba58c3, and the URL parsing is gone.

`PROTOCOL_ALTERNATIVES` is replaced by two plain constants, and the invalid prefix is now derived with `startsWith` + `indexOf`:

```ts
public getInvalidUrlPrefix(pageUrl: string): string | undefined {
  if (!pageUrl.startsWith(HTTPS_PROTOCOL)) {
    return;
  }
  const authorityEnd = pageUrl.indexOf('/', HTTPS_PROTOCOL.length);
  if (authorityEnd < 0) {
    return;
  }
  return HTTP_PROTOCOL + pageUrl.slice(HTTPS_PROTOCOL.length, authorityEnd + 1);
}
```

That simplified the rest as you expected: including the trailing slash in the prefix made the separate "is the origin followed by `/`, `?` or `#`" boundary check unnecessary, correcting a URL became a plain protocol swap, and the `hasInvalidUrl` scan now takes the prefix string instead of a predicate. The special case for documents without any Hydra controls could go too, since the prefix computation is no longer expensive enough to be worth skipping.

Measured on a realistic `fragments.dbpedia.org` fragment (200k iterations):

| | before | after |
| --- | --- | --- |
| deriving the prefix | 374 ns | 35 ns |
| valid metadata (common case) | 1037 ns | 566 ns |

One thing I kept: the correction is still scoped to the host of the page. Without that, `hydra:property` values such as `http://www.w3.org/1999/02/22-rdf-syntax-ns#subject` would be rewritten to https and `QuerySourceQpf.getSearchForm` would no longer recognise the mappings, so QPF sources would stop being detected entirely. There is now a test asserting those values are left untouched.


---

packages/actor-rdf-metadata-primary-topic/lib/ActorRdfMetadataPrimaryTopic.ts
rubensworks: I checked this by reverting only this line and re-running the TPF system test: the query returns **23 results instead of 2**, with all 21 metadata quads showing up as query results.

The reason the extractor correction does not cover it is that the two actors act on different things at different stages:

- `ActorRdfMetadataPrimaryTopic` (bus `rdf-metadata`) splits the raw quad stream into a data stream and a metadata stream. It runs *before* the metadata extract bus.
- `ActorRdfMetadataExtractHydraControls` (bus `rdf-metadata-extract`) only rewrites the extracted Hydra property hash in memory. It never touches the quads, and it runs *after* the split.

So the `void:subset` lookup here still compares `http://tpf.example.org/ds` against the page URL `https://tpf.example.org/ds`, no metadata graph is detected, and since `dataToMetadataOnInvalidMetadataGraph` defaults to true, every metadata quad is emitted as data as well. The Hydra controls are fine at that point (pagination works, the corrected search form is used), but the metadata triples end up in the results.

I did simplify this in the same spirit as your other comment: the regex-based helper is replaced by a single-direction check without allocation-heavy parsing, mirroring the extractor.

```ts
function equalsPageUrl(url: string, pageUrl: string): boolean {
  return url === pageUrl ||
    (url.startsWith(HTTP_PROTOCOL) && pageUrl === HTTPS_PROTOCOL + url.slice(HTTP_PROTOCOL.length));
}
```

It is only reached for `void:subset` quads (one or two per page), after the predicate check, and short-circuits on exact equality.

Happy to drop it if you would rather handle the leaking metadata differently, but as it stands something is needed here.


---

Issue #198, opened 2018-08-07
Add alternative source-based FROM/FROM NAMED actor


rubensworks:
#### Issue type:

- :heavy_plus_sign: Feature request

____
#### Description:

Currently, FROM/FROM NAMED is translated to GRAPH clauses, to identify the named graph.

In some cases, it might make sense to use a different semantics, for selecting the datasource/interface. This actor would then simply set/append the list of sources based on the FROM/FROM NAMED clauses.

rubensworks:
For reference, optional spec tests are relevant here: https://github.com/w3c/rdf-tests/pull/370

PR #1746, opened 2026-08-17
Fix dataset spec tests

noahvsb:
Part of #1736.

Fixed and unskipped the dataset spec tests.

Had to populate the stores with the dataset quads.


rubensworks:
If I understand correct, this inspects the query for FROM (NAMED) clauses, and forcefully ingests them into the store.
If so, I don't think this is the right way forward.
Instead, we should make the engine recognise these clauses according to what the spec test requires.
We should double-check what the spec actually requires for these.
Also, do these tests have any special modes/options on them to modify the way the engine should process FROM (NAMED)?

Related to https://github.com/comunica/comunica/issues/198

noahvsb:
The [specs](https://www.w3.org/TR/sparql11-query/#specifyingDataset) say: "If a query provides such a dataset description, then it is used in place of any dataset that the query service would use if no dataset description is provided in a query." So that's in line with the spec test requirements.

The [manifest.ttl](https://w3c.github.io/rdf-tests/sparql/sparql10/dataset/manifest.ttl) doesn't have any special modes or options turned on.

I think it's indeed best to add a new source actor for this.

rubensworks:
>  "If a query provides such a dataset description, then it is used in place of any dataset that the query service would use if no dataset description is provided in a query." So that's in line with the spec test requirements.

That's not how I understand this sentence.
Based on my understanding of the spec, it says nothing about whether or not FROM should actually dereference remote URLs.
A quick google showed me that engines like Jena and Virtuoso in fact also don't do this (but they have an option to enable it).
As such, I don't think we should implement #198 (at least not for making this spec test pass).

I actually think these spec tests are incorrectly defined.
I know there are other spec tests that specifically defined which named graphs are part of the dataset.
As such, I think we should create a PR to w3c/rdf-tests to add these named graph definitions to the manifest.

noahvsb:
It was decided to implement https://github.com/comunica/comunica/issues/198 anyway with a context option related to https://github.com/w3c/rdf-tests/pull/370.

noahvsb:
Replaced by #1768.

PR #1747, opened 2026-08-18
Restrict bind joins on `LEFT_JOIN` to fix a nested-optionals spec test

noahvsb:
Part of #1736.

In some cases, binding with `LEFT_JOIN` (which the `OPTIONAL` implementation uses) produces incorrect results. The failing spec test was one of those cases. Now on conflict, these binds are skipped.

AI also told me to implement it for `MINUS`, I'm not 100% sure about that, but it seems correct at a first glance and nothing breaks.

I also unskipped 2 other unrelated tests that I forgot to unskip in #1739.


rubensworks:
To be honest, I don't really understand why exactly this is necessary.
Could you explain why it's needed?

packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
```diff
@@ -181,17 +182,84 @@ export class ActorRdfJoinMultiBind extends ActorRdfJoin<IActorRdfJoinMultiBindTe
...
+        visitor: (op: Algebra.Extend) => {
+          addTermIfVar(op.variable);
+        },
+      },
+    });
+
+    return [ ...variables.values() ];
+  }
+
+  public static canBindWithOperation(operation: Algebra.Operation, boundVariables: RDF.Variable[]): boolean {
```
rubensworks: This added param is a breaking API change. (which we shouldn't do) So we need either to not add this param, or make it optional.

packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
```diff
@@ -181,17 +182,84 @@ export class ActorRdfJoinMultiBind extends ActorRdfJoin<IActorRdfJoinMultiBindTe
...
+    let valid = true;
+    const boundVarNames = new Set(boundVariables.map(v => v.value));
```
rubensworks: Isn't this just `Object.keys(boundVariables)`?
Is there even a reason why this should be a record?

packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
```diff
@@ -181,17 +182,84 @@ export class ActorRdfJoinMultiBind extends ActorRdfJoin<IActorRdfJoinMultiBindTe
     };
   }
 
-  public static canBindWithOperation(operation: Algebra.Operation): boolean {
-    let valid = true;
+  public static getVariables(operation: Algebra.Operation): RDF.Variable[] {
```
rubensworks: What should this method do _exactly_?

noahvsb:
> To be honest, I don't really understand why exactly this is necessary. Could you explain why it's needed?

Me neither, I spent a while trying to understand it yesterday afternoon, but I couldn't. Claude confidentially claims the specs say this is needed, however I couldn't find it in the specs. But the code makes the failing spec tests pass, so I pushed it anyway.

packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
noahvsb: It fetches all of the variable names of variables used in `operation`.

packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
noahvsb: It didn't work with `Object.keys(boundVariables)`.

packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
rubensworks: But what does _used in_ mean?
Does it include variables in nested operations?
Does it included projected out variables in subqueries?
...

Does it differ from the `inScopeVariables` helper?

rubensworks:
> Me neither, I spent a while trying to understand it yesterday afternoon, but I couldn't. Claude confidentially claims the specs say this is needed, however I couldn't find it in the specs. But the code makes the failing spec tests pass, so I pushed it anyway.

I'm afraid we can't merge this if we don't understand what it does.
Could you have another look at it.
Perhaps @jitsedesmet can help, as I think he touched similar code recently.

jitsedesmet:
@noahvsb can you share the spec test + the expected result + the result Comunica gave?

If I have this, I can take a look at the spec and maybe explain the issue :)

noahvsb:
I pasted the error output for the test below:

<details>

<summary>Expand</summary>

```
  Nested-optionals with a shared variable that does not appear in the middle pattern (a not well-formed query pattern as per "Semantics and Complexity" of SPARQL
  Error: Invalid query evaluation

  Query:

PREFIX :    <http://example/>

SELECT *
{ 
    :x1 :p ?v .
    OPTIONAL
    {
      :x3 :q ?w .
      OPTIONAL { :x2 :p ?v }
    }
}


  Data links: https://w3c.github.io/rdf-tests/sparql/sparql10/algebra/two-nested-opt.ttl

  Result Source: https://w3c.github.io/rdf-tests/sparql/sparql10/algebra/two-nested-opt.srx

  Expected: [QueryResultBindings:
    Variables: [
  "?v",
  "?w"
]
    Bindings:  [
  {
    "?v": {
      "termType": "Literal",
      "value": "1",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#integer"
      },
      "direction": ""
    }
  }
]
]

  Got: 
 [QueryResultBindings:
    Variables: [
  "?v",
  "?w"
]
    Bindings:  [
  {
    "?w": {
      "termType": "Literal",
      "value": "3",
      "language": "",
      "direction": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#integer"
      }
    },
    "?v": {
      "termType": "Literal",
      "value": "1",
      "language": "",
      "direction": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#integer"
      }
    }
  },
  {
    "?w": {
      "termType": "Literal",
      "value": "4",
      "language": "",
      "direction": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#integer"
      }
    },
    "?v": {
      "termType": "Literal",
      "value": "1",
      "language": "",
      "direction": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#integer"
      }
    }
  }
]
]

  More info: http://www.w3.org/2001/sw/DataAccess/tests/data-r2/algebra/manifest#nested-opt-1
  ``` 
  
  </details>

File: packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts (hunk 3 above)
Comment (noahvsb) [reply packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts]: No it does not.

**Comment by jitsedesmet** [comment]
I had a offline chat with @noahvsb where we discussed how the expected result is achieved:

```
SELECT *
{ 
    :x1 :p ?v .
    OPTIONAL
    {
      :x3 :q ?w .
      OPTIONAL { :x2 :p ?v }
    }
}

Inner: { ?v: 2 }
middle: { ?w: 3 }
outer: { ?v: 1 }

--- INNER OPTIONAL

JOIN: { ?v: 2; ?w: 3 }  
Middle \ Inner: no results because compatible (trivially so)

--- OUTER OPTIONAL:

JOIN: No results.  
OUTER \ MIDDLE:  { ?v: 1 } (since middle results are not compatible)
```

**Comment by noahvsb** [comment]
The multi-bind actor is an actor used purely for performance optimization, but the optimization breaks on 5 sub-operations: `EXTEND`, `GROUP`, `LEFT_JOIN`, `MINUS` and `FILTER`; because of their algebra definitions.

I didn't dig into detail, however there are 2 spec tests (`nested-opt-1` and `dawg-optional-filter-005-not-simplified`) that fail if we don't implement `LEFT_JOIN` and `FILTER` as exceptions. So I am confident that these are needed[^1]. To justify the `MINUS` exception I'd have to look into detail in the algebra, but I don't want to do that unless necessary. I could also just omit `MINUS` from this PR.

The reason I didn't just use `skipHandler` for the implementation and made it more complex is because there are cases where this optimization can be used if there isn't a variable conflict. I can make it use `skipHandler` and the spec tests would all still succeed (except for `FILTER`[^1]).

[^1]: Due to `dawg-optional-filter-005-not-simplified` needing a bug fix in Traqula and it also being more difficult than I imagined - it breaks other spec tests; I need a more proper invalid check - I'll leave the `FILTER` exception implementation to a different PR.

**Comment by rubensworks** [comment]
I'm not saying the fix is wrong, but I still don't fully understand the issue.

The `two-nested-opt` spec test only uses left-joins, while the BJ's `canBindWithOperation` is used by the inner-join and left-join. Only in the inner join actor are the `boundVariables` passed, so for left-join, they are never passed, so it defaults to the "skip" behaviour. Is this change even needed for inner-joins? From the spec test, it looks like it's only needed for left-joins?
In any case, if it should be for both (which I'm not certain of), then I would guess a change is also needed for the left-join actor that uses BJ.

(also see the merge conflict, given the EARL reporting changes)

**Comment by noahvsb** [comment]
> The `two-nested-opt` spec test only uses left-joins, while the BJ's `canBindWithOperation` is used by the inner-join and left-join. Only in the inner join actor are the `boundVariables` passed, so for left-join, they are never passed, so it defaults to the "skip" behaviour. Is this change even needed for inner-joins? From the spec test, it looks like it's only needed for left-joins? In any case, if it should be for both (which I'm not certain of), then I would guess a change is also needed for the left-join actor that uses BJ.

Yup that was indeed wrong, the tests just succeeded because the default behavior was to make it invalid. I changed the default behavior to be valid and passed `boundVariables` only for the left-join.

File: engines/query-sparql/package.json [hunk 4]
```diff
@@ -60,7 +60,7 @@
     "spec:base-1-0": "node ../../node_modules/rdf-test-suite/bin/Runner.js spec/sparql-engine.js https://w3c.github.io/rdf-tests/sparql/sparql10/manifest.ttl -c ../../.rdf-test-suite-cache/",
     "spec:base-1-1": "node ../../node_modules/rdf-test-suite/bin/Runner.js spec/sparql-engine.js http://w3c.github.io/rdf-tests/sparql/sparql11/manifest-all.ttl -c ../../.rdf-test-suite-cache/",
     "spec:base-1-2": "node ../../node_modules/rdf-test-suite/bin/Runner.js spec/sparql-engine.js https://w3c.github.io/rdf-tests/sparql/sparql12/manifest.ttl -c ../../.rdf-test-suite-cache/",
-    "spec:query-1-0": "yarn run --silent spec:base-1-0 --skip \"(date-[12]|open-eq-|nested-opt-1|dawg-optional|#graph-|dataset-|add-literals|limit-|offset-|slice-2|blabel|syntax-sparql4)\"",
+    "spec:query-1-0": "yarn run --silent spec:base-1-0 --skip --skip \"(date-[12]|open-eq-|dawg-optional|#graph-|dataset-|add-literals|limit-|blabel|syntax-sparql4)\"",
```
rubensworks: Two times --skip looks incorrect.

packages/actor-rdf-join-optional-bind/lib/ActorRdfJoinOptionalBind.ts
```diff
@@ -102,7 +102,8 @@ export class ActorRdfJoinOptionalBind extends ActorRdfJoin {
     const requestItemTimes = ActorRdfJoin.getRequestItemTimes(metadatas);
 
     // Reject binding on some operation types (including when reachable through e.g. a Project or Filter wrapper)
-    if (!ActorRdfJoinMultiBind.canBindWithOperation(action.entries[1].operation)) {
+    const boundVariables = inScopeVariables(action.entries[0].operation);
+    if (!ActorRdfJoinMultiBind.canBindWithOperation(action.entries[1].operation, boundVariables)) {
       return failTest(`Actor ${this.name} can not bind on Extend and Group operations`);
```
rubensworks: This message should probably be changed here now.

packages/actor-rdf-join-optional-bind/lib/ActorRdfJoinOptionalBind.ts
```diff
@@ -102,7 +102,8 @@ export class ActorRdfJoinOptionalBind extends ActorRdfJoin {
     const requestItemTimes = ActorRdfJoin.getRequestItemTimes(metadatas);
 
     // Reject binding on some operation types (including when reachable through e.g. a Project or Filter wrapper)
-    if (!ActorRdfJoinMultiBind.canBindWithOperation(action.entries[1].operation)) {
+    const boundVariables = inScopeVariables(action.entries[0].operation);
+    if (!ActorRdfJoinMultiBind.canBindWithOperation(action.entries[1].operation, boundVariables)) {
```
rubensworks, line 106: Can you also add unit test(s) for this actor?

packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
```diff
@@ -181,17 +182,59 @@ export class ActorRdfJoinMultiBind extends ActorRdfJoin<IActorRdfJoinMultiBindTe
...
+
+          const conflict = rightVars.some(v => !leftVars.has(v) && boundVarNames.includes(v));
+          if (conflict) {
+            valid = false;
+            return { shortcut: true };
+          }
+          return { shortcut: false };
+        },
+      },
+      [Algebra.Types.MINUS]: {
```
rubensworks, line 220: Still not sure about the MINUS case here.
Can you think of an example query where NOT having it would cause issues, but including this would fix it?
If you have this case, you can include it in the QuerySparql system tests file.

rubensworks (changes requested):

rubensworks (approved):
Everythings clear and looking good to me now, nice work!!

PR #1751, opened 2026-08-21
Implement nonLexical- and fullTermComparison for (in)equality

noahvsb:
For both equality and inequality, I implemented the `nonLexicalCompraison` and `fullTermComparison` options.

They're only used for literals, `nonLexicalComparison` allows comparison of non-lexical literals and `fullTermComparison` allows comparing 2 literals with unknown or different types that aren't exactly equal.

I left 4 `open-eq-*` tests skipped, because they require extensions that aren't useful or we have already implemented them in a different way. (https://w3c.github.io/rdf-tests/sparql/sparql10/open-world/index.html)


packages/actor-function-factory-term-equality/lib/TermFunctionEquality.ts
```diff
@@ -75,32 +93,77 @@ export class TermFunctionEquality extends TermFunctionBase {
             const right = _right.toRDF(exprEval.context.getSafe(KeysInitQuery.dataFactory));
             const val = left.equals(right);
             if (!val && (left.termType === 'Literal') && (right.termType === 'Literal')) {
-              throw new RDFEqualTypeError([ _left, _right ]);
+              this.fullTermComparisonCheck(exprEval, _left, _right);
```
jitsedesmet: Why do we need `this.fullTermComparisonCheck(exprEval, _left, _right);`?
We already tested that both left and right are literals. So the context option on fullComparison does not apply? Only the one on nonlexicals does? Or am I misinterpreting something?

packages/actor-function-factory-term-equality/lib/TermFunctionEquality.ts
```diff
@@ -75,32 +93,77 @@ export class TermFunctionEquality extends TermFunctionBase {
...
+  }
+
+  private nonLexicalCheck(exprEval: IInternalEvaluator, a: TermExpression, b: TermExpression): void {
+    const nonLexical = [ a, b ].find(arg => arg instanceof NonLexicalLiteral);
+    if (nonLexical && !exprEval.context.get(KeysExpressionEvaluator.nonLexicalComparison)) {
+      throw new InvalidLexicalForm(
+        nonLexical.toRDF(exprEval.context.getSafe(KeysInitQuery.dataFactory)),
+      );
+    }
+  }
```
jitsedesmet, line 166: So here then we also already know both are literal?

jitsedesmet (changes requested):
Looking at the code, I have just a small question

packages/actor-function-factory-term-equality/lib/TermFunctionEquality.ts
noahvsb: Because the default behaviour should still throw the `RDFEqualTypeError` even with non-lexical literals if `nonLexicalComparison` is true.

packages/actor-function-factory-term-equality/lib/TermFunctionEquality.ts
noahvsb: Yes, I tried to make the types of `a` and `b` be literal, but it didn't immediately work, so I decided not to pain myself with making it work.

packages/actor-function-factory-term-equality/lib/TermFunctionEquality.ts
jitsedesmet: But the context ke that should be checked is the compareNonLexical key right? Since the check care about comparing two literals in this branch?

jitsedesmet:
```suggestion
               if (!exprEval.context.get(KeysExpressionEvaluator.fullTermComparison)) {
               // Throw for two lexical but incompatible term types, e.g. `1` and `"string"`  
      throw new RDFEqualTypeError([ a, b ]);
    }
              this.fullTermComparisonCheck(exprEval, _left, _right);
```
I would inline the throw since it has only one usage anyways? That I can see.

packages/actor-function-factory-term-equality/lib/TermFunctionEquality.ts
```diff
@@ -75,32 +93,77 @@ export class TermFunctionEquality extends TermFunctionBase {
             const right = _right.toRDF(exprEval.context.getSafe(KeysInitQuery.dataFactory));
             const val = left.equals(right);
             if (!val && (left.termType === 'Literal') && (right.termType === 'Literal')) {
-              throw new RDFEqualTypeError([ _left, _right ]);
+              this.fullTermComparisonCheck(exprEval, _left, _right);
+              this.nonLexicalCheck(exprEval, _left, _right);
```
jitsedesmet, line 102:
```suggestion
              // Throw if any are nonLexical, unless when it is allowed.
              this.nonLexicalCheck(exprEval, _left, _right);
```

jitsedesmet (changes requested):
Since I stumbled 3 times, Let's just add some docs :) 😅

jitsedesmet (approved):

jitsedesmet:
Seems like some unrelated changes might have slipped in? To me it looks good, though. :) Nice work

engines/query-sparql/package.json
```diff
@@ -60,7 +60,7 @@
     "spec:base-1-0": "node ../../node_modules/rdf-test-suite/bin/Runner.js spec/sparql-engine.js https://w3c.github.io/rdf-tests/sparql/sparql10/manifest.ttl -c ../../.rdf-test-suite-cache/",
     "spec:base-1-1": "node ../../node_modules/rdf-test-suite/bin/Runner.js spec/sparql-engine.js http://w3c.github.io/rdf-tests/sparql/sparql11/manifest-all.ttl -c ../../.rdf-test-suite-cache/",
     "spec:base-1-2": "node ../../node_modules/rdf-test-suite/bin/Runner.js spec/sparql-engine.js https://w3c.github.io/rdf-tests/sparql/sparql12/manifest.ttl -c ../../.rdf-test-suite-cache/",
-    "spec:query-1-0": "yarn run --silent spec:base-1-0 --skip \"(date-[12]|open-eq-|dawg-optional|#graph-|dataset-|add-literals|limit-|blabel|syntax-sparql4)\"",
+    "spec:query-1-0": "yarn run --silent spec:base-1-0 --skip \"(date-[12]|open-eq-(08|1)|dawg-optional|#graph-|dataset-|add-literals|limit-|blabel|syntax-sparql4)\"",
```
rubensworks: For the 4 tests we deliberately skip now, could you write their names out in full (instead of using a regex)?
That way, there's a clearer correspondence to what you added to the comunica website.
In the end, I suspect these will be the only skipped tests in this list.

rubensworks (changes requested):
Just one minor suggestion. Good to go besides this!

PR #1753, opened 2026-08-23
Swap definitions of VOID_DISTINCT_OBJECTS and VOID_DISTINCT_SUBJECTS

balhoff:
I just happened to come across this; it seems like an accident.


rubensworks:
Oops, that seems wrong indeed. Thanks for the fix!

PR #1756, opened 2026-08-26
Don't bind-join on FILTER that aren't direct children of OPTIONAL

noahvsb:
Part of #1736

The issue had to do with these 2 queries:

[`dawg-optional-filter-004`](https://w3c.github.io/rdf-tests/sparql/sparql10/optional-filter/index.html#dawg-optional-filter-004):
```SPARQL
PREFIX  dc: <http://purl.org/dc/elements/1.1/>
PREFIX  x: <http://example.org/ns#>
SELECT  ?title ?price
WHERE
    { ?book dc:title ?title . 
      OPTIONAL
        { ?book x:price ?price . 
          FILTER (?price < 15 && ?title = "TITLE 2") .
        } .
    }
```

[`dawg-optional-filter-005-not-simplified`](https://w3c.github.io/rdf-tests/sparql/sparql10/optional-filter/index.html#dawg-optional-filter-005-not-simplified):
```SPARQL
PREFIX  dc: <http://purl.org/dc/elements/1.1/>
PREFIX  x: <http://example.org/ns#>
SELECT  ?title ?price
WHERE
    { ?book dc:title ?title . 
      OPTIONAL
        {
          { 
            ?book x:price ?price . 
            FILTER (?title = "TITLE 2") .
          }
        } .
    }
```

Notice that the first FILTER is nested in 1 pair of brackets, while the second one is nested in 2 pairs. The issue was that the second one doesn't allow a bind join since it isn't a direct child of OPTIONAL, so the second spec test failed. Related to #1747.

This also needed a Traqula update, so I unskipped a few more tests that needed a Traqula update as well.


packages/actor-query-operation-leftjoin/lib/ActorQueryOperationLeftJoin.ts
```diff
@@ -37,9 +37,11 @@ export class ActorQueryOperationLeftJoin extends ActorQueryOperationTypedMediate
         // we attach the expression to the right-hand operation,
         // and enforce a bind-join.
         if (operationOriginal.expression && index === 1) {
+          const filterOperation = algebraFactory.createFilter(subOperation, operationOriginal.expression);
+          (<Algebra.Filter & { isHoistedLeftJoinFilter?: true }> filterOperation).isHoistedLeftJoinFilter = true;
```
jitsedesmet: I think Comunica holds metadata for each key. I assume that will be cleaner then this casting.

jitsedesmet (changes requested):
I think everything looks clean as is. Functionality-wise, Ruben will have more to say.

packages/actor-query-operation-leftjoin/lib/ActorQueryOperationLeftJoin.ts
rubensworks: Indeed, we can attach metadata to operations.
See for example how sources are annotated to operations (this is done in some optimize actors).

jitsedesmet (changes requested):
tests seem to be failing?

noahvsb:
> tests seem to be failing?

Forgot to edit another test file, should be fixed now.

packages/actor-query-operation-leftjoin/lib/ActorQueryOperationLeftJoin.ts
```diff
@@ -37,9 +37,12 @@ export class ActorQueryOperationLeftJoin extends ActorQueryOperationTypedMediate
         // we attach the expression to the right-hand operation,
         // and enforce a bind-join.
         if (operationOriginal.expression && index === 1) {
+          const filterOperation: Algebra.Operation =
+            algebraFactory.createFilter(subOperation, operationOriginal.expression);
+          filterOperation.metadata = { isHoistedLeftJoinFilter: true };
```
jitsedesmet: This code would undo other metadata being present.

Could we do:
```suggestion
          filterOperation.metadata.isHoistedLeftJoinFilter = true;
```


or perhabs we need a algebra-util :
```ts
function withMetaData<T extends Algebra.Operation>(op: T): T & {metadata: TypeOfMetataDataIdk } {
   op.metadata ??= {};
    return op;
}
```

packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
```diff
@@ -180,6 +180,10 @@ describe('ActorQueryOperationLeftJoin', () => {
       const expression = AF.createTermExpression(DF.literal(''));
       const op: any = { operation: { type: 'leftjoin', input: [{}, {}], expression }, context };
       const output = getSafeBindings(await actor.run(op, undefined));
+
+      const expectedFilterOperation: any = AF.createFilter(<any>{}, <any>expression);
+      expectedFilterOperation.metadata = { isHoistedLeftJoinFilter: true };
```
jitsedesmet: same here

packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
```diff
@@ -180,6 +180,10 @@ describe('ActorQueryOperationLeftJoin', () => {
       const expression = AF.createTermExpression(DF.literal(''));
       const op: any = { operation: { type: 'leftjoin', input: [{}, {}], expression }, context };
       const output = getSafeBindings(await actor.run(op, undefined));
+
+      const expectedFilterOperation: any = AF.createFilter(<any>{}, <any>expression);
```
jitsedesmet: Since you touched this, can we cleanup and cast to the correct type instead of any?

jitsedesmet (changes requested):

packages/actor-query-operation-leftjoin/lib/ActorQueryOperationLeftJoin.ts
noahvsb: I just assumed there's no metadata present because the filter operation is freshly created.

packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
noahvsb: This is a test, so I know with 100% certainty that there's no metadata yet.

packages/actor-query-operation-leftjoin/lib/ActorQueryOperationLeftJoin.ts
noahvsb: I couldn't get both working, metadata can be undefined which messes up both of your suggestions.

jitsedesmet: This is very true. Woeps. You may revert, or do the withMetadata, up to you.

jitsedesmet (changes requested):

packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
jitsedesmet: Because you just created the nice helper, we should use it here too ;)

jitsedesmet (changes requested):
I like that. 😍 So clean

packages/actor-rdf-join-inner-multi-bind/test/ActorRdfJoinMultiBind-test.ts
```diff
@@ -123,6 +123,29 @@ IQueryOperationResultBindings
 
             expect(ActorRdfJoinMultiBind.canBindWithOperation(minusOp)).toBe(true);
           });
+
+          it('should allow binding on a right stream with safe FILTER', () => {
+            const filter: Algebra.Operation =
+              FACTORY.createFilter(<any>{}, FACTORY.createTermExpression(DF.literal('')));
```
jitsedesmet:
```suggestion
              FACTORY.createFilter(FACTORY.createNop(), FACTORY.createTermExpression(DF.literal('')));
```

packages/actor-rdf-join-inner-multi-bind/test/ActorRdfJoinMultiBind-test.ts
```diff
@@ -123,6 +123,29 @@ IQueryOperationResultBindings
...
+            filter.metadata = { isHoistedLeftJoinFilter: true };
+
+            expect(ActorRdfJoinMultiBind.canBindWithOperation(filter)).toBe(true);
+          });
+
+          it('should reject binding on a right stream with conflicting FILTER', () => {
+            const filter: any = FACTORY.createFilter(<any>{}, FACTORY.createTermExpression(DF.literal('')));
```
jitsedesmet: here to

packages/actor-rdf-join-inner-multi-bind/test/ActorRdfJoinMultiBind-test.ts
```diff
@@ -123,6 +123,29 @@ IQueryOperationResultBindings
...
+
+            expect(ActorRdfJoinMultiBind.canBindWithOperation(filter)).toBe(false);
+          });
+
+          it('should not reject a FILTER nested inside a FILTER EXISTS', () => {
+            const innerFilter: any = FACTORY.createFilter(<any>{}, FACTORY.createTermExpression(DF.literal('')));
```
jitsedesmet: and here

packages/actor-rdf-join-inner-multi-bind/test/ActorRdfJoinMultiBind-test.ts
```diff
@@ -123,6 +123,29 @@ IQueryOperationResultBindings
...
+            const existsExpr = FACTORY.createExistenceExpression(false, innerFilter);
+            const outerFilter: Algebra.Operation = FACTORY.createFilter(<any>{}, existsExpr);
```
jitsedesmet: and here

jitsedesmet (changes requested):

jitsedesmet (approved):

packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
```diff
@@ -235,6 +235,27 @@ export class ActorRdfJoinMultiBind extends ActorRdfJoin<IActorRdfJoinMultiBindTe
           return { shortcut: false };
         },
       },
+      [Algebra.Types.FILTER]: {
+        preVisitor: (op: Algebra.Filter) => {
+          // Conflict when FILTER is not a direct child of LEFT_JOIN
+          const conflict = !op.metadata?.isHoistedLeftJoinFilter;
+          if (conflict) {
```
rubensworks:
```suggestion
          if (!op.metadata?.isHoistedLeftJoinFilter) {
```

packages/actor-query-operation-leftjoin/test/ActorQueryOperationLeftJoin-test.ts
rubensworks: Yes, what Jitse says :-)

rubensworks (changes requested):
Very clean indeed! Just two minor things, and this one is good to go.

PR #1757, opened 2026-08-27
Add alternative source-based FROM/FROM NAMED actor

### LINKED ISSUE #198: Add alternative source-based FROM/FROM NAMED actor

noahvsb:
Resolves #198.


packages/actor-optimize-query-operation-set-sources-from-dataset/lib/ActorOptimizeQueryOperationSetSourcesFromDataset.ts
```diff
@@ -0,0 +1,87 @@
...
+    return { operation, context };
+  }
+
+  public static extractDatasetClauses(operation: Algebra.Operation): IDatasetClauses {
+    const defaultGraphs: string[] = [];
+    const namedGraphs: string[] = [];
+
+    algebraUtils.visitOperation(operation, {
+      from: {
+        visitor(op: Algebra.From) {
```
rubensworks: This will strip *all* of them, even for nested subqueries, which we probably want to keep (e.g. when they are contained within a SERVICE block that is sent to another endpoint).

packages/actor-optimize-query-operation-set-sources-from-dataset/lib/ActorOptimizeQueryOperationSetSourcesFromDataset.ts
```diff
@@ -0,0 +1,87 @@
...
+import { failTest, passTestVoid, type IActorTest, type TestResult } from '@comunica/core';
+import type { IActionContext, SourceType } from '@comunica/types';
+import type { Algebra } from '@comunica/utils-algebra';
+import { algebraUtils } from '@comunica/utils-algebra';
+import type * as RDF from '@rdfjs/types';
+
+/**
+ * A comunica Optimize Query Operation Set Sources From Dataset Optimize Query Operation Actor.
+ */
+export class ActorOptimizeQueryOperationSetSourcesFromDataset extends ActorOptimizeQueryOperation {
```
rubensworks, line 16: Didn't check the code in detail; but does this ensure that data targetted at with FROM NAMED is wrapped within a named graph?
We'll need to decide what to do when targetting TriG files with other named graphs (override them or keep the graphs?). Maybe @jitsedesmet can help you with thinking.
I feel like defaulting to completely dropping named graphs targetted by FROM NAMED might be the safest approach.

packages/actor-optimize-query-operation-set-sources-from-dataset/lib/ActorOptimizeQueryOperationSetSourcesFromDataset.ts
jitsedesmet: Why do you talk about subqueries? Are they not simply a projection operator? For SERVICE, I agree we should take special care.

packages/actor-optimize-query-operation-set-sources-from-dataset/lib/ActorOptimizeQueryOperationSetSourcesFromDataset.ts
jitsedesmet: Hello, my name is Jitse. I am here to help with thinking. I'd say start with looking at what other engines do.

But then again, looking at the spec, it says that you target a _dataset_ using `FROM` and `FROM NAMED` so that means that you would need a `GRAPH` operation to get the graphs from a dataset? And then using the options that say that `defaultGraph = unionGraph` consumers could select their desired behaviour?
https://www.w3.org/TR/sparql12-query/#specDataset

rubensworks:
@jitsedesmet Can you check first and ping me once you're happy?

packages/context-entries/lib/Keys.ts
```diff
@@ -342,6 +342,12 @@ export const KeysQueryOperation = {
   serviceSources: new ActionContextKey<Record<string, IQuerySourceWrapper>>(
     '@comunica/bus-query-operation:serviceSources',
   ),
+  /**
+   * A boolean denoting if datasets/graphs in FROM (NAMED) clauses should be set as sources (default: false).
+   */
```
jitsedesmet, line 347: Can you be specific? The `/` leaves interpretation? is it _or_, _and_, is actually one of the two correct?

packages/actor-optimize-query-operation-set-sources-from-dataset/README.md
```diff
@@ -0,0 +1,41 @@
...
+      "@id": "urn:comunica:default:optimize-query-operation/actors#set-sources-from-dataset",
+      "@type": "ActorOptimizeQueryOperationSetSourcesFromDataset"
+    }
+  ]
+}
+```
+
+### Config Parameters
+
+TODO: fill in parameters (this section can be removed if there are none)
```
jitsedesmet, line 39: TODO

packages/context-entries/lib/Keys.ts
jitsedesmet: Could you maybe also specify what would be the deafult bahaviour?

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -2323,6 +2323,72 @@ WHERE { }
...
+
+        globalThis.fetch = <typeof globalThis.fetch> jest.fn(async(input: string, init?: RequestInit) => {
+          if (input === datasetIri) {
+            return <Response> {
+              status: 200,
+              ok: true,
+              headers: new Headers({ 'content-type': 'text/turtle' }),
+              body: stringToStream(datasetTurtle),
+              url: input,
+            };
```
jitsedesmet, line 2354: If this is possible, it would be cleaner, but have a look whether it can work:



```suggestion
            return {
              status: 200,
              ok: true,
              headers: new Headers({ 'content-type': 'text/turtle' }),
              body: stringToStream(datasetTurtle),
              url: input,
            } satisfies Response;
```

packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts
```diff
@@ -198,6 +198,10 @@ export class CliArgsHandlerBase implements ICliArgsHandler {
           type: 'boolean',
           describe: 'If extension functions must always be pushed down',
         },
+        fromNamesAsSources: {
+          type: 'boolean',
+          describe: 'When true, sets datasets/graphs in FROM (NAMED) clauses as sources',
```
jitsedesmet, line 203: same interpretation of `/`

packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts
```diff
@@ -373,6 +377,11 @@ export class CliArgsHandlerBase implements ICliArgsHandler {
     if (args.extensionFunctionsAlwaysPushdown) {
       context[KeysInitQuery.extensionFunctionsAlwaysPushdown.name] = true;
     }
+
+    // Define if datasets/graphs in FROM (NAMED) clauses should be set as sources
```
jitsedesmet, line 381: here too

packages/actor-optimize-query-operation-set-sources-from-dataset/lib/ActorOptimizeQueryOperationSetSourcesFromDataset.ts
```diff
@@ -0,0 +1,87 @@
...
+    });
+
+    return { defaultGraphs, namedGraphs };
+  }
+
+  public static appendSources(context: IActionContext, clauses: IDatasetClauses): IActionContext {
+    const existingSources: SourceType[] = context.get(KeysInitQuery.querySourcesUnidentified) ?? [];
+
+    const mergedSources = [ ...existingSources, ...clauses.defaultGraphs, ...clauses.namedGraphs ].filter(
+      (source, index, self) => index === self.indexOf(source),
```
jitsedesmet, line 63: I think `SourceType` will be an object? So indexof will look for the same object? (not a deeped equlity) Is this desired?

packages/actor-optimize-query-operation-set-sources-from-dataset/lib/ActorOptimizeQueryOperationSetSourcesFromDataset.ts
```diff
@@ -0,0 +1,87 @@
...
+      },
+    });
+
+    return { defaultGraphs, namedGraphs };
+  }
+
+  public static appendSources(context: IActionContext, clauses: IDatasetClauses): IActionContext {
+    const existingSources: SourceType[] = context.get(KeysInitQuery.querySourcesUnidentified) ?? [];
+
+    const mergedSources = [ ...existingSources, ...clauses.defaultGraphs, ...clauses.namedGraphs ].filter(
```
jitsedesmet, line 62: Do we need a differentiation between `FROM Named` and `FROM`? I currently don't see one?

packages/actor-optimize-query-operation-set-sources-from-dataset/lib/ActorOptimizeQueryOperationSetSourcesFromDataset.ts
jitsedesmet: Oh, it might just be strings. In that case, a construction like `[ ...new Set(mergedSources) ]`

might be desired. Since the current implementation is O(n^2). 


(not that the set will be better neceserelly, but at least that way we have the illusion that there it is implemented in a smart fassion)

packages/actor-optimize-query-operation-set-sources-from-dataset/test/ActorOptimizeQueryOperationSetSourcesFromDataset-test.ts
```diff
@@ -0,0 +1,131 @@
...
+        operation: <any> {},
+      };
+      await expect(actor.test(action)).resolves.toFailTest(
+        'This actor can only be used when fromNamedAsSources is enabled.',
+      );
+    });
+    it('passes if fromNamedAsSources is true', async() => {
+      const action: IActionOptimizeQueryOperation = {
+        context: new ActionContext().set(KeysQueryOperation.fromNamedAsSources, true),
+        operation: <any> {},
```
jitsedesmet, line 33:
```suggestion
        operation:factory.createNop(),
```

packages/actor-optimize-query-operation-set-sources-from-dataset/test/ActorOptimizeQueryOperationSetSourcesFromDataset-test.ts
```diff
@@ -0,0 +1,131 @@
...
+  beforeEach(() => {
+    bus = new Bus({ name: 'bus' });
+    actor = new ActorOptimizeQueryOperationSetSourcesFromDataset({ name: 'actor', bus });
+  });
+
+  describe('test', () => {
+    it('fails if fromNamedAsSources is false', async() => {
+      const action: IActionOptimizeQueryOperation = {
+        context: new ActionContext(),
+        operation: <any> {},
```
jitsedesmet, line 24: here to

packages/actor-optimize-query-operation-set-sources-from-dataset/lib/ActorOptimizeQueryOperationSetSourcesFromDataset.ts
```diff
@@ -0,0 +1,87 @@
...
+      (source, index, self) => index === self.indexOf(source),
+    );
+
+    return context.set(KeysInitQuery.querySourcesUnidentified, mergedSources);
+  }
+
+  public static stripDatasetClauses(operation: Algebra.Operation): Algebra.Operation {
```
jitsedesmet, line 69: I think it is worth adding something like this because subSelects etc cannot have a FROM, nor can UpdateQueries.

```suggestion
   /**
   * According to SPARQL 1.2 grammar, a FROM and FROM named clause can only appear in the top level Query
   */
  public static stripDatasetClauses(operation: Algebra.Operation): Algebra.Operation {
```

jitsedesmet (changes requested):
Good stuff already

packages/actor-optimize-query-operation-set-sources-from-dataset/lib/ActorOptimizeQueryOperationSetSourcesFromDataset.ts
jitsedesmet: Looking at the grammar, From can only appear in the top level query. I even wonder whether we need to do this removal using a full tree traversal since we can assume valid SPARQL was used to construct the algebra tree? @rubensworks and @noahvsb

rubensworks: Yes, assume top-level.

packages/context-entries/lib/Keys.ts
noahvsb: I'm not sure what the difference between datasets and graphs actually are. I think in SPARQL 1.0 they're the same.

packages/actor-optimize-query-operation-set-sources-from-dataset/lib/ActorOptimizeQueryOperationSetSourcesFromDataset.ts
noahvsb: For the `extractDatasetClauses`-function, does that mean I also need to only check the top-level `FROM`'s?

noahvsb: Wait nevermind, this is the `extractDatasetClauses`-function.

packages/context-entries/lib/Keys.ts
jitsedesmet: Yes, I think for 1.0 they are, but for 1.2 they are not.

https://www.w3.org/TR/sparql12-query/#rdfDataset

engines/query-sparql/test/QuerySparql-test.ts
```diff
@@ -2323,6 +2323,72 @@ WHERE { }
...
+          {
+            sources: [],
+          },
+        )).rejects.toThrow('none of the configured actors were able to handle the operation type pattern');
+      });
+
+      it('appends the FROM IRI as a real source when fromNamedAsSources is true', async() => {
+        const originalFetch = globalThis.fetch;
+
+        globalThis.fetch = <typeof globalThis.fetch> jest.fn(async(input: string, init?: RequestInit) => {
```
rubensworks, line 2346: Let's not override the global fetch. You can instead just pass a custom fetch function through the comunica context.

packages/actor-init-query/lib/cli/CliArgsHandlerBase.ts
```diff
@@ -198,6 +198,10 @@ export class CliArgsHandlerBase implements ICliArgsHandler {
           type: 'boolean',
           describe: 'If extension functions must always be pushed down',
         },
+        fromNamesAsSources: {
```
rubensworks, line 201: `dereferenceFromNamed` sounds a bit nicer IMO (probably applies to other places too) :-)

rubensworks (changes requested):

packages/actor-optimize-query-operation-set-sources-from-dataset/lib/ActorOptimizeQueryOperationSetSourcesFromDataset.ts
jitsedesmet: I only see the difference now. `FROM` says merge this dataset with the defaultGraph, while `FROM NAMED` says add this graph to the dataset but keep it within the named graph.

So the cases are:
1. `FROM` on a file with triples in default graph: Merge them in the target default graph
2. `FROM` on a file with named graph triples: choice we need to make
3. `FROM NAMED` on a file with triples in default graphs: choice we need to make
4. `FROM NAMED` on a file with named graph triples: Keep the named graphs.

I'd say only (1) is truly natural. (4) is already an opinion but one I am willing to defend. (2) and (3) have no right answer.

Starting with (3), I would also keep the triples within a named graph.

For (2) I think I would keep the named graphs within the target dataset. I do envision use cases that would benefit from being able to discover all graphs within a file (to get full functional coverage). To this end, we should implement support for a SERVICE operation that targets a dataset just like we are doing here (which is something Comunica can already do I think). So you would be able to write:
```sparql
  # All triples (in A and B) expilicitly referencing the named graphs in A.
SELECT ?tt
FROM <ex://a>
FROM <ex://b> {
  SERVICE <ex://a> { GRAPH ?g { } }
  {
    { ?g ?p ?o . BIND(<<(?g ?p ?o )>> AS ?tt ) }
    UNION { ?s ?g ?o . BIND(<<(?s ?g ?o )>> AS ?tt ) }
    UNION { ?s ?p ?g . BIND(<<(?s ?p ?g )>> AS ?tt ) }
  }
}
```
(I write too much weird SPARQL, lol)

rubensworks:
For reference, #1769 continues upon this. I still need to critically go through it myself.

packages/actor-optimize-query-operation-set-sources-from-dataset/lib/ActorOptimizeQueryOperationSetSourcesFromDataset.ts
Bavop: @jitsedesmet I added a new flag `dereferenceFromNamedConflictMode` with 3 modes for when a `FROM NAMED` source already contains its own named graph(s):

- `error` (default): throws, refusing to load the source, avoids silently losing the existing named graph data.
- `overwrite`: discards the source's existing named graphs, rewriting everything into the graph named after the `FROM NAMED` IRI.
- `merge`: leaves the source's existing named graphs untouched, and only rewrites its default-graph quads into the `FROM NAMED` IRI's graph.

jitsedesmet: @Bavop I agree that adding a context option is a good idea, I do suggest, however, that we make it a callback (sync or async, I do not care too much; maybe @rubensworks does?)
The callback will give configurers more fine grained control.
I can for example, see someone make a language extension that would encapsulate the information we talk about here and then they can still rewrite, pipe to Comunica and provide the callbacks needed to control each source.
The `error` is default and sounds good.
`overrite` is rather vague for me; I'd go for `mergeNamedInSourceGraph`
and for `merge` -> `keepSourceGraphs`

rubensworks: How would such a callback look @jitsedesmet? I can not immediately see the value of it myself. It sounds like it could complicate this PR quite a bit, for little benefit, so maybe we keep it as future work?

jitsedesmet: We are already adding the options above as part of the context, right? So you'd just have a callback in the context like:
`(name: RDF.Term) => 'error' | 'mergeNamedInSourceGraph' | 'keepSourceGraphs'`

If we are adding a context option, I'd make it a callback. I see little value in a raw option that holds for every source. The callback allows finer control at no cost. (I presume the FROM NAMED code still performs a for loop over all datasets in the `from (named)` operation and assigns as such. A callback in this loop required minimal changes.

rubensworks: > So you'd just have a callback in the context

Sure, after looking at the code, this does make sense indeed. (but won't work as CLI option).

@Bavop can you make the necessary changes?

rubensworks:
Followed up in #1824

PR #1761, opened 2026-08-27
Fix deskolemization for destinations wrapped in a source object

### LINKED ISSUE #985: bus-rdf-update-quads could not find the source for skolemization
#### Issue type:

- :bug: Bug 

____
#### Description:

I am trying to create a tree of widgets for a form system.
The tree of widgets starts with a LDflex path. 
Children and grand children of this LDflex path are passed onto widgets.
The widgets have a input element onchange event and that changes the term inside the LDflex.
For the widgets that are in the root of the tree it works.
For widgets that are in a sub tree the skolemization seems to fail.

I am using LDflex, @LDflex/comunica.

I found two things: 
- [The comunica bus could not find the source, if hardcoded to '0' it works](https://github.com/danielbeeke/frm/blob/52e89915cd99ef23f87e5f754b5ffe826311d0e2/patches/%40comunica%2Bbus-rdf-update-quads%2B2.2.0.patch)
- [LDflex did not skolemize the subject inside the UPDATE](https://github.com/danielbeeke/frm/blob/52e89915cd99ef23f87e5f754b5ffe826311d0e2/patches/ldflex%2B2.15.0.patch)

The first patch is incorrect ofcourse, but it may give a pointer to where things go wrong.
The second patch I do not know, I would think skolemization needs to happen but I am just starting to learn about all these things.

With these two patches applied the project works.

____
#### Environment:
I working on a browser app.

    "@comunica/query-sparql": "^2.1.0",
    "@ldflex/async-iteration-handlers": "^1.1.1",
    "@ldflex/comunica": "^5.0.1",
    "ldflex": "^2.15.0",

#### Crash log:

https://pastebin.com/8TyPYVfa

The error is inside sparql.js where it finds a blank node.

rubensworks:
Closes #985

## Problem

`deskolemize()` in `ActorRdfUpdateQuadsDestination` looked up the destination's source id using the **raw** `KeysRdfUpdateQuads.destination` context value as the map key:

```ts
const destination = action.context.get(KeysRdfUpdateQuads.destination);
const id = action.context.get<Map<any, string>>(KeysQuerySourceIdentify.sourceIds)?.get(destination);
if (!id) {
  return action;   // silently skips deskolemization
}
```

But `getSourceId()` keys that map on `source.referenceValue`. Those are the same JS object only when the destination is a bare RDF/JS store or a bare URL string. Any `IDataDestination` wrapper — which is exactly what `ActorContextPreprocessSourceToDestination` copies straight through from `sources[0]` — misses the lookup, and the miss is silent.

The result: deletes carry skolemized terms that match nothing, and inserts write skolemized labels into the destination.

```js
// sources: [{ type: 'rdfjs', value: store }]
// DELETE { ?s <ex:first> ?o } INSERT { ?s <ex:firstNew> ?o } WHERE { ?s <ex:first> ?o }

_:b1  ex:first     ex:One       // should have been deleted
_:bc_0_b1  ex:firstNew  ex:One  // inserted with the skolemized label
```

And for the LDflex pattern from the issue — take a blank node's skolemized IRI from a `SELECT`, use it as a subject in a separate update:

```js
urn:comunica_skolem:source_0:b1  ex:label  "hello"   // raw skolem IRI written to the store
```

This is what @danielbeeke's [`'0'` hardcode patch](https://github.com/comunica/comunica/blob/52e89915cd99ef23f87e5f754b5ffe826311d0e2/patches/%40comunica%2Bbus-rdf-update-quads%2B2.2.0.patch) was masking: the id was `undefined`, and `'0'` happened to be the right answer for a single source.

## Fix

Unwrap the destination with `getDataDestinationValue()` before the lookup, mirroring how [`FragmentSelectorShapes.ts`](https://github.com/comunica/comunica/blob/master/packages/utils-query-operation/lib/FragmentSelectorShapes.ts#L226) already compares a destination against a source's `referenceValue`. Since that helper cannot handle `undefined`, the no-destination case now returns early.

This adds `@comunica/utils-query-operation` as a dependency of `@comunica/bus-rdf-update-quads`; it introduces no dependency cycle (`utils-query-operation` only depends on `context-entries`, `core`, `types`, `utils-algebra` and `utils-bindings-factory`), and the package's tests already imported from it.

## Tests

Unit tests in `bus-rdf-update-quads` (each verified to fail without the fix):

- deskolemizes for a destination wrapped in an `IDataDestination` object;
- still does *not* deskolemize a wrapped destination belonging to a different source;
- still does not deskolemize when the destination has no source id at all (this path was previously only reached via the no-destination case, which now returns earlier).

System tests in `query-sparql`:

- `INSERT … WHERE` over a blank node with `sources: [{ type: 'rdfjs', value: store }]`, asserting the original blank node label is used and no `bc_0_*` label leaks in;
- the LDflex pattern: a `SELECT` to obtain `urn:comunica_skolem:source_0:b1`, then a separate `INSERT DATA` using that IRI, asserting it resolves back to `_:b1`.

Full suite passes with coverage thresholds intact.

## Note on the second half of #985

The issue also reported that [LDflex did not skolemize the subject inside the UPDATE](https://github.com/danielbeeke/frm/blob/52e89915cd99ef23f87e5f754b5ffe826311d0e2/patches/ldflex%2B2.15.0.patch). That remains an LDflex-side matter, as @rubensworks noted at the time — this PR only covers the Comunica half.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PRtFXm1gq9aatHUN8KRfuF)_

### DIFF
--- a/packages/bus-rdf-update-quads/lib/ActorRdfUpdateQuadsDestination.ts
+++ b/packages/bus-rdf-update-quads/lib/ActorRdfUpdateQuadsDestination.ts
@@ -25,8 +25,6 @@ export function deskolemize(action: IActionRdfUpdateQuads): IActionRdfUpdateQuad
   if (!destination) {
     return action;
   }
-  // Source ids are keyed by the source's reference value, so the destination must be unwrapped
-  // before looking it up, as it may be wrapped in an IDataDestination object.
   const id = action.context.get<Map<any, string>>(KeysQuerySourceIdentify.sourceIds)
     ?.get(getDataDestinationValue(destination));
   if (!id) {
--- a/packages/bus-rdf-update-quads/lib/ActorRdfUpdateQuadsDestination.ts
+++ b/packages/bus-rdf-update-quads/lib/ActorRdfUpdateQuadsDestination.ts
@@ -3,6 +3,7 @@ import { KeysInitQuery, KeysQuerySourceIdentify, KeysRdfUpdateQuads } from '@com
 import type { IActorTest, TestResult } from '@comunica/core';
 import { passTestVoid } from '@comunica/core';
 import type { ComunicaDataFactory, IActionContext } from '@comunica/types';
+import { getDataDestinationValue } from '@comunica/utils-query-operation';
 import type * as RDF from '@rdfjs/types';
 import type { AsyncIterator } from 'asynciterator';
 import type { IActionRdfUpdateQuads, IActorRdfUpdateQuadsOutput } from './ActorRdfUpdateQuads';
@@ -21,7 +22,13 @@ AsyncIterator<RDF.Quad> | undefined {
 export function deskolemize(action: IActionRdfUpdateQuads): IActionRdfUpdateQuads {
   const dataFactory: ComunicaDataFactory = action.context.getSafe(KeysInitQuery.dataFactory);
   const destination = action.context.get(KeysRdfUpdateQuads.destination);
-  const id = action.context.get<Map<any, string>>(KeysQuerySourceIdentify.sourceIds)?.get(destination);
+  if (!destination) {
+    return action;
+  }
+  // Source ids are keyed by the source's reference value, so the destination must be unwrapped
+  // before looking it up, as it may be wrapped in an IDataDestination object.
+  const id = action.context.get<Map<any, string>>(KeysQuerySourceIdentify.sourceIds)
+    ?.get(getDataDestinationValue(destination));
   if (!id) {
     return action;
   }
--- a/packages/bus-rdf-update-quads/test/ActorRdfUpdateQuadsDestination-test.ts
+++ b/packages/bus-rdf-update-quads/test/ActorRdfUpdateQuadsDestination-test.ts
@@ -163,6 +163,93 @@ describe('ActorRdfUpdateQuadsDestination', () => {
       expect(store.getQuads(null, null, null, null)).toEqual([ q1 ]);
     });
   });
+
+  // https://github.com/comunica/comunica/issues/985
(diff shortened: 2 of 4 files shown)


packages/bus-rdf-update-quads/lib/ActorRdfUpdateQuadsDestination.ts
```diff
@@ -21,7 +22,13 @@ AsyncIterator<RDF.Quad> | undefined {
 export function deskolemize(action: IActionRdfUpdateQuads): IActionRdfUpdateQuads {
   const dataFactory: ComunicaDataFactory = action.context.getSafe(KeysInitQuery.dataFactory);
   const destination = action.context.get(KeysRdfUpdateQuads.destination);
-  const id = action.context.get<Map<any, string>>(KeysQuerySourceIdentify.sourceIds)?.get(destination);
+  if (!destination) {
+    return action;
+  }
+  // Source ids are keyed by the source's reference value, so the destination must be unwrapped
+  // before looking it up, as it may be wrapped in an IDataDestination object.
```
rubensworks:
```suggestion
```

packages/bus-rdf-update-quads/test/ActorRdfUpdateQuadsDestination-test.ts
```diff
@@ -163,6 +163,93 @@ describe('ActorRdfUpdateQuadsDestination', () => {
       expect(store.getQuads(null, null, null, null)).toEqual([ q1 ]);
     });
   });
+
+  // https://github.com/comunica/comunica/issues/985
```
rubensworks:
```suggestion
```

PR #1762, opened 2026-08-27
Only inject VALUES clauses in filters that use the bound variables

### LINKED ISSUE #1759: initialBindings adds a VALUES join at every FILTER, including sub-operations that don't use the bound variables
When a query is run with `initialBindings` against a SPARQL endpoint, `materializeOperation` joins a
`VALUES` clause into every `FILTER` operation:

https://github.com/comunica/comunica/blob/master/packages/utils-query-operation/lib/MaterializeBindings.js#L120-L129

```js
// Make a values clause using all the variables from originalBindings.
const values = createValuesFromBindings(algebraFactory, originalBindings);
...
recursionResultInput = algebraFactory.createJoin([...values, recursionResultInput]);
```

It uses *all* the original bindings, so the clause also lands in sub-operations that never mention
the bound variables. A query of ours with two bindings (`?query`, `?datasetUri`) comes out with six
copies of each. This one, for instance, gets both injected even though neither appears in the branch:

```sparql
OPTIONAL {
  ?uri ?parents ?broader .
  VALUES ?parents { gn:parentADM1 gn:parentADM2 }
  { VALUES ?query { "depot" } VALUES ?datasetUri { <https://www.geonames.org> }   # <-- added
    ?broader gn:officialName ?label . FILTER(LANG(?label) IN ("nl", "en")) }
  UNION
  { ?broader gn:name ?label }
  FILTER(?uri != ?broader)
}
```

That's semantically harmless, but it is expensive on the endpoint. Apache Jena stops pushing
`?broader` into the UNION branch once the branch contains a `VALUES` join, and scans the predicate
instead of probing the bound parents. Measured against Fuseki 6.2.0 over a 134M-triple store, the
same query with and without that one injected clause:

```
without:  200 in  0.21 s
with:     503 in 30.12 s   (cut off by arq:queryTimeout)
```

Everything else about the two queries is byte-identical, and removing any other injected clause
changes nothing – it's specifically the one inside the UNION branch.

The bindings are already joined in at the top of the query, so as far as I can tell the extra
clauses aren't needed for correctness: the values propagate by join anyway. Restricting the
injection to sub-operations that actually reference the bound variables (or hoisting it once to the
top) would keep the semantics and stop the planner falling over. `EXISTS`/`NOT EXISTS` filters are
already skipped a few lines above, so there's precedent for being selective here.

Version: `@comunica/query-sparql` 5.3.0, endpoint is Jena Fuseki 6.2.0 with a `text:query` index.

Possibly unrelated, but in case it rings a bell: I couldn't build a runnable reproduction against a
local endpoint on 5.3.0, because `initialBindings` + a `type: 'sparql'` source fails with
`fetch failed` / `invalid content-length header` before any request goes out. The same query without
`initialBindings` is sent fine. The evidence above comes from captured production traffic plus the
source, not from that script.

rubensworks:
Fixes #1759.

## Problem

`materializeOperation` joined a `VALUES` clause for **every** bound variable into **every** `FILTER` operation, including sub-operations that never refer to those variables. For a query run with two `initialBindings` against a SPARQL endpoint, this results in several copies of each clause spread over the query, e.g. inside a UNION branch that mentions neither variable:

```sparql
OPTIONAL {
  ?uri ?parents ?broader .
  VALUES ?parents { gn:parentADM1 gn:parentADM2 }
  { VALUES ?query { "depot" } VALUES ?datasetUri { <ex:d> }   # <-- added, but unused here
    ?broader gn:officialName ?label . FILTER(LANG(?label) IN ("nl", "en")) }
  UNION
  { ?broader gn:name ?label }
  FILTER(?uri != ?broader)
}
```

This is semantically harmless, but it is expensive on the endpoint: as reported in the issue, Apache Jena stops pushing `?broader` into the UNION branch once that branch contains a `VALUES` join, turning a 0.21s query into a timeout.

Since `ActorRdfJoinMultiBind` re-materializes its sub-operations for every binding of the left-hand stream, the same clauses also end up in bind-join sub-queries, not only in queries using `initialBindings`.

## Change

The `VALUES` clauses in a filter are now only created for the bound variables that the filter actually uses: the variables in scope of its input, plus the variables its expression refers to.

```ts
const values: Algebra.Operation[] = createValuesFromBindings(algebraFactory, originalBindings, [
  ...algebraUtils.inScopeVariables(filterOp.input),
  ...getExpressionVariables(filterOp.expression),
]);
```

Bound variables that are not used there don't have to be re-injected, since they are joined in at the operations that do refer to them, and at the projection of the query. When none of them is used, no join is added at all (restoring the `values.length > 0` guard from a531e81, which became unreachable and was dropped in fb27fc7).

The expression variables are needed next to the in-scope ones because in-scope variables deliberately exclude the variables that expressions *refer to* (`inScopeVariables` only reports aggregate and extend *targets*). Those are the ones that still matter with `bindFilter: false`, where the expression keeps referring to the variable instead of being materialized.

Both are pre-existing utils, so this adds no traversal code, and it is cheaper than injecting everywhere: the in-scope visitor prunes inside terms and stops at projections.

Verified on the query from the issue: the two injected clauses inside the UNION branch are gone, and the rest of the query is unchanged.

## SHACL pre-binding compatibility

`initialBindings` is what [shacl-engine](https://github.com/rdf-ext/shacl-engine) uses for SHACL pre-binding (`lib/sparql.js` calls `queryBindings(query, { sources, initialBindings })`), and the cases it depends on are captured in [bergos/comunica-tests](https://github.com/bergos/comunica-tests). All six were run against this branch and against the base commit: identical results, and each matches what the corresponding SHACL test expects.

| comunica-tests case | expected | before | after |
| --- | --- | --- | --- |
| pre-binding-002 (union of filter-only branches) | `$this` bound | ✅ | ✅ |
| pre-binding-004 (`BIND ($this AS ?that)`) | `$this` bound | ✅ | ✅ |
| pre-binding-005 (`bound($this)`) | `$this` bound | ✅ | ✅ |
| pre-binding-006 (sub-select, both projections) | `$this` bound | ✅ | ✅ |
| property-sparql-001 (`$PATH` not projected) | empty | ✅ | ✅ |
| unsupported-sparql-005 (`BIND (true AS $this)`) | error | ✅ | ✅ |

Each existing test now names the case it mirrors, and the one case that had no test — pre-binding-002, a union whose branches only contain a filter, so the pre-bound variable is never matched against a triple — has been added.

## Tests

* `MaterializeBindings-test.ts`:
  * `should not modify a filter expression without matching variables` now asserts that the operation is left untouched (matching what its name always claimed).
  * New: only a values clause for bound variables occurring in the filter expression.
  * New: only a values clause for bound variables occurring in the filter input.
  * New (regression for this issue): no values clauses in union branches that do not use the bound variables.
  * `should use the original InitialBindings in recursive calls with nested filter expressions` now uses filters that refer to the bound variable, so it still covers the propagation of the original bindings into recursive calls.
* `QuerySparql-test.ts`: new end-to-end `initialBindings` test with a filter inside a UNION branch that does not use the bound variable, asserting the results are unaffected, plus the pre-binding-002 test and the case references described above.

Full unit test suite (6726 tests) and the complete spec suite of `@comunica/query-sparql` (SPARQL 1.0/1.1/1.2 query, update, CSV/TSV, JSON, xsd datetime duration) pass.

### DIFF
--- a/packages/utils-query-operation/lib/MaterializeBindings.ts
+++ b/packages/utils-query-operation/lib/MaterializeBindings.ts
@@ -5,7 +5,7 @@ import type { BindingsFactory } from '@comunica/utils-bindings-factory';
 import type * as RDF from '@rdfjs/types';
 import type { Variable } from 'rdf-data-factory';
 import { termToString } from 'rdf-string';
-import { mapTermsNested, someTermsNested } from 'rdf-terms';
+import { getTermsNested, mapTermsNested, someTermsNested, uniqTerms } from 'rdf-terms';
 
 /**
  * Materialize a term with the given binding.
@@ -167,8 +167,17 @@ export function materializeOperation(
           return filterOp;
         }
 
-        // Make a values clause using all the variables from originalBindings.
-        const values: Algebra.Operation[] = createValuesFromBindings(algebraFactory, originalBindings);
+        // Make a values clause for the variables from originalBindings that occur inside this filter operation.
+        // Bound variables that don't occur here don't have to be re-injected,
+        // since they are joined in at the operations that do refer to them,
+        // and at the projection of the query.
+        // Injecting them everywhere is semantically harmless,
+        // but can significantly degrade the plans of (remote) query engines.
+        const values: Algebra.Operation[] = createValuesFromBindings(
+          algebraFactory,
+          originalBindings,
+          getOperationVariables(filterOp),
+        );
 
         // Recursively materialize the filter expression
         const recursionResultExpression: Algebra.Expression = <Algebra.Expression> materializeOperation(
@@ -188,7 +197,9 @@ export function materializeOperation(
           options,
         );
 
-        recursionResultInput = algebraFactory.createJoin([ ...values, recursionResultInput ]);
+        if (values.length > 0) {
+          recursionResultInput = algebraFactory.createJoin([ ...values, recursionResultInput ]);
+        }
 
         return algebraFactory.createFilter(recursionResultInput, recursionResultExpression);
       },
@@ -330,4 +341,53 @@ function createValuesFromBindings(
   }
 
   return values;
+}
+
+/**
+ * Get all variables that occur inside the given operation.
+ * Contrary to {@link algebraUtils.inScopeVariables}, this also considers variables that are not in scope,
+ * such as variables inside expressions and sub-queries.
+ * @param operation An algebra operation.
+ * @returns The variables occurring inside the given operation.
+ */
+function getOperationVariables(operation: Algebra.Operation): RDF.Variable[] {
+  const variables: RDF.Variable[] = [];
+  collectVariables(operation, variables);
+  return uniqTerms(variables);
+}
+
+/**
+ * Recursively collect all variables inside the given algebra value into the given array.
+ * @param value Part of an algebra operation, which can be an operation, an RDF term, an array, or any other value.
+ * @param variables The array to append the discovered variables to.
+ */
+function collectVariables(value: unknown, variables: RDF.Variable[]): void {
+  if (Array.isArray(value)) {
+    for (const entry of value) {
+      collectVariables(entry, variables);
+    }
+    return;
+  }
+  if (typeof value !== 'object' || value === null) {
+    return;
+  }
+  if ('termType' in value) {
+    const term = <RDF.Term> value;
+    if (term.termType === 'Variable') {
+      variables.push(term);
+    } else if (term.termType === 'Quad') {
+      for (const subTerm of getTermsNested(term)) {
+        if (subTerm.termType === 'Variable') {
+          variables.push(subTerm);
+        }
+      }
+    }
+    return;
+  }
+  for (const [ key, entry ] of Object.entries(value)) {
+    // Metadata can contain arbitrary (possibly cyclic) values that are not part of the query itself.
+    if (key !== 'metadata') {
+      collectVariables(entry, variables);
+    }
+  }
 }
--- a/packages/utils-query-operation/lib/MaterializeBindings.ts
+++ b/packages/utils-query-operation/lib/MaterializeBindings.ts
@@ -5,7 +5,8 @@ import type { BindingsFactory } from '@comunica/utils-bindings-factory';
 import type * as RDF from '@rdfjs/types';
 import type { Variable } from 'rdf-data-factory';
 import { termToString } from 'rdf-string';
-import { mapTermsNested, someTermsNested, uniqTerms } from 'rdf-terms';
+import { mapTermsNested, someTermsNested } from 'rdf-terms';
+import { getExpressionVariables } from './Expressions';
 
 /**
  * Materialize a term with the given binding.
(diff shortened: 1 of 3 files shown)


packages/utils-query-operation/lib/MaterializeBindings.ts
```diff
@@ -331,3 +342,21 @@ function createValuesFromBindings(
...
+function getOperationVariables(operation: Algebra.Operation): RDF.Variable[] {
+  const variables: RDF.Variable[] = [];
+  // The operation visitors skip the terms inside operations, so we visit all objects instead.
+  algebraUtils.transformer.visitObject(operation, (value) => {
+    if ((<RDF.Term> value).termType === 'Variable') {
+      variables.push(<RDF.Variable> value);
+    }
+  });
+  return uniqTerms(variables);
+}
```
jitsedesmet: I am thinking `inScopeVariables` would get the same thing done but cheaper?

jitsedesmet (changes requested):

packages/utils-query-operation/lib/MaterializeBindings.ts
rubensworks: Good call — done in 89e729a, the helper is gone entirely.

`inScopeVariables` alone isn't quite enough though: by design it only reports the aggregate/extend *target* variables of expressions, not the variables an expression *refers to* (`FILTER(?a = ex:x)` yields nothing). Those are exactly the ones that still matter when `bindFilter: false`, since then the expression keeps referring to the variable instead of being materialized. So it's the in-scope variables of the input plus the expression's own variables:

```ts
const values: Algebra.Operation[] = createValuesFromBindings(algebraFactory, originalBindings, [
  ...algebraUtils.inScopeVariables(filterOp.input),
  ...getExpressionVariables(filterOp.expression),
]);
```

Both are existing utils, so no new traversal code at all. It's also cheaper than what I had, as you expected: the in-scope visitor prunes inside terms and stops at projections, and expressions are small. Micro-benchmarked on a filter over a 7-pattern BGP with an OPTIONAL and a UNION, 20k iterations: 1405ms for the full object walk vs 662ms for this, so ~2.1x faster, with the same resulting variable set.

Duplicates between the two lists are harmless — `createValuesFromBindings` only does a `.some()` membership check — so no `uniqTerms` needed.

One behavioural nuance worth noting: variables that occur in the input but are *not* in scope there (inside a sub-select that doesn't project them, or on the right-hand side of a MINUS) no longer get a values clause. They're not visible at the filter's level anyway, so this is if anything more correct than what I had.

Full unit suite (6725 tests) and the complete spec suite pass, and the query from #1759 still comes out without the injected clauses.


---

packages/utils-query-operation/lib/MaterializeBindings.ts
```diff
@@ -167,8 +168,17 @@ export function materializeOperation(
...
 
-        // Make a values clause using all the variables from originalBindings.
-        const values: Algebra.Operation[] = createValuesFromBindings(algebraFactory, originalBindings);
+        // Make a values clause for the variables from originalBindings that are used by this filter operation:
+        // the variables in scope of its input, and the variables its expression refers to.
+        // Bound variables that are not used here don't have to be re-injected,
+        // since they are joined in at the operations that do refer to them,
+        // and at the projection of the query.
+        // Injecting them everywhere is semantically harmless,
+        // but can significantly degrade the plans of (remote) query engines.
```
rubensworks:
```suggestion
        // Bound variables that are not used here don't have to be re-injected.
```

PR #1768, opened 2026-08-28
Fixed three `FROM NAMED`-related bugs

noahvsb:
Part of #1736.

## bug 1

dataset-02 returned `bindings: [{}]`, while it expected `bindings: []`

## bug 2

When using `FROM` and `GRAPH` in the same query, it results in an error:

```
Error: A union can only be applied on at least one operation
....
```

## bug 3

When using `FROM NAMED` and `GRAPH` in the same query, it results in an error:

```
Error: Query operation processing failed: none of the configured actors were able to handle the operation type pattern
    Error messages of failing actors:
        Actor urn:comunica:default:query-operation/actors#source requires an operation with source annotation.
        ....
```


noahvsb:
The only bug left is the 3rd one, I haven't figured out what causes it, but I believe it's also to be fixed in `ActorQueryOperationFromQuad.ts`.

PR #1769, opened 2026-08-28
Fix the FROM (NAMED) dataset spec tests of SPARQL 1.0

rubensworks:
Part of #1736, and branched off https://github.com/comunica/comunica/pull/1768 (the first five commits are @noahvsb's, so that PR can be closed in favour of this one, or this one rebased once it lands).

With this, **all `dataset-*` tests of the SPARQL 1.0 suite pass**, and `dataset-` is dropped from the `spec:query-1-0` skip list.

## The remaining bugs on top of #1768

#1768 fixed the empty-solutions bug of `dataset-02` and the crash on `FROM` + `GRAPH` (`dataset-04`). Three more problems remained, all in the construction of the RDF dataset:

### 1. The source annotation was dropped when rewriting graphs

`ActorQueryOperationFromQuad` recreates patterns and paths with another graph, but only `applyOperationDefaultGraph` carried over their `metadata`. That metadata holds the `scopedSource` annotation, so every rewritten `GRAPH` pattern lost its source and failed with:

```
Actor urn:comunica:default:query-operation/actors#source requires an operation with source annotation.
```

This was the unresolved 3rd bug of #1768, and made `dataset-03`, `dataset-06`, `dataset-08`, `dataset-09b` and `dataset-10b` fail.

### 2. Operations within a FROM were pruned before their graphs were rewritten

`ActorOptimizeQueryOperationPruneEmptySourceOperations` asks the source whether a pattern has results, during query optimization. The graphs of patterns within a `FROM` are only rewritten when that `FROM` operation is executed, which happens *after* optimization, so those emptiness checks run against the wrong graphs.

For `dataset-07`, `dataset-11` and `dataset-12b`, this pruned the left-hand side of

```sparql
SELECT * FROM <data-g1.ttl> FROM NAMED <data-g2.ttl>
{ { ?s ?p ?o } UNION { GRAPH ?g { ?s ?p ?o } } }
```

because `?s ?p ?o` was still scoped to the (empty) default graph of the source at that point. The optimizer now no longer descends into `FROM` operations, just like it already did for `SERVICE`.

This is not limited to queries: the `USING` and `USING NAMED` clauses of an update are translated into a `FROM` operation *nested* within the update operation, so the same pruning made

```sparql
DELETE { GRAPH <g1> { ?s ?p ?o } } USING <g1> WHERE { { ?s ?p ?o } UNION { ?s ?p ?o } }
```

silently delete nothing. Both cases are covered by tests.

### 3. Graphs selected by FROM were matched by GRAPH clauses

The named graph transformation considered a pattern graph available when it occurred in `FROM` *or* `FROM NAMED`, because it ran after the default graph transformation had already rewritten default graph patterns to the `FROM` graphs. So `FROM <g1> { GRAPH <g1> { ?s ?p ?o } }` returned results, while SPARQL 1.0 (8.2) and SPARQL 1.1 (13.2) define the named graphs of the dataset as exactly those of `FROM NAMED` (and `FROM <g1> { GRAPH ?g { … } }` already correctly returned nothing, so the engine was also inconsistent with itself here).

The two transformations are swapped, so that the named graph transformation can still tell patterns over the default graph apart from patterns explicitly scoped via `GRAPH`, and only `FROM NAMED` graphs are exposed as named graphs. This matches the reading in https://github.com/comunica/comunica/pull/1757#discussion_r3886066906: `FROM` merges into the default graph, `FROM NAMED` keeps the graph named.

## Testing

- `dataset` manifest of SPARQL 1.0: 12 / 12 (was 4 / 12 on #1768)
- `spec:query-1-0`: 464 / 464
- `spec:query-1-1`: 326 / 326, `spec:update-1-1`: 157 / 157, `spec:csv-tsv-1-1`: 6 / 6, `spec:json-1-1`: 4 / 4, `spec:query-1-2`: 263 / 263
- Full unit test suite passes, both changed files remain at 100% coverage.
- Added unit tests for the metadata retention and the `FROM` skip, and a `RDF dataset construction with FROM and FROM NAMED` block of end-to-end tests in `QuerySparql-test.ts` covering the dataset rules of the spec (including property paths within `GRAPH`, and the `USING` case above).

### DIFF
--- a/engines/query-sparql/package.json
+++ b/engines/query-sparql/package.json
@@ -60,7 +60,7 @@
     "spec:base-1-0": "node ../../node_modules/rdf-test-suite/bin/Runner.js spec/sparql-engine.js https://w3c.github.io/rdf-tests/sparql/sparql10/manifest.ttl -c ../../.rdf-test-suite-cache/",
     "spec:base-1-1": "node ../../node_modules/rdf-test-suite/bin/Runner.js spec/sparql-engine.js http://w3c.github.io/rdf-tests/sparql/sparql11/manifest-all.ttl -c ../../.rdf-test-suite-cache/",
     "spec:base-1-2": "node ../../node_modules/rdf-test-suite/bin/Runner.js spec/sparql-engine.js https://w3c.github.io/rdf-tests/sparql/sparql12/manifest.ttl -c ../../.rdf-test-suite-cache/",
-    "spec:query-1-0": "yarn run --silent spec:base-1-0 --skip \"(date-1|date-2|open-eq-08|open-eq-10|open-eq-11|open-eq-12|#graph-|dataset-)\"",
+    "spec:query-1-0": "yarn run --silent spec:base-1-0 --skip \"(date-1|date-2|open-eq-08|open-eq-10|open-eq-11|open-eq-12|#graph-)\"",
     "spec:query-1-1": "yarn run --silent spec:base-1-1 -s http://www.w3.org/TR/sparql11-query/ --skip \"(agg-empty-group-count-graph|bindings/manifest#graph)\"",
     "spec:query-1-2": "yarn run --silent spec:base-1-2",
     "spec:update-1-1": "yarn run --silent spec:base-1-1 -s http://www.w3.org/TR/sparql11-update/",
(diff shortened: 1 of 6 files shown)


rubensworks:
Note to self: check all comments in #1757 as well.

PR #1772, opened 2026-08-30
Run the SPARQL 1.1 federated query spec tests

voidten:
This PR runs the SPARQL 1.1 federated query test suite as part of `yarn spec`.

> **Blocked:** depends on rubensworks/rdf-test-suite.js#130.

`service3`, `service5`, `service6` and `service7` are skipped. As mentioned in https://github.com/comunica/comunica/issues/721#issuecomment-4787164374, problems in the implementation should be out of scope, so I skipped those four.

PR #1779, opened 2026-09-01
Support SERVICE SILENT and nested SERVICE clauses

rubensworks:
Enables three of the four SPARQL 1.1 Federated Query spec tests that were skipped in https://github.com/comunica/comunica/pull/1772: `service3`, `service6` and `service7`. Contributes to https://github.com/comunica/comunica/issues/721.

`service5` uses a variable as SERVICE target and stays skipped here; it is handled separately in https://github.com/comunica/comunica/pull/1810.

## Background

Since [`04b51f8`](https://github.com/comunica/comunica/commit/04b51f8b45a4ef0cbc870fe132b102c237d58e17) ("Rewrite SERVICE clauses as query operation with annotated source"), SERVICE is handled at optimize time by `assign-sources-exhaustive` instead of by a runtime actor. That rewrite only covers `SERVICE <iri>` at the top level of the operation tree; anything it does not rewrite leaves a bare `service` node that nothing on the query operation bus can execute:

```
none of the configured actors were able to handle the operation type service
```

Two cases fell into that gap, and `SERVICE SILENT` regressed at the same time.

## Changes

**Nested SERVICE clauses (`service3`)** — these were deliberately left untransformed, but were then unexecutable. They are now annotated with the source of the *enclosing* SERVICE clause and passed along verbatim, so the parent target is responsible for resolving them, as the spec intends. `assignExhaustive` gained a `withinService` flag for this; `query-source-identify` still does not descend into SERVICE bodies.

**`SERVICE SILENT` (`service6`, `service7`)** — this only set `KeysInitQuery.lenient` on the target source, which is read solely by `ActorDereferenceBase`; errors raised by the source itself still propagated, so `SERVICE SILENT` behaved identically to non-silent. Silent targets now additionally carry a new `KeysQueryOperation.silent` flag on their source context, which `ActorQueryOperationSource` acts on by swallowing errors and producing a single empty solution, as required by [§3.2](https://www.w3.org/TR/sparql11-federated-query/#serviceFailure).

Actors that push bindings of one entry into the source of another declare this through the new `pushesBindingsToSource` option, and `ActorRdfJoin` rejects them for a silent target. They invoke the source once per chunk of bindings, bypassing `ActorQueryOperationSource` entirely, so such a clause would otherwise throw rather than produce its empty solution — and since an empty solution joins with everything, silencing per chunk would multiply the other side by the number of chunks. Emptiness probes in `prune-empty-source-operations` skip silent targets for the same reason: a failing target must produce an empty solution rather than be pruned away.

The spec test harness now passes its service fetch function into the nested `engine.query` call, so a mocked endpoint can resolve nested SERVICE clauses in turn.

## Testing

Spec suites, compared against a baseline run on `b5afaa3`:

| Suite | Before | After |
|---|---|---|
| SPARQL 1.1 federated query | 5 / 9 | **9 / 9** (`service5` skipped) |
| SPARQL 1.0 | 467 / 476 | 467 / 476 |
| SPARQL 1.1 query | 326 / 328 | 326 / 328 |
| SPARQL 1.1 update | 157 / 157 | 157 / 157 |
| SPARQL 1.1 csv/tsv, json, service description | all pass | all pass |
| SPARQL 1.2 | 263 / 263 | 263 / 263 |

The pre-existing SPARQL 1.0 and 1.1 failures are the ones the `spec:*` scripts `--skip`; they are unchanged by this PR.

Unit tests pass across the touched packages, with full coverage of the silencing iterator and every changed branch. Beyond the spec tests, a `SERVICE SILENT` against a broken endpoint joined with 40 local bindings returns exactly 40 rows, where the same query throws without the pushdown check.

## Notes for review

- **A behaviour difference worth a decision.** The `serviceAllowFileTargets` test added in https://github.com/comunica/comunica/pull/1808 asserts that a silent SERVICE against a blocked local file yields *no* results, and it still passes: `lenient` turns the dereference failure into an empty document before any error reaches the bindings stream, so the new silencing never sees it. A failing *endpoint* now yields the spec's single empty solution instead. Both are "the service failed", so the two paths disagree. Unifying them means not setting `lenient` for silent targets and letting the error reach `SilencedBindingsIterator`, which would change that test from 0 rows to 1. I left the merged behaviour alone — happy to unify if you prefer.
- Delegating nested SERVICE clauses assumes the parent endpoint supports nested SERVICE. That matches the spec and the manifest's own comment on `service3` ("This query depends in the capabilities of the example1.org endpoint"), but it does mean such queries fail against endpoints that do not support it.
- The delegated query serializes with a redundant `SELECT … WHERE { SELECT … WHERE { … } }` nesting. Harmless, but worth a follow-up.
- Refusing binding pushdown into silent sources costs that optimisation for `SERVICE SILENT` clauses; `multi-bind` takes over and silences per binding, which is correct.
- The empty solution is built with `new BindingsFactory(dataFactory)` rather than through `mediatorMergeBindingsContext`, to avoid adding a mediator parameter to `ActorQueryOperationSource`. It carries no values, so it needs no context merge handlers; a `TODO` marks this for the next major.
- `SilencedBindingsIterator` ends the stream quietly if a silent source errors after results were already emitted, since the empty solution can no longer be substituted. The spec only defines the all-or-nothing case.

https://claude.ai/code/session_01X4QghLG8yMWpYmoYHdckv9

### DIFF
--- a/packages/actor-rdf-join-inner-multi-smallest-filter-bindings/lib/ActorRdfJoinMultiSmallestFilterBindings.ts
+++ b/packages/actor-rdf-join-inner-multi-smallest-filter-bindings/lib/ActorRdfJoinMultiSmallestFilterBindings.ts
@@ -40,6 +40,7 @@ export class ActorRdfJoinMultiSmallestFilterBindings extends ActorRdfJoin {
     super(args, {
       logicalType: 'inner',
       physicalName: 'multi-smallest-filter-bindings',
+      pushesBindingsToSource: true,
       limitEntries: 2,
       limitEntriesMin: true,
       isLeaf: false,
--- a/packages/context-entries/lib/Keys.ts
+++ b/packages/context-entries/lib/Keys.ts
@@ -331,6 +331,12 @@ export const KeysQueryOperation = {
    */
   readOnly: new ActionContextKey<boolean>('@comunica/bus-query-operation:readOnly'),
   /**
+   * Flag on a query source context indicating that this source is the target of a `SERVICE SILENT` clause.
+   * Errors from such a source must be swallowed, and replaced by a single empty solution,
+   * as mandated by SPARQL 1.1 Federated Query.
+   */
+  silent: new ActionContextKey<boolean>('@comunica/bus-query-operation:silent'),
+  /**
    * An internal context entry to mark that a property path with arbitrary length and a distinct key is being processed.
    */
   isPathArbitraryLengthDistinctKey: new ActionContextKey<boolean>(
--- a/engines/query-sparql/package.json
+++ b/engines/query-sparql/package.json
@@ -66,7 +66,7 @@
     "spec:update-1-1": "yarn run --silent spec:base-1-1 -s http://www.w3.org/TR/sparql11-update/",
     "spec:csv-tsv-1-1": "yarn run --silent spec:base-1-1 -s http://www.w3.org/TR/sparql11-results-csv-tsv/",
     "spec:json-1-1": "yarn run --silent spec:base-1-1 -s http://www.w3.org/TR/sparql11-results-json/",
-    "spec:fed-1-1": "yarn run --silent spec:base-1-1 -s http://www.w3.org/TR/sparql11-federated-query/ --skip \"service/manifest#service(3|5|6|7)\"",
+    "spec:fed-1-1": "yarn run --silent spec:base-1-1 -s http://www.w3.org/TR/sparql11-federated-query/ --skip \"service/manifest#service5\"",
     "spec:sd-1-1": "yarn run --silent spec:base-1-1 -s http://www.w3.org/TR/sparql11-service-description/",
     "spec:prot-1-1": "yarn run --silent spec:base-1-1 -s http://www.w3.org/TR/sparql11-protocol/ --skip \"protocol/manifest#bad_update_dataset_conflict\"",
     "spec:graphstore-1-1": "yarn run --silent spec:base-1-1 -s http://www.w3.org/TR/sparql11-http-rdf-update/",
--- a/packages/actor-init-query/spec/sparql-engine-base.js
+++ b/packages/actor-init-query/spec/sparql-engine-base.js
@@ -113,7 +113,8 @@ function createServiceFetch(engine, serviceData) {
       return new Response(null, { status: 404 });
     }
 
-    const result = await engine.query(query, { sources: [ store ]});
+    // Pass the same fetch function, so that this endpoint can resolve nested SERVICE clauses.
+    const result = await engine.query(query, { sources: [ store ], fetch: serviceFetch });
     const mediaType = 'application/sparql-results+json';
     const body = await stringifyStream((await engine.resultToString(result, mediaType)).data);
     return new Response(body, { status: 200, headers: { 'content-type': mediaType }});
--- a/packages/actor-optimize-query-operation-assign-sources-exhaustive/lib/ActorOptimizeQueryOperationAssignSourcesExhaustive.ts
+++ b/packages/actor-optimize-query-operation-assign-sources-exhaustive/lib/ActorOptimizeQueryOperationAssignSourcesExhaustive.ts
@@ -58,12 +58,14 @@ export class ActorOptimizeQueryOperationAssignSourcesExhaustive extends ActorOpt
    * @param operation The input operation.
    * @param sources The sources to assign.
    * @param serviceSources Mapping of SERVICE names to sources.
+   * @param withinService If we are assigning sources within the body of a SERVICE clause.
    */
   public assignExhaustive(
     factory: AlgebraFactory,
     operation: Algebra.Operation,
     sources: IQuerySourceWrapper[],
     serviceSources: Record<string, IQuerySourceWrapper>,
+    withinService = false,
   ): Algebra.Operation {
     return algebraUtils.mapOperation(operation, {
       [Algebra.Types.PATTERN]: {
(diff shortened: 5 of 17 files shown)


packages/actor-query-operation-join/lib/ActorQueryOperationJoin.ts
```diff
@@ -41,6 +41,9 @@ export class ActorQueryOperationJoin extends ActorQueryOperationTypedMediated<Al
       .map(({ output, operation }) => ({
         output: getSafeBindings(output),
         operation,
+        // SERVICE clauses with a variable target can only be evaluated once that variable is bound,
+        // so we enforce a bind-join for these entries.
+        ...containsUnboundService(operation) && { operationRequired: <const> true },
```
rubensworks: It feels weird to hack this in here.
Can this not be done in a more clean manner?
For example within the bind join actor itself?
Or somewhere else?

packages/actor-rdf-join-inner-multi-smallest-filter-bindings/lib/ActorRdfJoinMultiSmallestFilterBindings.ts
```diff
@@ -216,6 +216,12 @@ export class ActorRdfJoinMultiSmallestFilterBindings extends ActorRdfJoin {
     if (!sourceWrapper) {
       return failTest(`Actor ${this.name} can only process if entries[1] has a source`);
     }
+    // Reject pushing bindings into the target of a SERVICE SILENT clause.
+    // Its bindings are pushed down in chunks, with one source call per chunk,
+    // so a failing source would produce one empty solution per chunk instead of exactly one.
+    if (sourceWrapper.context?.get(KeysQueryOperation.silent)) {
```
rubensworks: I don't understand why this is necessary here in this actor.

packages/context-entries/lib/Keys.ts
```diff
@@ -318,6 +318,14 @@ export const KeysQueryOperation = {
    * Flag for indicating that only read operations are allowed, defaults to false.
    */
   readOnly: new ActionContextKey<boolean>('@comunica/bus-query-operation:readOnly'),
+  /**
+   * Flag on a query source context indicating that this source is the target of a `SERVICE SILENT` clause.
+   * Errors from such a source must be swallowed, and replaced by a single empty solution,
+   * as mandated by SPARQL 1.1 Federated Query.
+   * Unlike `KeysInitQuery.lenient`, which makes a failing source contribute nothing,
+   * this makes a failing source contribute the join identity.
```
rubensworks:
```suggestion
```

packages/actor-query-operation-join/lib/ActorQueryOperationJoin.ts
rubensworks: Agreed, moved out of this actor entirely — it is now reverted to its original state.

The detection lives in `ActorRdfJoin.isOperationRequired()` in the join bus, which every join actor already funnels through in `test()`, next to the existing `operationRequired` check:

```ts
public static isOperationRequired(entry: IJoinEntry): boolean {
  if (entry.operationRequired) {
    return true;
  }
  // ...true for a SERVICE clause with a variable target
}
```

`ActorRdfJoinMultiBind` uses the same helper for its own "must not be the first entry" check. Two bonuses: it now also applies to optional and minus joins, which the old placement in `ActorQueryOperationJoin` missed entirely (so `OPTIONAL { SERVICE ?s { … } }` was broken), and the check moved below the entry-count and entry-type validations, so malformed entries are rejected before it runs.

One trade-off worth flagging: this traverses the entry's operation once per actor test rather than once per join. It short-circuits on the first SERVICE clause, and it is cheap next to the metadata resolution the bus already does, but if you would rather pay it once, the alternative is to compute it where the entries are built — which is what you just objected to, so I went this way.

---

packages/actor-rdf-join-inner-multi-smallest-filter-bindings/lib/ActorRdfJoinMultiSmallestFilterBindings.ts
rubensworks: It is needed, but you are right that it did not belong here. It is gone from this actor; instead this actor now declares what it does:

```ts
physicalName: 'multi-smallest-filter-bindings',
pushesBindingsToSource: true,
```

and the join bus rejects such actors for a silent target, in one place, alongside the `operationRequired` check.

Why it is needed at all: this actor chunks the first stream and issues **one `source.queryBindings` call per chunk**. Silencing lives in `ActorQueryOperationSource`, which this path bypasses completely, so without the check a `SERVICE SILENT` against a failing endpoint throws instead of yielding the empty solution. Concretely, a silent clause joined with 40 local bindings:

```
with the check:    40 rows
without the check: Error: Invalid SPARQL endpoint response from http://broken.example/sparql (HTTP status 404)
```

Even if the silencing were pushed down into the source object so the error were caught here, it would be wrong: each chunk would produce its own empty solution, and an empty solution joins with everything, so N chunks would multiply the left side by N. The spec wants exactly one. Hence refusing the pushdown rather than trying to silence it per chunk.

It does cost the pushdown optimisation for silent clauses — `multi-bind` takes over, which materialises per binding and therefore silences per binding correctly. If you see a way to keep the pushdown and still produce exactly one empty solution, I am happy to revisit.

---

packages/context-entries/lib/Keys.ts
rubensworks: Done, the two lines comparing against `lenient` are removed. The doc comment is now just:

```ts
/**
 * Flag on a query source context indicating that this source is the target of a `SERVICE SILENT` clause.
 * Errors from such a source must be swallowed, and replaced by a single empty solution,
 * as mandated by SPARQL 1.1 Federated Query.
 */
```

---

packages/actor-query-operation-service/README.md
```diff
@@ -0,0 +1,55 @@
...
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-query-operation-service.svg)](https://www.npmjs.com/package/@comunica/actor-query-operation-service)
+
+A [Query Operation](https://github.com/comunica/comunica/tree/master/packages/bus-query-operation) actor
+that handles SPARQL `SERVICE` operations that survive the optimization phase.
+
+Most `SERVICE` clauses are rewritten into source-annotated operations by
+[`@comunica/actor-optimize-query-operation-assign-sources-exhaustive`](https://github.com/comunica/comunica/tree/master/packages/actor-optimize-query-operation-assign-sources-exhaustive).
+This actor handles the remaining cases:
+
+* `SERVICE SILENT`, where errors must be swallowed and replaced by a single empty solution,
```
rubensworks: Are these docs up to date?

PR #1781, opened 2026-09-01
Make physical query plans complete and deterministic

rubensworks:
Physical query plans were built as a side effect of running, from three unrelated base classes, keyed on the algebra operation or join action object that the engine was executing. That single decision caused most of what was wrong with them: an actor that re-dispatched the same object lost its node, an actor that built a new object for an already-executed subtree could never refer back to it, and the tree was then patched up with a `stashChildren`/`unstashChild` pair whose failures were silent.

Each phase below is one commit.

## What was wrong

Every one of these was reproduced against the engine before being fixed:

- **`explain physical` crashed on queries that execute fine.** `ActorRdfJoin.run` dereferenced `sideData.metadatas` whenever a plan logger was present, while `ActorRdfJoinNone` passed none. `SELECT * WHERE { }` and `SELECT * WHERE { BIND(1 AS ?x) }` returned results normally but threw when explained.
- **Whole subtrees disappeared.** ALP path evaluation and `ActorRdfJoinWrapStream` both re-dispatch the same object, which was also the plan node key; a two-pattern BGP under the wrap-stream join actor explained as `join-inner(wrap-stream)` with nothing below it. A three-pattern BGP reported one of its two join inputs. The operation driving a bind join never appeared at all.
- **Sources were a black box.** A query over a SPARQL endpoint explained as a single line, `slice src:0`.
- **Statistics were join-only and sometimes fabricated.** Sharing a stream between nested operators made the outer one report `0`, which the compact renderer then hid because zero is falsy. Metadata was appended from a floating promise nothing awaited, so repeated runs of the same query over the same data produced different plans.
- **Collected information was dropped.** The actor of each node was recorded and never serialized, so `foaf:knows+` and `foaf:knows?` explained as identical, uninformative `path` chains.
- **Repetition was flattened.** A `FILTER EXISTS` added one child per binding next to the filter's real input; summarizing was hard-coded to bind joins and kept only the first occurrence.

## What changed

- **Phase 0** pins down the plan for a corpus of queries against an in-memory source, so every later change is visible. Cases that were wrong are marked, the two crashes are asserted, and the determinism test starts out as `it.failing`.
- **Phase 1** fixes the crash and reports duplicate node keys instead of silently overwriting them.
- **Phase 2** gives plan nodes their own identity: `logOperation` returns a handle, and that handle travels through the context. Join entries are found by the output they produced and adopted as inputs, which retires `stashChildren`/`unstashChild`. `isLeaf` is gone; a single-entry join now reports as `join-inner(single)` and a short-circuited one as `join-inner(empty)`.
- **Phase 3** makes measuring part of associating an output with a node, so every operator reports its cardinality and timings, and the explain actor awaits the logger before serializing. `instrumentIterator` now lets several measurements observe the same iterator, and no longer patches the iterators it reads from — doing so disturbed streams that were already flowing, to the point where a three-pattern BGP returned three of its five results while being explained.
- **Phase 4** records the operation tree a source was handed, marked as `delegated`, and has `QuerySourceSparql` report the query it sends and the HTTP requests it makes.
- **Phase 5** has operations that evaluate a sub-operation per binding — bind joins, `EXISTS`, ALP paths — group those evaluations under a node of their own, and summarizes repetitions only within such a group, with the totals they stand for. The actor of every node is serialized, interned into a legend.
- **Phase 6** expands the `explain` context shortcut in `QueryEngineBase`, so actors read `KeysInitQuery.explain` alone, and documents that this mode executes the query.
- **Review follow-ups**, in a final commit, are listed at the bottom of this description.

## What a plan looks like now

`SELECT * WHERE { ?s foaf:name ?n OPTIONAL { ?s foaf:knows ?f } }`, before:

```
project (f,n,s)
  leftjoin
    join-optional(bind) cardReal:6 timeSelf:2.18ms timeLife:4.949ms
      pattern (http://example.org/alice http://xmlns.com/foaf/0.1/knows ?f) src:0 compacted-occurrences:5
```

and after:

```
project (f,n,s) cardEst:13.648 cardReal:6 timeSelf:1.2ms timeLife:3.5ms actor:0
  leftjoin cardEst:13.648 cardReal:6 timeSelf:1.1ms timeLife:3.7ms actor:1
    join-optional(bind) cardEst:13.648 cardReal:6 timeSelf:1.1ms timeLife:4.5ms actor:2
      pattern (?s foaf:name ?n) cardEst:5 src:0 cardReal:5 timeSelf:0.9ms timeLife:9.2ms actor:3
      pattern (?s foaf:knows ?f) cardEst:5 src:0 cardReal:0 timeSelf:0.4ms timeLife:11.4ms actor:3
      bindings actor:2
        pattern (ex:alice foaf:knows ?f) cardEst:2 src:0 cardReal:2 timeSelf:0.3ms timeLife:1.1ms actor:3 compacted-occurrences:5 cardRealSum:5 timeSelfSum:1.4ms timeLifeSum:4.8ms

sources:
  0: QuerySourceRdfJs(N3Store)(SkolemID:0)

actors:
  0: urn:comunica:default:query-operation/actors#project
  1: urn:comunica:default:query-operation/actors#leftjoin
  2: urn:comunica:default:rdf-join/actors#optional-bind
  3: urn:comunica:default:query-operation/actors#source
```

## Breaking changes

`IJoinEntry` is unchanged. Everything that does break:

- **`IPhysicalQueryPlanLogger` (`@comunica/types`) is a different interface.** `logOperation` takes a single `ILogOperationArgs` object and returns an `IPhysicalQueryPlanNode` handle; `stashChildren`, `unstashChild` and `appendMetadata` are gone; `getNodeForOutput` and `finalize` are new. Anyone implementing their own logger has to write all four methods.
- **`IPhysicalQueryPlanNode` (`@comunica/types`) is new and has three required methods**: `appendMetadata`, `adoptInput` and `setOutput`. It replaces `IPlanNode`, which is removed.
- **`KeysInitQuery.physicalQueryPlanNode`** now holds one of those handles instead of an arbitrary object, so code that put its own object there, or read one back out, no longer type-checks.
- **`isLeaf` is removed from `IActorRdfJoinInternalOptions`.** This is a compile error for any join actor outside this repository that sets it; the plan now shows single-entry and zero-entry joins as `join-inner(single)` and `join-inner(empty)` instead of hiding them.
- **`instrumentIterator` (`@comunica/utils-iterator`) returns `{ counters, finish }`** instead of a promise of counters, and the counters carry a `state` of `ended`, `destroyed` or `unfinished`.
- **`physicalPlanMetadata` of the three bind-join actors no longer carries `bindOperation` or `bindOperationCardinality`**, and `ActorRdfJoinOptionalBind` no longer reports a constant `bindIndex: 0`. The bound operation is now a child node of its own, which reports the same pattern, source and cardinality.
- **The `physical-json` output gains fields and changes the shape of compacted children.** Anything parsing it needs updating.

## Deviations from the original design

- Join entries are located by the output they produced, via `getNodeForOutput`, rather than by adding a field to `IJoinEntry`, which had to stay unchanged.
- `timeSelf` is the time spent in an operator's own output iterator, not a figure that is strictly exclusive of its descendants. An operator that hands its input's stream through unchanged is measured on that same stream and reports the same `timeSelf` as the operator below it. The README says so rather than implying the numbers add up.

## Review follow-ups

A fresh-context review of this branch raised the following, all addressed in the last commit:

- **Plans of cut-short queries were less deterministic than they could be.** `finalize` waited a scheduler tick and then stopped every measurement, so an operator that was still running landed in `unfinished`, and its estimated cardinality was skipped. It now destroys what is still running immediately and always records the estimate; waiting longer was measured to make this worse, not better. `instrumentIterator` no longer reports an iterator that had already ended as `destroyed`. What remains is the engine's own: how far a branch below a `LIMIT` got before teardown genuinely varies between runs, which the README now states and a new corpus test pins down.
- **A live web page was asserted for its exact triple count.** `QuerySparql-test.ts` no longer hard-codes `523` for `https://www.rubensworks.net/`.
- **Templates were reported as evaluated operations.** `ActorQueryOperationSource.getSubOperations` listed the `template` of a `CONSTRUCT` and the `delete`/`insert` of a `DELETE INSERT` as sub-operations a delegating source had evaluated.
- **`ActorRdfJoin.run` guarded `sideData` in one place and dereferenced it two lines later.** The guard hid that `ActorRdfJoinSingle` passed `undefined!` from its `test`; it now passes the metadata of its single entry, and the guard is gone.
- **`REFACTOR-PLAN.md` was a planning artefact, not documentation**, and has been dropped from the package; what it recorded lives in this description and in the package README.
- Also: the two remaining stale `KNOWN ISSUE` comments in the corpus (both fixed by phase 5), the README's claim that a `LIMIT` query is measured as if drained, a doc comment for the last-writer-wins behaviour of `registerOutput`, and a note on how `httpRequests` are attributed when one source serves several operations at once.

## Verification

`yarn lint`, `yarn build`, `yarn test` and `yarn run test:config-compat` all pass; coverage stays at 100%.

https://claude.ai/code/session_01Y8oyzRck27fdrwfsqTre2W


packages/actor-query-process-explain-physical/lib/MemoryPhysicalQueryPlanLogger.ts
```diff
@@ -1,135 +1,205 @@
...
+    return typeof output === 'object' && output !== null ? this.nodesByOutput.get(output) : undefined;
+  }
+
+  /**
+   * Associate a query operation output with the given node, and start measuring it.
+   *
+   * An operation that passes its input's output through unchanged, such as a wrapping join,
+   * registers an output that was already registered by the operation below it. The last
+   * registration wins, which is what consumers need: they look an output up to find the
+   * operation that handed it to them, which is the outermost one.
```
rubensworks:
```suggestion
   * registration wins.
```

packages/utils-algebra/lib/utils.ts
```diff
@@ -406,6 +406,57 @@ export const visitOperation = transformer.visitNode.bind(transformer);
...
 export const visitOperationSub = transformer.visitNodeSpecific.bind(transformer);
 
+/**
+ * The keys that hold something other than the operations an operation is composed of.
+ *
+ * These come on top of the keys that {@link defaultNodePreVisitor} already leaves alone: the path a
+ * path traverses, and the templates a construct or an update produces, describe what to do rather
+ * than name an operation that is evaluated to obtain results.
+ */
+const nonOperationKeys: Partial<Record<string, Set<string>>> = {
```
rubensworks: Why is this needed? Does this indicate that there's a bug somewhere?

packages/utils-algebra/lib/utils.ts
```diff
@@ -406,6 +406,57 @@ export const visitOperation = transformer.visitNode.bind(transformer);
...
+ * Obtain the operations that the given operation is directly composed of, in the order they occur.
+ *
+ * Expressions are not included, as an expression is evaluated against a binding rather than being an
+ * operation that produces results of its own. Neither are the keys that {@link defaultNodePreVisitor}
+ * and {@link nonOperationKeys} describe as holding something other than an operation.
+ *
+ * @param operation An operation.
+ * @return The operations directly nested within it.
+ */
+export function getSubOperations(operation: Operation): Operation[] {
```
rubensworks: Can this not be done using something in algebraUtils instead?

packages/utils-query-operation/lib/PhysicalQueryPlan.ts
```diff
@@ -0,0 +1,42 @@
...
+ * arbitrary-length property paths, would otherwise report every evaluation as a separate child, next
+ * to their actual inputs. Grouping them makes the repetition explicit, and lets the plan summarize
+ * the repeated evaluations without hiding anything else.
+ *
+ * Calling this again for the same repetition yields the same group, so an operation that reaches for
+ * one per evaluation, rather than once up front, need not remember it. Which group that is follows
+ * from what is being repeated, so pass `operation` where one operation has several repetitions to
+ * keep apart, such as a filter with more than one `EXISTS`.
+ *
+ * The given context is returned unchanged when no physical query plan is being logged.
```
rubensworks: This comment is much longer than needed.

packages/actor-query-process-explain-physical/lib/MemoryPhysicalQueryPlanLogger.ts
rubensworks: Applied.

---

packages/utils-query-operation/lib/PhysicalQueryPlan.ts
rubensworks: Shortened to:

```ts
/**
 * Group the sub-operations that will be executed within the returned context under a single node of
 * the physical query plan, so that repeated evaluations are summarized rather than listed next to
 * the actual inputs.
 *
 * Calling this again for the same repetition yields the same group, so callers need not remember it.
 * The given context is returned unchanged when no physical query plan is being logged.
 */
```

---

packages/utils-algebra/lib/utils.ts
rubensworks: No bug — the canonical visitor traverses these keys on purpose, and it has to.

`defaultNodePreVisitor` answers "what should a traversal walk into", which is deliberately wider than "which nested operations get evaluated". A mapper must be able to rewrite a path expression and a quad template, so those keys stay traversable:

```js
const op = translate('CONSTRUCT { ?s ex:nick ?n } WHERE { ?s ex:name ?n }', { quads: true });
materializeOperation(op, bindings({ n: "Alice" }), AF, BF);
// CONSTRUCT { ?s ex:nick "Alice". } WHERE { ?s ex:name "Alice". }
```

That substitution into the template only works because `visitOperation`/`mapOperation` descend into `construct.template`. Same for `deleteinsert.delete`/`insert`, and for `path.predicate` (the path symbols are algebra nodes that mapping rewrites). Adding these to `defaultNodePreVisitor.ignoreKeys` would break materialization.

`getSubOperations` asks the narrower question — which nested operations are evaluated to produce results — so it subtracts the keys that hold data describing *what to do* rather than an operation to run. Happy to rename it to something that reads less like a general-purpose exclusion list if that helps.

---

packages/utils-algebra/lib/utils.ts
rubensworks: I looked, and there is nothing that fits: every helper in `algebraUtils` (ours and Traqula's) is a *recursive* map/visit over a whole subtree — `mapOperation`, `visitOperation`, `visitOperationSub` and their async/sub variants. This needs the direct children of one operation, one level deep.

The two candidates both come out worse:

- `transformer.visitNode` applies the per-type defaults, but dispatches by type, so collecting every child would mean a callback per operation type — the hardcoded list this change removed.
- `transformer.visitObject` is generic, but it walks the entire subtree and does not apply the per-type defaults, so it would need the same ignore keys passed in by hand plus depth tracking and array handling on top. I tried it; it is longer than the loop.

So the loop stays unless you would rather I add a single-level traversal to `utils-algebra` for it. Let me know and I'll do that instead.

---

packages/utils-algebra/lib/utils.ts
```diff
@@ -406,6 +406,81 @@ export const visitOperation = transformer.visitNode.bind(transformer);
...
 
+/**
+ * Visit the values held directly by the given operation, without descending into them.
+ *
+ * The other visitors walk a whole subtree, which is what you want to reach every operation of a
+ * kind. This is the single-level counterpart, for taking one operation apart: it visits the values
+ * of its own keys, and an array is visited element by element rather than as a whole.
+ *
+ * The keys that {@link defaultObjectContext} and {@link defaultNodePreVisitor} leave alone are
+ * skipped, so a caller sees the same members a traversal would descend into.
```
rubensworks:
```suggestion
```

packages/utils-algebra/lib/utils.ts
```diff
@@ -406,6 +406,81 @@ export const visitOperation = transformer.visitNode.bind(transformer);
...
+    }
+  }
+}
+
+/**
+ * The keys that hold something other than the operations an operation is composed of.
+ *
+ * These come on top of the keys that {@link defaultNodePreVisitor} already leaves alone: the path a
+ * path traverses, and the templates a construct or an update produces, describe what to do rather
+ * than name an operation that is evaluated to obtain results.
```
rubensworks:
```suggestion
 * These come on top of the keys that {@link defaultNodePreVisitor} already leaves alone.
```

packages/utils-algebra/lib/utils.ts
```diff
@@ -406,6 +406,81 @@ export const visitOperation = transformer.visitNode.bind(transformer);
...
+  [Types.CONSTRUCT]: new Set([ 'template' ]),
+  [Types.DELETE_INSERT]: new Set([ 'delete', 'insert' ]),
+};
+
+/**
+ * Obtain the operations that the given operation is directly composed of, in the order they occur.
+ *
+ * Expressions are not included, as an expression is evaluated against a binding rather than being an
+ * operation that produces results of its own. Neither are the keys that {@link nonOperationKeys}
+ * describes as holding something other than an operation.
```
rubensworks:
```suggestion
 * Expressions are not included.
```

PR #1782, opened 2026-09-01
Fix DESCRIBE queries with a FROM (NAMED) dataset

redreye:
A DESCRIBE query that defines its own dataset fails:

```js

const sources = [{
  type: 'serialized',
  mediaType: 'application/trig',
  baseIRI: 'http://example.org/',
  value: `
    <http://example.org/s> <http://example.org/p> <http://example.org/o> .
    <http://example.org/g> { <http://example.org/s> <http://example.org/p> <http://example.org/o> }
  `,
}];

async function run(query) {
  try {
    const quads = await new QueryEngine().queryQuads(query, { sources });
    console.log(`${query}\n  => ${(await quads.toArray()).length} quads`);
  } catch (error) {
    console.log(`${query}\n  => ${error.message.split('\n')[0]}`);
  }
}

(async() => {
  await run('DESCRIBE <http://example.org/s>');
  await run('CONSTRUCT { ?s ?p ?o } FROM <http://example.org/g> WHERE { ?s ?p ?o }');
  await run('DESCRIBE <http://example.org/s> FROM <http://example.org/g>');
})();
```

```
DESCRIBE <http://example.org/s>
  => 1 quads
CONSTRUCT { ?s ?p ?o } FROM <http://example.org/g> WHERE { ?s ?p ?o }
  => 1 quads
DESCRIBE <http://example.org/s> FROM <http://example.org/g>
  => Query operation processing failed: none of the configured actors were able to handle the operation type describe
```

A describe on its own works, and so does a FROM dataset on a construct, but their combination does not.

### DIFF
--- a/packages/actor-optimize-query-operation-describe-to-constructs-subject/lib/ActorOptimizeQueryOperationDescribeToConstructsSubject.ts
+++ b/packages/actor-optimize-query-operation-describe-to-constructs-subject/lib/ActorOptimizeQueryOperationDescribeToConstructsSubject.ts
@@ -8,7 +8,7 @@ import { KeysInitQuery } from '@comunica/context-entries';
 import type { IActorTest, TestResult } from '@comunica/core';
 import { failTest, passTest } from '@comunica/core';
 import type { ComunicaDataFactory } from '@comunica/types';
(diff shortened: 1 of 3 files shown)


rubensworks:
Good catch, and a clean fix. Thanks @redreye!

PR #1784, opened 2026-09-02
Respond with 400 to invalid SPARQL endpoint requests

redreye:
Follow-up to https://github.com/comunica/comunica/pull/1778.

A POST or QUERY request whose body the endpoint cannot interpret, such as one sent without a content type, makes parseBody reject. Nothing handles that rejection, so the client never receives a response at all, and the unhandled rejection terminates the process.

```js

(async() => {

  const service = new HttpServiceSparqlEndpoint({ engine, port: 3000 });

    method: 'POST',
    headers: { 'content-type': 'text/plain' },
    body: 'ASK {}',
  });
})();
```

Before, no status is ever printed, because the process is gone:

```
Error: Invalid request body received, query type could not be determined
```

After:

```
400 Invalid request body received, query type could not be determined
```

### DIFF
--- a/packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
+++ b/packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
@@ -1437,7 +1437,7 @@ describe('HttpServiceSparqlEndpoint', () => {
         );
 
         await expect(endCalledPromise).resolves.toBe(
-          'The response for the given query could not be serialized for the requested media type\n',
+          'The response for the given query could not be serialized for the requested media type',
         );
         expect(response.writeHead).toHaveBeenLastCalledWith(
           400,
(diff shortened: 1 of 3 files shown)


rubensworks (changes requested):
Thanks, looks great!

I like the new `writeBadRequest` method.
However, there are a couple of existing places that emit a `[400]`; could you modify those to also make use of this new method?

redreye:
I updated the existing 400 responses to use the new `writeBadRequest` method as well.

packages/actor-init-query/test/HttpServiceSparqlEndpoint-test.ts
```diff
@@ -1398,7 +1437,7 @@ describe('HttpServiceSparqlEndpoint', () => {
         );
 
         await expect(endCalledPromise).resolves.toBe(
-          'The response for the given query could not be serialized for the requested media type\n',
```
rubensworks, line 1401: Pretty sure the terminating newline is intentional. Can we keep it for 400 responses?

redreye: Sure.

rubensworks:
Thanks @redreye!

PR #1785, opened 2026-09-02
Fix the SPARQL endpoint hanging on updates with a WHERE clause

redreye:
An update with a WHERE clause sent to the endpoint never receives a response. The status line and headers are written, but the body never ends, so the client waits until it gives up.

```js
const { RdfStore } = require('rdf-stores');

(async() => {
  const store = RdfStore.createDefault(true);

  const service = new HttpServiceSparqlEndpoint({
    engine,
    port: 3000,
    context: { sources: [{ type: 'rdfjs', value: store }], destination: store },
  });

    method: 'POST',
    headers: { 'content-type': 'application/sparql-update' },
    body: 'INSERT { GRAPH <http://example.org/g> { ?s a <http://example.org/T> } } WHERE { ?s ?p ?o }',
  });
})();
```

The execution is now memoized, so awaiting it and serializing it share a single run.

The two `process.send!` calls in the same method are also guarded. Outside a cluster worker `process.send` is `undefined`, so the snippet above throws `TypeError: process.send is not a function` before it reaches the engine, which makes the endpoint class unusable when it is embedded rather than started through `comunica-sparql-http`.

### DIFF
--- a/packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
+++ b/packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
@@ -491,7 +491,7 @@ export class HttpServiceSparqlEndpoint {
     stdout.write(`      Received ${queryBody.type} query: ${queryBody.value}\n`);
 
     // Send message to master process to indicate the start of an execution
-    process.send!({ type: 'start', queryId });
+    process.send?.({ type: 'start', queryId });
 
     // Determine context
     let context = {
@@ -506,9 +506,12 @@ export class HttpServiceSparqlEndpoint {
     try {
       result = await engine.query(queryBody.value, context);
 
-      // For update queries, also await the result
+      // For update queries, also await the result, so that failures are reported as a bad request.
+      // The execution is memoized, as updates may not be executed again while serializing the result.
       if (result.resultType === 'void') {
-        await result.execute();
+        const executed = result.execute();
+        await executed;
+        result = { ...result, execute: () => executed };
       }
     } catch (error: unknown) {
       stdout.write('[400] Bad request\n');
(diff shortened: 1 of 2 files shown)


packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -506,9 +506,12 @@ export class HttpServiceSparqlEndpoint {
...
       result = await engine.query(queryBody.value, context);
 
-      // For update queries, also await the result
+      // For update queries, also await the result, so that failures are reported as a bad request.
+      // The execution is memoized, as updates may not be executed again while serializing the result.
       if (result.resultType === 'void') {
-        await result.execute();
+        const executed = result.execute();
+        await executed;
+        result = { ...result, execute: () => executed };
```
rubensworks, line 514: I'm not sure I understand this change.
What problem does it fix?
Is it needed?
I don't see a corresponding unit test for it.

rubensworks (changes requested):
Thanks! Just one question.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
redreye: `writeQueryResult` awaits `result.execute()` so a failing update comes back as a 400, and then hands the same result to `resultToString`, whose simple serializer calls `execute()` a second time. For a `deleteinsert` that execution comes from the update-quads mediator and consumes the insert/delete quad streams, so the second call never settles: headers go out, the body never ends, and the client waits until it gives up. A `compositeupdate` did respond, but only because its `execute` replays every sub-operation so those were applied twice per request.

I've replaced the unit test with one that shows this: the second execution never settles and the serializer awaits it, so without the memoization the test times out instead of ending the response.

rubensworks (approved):
I understand now, thanks @redreye!

PR #1786, opened 2026-09-03
Resolve relative IRIs in endpoint requests against the endpoint

redreye:
Any request to the endpoint that contains a relative IRI is rejected, because the service does not define a base IRI to resolve them against.

```js
const { RdfStore } = require('rdf-stores');

(async() => {
  const store = RdfStore.createDefault(true);

  const service = new HttpServiceSparqlEndpoint({
    engine,
    port: 3000,
    context: { sources: [{ type: 'rdfjs', value: store }], destination: store },
  });

    method: 'POST',
    headers: { 'content-type': 'application/sparql-update' },
    body: 'INSERT DATA { <s> <p> <o> }',
  });
})();
```

Before:

```
400 Cannot resolve relative IRI s because no base IRI was set.
```

After:

```
200 ok
```

with `<s>` stored as `http://127.0.0.1:3000/s`.

### DIFF
--- a/packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
+++ b/packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
@@ -7,6 +7,7 @@ import type { IncomingMessage, ServerResponse } from 'node:http';
 import * as path from 'node:path';
 import * as querystring from 'node:querystring';
 import type { Writable } from 'node:stream';
+import type { TLSSocket } from 'node:tls';
 import * as url from 'node:url';
 import { KeysInitQuery, KeysQueryOperation } from '@comunica/context-entries';
 import { ActionContext } from '@comunica/core';
@@ -474,7 +475,8 @@ export class HttpServiceSparqlEndpoint {
    * @return {string} The base IRI.
    */
   public static getBaseIRI(request: http.IncomingMessage, port: number): string {
-    return `http://${request.headers.host ?? `localhost:${port}`}/sparql`;
+    const protocol = (<TLSSocket> request.socket).encrypted ? 'https' : 'http';
+    return `${protocol}://${request.headers.host ?? `localhost:${port}`}/sparql`;
   }
 
   /**
--- a/packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
+++ b/packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
@@ -468,6 +468,16 @@ export class HttpServiceSparqlEndpoint {
   }
 
   /**
+   * Determine the base IRI that relative IRIs in a request are resolved against.
+   * @param {module:http.IncomingMessage} request Request object.
+   * @param {number} port The port this service is running on.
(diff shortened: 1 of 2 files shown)


packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -467,6 +467,16 @@ export class HttpServiceSparqlEndpoint {
...
   }
 
+  /**
+   * Determine the base IRI that relative IRIs in a request are resolved against.
+   * @param {module:http.IncomingMessage} request Request object.
+   * @param {number} port The port this service is running on.
+   * @return {string} The base IRI.
+   */
+  public static getBaseIRI(request: http.IncomingMessage, port: number): string {
+    return `http://${request.headers.host ?? `localhost:${port}`}/sparql`;
```
rubensworks: The server is not always necessarily running on `http` though, so we'd also need to check the protocol.

But even then, there might be cases where the server is behind a proxy.
So for those cases, I think we should also add a CLI option to allow the baseIRI to be defined by the user.

rubensworks (changes requested):
Cool!
Just one suggestion.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
redreye: Good points.

Scheme: now taken from `request.socket.encrypted` rather than hardcoded.

Proxy: `--baseIRI` already exists in `CliArgsHandlerBase`, and `generateConstructorArguments` includes it for the HTTP endpoint, so `comunica-sparql-http --baseIRI https://example.org/sparql` already works.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -467,6 +468,17 @@ export class HttpServiceSparqlEndpoint {
...
   }
 
+  /**
+   * Determine the base IRI that relative IRIs in a request are resolved against.
+   * @param {module:http.IncomingMessage} request Request object.
+   * @param {number} port The port this service is running on.
+   * @return {string} The base IRI.
+   */
+  public static getBaseIRI(request: http.IncomingMessage, port: number): string {
+    const protocol = (<TLSSocket> request.socket).encrypted ? 'https' : 'http';
```
rubensworks, line 478: Doesn't `request.protocol` exist?

redreye: It doesn't.

rubensworks:
Great, thanks @redreye!

(side note: thanks for all of these PRs! I was curious to see what things you are working on, but your profile is private. If you'd like to satisfy my curiosity in sharing how you are making use of Comunica, feel free to reach out to me in private.:-) )

redreye:
Thanks @rubensworks! I’m currently evaluating Comunica for a future project, which has also led me to fix a few bugs along the way.

Issue #1266, opened 2023-09-13
Use bignumber.js in internal Decimal Literal


jitsedesmet:
The expression evaluator already uses `bignumber.js` to perform evaluations. However, this is not enough, an [XSD decimal](https://www.w3.org/TR/xmlschema-2/#decimal) is unbounded, that means that after the calculations, we need to keep the value as a `bignumber`.

This means the `DecimalLiteral` typed value should be a `bignumber` in order to be spec compliant.

This is originally mentioned in an issue of the sparqlee repo: https://github.com/comunica/sparqlee/issues/45.

PR #1787, opened 2026-09-03
Fixes to `xsd:string` function and numeric literal representations

surilindur:
Any feedback would be welcome on this. :slightly_smiling_face: The current version is the minimal, least invasive and most readable way to address the interconnected issues of
1. `xsd:string(?numeric_literal)` not following the [XPath string cast specification](https://www.w3.org/TR/xpath-functions/#casting-to-string), and
2. the string values of numeric literals, when derived from the numeric value in absence of lexical, not always following the [XSD datatype](https://www.w3.org/TR/xmlschema-2/#built-in-primitive-datatypes) lexical forms (for example, decimals could end up represented as exponentials).

The cleanest solution to this that I could find was to
1. change the numeric literal string formatters, that are used as fallback when lexical value is not defined, to always output canonical representations following the XSD datatype spec, and
2. update the `xsd:string` implementation so that it considers each datatype separately (previously, all were treated as floats), and also takes into consideration the overriding rules from the XPath cast specification.

I also expanded the test coverage for the different numeric literals, because there were come edge cases that were not considered, such as double/float values around `1e6` and `1e-6`, and large or small decimal values.

I had to update some other tests, as well, because they contained hardcoded numeric literal values in their expectations.

The only downside that I can see, is that the canonical representations can be less readable, but no human is likely to read them anyway, I guess. :grimacing: 

Here is an example query that can be run on the current web version of Comunica to observe the issues with the current implementation:

```sparql
SELECT
  ?number
  (STR(?number) AS ?number_str)
  (xsd:string(?number) AS ?number_cast)
WHERE {
  VALUES ?number {
    "0.0000001"^^xsd:decimal # cast becomes exponential
    "0.0000001"^^xsd:double # cast format does not follow spec
    "0.0000001"^^xsd:float # cast format does not follow spec
    "0.000001"^^xsd:decimal
    "0.000001"^^xsd:double
    "0.000001"^^xsd:float
    "100000"^^xsd:decimal
    "100000"^^xsd:double
    "100000"^^xsd:float
    "1000000"^^xsd:decimal
    "1000000"^^xsd:double # cast remains decimal when should be exponential
    "1000000"^^xsd:float # cast remains decimal when should be exponential
    "INF"^^xsd:double # cast format does not follow spec
    "INF"^^xsd:float # cast format does not follow spec
  }
}
```


rubensworks:
Thanks for the PR @surilindur!

@jitsedesmet You probably know more about this part of the spec(s) than me, could you have a first look at this?

packages/actor-function-factory-term-xsd-to-string/test/op.xsd-to-string-test.ts
```diff
@@ -15,19 +15,71 @@ describe('to string', () => {
...
         <http://example.org/z> = "http://example.org/z"
-        "true"^^xsd:boolean = "true"
-        "false"^^xsd:boolean = "false"
+
         "1"^^xsd:boolean = "true"
         "0"^^xsd:boolean = "false"
-        "1"^^xsd:integer = "1"
-        "0.0"^^xsd:decimal = "0"
+        "true"^^xsd:boolean = "true"
+        "false"^^xsd:boolean = "false"
```
jitsedesmet, line 20: Some tests were lost here? like 1 xsd:integer?

packages/actor-function-factory-term-xsd-to-string/test/op.xsd-to-string-test.ts
```diff
@@ -15,19 +15,71 @@ describe('to string', () => {
...
-        "1.25"^^xsd:float = "1.25"
-        "2.5"^^xsd:decimal = "2.5"
-        "-2.5"^^xsd:decimal = "-2.5"
+        "+0.0"^^xsd:decimal = "0"
+        "-0.0"^^xsd:decimal = "0"
+        "+1000000.0"^^xsd:decimal = "1000000"
+        "-1000000.0"^^xsd:decimal = "-1000000"
+        "+1000000.1"^^xsd:decimal = "1000000.1"
+        "-1000000.1"^^xsd:decimal = "-1000000.1"
+
```
jitsedesmet, line 48: Here again, we also seem to have lost some tests? Is something provably wrong with the input values we had at first? (Adding tests is still okay ofcourse)

packages/utils-expression-evaluator/test/unit/expressions/Term-test.ts
```diff
@@ -29,35 +29,96 @@ describe('Term', () => {
 
   describe('the string representation of numeric literals', () => {
     describe('like integers', () => {
-      it('should not have a leading +', () => {
-        const num = new IntegerLiteral(1);
-        expect(num.toRDF(DF).value).toBe('1');
+      it('should properly express zero', () => {
+        const num = new DecimalLiteral(0);
+        expect(num.toRDF(DF).value).toBe('0');
```
jitsedesmet, line 34: Same here, we lose tests.

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
```diff
@@ -20,7 +23,35 @@ export class TermFunctionXsdToString extends TermFunctionBase {
...
+          }
+          if (val.dataType === TypeURL.XSD_DECIMAL) {
+            return string(decimal(val.typedValue).str());
+          }
+          if (val.dataType === TypeURL.XSD_DOUBLE || val.dataType === TypeURL.XSD_FLOAT) {
+            // Exact zero should always be returned as "0" which differs from the
+            // XSD datatype canonical representation
+            if (val.typedValue === 0) {
+              return string('0');
+            }
```
jitsedesmet: Ca you link the spec here?

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
```diff
@@ -20,7 +23,35 @@ export class TermFunctionXsdToString extends TermFunctionBase {
       arity: 1,
       operator: TypeURL.XSD_STRING,
       overloads: declare(TypeURL.XSD_STRING)
-        .onNumeric1(() => (val: NumericLiteral) => string(float(val.typedValue).str()))
+        .onNumeric1(() => (val: NumericLiteral) => {
```
jitsedesmet, line 30:
```suggestion
        // the spec for numerics is really specific: https://www.w3.org/TR/xpath-functions/#casting-to-string . Also note that we need to go through .onNumeric1 since the spec does NOT talk about type derivations SPECIFICALLY when checking the types in the numerc check, e.g. for `xs:string` they do consider 'or a type derived from `xs:string`)
        .onNumeric1(() => (val: NumericLiteral) => {
```

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
```diff
@@ -20,7 +23,35 @@ export class TermFunctionXsdToString extends TermFunctionBase {
...
+          if (val.dataType === TypeURL.XSD_INTEGER) {
+            return string(integer(val.typedValue).str());
+          }
+          if (val.dataType === TypeURL.XSD_DECIMAL) {
+            return string(decimal(val.typedValue).str());
+          }
```
jitsedesmet, line 43: Please add some comments on why this covers the conditional defiend in the spec:
> If SV is in the value space of xs:integer, that is, if there are no significant digits after the decimal point, then the value is converted from an xs:decimal to an xs:integer and the resulting xs:integer is converted to an xs:string using the rule above.

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
jitsedesmet: Did we already migrate the evaluator to bigNumbers.js? (or something like that?)

If we did, do we have a way of handling positive and negative 0 and infty?

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
```diff
@@ -20,7 +23,35 @@ export class TermFunctionXsdToString extends TermFunctionBase {
...
+            // should be converted to decimal before casting to string
+            if (
+              (val.typedValue > -1e6 && val.typedValue <= -1e-6) ||
+              (val.typedValue >= 1e-6 && val.typedValue < 1e6)
+            ) {
+              return string(decimal(val.typedValue).str());
+            }
+            if (val.dataType === TypeURL.XSD_DOUBLE) {
+              return string(double(val.typedValue).str());
+            }
```
jitsedesmet, line 72: I wonder whether it is this simple, given the fact that the spec about this section is quite lengthy and does in fact _not_ refer to the canonical representation of float, decimal, or double.

jitsedesmet (changes requested):
I will read the change in canonical representation soon. I already have some questions though.

packages/actor-function-factory-term-xsd-to-string/test/op.xsd-to-string-test.ts
surilindur: I rearranged them by datatype, so the 1 xsd:integer is now grouped with the other xsd:integer tests. The exact value `1` is not tested anymore, though, actually, which is a good point (explained in a comment below).

packages/actor-function-factory-term-xsd-to-string/test/op.xsd-to-string-test.ts
surilindur: Good catch, thank you! I completely missed the small decimal values here. They are tested in the term tests, but should also still be included here. There was nothing wrong with the values themselves, I just replaced them all to make them uniform and maybe more readable (especially the values for 1 and 0, except 0 without a sign is missing).

I will change the test cases to also include the original inputs here.

packages/utils-expression-evaluator/test/unit/expressions/Term-test.ts
surilindur: The specific test, yes, but the thing the test tests is still there (no leading + in positive numbers, which is tested by the `1234` test). I can add back a test for the value 1 specifically, though.

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
surilindur: I need to check this, but as far as I can tell, the numbers are just normal JS numbers. There exists a function to check the sign of a number, though, but when I tried it, I could not make it work. I will have another look at it.

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
surilindur: The spec does not directly say that it is the canonical representation of a double/float, but the requirements it lays out for the format are actually the same as in the canonical double/float spec. :sweat_smile:

Specifically, the cast requires a format with the following:
* Mantissa: lexical decimal, but no leading +, must have decimal point, must have exactly one non-zero digit to the left, and at least one digit to the right (same requirements as in canonical double/float, but not in lexical double/float)
* Exponent marked by uppercase `E` (required in canonical, not in lexical)
* Exponent: lexical integer, but no leading + or leading zeroes (same requirements as in canonical double/float, but not in lexical double/float)

Only the double is returned here on this line (as its canonical representation), and the case for equivalent float is covered by the fallback value (canonical float representation) at the end of the function, which will be the next statement after this, if the function does not return.

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
jitsedesmet: Found the issue. It is still open :) So you can ignore this comment:
https://github.com/comunica/comunica/issues/1266

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
jitsedesmet: Aha, I see. Okay. I will now look to the spec on canonical representation. I do want to point out that ineed the float case is handled in the fallback, but then I would remove the comment 

>  // The original code used float for everything,
          // so keeping float as fallback makes sense.
          
Simply because it is not only a fallback but only serves as the last case of our weird spec definition on how to handle numeric

packages/utils-expression-evaluator/lib/expressions/Term.ts
```diff
@@ -206,23 +206,20 @@ export class DecimalLiteral extends NumericLiteral {
     super(typedValue, dataType ?? TypeURL.XSD_DECIMAL, strValue, language);
   }
 
+  /**
+   * Fallback formatter to produce canonical representation when no lexical value exists.
+   * This follows the specification for decimal: https://www.w3.org/TR/xmlschema-2/#double
```
jitsedesmet:
```suggestion
   * This follows the specification for decimal: https://www.w3.org/TR/xmlschema-2/#decimal:
   > Specifically, the preceding optional "+" sign is prohibited. The decimal point is required. Leading and trailing zeroes are prohibited subject to the following: there must be at least one digit to the right and to the left of the decimal point which may be a zero.
```

packages/utils-expression-evaluator/lib/expressions/Term.ts
```diff
@@ -206,23 +206,20 @@ export class DecimalLiteral extends NumericLiteral {
...
-    dataType?: string,
-    public override strValue?: string,
-    public override language?: string,
-  ) {
-    super(typedValue, dataType ?? TypeURL.XSD_FLOAT, strValue, language);
-  }
+    // Only return the JS representation if it consists of decimal digits and a decimal point
+    if (/^-?[.0-9]+$/u.test(jsString)) {
+      return jsString;
+    }
```
jitsedesmet: This is not correct right? We need to have a `.`? for decimal

packages/utils-expression-evaluator/lib/expressions/Term.ts
```diff
@@ -206,23 +206,20 @@ export class DecimalLiteral extends NumericLiteral {
...
 
-  protected specificFormatter(val: number): string {
-    return val.toString();
+    // Use the longest possible decimal format with trailing zeroes removed as fallback
+    return val.toFixed(20).replace(/([0-9])0*$/u, '$1');
```
jitsedesmet: I do not understand how this works 😅 

Maybe I confused myself, but I feel like we should check the `jsString` for a dot, and if it does not have one, add `+".0"`.

packages/utils-expression-evaluator/lib/expressions/Term.ts
```diff
@@ -236,30 +233,45 @@ export class DoubleLiteral extends NumericLiteral {
...
-    // SPARQL test suite prefers trailing zero's
-    const mantisse = jsMantisse.includes('.') ?
-      jsMantisse :
-      `${jsMantisse}.0`;
+    return 'NaN';
+  }
+}
 
-    return `${mantisse}E${exponent}`;
+export class FloatLiteral extends DoubleLiteral {
```
jitsedesmet, line 280:
```suggestion
/**
 * Every Float (32bit) is a Double (64bit). Canonical representation is created in the same way.
*/
export class FloatLiteral extends DoubleLiteral {
```

packages/utils-expression-evaluator/lib/expressions/Term.ts
```diff
@@ -236,30 +233,45 @@ export class DoubleLiteral extends NumericLiteral {
...
   protected specificFormatter(val: number): string {
-    if (!Number.isFinite(val)) {
-      if (val > 0) {
-        return 'INF';
-      }
-      if (val < 0) {
-        return '-INF';
-      }
-      return 'NaN';
+    if (Number.isFinite(val)) {
```
jitsedesmet: Number.isFinite. We should add a comment to our string constructor handling of `0` that pos and neg infty are handled by the canonical representations themself.

packages/utils-expression-evaluator/lib/expressions/Term.ts
```diff
@@ -236,30 +233,45 @@ export class DoubleLiteral extends NumericLiteral {
...
+      // This produces the JS version in form "<mantissa>e(+|-)<exponent>"
+      const [ jsMantissa, jsExponent ] = val.toExponential().split('e');
+
+      // Optional "+" is prohibited in the exponent
+      const exponent = jsExponent.replace(/^\+/u, '');
```
jitsedesmet: Quick question, will JS ever generate the `+`?

packages/utils-expression-evaluator/lib/expressions/Term.ts
```diff
@@ -236,30 +233,45 @@ export class DoubleLiteral extends NumericLiteral {
...
-      if (val > 0) {
-        return 'INF';
-      }
-      if (val < 0) {
-        return '-INF';
-      }
-      return 'NaN';
+    if (Number.isFinite(val)) {
+      // This produces the JS version in form "<mantissa>e(+|-)<exponent>"
+      const [ jsMantissa, jsExponent ] = val.toExponential().split('e');
```
jitsedesmet: Nit pick , but can you add that this function covers the requirement:

> that there is a single digit which is non-zero to the left of the decimal point and at least a single digit to the right of the decimal

packages/actor-function-factory-term-xsd-to-string/test/op.xsd-to-string-test.ts
surilindur: This should now be fixed, I added the missing cases back.

packages/actor-function-factory-term-xsd-to-string/test/op.xsd-to-string-test.ts
surilindur: This is also fixed now!

jitsedesmet (changes requested):
Okay. Nice work! The reaseon for the nit picks is because I have frustrations with expression evaluator work and with everything spec related that requires you to go 2 specs down. As such, good documentation is key such that no successor has to go through this again xp 

Thank you for your service ;)

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: The `.` is included in the regex, in the brackets next to `0-9`. It is definitely not ideal, because it also matches `0.1.2.3.4.5.` and equivalent, but I am *assuming* that `val.toString()` produces either a valid decimal or something that should be ignored as a decimal. :grin:

I will change it to specifically enforce `^-?[0-9]\.[0-9]+$` so it is not ambiguous.

Edit: with forbidden trailing zeroes.

jitsedesmet: No, my issue is that the regex above also accepts a pattern that does not contain a dot? like `5`

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: It also took a while for me to figure this out. xd

Basically, `.toFixed()` is going to produce a decimal with a dot, and *always* use 20 decimal places for it. It will completely mangle `33.33` to something unreadable with 20 forced decimal places, but that is why this is the fallback *after* trying `.toString()` which *will* make `33.33` into `"33.33"`.

The regex will remove all trailing zeroes only, and will do it in the decimal part (because of the `$` anchor at the end). It will start (conceptually, not necessarily actually) from the end of the string, and find all (zero or more) consecutive `0`, moving left until it finds either the first non-zero number `[1-9]` *or* the zero immediately after the decimal place `[0]`. It then replaces the match that consists of `<first non-zero or the zero after decimal><trailing zeroes if any>$` with `<first non-zero or the zero after decimal>$`.

For example, given `0.0000001`, JS will try to convert it into `1e-7` which is not a decimal, and is rejected by the regex. The `.toFixed(20)` call will turn it into `0.00000010000000000000` with trailing zeroes. From this, the regex will match `10000000000000` (where `1` is for `[0-9]`, and the rest is for `0*`), and replace it with the match of `[0-9]`, which is `1`, doing essentially:
1. `0.00000010000000000000`
2. `0.000000` + `1` + `0000000000000`
3. `0.000000` + `1`
4. `0.0000001`

It is not the most amazing way to do it, but it is a one-liner. :disappointed:

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: Oh, I actually have a question about this! The reason I changed the inheritance was to avoid duplicating the `specificFormatter` code. Do you think it will have significant consequences to the handling of the *types* themselves in the engine, wrt. class inheritance, and should I instead just duplicate the formatter code? :thinking:

JavaScript handles everything as 64-bit doubles, so in that sense it should not matter, as in the suggested comment.

Edit: And the canonical representations are identical. So I am mostly asking about the JS classes and inheritance.

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: That is what I was also wondering when I saw the original code, because it feels very redundant, but it actually does. For example, `1.0.toExponential()` produces `"1e+0"`.

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: Oh, that is definitely true. I need to fix it, for sure. Thank you!

surilindur: Actually, this was a very good observation, because the canonical representation *requires* a decimal point. I will update the `xsd:string` function to handle whole number decimals separately, then, because they cannot be handled by the canonical decimal representation, after all. Amazing. Thank you again. :slightly_smiling_face:

packages/utils-expression-evaluator/lib/expressions/Term.ts
jitsedesmet: I do not think class inheritance is or should be used anywhere. These classes provide us with utilities that ease the handling of what actually matters to the engine, namely, the datatype.

packages/utils-expression-evaluator/lib/expressions/Term.ts
jitsedesmet: Okay. Crazy. thanks

packages/utils-expression-evaluator/lib/expressions/Term.ts
jitsedesmet: Ah yes: 

> If SV is in the value space of xs:integer, that is, if there are no significant digits after the decimal point, then the value is converted from an xs:decimal to an xs:integer and the resulting xs:integer is converted to an xs:string using the rule above.

Feel free to just cite these in the implementation. :)

packages/utils-expression-evaluator/lib/expressions/Term.ts
jitsedesmet: Ah, I see. this would make sense. However, since decimal, according to the spec, is infinite length. This would not work. (I know it would now because we use Number), but still. I'd change it to something more stable and understandable to be sure.

jitsedesmet: Also, we know the regex failed. Given that it failed on a string we got from number.toString, I think we know the problem would be the lack of `.0`. I would just replace the return with `jsString + '.0'`

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
surilindur: This should now be clarified in the comment on the final return statement!

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: This is fixed now, I also added links to the exact spec sections for the classes, and a comment to clarify what the regex tests.

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: There is a comment that explains it now in `xsd:string`.

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: There is a comment now for this, above the line where the mantissa is modified if necessary.

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: > we know the problem would be the lack of `.0`

The problem could actually be one of:
1. The number being integer, and JS dropping `.0` from behind it. This would be easy to fix as you described, and is already handled by the regex replacement: `toFixed` forces the trailing zeroes, and all except one are then removed.
2. The number being so small that it ends up represented as exponential. For example, `0.0000001.toString()` produces an exponential, which is not a valid decimal representation (as far as I can tell, based on the spec). Using `.toFixed` avoids this, by forcing it into a decimal-like representation instead of an exponent.

So, the only way to force a decimal format is to use `toFixed`, as far as I can tell (currently, anyway, maybe with the different number library it can be fixed). And even then, using `.toFixed` requires a decimal slot argument, because without one, `0.0000001.toFixed()` produces `0` which is also not good. I would have imagined it to be smart, somehow, but it is apparently not. 😞

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: This is no longer needed, since the formatter handles this correctly. I have removed all regexes and stuff. :partying_face:

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: This fixed with the number formatter, and I have removed all regexes.

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: I have moved the description of canonical form to the class docstring, since the formatter handles all of this now correctly.

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
```diff
@@ -20,7 +24,57 @@ export class TermFunctionXsdToString extends TermFunctionBase {
       arity: 1,
       operator: TypeURL.XSD_STRING,
       overloads: declare(TypeURL.XSD_STRING)
-        .onNumeric1(() => (val: NumericLiteral) => string(float(val.typedValue).str()))
+        // We need to go through .onNumeric1 since the spec does NOT talk about type derivations
+        // SPECIFICALLY when checking the types in the numeric check,
+        // e.g. for `xs:string` they do consider "or a type derived from `xs:string`".
```
jitsedesmet, line 29:
```suggestion
        // We need to go through .onNumeric1 since the spec () does NOT talk about type derivations
        // SPECIFICALLY when checking the types in the numeric check,
        // e.g. for `xs:string` they do consider "or a type derived from `xs:string`".
```

packages/utils-expression-evaluator/lib/expressions/Term.ts
```diff
@@ -161,8 +163,6 @@ export abstract class NumericLiteral extends Literal<number> {
     super(typedValue, dataType, strValue, language);
   }
 
-  protected abstract specificFormatter(val: number): string;
-
```
jitsedesmet, line 165: @rubensworks Strictly spoken this is a breaking change right?

packages/utils-expression-evaluator/lib/expressions/Term.ts
```diff
@@ -176,12 +176,31 @@ export abstract class NumericLiteral extends Literal<number> {
   }
 
   public override str(): string {
-    return this.strValue ??
-      this.specificFormatter(this.typedValue);
+    if (this.strValue) {
+      return this.strValue;
+    }
+    let representation = (<typeof NumericLiteral> this.constructor).formatter.format(this.typedValue);
```
jitsedesmet: I am not a huge fan of this `this.constructor` notation. I wonder, is it possible to use the abstract function/ abstract field syntax? if yes, I'd use that instead.

jitsedesmet: Thinking about it, that might also be breaking? Because if it is an abstract field, implementers that were valid are now invalid.

jitsedesmet: Would the issue be fixed if we would simply not make this static?

jitsedesmet: and keep `specificFormatter` as a base implementation that does the infty check for us

jitsedesmet (changes requested):
Clean!

packages/utils-expression-evaluator/lib/expressions/Term.ts
rubensworks: Yes indeed. Since we don't know who is using this for what purposes, best to not break in a non-major release.

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: > Would the issue be fixed if we would simply not make this static?

Yes, but then the formatter would be constructed for every instance of the class, unless it lives somewhere in a global scope, or is in a context, which could get messy. :thinking: 

> and keep `specificFormatter` as a base implementation that does the infty check for us

I think I will do that: restore the `specificFormatter` and make it use a static formatter directly on the specific literal class. Then it will not be a breaking change anymore.

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
surilindur: Do the braces mean something specific here? I am trying to figure out what the suggested change does. :sweat_smile:

jitsedesmet: Oh, sorry, I wanted to do something but removed it I thought, please ignore.

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: The breaking change is reverted now, and the formatter is used inside the `specificFormatter` method of each class.

jitsedesmet (approved):

packages/utils-expression-evaluator/lib/expressions/Term.ts
```diff
@@ -181,7 +181,20 @@ export abstract class NumericLiteral extends Literal<number> {
...
 }
 
+/**
+ * Integer datatype from XSD: https://www.w3.org/TR/xmlschema-2/#integer
+ *
+ * The canonical representation consists of a finite-length sequence of decimal digits (#x30-#x39),
+ * with leading + and leading zeroes prohibited.
+ */
 export class IntegerLiteral extends NumericLiteral {
+  protected static readonly formatter = Intl.NumberFormat(undefined, {
```
rubensworks: ` Intl.NumberFormat` is locale-dependent.
You're passing `undefined`, which resolves to the system default.
So using this on different machines can give different behaviour.

For example, `LC_ALL=de_DE.UTF-8`, `DecimalLiteral(1.5)` serializes as `"1,5"` and `DoubleLiteral(1100) `as `"1,1E3"`.

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
```diff
@@ -20,7 +24,57 @@ export class TermFunctionXsdToString extends TermFunctionBase {
...
+        .onNumeric1(() => (val: NumericLiteral) => {
+          if (
+            val.dataType === TypeURL.XSD_INTEGER ||
```
rubensworks: What should happen with `xsd:nonNegativeInteger` and such?
Right now, it will fall back to float.
Is that what we want here?
(asking because I don't know this part of the spec that well)

packages/utils-expression-evaluator/lib/expressions/Term.ts
```diff
@@ -191,12 +204,27 @@ export class IntegerLiteral extends NumericLiteral {
     super(typedValue, dataType ?? TypeURL.XSD_INTEGER, strValue, language);
   }
 
-  protected specificFormatter(val: number): string {
-    return val.toFixed(0);
+  protected override specificFormatter(val: number): string {
+    return IntegerLiteral.formatter.format(val);
```
rubensworks: Ok, so this is where the SPARQL spec is not accurate on, namely which value-to-lexical mapping should be used; the one from XSD (what you chose in this PR), or the one from F&O ([the one we use in master(https://github.com/comunica/comunica/blob/95f2d4907a5e503cbb5a7e402024966caaa3628c/packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts#L14-L16)).

Since both are valid according to the spec, this PR does not really "fix" a problem.
That being said, I don't mind us adding support for it, but I would not replace our current behaviour (which still is spec-compliant).
I also think the XSD interpretation probably is a bit less efficient.

As such, I would propose adding a mode (context option) to choose between the two modes, which should default to F&O.

Since I don't know these specs that well, I first would like @jitsedesmet's input on this as well.

rubensworks (changes requested):

packages/utils-expression-evaluator/lib/expressions/Term.ts
jitsedesmet: Ahh yes. So it seems that if you look at SPARQL 1.1 and then read the text (dont click the links!) you get to:
https://www.w3.org/TR/2007/REC-xpath-functions-20070123/#casting-from-primitive-to-primitive

jitsedesmet: Aha! luckily for us, that is still the same :)

```
If ST is a numeric type, the following rules apply:

If ST is xs:integer, TV is the canonical lexical representation of SV as defined in [[XML Schema Part 2: Datatypes Second Edition]](https://www.w3.org/TR/2007/REC-xpath-functions-20070123/#xmlschema-2). There is no decimal point.

If ST is xs:decimal, then:

If SV is in the value space of xs:integer, that is, if there are no significant digits after the decimal point, then the value is converted from an xs:decimal to an xs:integer and the resulting xs:integer is converted to an xs:string using the rule above.

Otherwise, the canonical lexical representation of SV is returned, as defined in [[XML Schema Part 2: Datatypes Second Edition]](https://www.w3.org/TR/2007/REC-xpath-functions-20070123/#xmlschema-2).

If ST is xs:float or xs:double, then:

TV will be an xs:string in the lexical space of xs:double or xs:float that when converted to an xs:double or xs:float under the rules of [17.1.1 Casting from xs:string and xs:untypedAtomic](https://www.w3.org/TR/2007/REC-xpath-functions-20070123/#casting-from-strings) produces a value that is equal to SV, or is "NaN" if SV is NaN. In addition, TV must satisfy the constraints in the following sub-bullets.

If SV has an absolute value that is greater than or equal to 0.000001 (one millionth) and less than 1000000 (one million), then the value is converted to an xs:decimal and the resulting xs:decimal is converted to an xs:string according to the rules above, as though using an implementation of xs:decimal that imposes no limits on the totalDigits or fractionDigits facets.

If SV has the value positive or negative zero, TV is "0" or "-0" respectively.

If SV is positive or negative infinity, TV is the string "INF" or "-INF" respectively.

In other cases, the result consists of a mantissa, which has the lexical form of an xs:decimal, followed by the letter "E", followed by an exponent which has the lexical form of an xs:integer. Leading zeroes and "+" signs are prohibited in the exponent. For the mantissa, there must be a decimal point, and there must be exactly one digit before the decimal point, which must be non-zero. The "+" sign is prohibited. There must be at least one digit after the decimal point. Apart from this mandatory digit, trailing zero digits are prohibited.

Note:

The above rules allow more than one representation of the same value. For example, the xs:float value whose exact decimal representation is 1.26743223E15 might be represented by any of the strings "1.26743223E15", "1.26743222E15" or "1.26743224E15" (inter alia). It is implementation-dependent which of these representations is chosen.
```

jitsedesmet: https://www.w3.org/TR/xmlschema-2/#decimal :)

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
jitsedesmet: I also Do not like this and I think handling through the overloadTree would be wayyyy beter. The spec actually seems to not decide on these types, so maybe we can do what we please and indeed interpret them as type derivations would allow us to fo.

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: That is a good point. I tried to see if there is a `C` locale, but Node claims it is invalid. If there is no C locale or equivalent, I will just drop the internationalization approach.

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
surilindur: Is the overload tree something Comunica/Traqula-specific? I can see if there is an example somewhere of an already existing implementation.

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: > the one from XSD (what you chose in this PR), or the one from F&O

The XPath cast refers to the XSD spec for integer representation, so I thought they would be the same? :confused:

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
jitsedesmet: the builder you use now constructs an `overloadTree` So you can simply use things like `onDecimal` I assume. If that does not exists, use something lie `.set([type, type])`. All operations are constructed like that

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: To address this point:

> I also think the XSD interpretation probably is a bit less efficient.

I ran a small benchmark script to measure the conversions - the `str()` method only - between the current master branch and this PR here, for 10 million different values (positive and negative), with a replication of 10 per value. The numbers in the table are average per-literal conversion times.

| Type | master (ms) | canonical (ms) | change (%) |
| -- | --:| --:| --: |
| `xsd:integer` | 0.00012 | 0.00012 | 0 |
| `xsd:decimal` | 0.00011 | 0.00014 | +33 |
| `xsd:double` | 0.00039 | 0.00038 | -2 |
| `xsd:float` | 0.00011 | 0.00039 | +242 |

There is a difference in the time it takes to produce the string representation per-literal, but it is not an order of magnitude, and I would think it is not significant enough to introduce a context flag just to disable canonical formats? The only exception is `xsd:float`, because the current implementation of float in master only calls `.toString()` without any additional checks, which happens to produce the lexical value for finite numbers, but not for infinite. The PR changes it to follow the double implementation, including fix-ups for infinite value (`Infinity` -> `INF`).

Also, the decimals before this PR do not even follow the lexical decimal format for small values (produced as exponentials), so the actual output was invalid.

So, assuming:
1. the XPath and XSD representations are identical (which they are, except for the few exceptions handled separately), and
2. the performance impact is sufficiently small,

I would propose the canonical (and/or just fixed) formats in the PR to be the default? That way, the output formats would be more constrained/well-defined, and easier to reason around. The SPARQL spec may be more flexible, but since the implementation has to choose something as the fallback representation, the canonical representation feels like the best for that intuitively? :blush: 

Edit: I have updated the numbers after fixing the logic issues Jitse pointed out.

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
surilindur: This should be fixed now! I hope the implementation follows your suggestions, if not, I can modify it further. :slightly_smiling_face:

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: There is no `C` locale, it seems, and the supported locales seem to depend on the specific system and/or JS runtime. So I have reverted the approach back to normal string conversions, and removed the internationalization API usage. :cry:

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
```diff
@@ -20,10 +27,47 @@ export class TermFunctionXsdToString extends TermFunctionBase {
...
+        .set<IntegerLiteral>([ TypeURL.XSD_INTEGER ], () => TermFunctionXsdToString.castAsInteger)
+        .set<DoubleLiteral>([ TypeURL.XSD_DOUBLE ], () => TermFunctionXsdToString.castAsDouble)
+        .set<FloatLiteral>([ TypeURL.XSD_FLOAT ], () => TermFunctionXsdToString.castAsDouble)
         .onBoolean1Typed(() => val => string(bool(val).str()))
         .onTerm1(() => (val: StringLiteral) => string(val.str()))
         .collect(),
     });
   }
+
+  private static castAsInteger<T extends NumericLiteral>([ val ]: [T]): StringLiteral {
```
jitsedesmet: can this be simplified?
```suggestion
  private static castAsInteger(val: NumericLiteral): StringLiteral {
```

Same below

packages/utils-expression-evaluator/lib/expressions/Term.ts
```diff
@@ -176,11 +176,26 @@ export abstract class NumericLiteral extends Literal<number> {
...
-      this.specificFormatter(this.typedValue);
+    if (this.strValue) {
+      return this.strValue;
+    }
+
+    const str = this.specificFormatter(this.typedValue);
+
+    if (!Number.isFinite(this.typedValue)) {
+      return str.replace('Infinity', 'INF');
+    }
```
jitsedesmet: I feel like this is much more part of the contract of `specificFormatter`? I know it's to educe replication but I liked it better when they all did there own value space handling. here numeric steps out of bounds imo

packages/utils-expression-evaluator/lib/expressions/Term.ts
```diff
@@ -191,11 +206,27 @@ export class IntegerLiteral extends NumericLiteral {
...
+  protected override specificFormatter(val: number): string {
+    if (Number.isFinite(val)) {
+      // Force the number to not be represented as an exponential,
+      // even when large enough for JS to automatically try it.
+      return val.toFixed(0);
+    }
+
+    // The integer in XSD spec cannot be infinity or NaN, because they are not real numbers,
+    // but this is a value space and not a string representation issue.
+    // Thus, return a fallback string representation here.
```
jitsedesmet: Yes, but then why are we checking isInfinite if we know this will not be the case, or rather, should not be the case?

packages/utils-expression-evaluator/lib/expressions/Term.ts
```diff
@@ -206,60 +237,88 @@ export class DecimalLiteral extends NumericLiteral {
...
+  protected override specificFormatter(val: number): string {
+    if (Number.isFinite(val)) {
+      let str = val.toString(10);
+
+      // When the number is so small that JavaScript forces exponential representation,
+      // the value must be forced into decimal format, and trailing zeroes must be stripped.
+      // This does not address accuracy issues, but it does ensure the output is a valid decimal.
+      if (str.includes('e')) {
+        str = val.toFixed(20).replace(/([0-9])0*$/u, '$1');
+      }
```
jitsedesmet: I'm sory. Can you explain this?

When we have 'e' we remove trailing `0`? Should we count how many should be removed?

packages/utils-expression-evaluator/lib/expressions/Term.ts
```diff
@@ -206,60 +237,88 @@ export class DecimalLiteral extends NumericLiteral {
...
+
+      // Ensure there is at least one decimal place.
+      if (!str.includes('.')) {
+        str += '.0';
+      }
+
+      return str;
+    }
+
+    // The decimal in XSD spec cannot be infinity or NaN, because they are not real numbers,
```
jitsedesmet: Same here, then why check? Can we not just assume the number is not Nan or Infty?

jitsedesmet (changes requested):

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
surilindur: Hopefully yes, I cannot remember why I left it like that, but will check it.

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: Okay, I will move that part to the formatters.

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: It should not be the case, but in the engine it currently can be the case. I can remove the checks, though.

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: When there is an `e`, it means the number is formatted as an exponential (JS will do this for very small numbers, for example). In that case, it is forced into decimal with `toFixed`, and the trailing zeroes are removed. The greedy `*` matches all zeroes, starting from the end, but stops before the last zero (the decimal place that needs to be kept) *or* a non-zero number - this is what ends up in the capture group `([0-9])`. The last relevant digit `[0-9]` and all trailing zeroes (the matching part) is replaced by the capture group (last relevant digit).

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: Yes, I will remove the check, even though you can have an infinite integer currently. :sweat_smile: But it is outside the scope of this part, for sure.

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
surilindur: This is now fixed.

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: This is now fixed, too.

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: This is now fixed, too.

packages/utils-expression-evaluator/lib/expressions/Term.ts
surilindur: This is now fixed, too.

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
```diff
@@ -20,10 +27,47 @@ export class TermFunctionXsdToString extends TermFunctionBase {
...
+        .set<FloatLiteral>([ TypeURL.XSD_FLOAT ], () => TermFunctionXsdToString.castAsDouble)
         .onBoolean1Typed(() => val => string(bool(val).str()))
         .onTerm1(() => (val: StringLiteral) => string(val.str()))
         .collect(),
     });
   }
+
+  private static castAsInteger([ val ]: [NumericLiteral]): StringLiteral {
+    return string(integer(val.typedValue).str());
+  }
```
jitsedesmet: I dont see why we need an array?


```suggestion
  private static castAsInteger(val: NumericLiteral): StringLiteral {
    return string(integer(val.typedValue).str());
  }
```

jitsedesmet (changes requested):

packages/actor-function-factory-term-xsd-to-string/lib/TermFunctionXsdToString.ts
surilindur: It is there to avoid having the overloadtree `set` methods become more ugly, because they take an array as the input. I will change the ugliness to be there, instead, then.

surilindur: This should now be fixed.

jitsedesmet (approved):

rubensworks:
Thanks @surilindur!

PR #1789, opened 2026-09-04
Bound the ORDER BY buffer by the LIMIT above it

rubensworks:
Follow-up to #1788, which listed this as the remaining half of the ORDER BY problem: `SELECT ?l WHERE { ?s rdfs:label ?l } ORDER BY ?l LIMIT 10` still materialised and sorted the whole result set to return 10 rows.

## Change

ORDER BY has to see all of its input before it can emit anything, but when only the first results are ever read, it does not have to *keep* all of them.

**`ActorQueryOperationSlice`** annotates the ORDER BY that produces its results with `start + length`, via a new `assignOperationSortLimit` next to the existing `assignOperationSource` in `@comunica/utils-query-operation`. It traverses only operations that map every input result to exactly one output result in the same order (`project`, `extend`), so `SELECT DISTINCT ?x WHERE { ... } ORDER BY ?x LIMIT 10` is deliberately left alone. Without a LIMIT nothing is annotated.

**`SortIterator`** uses that annotation to keep a bounded max-heap of the smallest items instead of buffering everything: O(limit) memory and O(n log limit) comparisons instead of O(n) and O(n log n). It fills the buffer first and heapifies once when it overflows, so a result set that fits within the limit costs exactly what it does today.

**`ActorQueryOperationOrderBy`** now evaluates all ORDER BY expressions in a single pass and sorts once with a lexicographic comparator, instead of chaining one `SortIterator` per expression. Without this the bound could only be applied to the outermost sort of the chain, and the inner ones would still buffer everything. The two are equivalent — a chain of stable sorts by the last key, then the second-to-last, and so on, *is* a lexicographic sort — and doing it in one pass also makes multi-expression ORDER BY faster on its own.

### Correctness

The comparator breaks ties by input position, so a bounded sort returns **exactly** the first `limit` results of the unbounded one rather than an arbitrary subset of the tied ones. That makes the change observationally invisible apart from the timing.

Beyond the unit tests, I checked that property directly: 12 query shapes (single key, descending, ties on the leading key, all-ties, three keys with mixed directions, an unbound sort variable, a partly unbound one, an expression key, an erroring expression, a select expression, and DISTINCT/REDUCED as negative controls) × 10 slices (`LIMIT 0`, first, first 10, past the end, exactly the length, `OFFSET` at and beyond the boundary). All 120 combinations return byte-identical rows to the unbounded sort sliced afterwards. The DISTINCT and REDUCED rows are the ones that would break if the traversal were too permissive.

`jest` (6 826 tests, 100% coverage) and the SPARQL 1.0 / 1.1 / 1.2 / update / csv-tsv / json / federation spec suites all pass.

## Results

200k solutions from an in-memory store. This machine has ±10% run-to-run drift, so both versions were run alternately (A, C, C, A) and the two runs per version averaged:

| query | before | after | |
|---|---|---|---|
| `ORDER BY ?o` | 6 049 ms | 6 096 ms | unchanged |
| `ORDER BY ?o LIMIT 10` | 5 567 ms | **1 362 ms** | 4.1× |
| `ORDER BY ?o LIMIT 10 OFFSET 90` | 5 630 ms | **1 414 ms** | 4.0× |
| `ORDER BY ?o LIMIT 10000` | 5 477 ms | **3 205 ms** | 1.7× |
| `ORDER BY ?o ?w` | 12 486 ms | **8 448 ms** | 1.5× |
| `ORDER BY ?o ?w LIMIT 10` | 12 492 ms | **2 998 ms** | 4.2× |

Reading the same result set without ORDER BY costs 325 ms, so `ORDER BY ?o LIMIT 10` goes from ~5.2 s of sorting to ~1.0 s.

Memory is the point of this at scale. With `--max-old-space-size=900` and 400k solutions, `ORDER BY ?o LIMIT 10` dies with `FatalProcessOutOfMemory` before this change and completes after it.

The `benchmark-web` query added in #1788 (`people-cambridge-pairs-orderby`, ~86k solutions ordered, `LIMIT 10`), alternated the same way, mean of 3 replications after a warmup:

| | mean |
|---|---|
| before | 3 801 ms |
| after | **2 147 ms** |

1.8× on a query where ~4 HTTP round trips to `fragments.dbpedia.org` are part of the total.

### On the unchanged row

`ORDER BY` without a LIMIT is the shape that had to not regress, and it is the one the restructure puts at risk: one sort over annotated solutions instead of one per expression means the annotation has to carry a *list* of results. Allocating that list per solution costs a measurable 5-8% on 200k rows, which I confirmed over 10 paired runs in both orderings before deciding it mattered. Sorting on a single expression therefore gets its own annotation without the list — hence the two branches in `runOperation`. With that, the medians land within 1% of master, which is inside this machine's noise.

## Notes

- The `benchmark-web` query from #1788 now exercises the bounded path rather than the full sort it was added to guard. Guarding the unbounded path again needs a separate query without a LIMIT; I left that out rather than adding a second heavy query to CI unasked, but happy to add one.
- The annotation lives in the slice actor rather than in an optimize-query-operation actor, which is where algebra rewriting normally belongs. A new actor package cannot reach the shipped engine without a new config version snapshot (`config-default-v5-3-0.json` pins `optimize-query-operation/actors-v5-1-3.json`), which seemed like the wrong thing to introduce here. Moving it later is a small change: the annotation itself is already a plain algebra-level operation.
- Unrelated, but found while working on this: the `window` config parameter of `ActorQueryOperationOrderBy` does not produce a sorted stream. With `window: 3`, `[5,4,3,2,1]` comes out as `[3,2,1,4,5]` — it emits the smallest item of a sliding window as soon as the window is full, without knowing whether something smaller follows. It defaults to unbounded, so nothing hits it by default, and this PR does not touch that path. It also means the note in #1788 about reusing `window` for top-k was wrong; that is why this adds a separate bounded-heap path instead.

https://claude.ai/code/session_019fVTUvQ2uCR1p4KQxAYrgs

---
_Generated by [Claude Code](https://claude.ai/code/session_019fVTUvQ2uCR1p4KQxAYrgs)_

### DIFF
--- a/packages/actor-query-operation-slice/lib/ActorQueryOperationSlice.ts
+++ b/packages/actor-query-operation-slice/lib/ActorQueryOperationSlice.ts
@@ -13,13 +13,24 @@ import type {
   IMetadata,
   IActionContext,
 } from '@comunica/types';
-import { Algebra } from '@comunica/utils-algebra';
+import { Algebra, isKnownOperation } from '@comunica/utils-algebra';
+import { assignOperationSortLimit } from '@comunica/utils-query-operation';
 import type { AsyncIterator } from 'asynciterator';
 
 /**
  * A comunica Slice Query Operation Actor.
  */
 export class ActorQueryOperationSlice extends ActorQueryOperationTypedMediated<Algebra.Slice> {
+  /**
+   * Operations that map every input result to exactly one output result in the same order, so that
+   * a bound on their output is also a bound on their input.
+   * Notably, DISTINCT and REDUCED are not among those.
+   */
+  private static readonly ORDER_PRESERVING_TYPES = new Set<string>([
+    Algebra.Types.PROJECT,
+    Algebra.Types.EXTEND,
+  ]);
+
   public constructor(args: IActorQueryOperationTypedMediatedArgs) {
     super(args, Algebra.Types.SLICE);
   }
@@ -36,9 +47,10 @@ export class ActorQueryOperationSlice extends ActorQueryOperationTypedMediated<A
       context = context.set(KeysQueryOperation.limitIndicator, operation.length);
     }
 
-    // Resolve the input
+    // Resolve the input, telling the ORDER BY below us (if any) how many results will be read
+    const input = ActorQueryOperationSlice.pushSortLimit(operation) ?? operation.input;
     const output: IQueryOperationResult = await this.mediatorQueryOperation
-      .mediate({ operation: operation.input, context });
+      .mediate({ operation: input, context });
 
     if (output.type === 'bindings') {
       const bindingsStream = this.sliceStream(output.bindingsStream, operation);
@@ -60,6 +72,41 @@ export class ActorQueryOperationSlice extends ActorQueryOperationTypedMediated<A
 
     // In all other cases, return the result as-is.
     return output;
+  }
+
+  /**
+   * Annotate the ORDER BY that produces the results of the given slice with the number of results
+   * that will actually be read, so that it can sort within a bounded buffer instead of
+   * materializing all of its input.
+   * @param operation A slice operation.
+   * @return The rewritten input of the slice, or undefined if there is no ORDER BY to annotate.
+   */
+  private static pushSortLimit(operation: Algebra.Slice): Algebra.Operation | undefined {
+    // Without a LIMIT, the number of results that will be read is unbounded
+    if (operation.length === undefined) {
+      return;
+    }
+    return ActorQueryOperationSlice.annotateSortLimit(operation.input, operation.start + operation.length);
+  }
+
+  /**
+   * Annotate the ORDER BY that determines the first `sortLimit` results of the given operation.
+   * @param operation The input of a slice, or of an operation that was traversed from one.
+   * @param sortLimit The number of results that will be read from that operation.
+   * @return The annotated operation, or undefined if there is no ORDER BY to annotate.
+   */
+  private static annotateSortLimit(operation: Algebra.Operation, sortLimit: number): Algebra.Operation | undefined {
+    if (isKnownOperation(operation, Algebra.Types.ORDER_BY)) {
+      return assignOperationSortLimit(operation, sortLimit);
+    }
+    if (ActorQueryOperationSlice.ORDER_PRESERVING_TYPES.has(operation.type)) {
+      const parent = <Algebra.Project | Algebra.Extend> operation;
+      const input = ActorQueryOperationSlice.annotateSortLimit(parent.input, sortLimit);
+      if (input) {
+        const copy: Algebra.Project | Algebra.Extend = { ...parent, input };
+        return copy;
+      }
+    }
   }
 
   // Slice the stream based on the pattern values
--- a/packages/actor-query-operation-slice/lib/ActorQueryOperationSlice.ts
+++ b/packages/actor-query-operation-slice/lib/ActorQueryOperationSlice.ts
@@ -13,24 +13,13 @@ import type {
   IMetadata,
   IActionContext,
 } from '@comunica/types';
-import { Algebra, isKnownOperation } from '@comunica/utils-algebra';
-import { assignOperationSortLimit } from '@comunica/utils-query-operation';
+import { Algebra } from '@comunica/utils-algebra';
 import type { AsyncIterator } from 'asynciterator';
 
 /**
  * A comunica Slice Query Operation Actor.
  */
 export class ActorQueryOperationSlice extends ActorQueryOperationTypedMediated<Algebra.Slice> {
-  /**
-   * Operations that map every input result to exactly one output result in the same order, so that
-   * a bound on their output is also a bound on their input.
-   * Notably, DISTINCT and REDUCED are not among those.
-   */
-  private static readonly ORDER_PRESERVING_TYPES = new Set<string>([
-    Algebra.Types.PROJECT,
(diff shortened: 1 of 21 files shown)


packages/actor-query-operation-slice/lib/ActorQueryOperationSlice.ts
```diff
@@ -13,13 +13,24 @@ import type {
...
 } from '@comunica/types';
-import { Algebra } from '@comunica/utils-algebra';
+import { Algebra, isKnownOperation } from '@comunica/utils-algebra';
+import { assignOperationSortLimit } from '@comunica/utils-query-operation';
 import type { AsyncIterator } from 'asynciterator';
 
 /**
  * A comunica Slice Query Operation Actor.
  */
 export class ActorQueryOperationSlice extends ActorQueryOperationTypedMediated<Algebra.Slice> {
```
rubensworks, line 22: Could these changes be moved to a new optimize actor instead?
Be sure to create the appropriate versioned configs.

packages/actor-query-operation-slice/lib/ActorQueryOperationSlice.ts
```diff
@@ -13,13 +13,24 @@ import type {
...
+  /**
+   * Operations that map every input result to exactly one output result in the same order, so that
+   * a bound on their output is also a bound on their input.
+   * Notably, DISTINCT and REDUCED are not among those.
+   */
+  private static readonly ORDER_PRESERVING_TYPES = new Set<string>([
```
rubensworks: I don't understand the need for this. Is it really needed? And if so, are you sure that just these two operator types are sufficient?

packages/utils-query-operation/lib/Utils.ts
```diff
@@ -101,6 +101,28 @@ export function assignOperationSource<O extends Algebra.Operation>(operation: O,
...
+ * The operation is copied and returned.
+ * @param operation An operation.
+ * @param sortLimit The number of smallest results that will be consumed.
+ */
+export function assignOperationSortLimit<O extends Algebra.Operation>(operation: O, sortLimit: number): O {
+  operation = { ...operation };
+  operation.metadata = { ...operation.metadata, sortLimit };
+  return operation;
+}
+
```
rubensworks: This should be moved into this new optimize actor.

rubensworks (commented):
Also, for benchmark-web, also add a query unguarded by LIMIT.

packages/actor-query-operation-slice/lib/ActorQueryOperationSlice.ts
rubensworks: Done. Moved into a new `@comunica/actor-optimize-query-operation-sort-limit-pushdown` actor on the optimize-query-operation bus; `ActorQueryOperationSlice` is byte-identical to master again.

For the versioned configs I followed the #1778 precedent: a new `optimize-query-operation/actors-v5-3-0.json` (which is `actors-v5-1-3.json` plus the new actor), with `config-default-v5-3-0.json` repointed at it. I also added the actor to the unversioned `actors.json`, which #1778 did not do for `rdf-metadata-extract` — happy to drop that if the omission there was deliberate.

I first named the new file `actors-v5-4-0.json` with a matching `config-default-v5-4-0.json`, but that makes `yarn test:config-compat` fail: it then caps the `config-default-v5-3-0.json` test range at 5.4.0 and tries `npm install @comunica/query-sparql@5.4.0`, which does not exist yet. With the v5-3-0 naming, `test:config-compat` passes.

---

packages/actor-query-operation-slice/lib/ActorQueryOperationSlice.ts
rubensworks: It is needed, but you were right that the set was wrong. It is now `PROJECT` only.

I enumerated what actually ends up between the `slice` and the `orderby`, over plain SELECT, `SELECT *`, `BIND` in the WHERE clause, OFFSET+LIMIT, GROUP BY, ORDER BY on an expression, projected expressions, VALUES, DISTINCT, REDUCED, CONSTRUCT, DESCRIBE, subqueries, nested subqueries and FROM:

- `slice → project → orderby` (most of the above)
- `slice → project → project → orderby` (nested subquery)
- `project → slice → project → orderby` (subquery)
- `slice → distinct → project → orderby`, `slice → reduced → project → orderby`
- `slice → construct → orderby`, `slice → describe → orderby`
- `from → slice → project → orderby`

So a `PROJECT` is always in between (twice for subqueries), which is why the traversal cannot just look at `slice.input`. `EXTEND` never occurs there, because `BIND` is applied below the `ORDER BY`, so I removed it. `DISTINCT`, `REDUCED`, `CONSTRUCT` and `DESCRIBE` do occur there and must stay excluded, since each of them can read more results than it produces, which would make the bound unsound.

There is a test for each of these cases, including one asserting that DISTINCT is not traversed.

---

packages/utils-query-operation/lib/Utils.ts
rubensworks: Done, `utils-query-operation/lib/Utils.ts` is back to master.

`assignOperationSortLimit` is now inlined in the new optimize actor, which is the only writer. For the read side I did not export a getter from the actor package, since that would make `actor-query-operation-orderby` depend on an optimize actor; instead it reads `operation.metadata?.sortLimit` directly, which is what the helper did anyway. Let me know if you would rather have a shared accessor somewhere.

---

PR #1794, opened 2026-09-06
Keep endpoint logs out of the generated EARL reports

redreye:
`spec-earl:sd-1-1` and `spec-earl:prot-1-1` redirect stdout into the EARL report, but the spec harness binds the SPARQL endpoint's log stream to stdout as well. Every request the endpoint serves is prepended to the report, so the file it
produces is not valid Turtle:

```
earl-sd-1-1.ttl
[200] GET to /sparql
      Requested media type: application/ld+json
      Received query for service description.
```


rubensworks:
Aha, good catch!

PR #1795, opened 2026-09-06
Parse SPARQL endpoint request bodies with async/await

redreye:
In a promise executor, `resolve`/`reject` settle the promise but don't return, so `parseBody` had to write `return resolve(...)` in every branch. The `catch` around `JSON.parse` of the `context` parameter omits that `return`: it rejects and then falls through to the checks below, which call `resolve` on a promise that has already settled. Those calls are no-ops, so the request is still rejected as intended but only because a promise ignores a second settle, not because the code stops.

This splits reading the body out of interpreting it. `readBody` collects the UTF-8 string, and `parseBody` becomes a plain `async` function that returns or throws.


rubensworks:
Looks good, thanks!

PR #1796, opened 2026-09-07
Use preOrder and async mapOperation

jitsedesmet:
This exposes the new Transformer functions as part of Comunica's self maintained `utils-algebra`. Afterward, I checked what code can be cleaned up easily using the newly introduced features of the `mapOperation`-family.

While checking the packages, a bug was discovered in `ActorOptimizeQueryOperationRewriteAdd` and fixed + tested.

@rubensworks The current version is still a draft since it works by linking a local version of Traqula still under review: https://github.com/comunica/traqula/pull/187 . 

I intend to finalize both commits at the same time, so if you are happy with the changes here, I will release the new Traqula. I want the main consumer of Traqula to be happy with the API this currently exposes.


jitsedesmet:
Something is wrong with the dif rendering. Let me merge with main first

packages/actor-optimize-query-operation-assign-sources-exhaustive/lib/ActorOptimizeQueryOperationAssignSourcesExhaustive.ts
```diff
@@ -101,11 +101,8 @@ export class ActorOptimizeQueryOperationAssignSourcesExhaustive extends ActorOpt
         },
       },
       [Algebra.Types.CONSTRUCT]: {
-        preVisitor: () => ({ continue: false }),
-        transform: constructOp => factory.createConstruct(
-          this.assignExhaustive(factory, constructOp.input, sources, serviceSources),
-          constructOp.template,
-        ),
+        // The template holds quad patterns to produce, not patterns to match, so it gets no source.
+        preVisitor: () => ({ ignoreKeys: new Set([ 'template', 'metadata' ]) }),
```
rubensworks, line 105: This seems wrong. Sources should be kept? (also below)

rubensworks (commented):
I haven't gone through it in detail. But the diff seems massive. Can you go over all the changes yourself first to see if they are all really necessary, and ping me when done?

packages/actor-optimize-query-operation-assign-sources-exhaustive/lib/ActorOptimizeQueryOperationAssignSourcesExhaustive.ts
jitsedesmet: No, so basically, the previous code did the recursion, but only on the `input` key. This does the same, we recurse on the things in input, except 'template'

rubensworks: But where is `this.assignExhaustive` called then? This must be called somewhere, otherwise `constructOp.input` has no source annotations.

jitsedesmet:
@rubensworks Something was wrong with the diff rendering before. After merging master that should be fixed.
I reviewed all changes and they seem correct.
However, special attention from you should go to:
1. transformFilterFalse no longer has `maxIterations`.  + The order seems to have changed. As such, I am also rather hesitant to accept this.
2. `ActorOptimizeQueryOperationPruneEmptySourceOperations.ts` has large potential (the async handling seems to reduce complexity a lot), however, I don't even understand the original code. The original code has about 3 three-iterations going on, some within each other, others sequentially.
3. `ActorOptimizeQueryOperationRewriteAdd` is a bugfix that seems needed to me, indeed SPARQL can have composite queries.

The other classes, I understand the diff and the original code completely, and they seem correct.

packages/actor-optimize-query-operation-assign-sources-exhaustive/lib/ActorOptimizeQueryOperationAssignSourcesExhaustive.ts
jitsedesmet: Our mapOperation function is a post-order transformation, so, since the preVisitor no longer says `continue: false`, we have already updated the input at the time the `transform` is called.

ACKNOWLEDGEMENTS.md
```diff
@@ -0,0 +1,11 @@
+# Acknowledgements
```
rubensworks: I would not add this here.

packages/actor-optimize-query-operation-assign-sources-exhaustive/lib/ActorOptimizeQueryOperationAssignSourcesExhaustive.ts
jitsedesmet: We traverse down the tree using `preVisitor`, this provided context to the post-order transformer. `continue: false` would mean that we need not iterate further down, by default, `continue` is true. 

The previous code basically stopped the execution of the current mapper and then recursively called the mapper again (starting a new mapping process).

ACKNOWLEDGEMENTS.md
jitsedesmet: Agreed

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -554,25 +554,26 @@ export class HttpServiceSparqlEndpoint {
...
-        }
-        return algebraFactory.createDeleteInsert(
-          deleteInsert.delete,
-          deleteInsert.insert,
-          algebraFactory.createFrom(deleteInsert.where, dataset.default, dataset.named),
-        );
-      }
-    }
-    return operation;
+    return algebraUtils.mapOperation(operation, {
```
rubensworks, line 557: This pattern seems to occur a lot:
Previous code specifically checks for some operation types, while new code replaces it with a mapOperation.
Does mapOperation still lead to a deep copy of operations?
If so, I am afraid all of these replacements may have a negative impact on performance.
Before I go further in my review, can you either confirm that no deep copies are done, or that there is no performance impact?

jitsedesmet: Deep copies are indeed made. I thought this was desired behavior? Keeping manipulations stateless. How can I benchmark best?

jitsedesmet: And what should I benchmark? The whole Comunica pipeline? Or this one function? I doubt making the copy or not will impact the system as a whole.

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -934,7 +934,10 @@ export class HttpServiceSparqlEndpoint {
         }
         // A request may only contain a single query or update
         if (Array.isArray(bodyStructure.query) || Array.isArray(bodyStructure.update)) {
-          throw new TypeError(`Invalid request body received, it can only contain a single query or update parameter`);
+          // The class of this error is observable to callers of this public method, so it stays the plain
+          // Error the promise rejected with before the body parsing became async.
+          // eslint-disable-next-line unicorn/prefer-type-error
+          throw new Error(`Invalid request body received, it can only contain a single query or update parameter`);
```
rubensworks: Why is this change needed?

packages/actor-optimize-query-operation-distinct-terms-pushdown/lib/ActorOptimizeQueryOperationDistinctTermsPushdown.ts
```diff
@@ -28,19 +28,25 @@ export class ActorOptimizeQueryOperationDistinctTermsPushdown extends ActorOptim
...
+        shape = source.source.getSelectorShape(
+          source.context ? action.context.merge(source.context) : action.context,
+        );
+        sourceShapes.set(source, shape);
+      }
+      return shape;
+    };
 
-    const operation = algebraUtils.mapOperation(action.operation, {
+    const operation = await algebraTransformer().transformNodeAsync(action.operation, {
```
rubensworks, line 46: This is really nice :-)

packages/utils-query-operation/lib/MaterializeBindings.ts
```diff
@@ -137,72 +141,40 @@ export function materializeOperation(
       );
     } },
     [Algebra.Types.FILTER]: {
```
rubensworks, line 143: Are you certain about the changes below here?
Why are they needed?
I'm asking because this piece of code has been the source of many issues, so I want to make sure we don't introduce regressions (we should already have extensive unit tests for it though).

rubensworks (changes requested):
Looks very exciting!

I don't have many comments.

The traqula-side changes seem ok to me, so let's get that part released so we can inspect if the CI here passes properly.

Can you also do a quick before/after performance check on watdiv and bsbm (just the scales our CI uses, so should be pretty quick)?

packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
jitsedesmet: Good catch!

packages/utils-query-operation/lib/MaterializeBindings.ts
jitsedesmet: I had another read, changing only one line that actually resulted in different behavior. For the other things, the behaviour should be the same

jitsedesmet: None of them are needed. They just clean up the system

rubensworks:
Looks perfect, thanks @jitsedesmet.

PR #1801, opened 2026-09-09
Add a BSBM benchmark at 10 000 products

rubensworks:
Adds a second BSBM-File experiment, `performance/benchmark-bsbm-file-10k`, at 10 000 products, and wires it into the CI performance matrix as `BSBM-File-10k`.

## Why

The existing `benchmark-bsbm-file` uses 1 000 products, which is 374 911 triples. A full query pass over it takes about 0.2 s. At that size what the numbers measure is dominated by per-query and per-request overhead, not by the joins and scans, so a change in the join stack is essentially invisible in it. Several join changes I measured recently moved BSBM-File by less than the run-to-run noise while moving the same queries substantially at larger scales.

At 10 000 products the dataset is 3 564 773 triples and the join work dominates.

## Measured budget

Run end to end with `jbr prepare -c 0 && jbr run -c 0` on a 4-core machine:

| step | time |
| --- | --- |
| `jbr prepare` (BSBM generation) | 27 s |
| `jbr run` (5 warmup + 10 measured mixes, including the lazy load on the first query) | 246 s |
| **total** | **273 s** |

4.6 min. For comparison the existing `BSBM-File` job takes 2.5 min end to end and the slowest job in the matrix, `web`, takes 20.1 min, so there is plenty of room under the runner limit.

Measured separately, loading the dataset into a file source takes 103 s and peaks at **5.72 GB RSS**, which the inherited `NODE_OPTIONS=--max-old-space-size=8192` covers, though not by a wide margin. That works out to about 1.7 KB per triple, which is what sets the ceiling on this approach: see the note on 100 000 products below.

## Per-query results

Means over the 10 measured mixes, from `query-times.csv`:

| query | ms | query | ms | query | ms |
| --- | --- | --- | --- | --- | --- |
| Q1 | 19.8 | Q5 | **3 240.3** | Q9 | 617.9 |
| Q2 | 49.5 | Q7 | 173.8 | Q10 | 106.6 |
| Q3 | 49.0 | Q8 | 99.6 | Q11 | 4.6 |
| Q4 | 49.7 | | | Q12 | 17.3 |

Result counts match the HDT run over the same dataset exactly. (Q6 is not part of the BSBM default mix and reports as undefined, same as in the existing experiment.)

## One thing the run turned up

The experiment does not work without raising the endpoint's per-query timeout, so one of the commits does that. The file source loads its 3.5M triples lazily on the first query, which takes about 100 s, and the endpoint's default timeout is 60 s, so the master killed the worker mid-load:

```
Worker 7185 got assigned a new query (1).
Worker 7185 timed out for query 1.
Worker 7185 died with 15. Starting new worker.
```

after which every BSBM query got a connection refused. `-t 300` covers the load with room to spare while still capping a runaway query well inside the job budget. The 1 000-product experiment does not hit this because its load is an order of magnitude shorter. Combination 1 already sets `queryTimeout` to 500 through the Docker hook, so only combination 0 needed it.

## Why not 100 000 products

Out of reach over a file source, and not marginally. The store costs about 1.7 KB per triple (5.72 GB measured for 3 564 773 triples), so 35 647 730 triples would need roughly 57 GB. A GitHub `ubuntu-latest` runner has 16 GB of RAM. Even raising `--max-old-space-size` past the 8 GB currently configured, the ceiling for this experiment is somewhere around 20 000 products, which is not far enough past 10 000 to be worth a third experiment. The 9.2 GB N-Triples file would also be awkward on the runner's disk.

A larger BSBM is viable over HDT, where the dataset is memory-mapped rather than loaded into the JS heap, but there the cost moves to preparation: `rdf2hdt` runs at about 9.5 MB/s in these images, so converting 9.2 GB would take roughly 17 min on top of generation, which eats most of the budget on its own. Shipping a prebuilt HDT as an asset would avoid that, at the price of a roughly 1.5 GB LFS object fetched on every run.

## Why the small variant stays

`benchmark-bsbm-file` has a long history of results on https://github.com/comunica/comunica-performance-results, and it isolates overhead-bound behaviour, which the larger experiment no longer does. The two answer different questions, so this adds rather than replaces.

## Why not WatDiv at a larger scale here

I measured that as well and it does not fit over a file source either. WatDiv at 10.9M triples needs 273.7 s to load at 9.2 GB RSS, and individual queries take up to 69.7 s (C3) and 53.5 s (C1). A larger WatDiv is viable over HDT: comunica/comunica-feature-hdt#39.

## Changes

- `performance/benchmark-bsbm-file-10k/`: same layout and settings as `benchmark-bsbm-file` with `productCount` 10 000 instead of 1 000, `-t 300` on the combination 0 endpoint, and `@id`s renamed accordingly. `warmupRuns`/`runs` are unchanged so the two are directly comparable per query.
- `.github/workflows/ci.yml`: `bsbm-file-10k` added to both arrays of the matrix expression, and its artifact, output path and `BSBM-File-10k` label added to `performance-consolidate` and to the `performance-compare` job that #1802 introduced.

No lockfile change: every dependency of the new workspace is already in `yarn.lock` at the same range, so `yarn install --frozen-lockfile` is satisfied.

Since #1802, the `performance` job also runs on pull requests, so this PR's CI does exercise the new job and `performance-compare` will report it.

https://claude.ai/code/session_019fVTUvQ2uCR1p4KQxAYrgs


performance/benchmark-bsbm-file-10k/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+This internal package benchmarks Comunica File using the [BSBM](http://wbsg.informatik.uni-mannheim.de/bizer/berlinsparqlbenchmark/) benchmark,
+at 10 000 products instead of the 1 000 used by [`benchmark-bsbm-file`](../benchmark-bsbm-file),
+which is 3 564 773 triples instead of 374 911.
+
+The smaller variant stays as it is. It gives fast feedback and a long history, and the two answer
+different questions: at 1 000 products the whole query set runs in a fifth of a second, so what it
+measures is dominated by per-query overhead, while at 10 000 products the joins and the scans they
+read are what takes the time.
+
```
rubensworks:
```suggestion
```

PR #1804, opened 2026-09-09
Send the query end event to the master on every response close

rubensworks:
A SPARQL endpoint worker announces `{ type: 'start' }` to the master before running a query, and the master arms a timeout that shuts the worker down if no matching `{ type: 'end' }` arrives. The worker only attached the listener that sends `end` after `resultToString` had resolved and the result was being piped:

```ts
data.pipe(response);
eventEmitter = data;
// ...
response.on('close', () => {
  process.send?.({ type: 'end', queryId });
});
```

Every way out of `writeQueryResult` before that point therefore leaves the master with an armed timeout for a query that is already over: a client that disconnects while the query is still executing, a query that throws and takes the `writeBadRequest` path, a `HEAD` request, or a result that cannot be serialized for the requested media type.

When that stale timeout later fires, the master finds the worker connected (it is healthy, just busy with unrelated queries), sends it `shutdown`, and the worker terminates its open connections with `!TIMEDOUT!` and exits. The endpoint then refuses connections until the replacement worker has reloaded its sources, which for a large dataset is tens of seconds.

## How this shows up

I hit this while benchmarking. My script probed the endpoint for readiness with `curl -m 5`; the first query also triggers the source load, so curl gave up before the response started streaming. Ten minutes later, in the middle of the benchmark, the master killed the worker over that long-finished probe:

```
Worker 18247 got assigned a new query (0).
...
Worker 18247 timed out for query 0.
Shutting down worker 18247 with 1 open connections.
Worker 18247 died with 15. Starting new worker.
```

and the rest of the run got connection refused. The gap between `-t` and the client's own timeout is all it takes, so any client that aborts a request can arm this.

Reproduced against a file endpoint with `-t 15`: abort one request during startup, then serve short queries for 25s.

| | queries started | queries finished | workers killed by timeout |
| --- | --- | --- | --- |
| before | 21 | 20 | 1 |
| after | 27 | 27 | 0 |

## Changes

- Attach the `close` listener immediately after sending `start`, so that every exit path from a query execution reports its end. The response can close at any point after the start is announced, and `close` fires once, so this covers the disconnect, failure, `HEAD` and serialization paths without changing the normal one.
- Clear a worker's pending timeouts when it exits. They can only ever fire on a dead worker, where they are no-ops that keep the master's event loop alive for up to the timeout.

A query that genuinely runs past the timeout is still shut down: with `-t 3` and a query that outlives it, the timeout fires exactly once and the endpoint recovers.

## Tests

Two tests in `HttpServiceSparqlEndpoint-test.ts` covering a client that disconnects mid-execution and a query that fails. Both fail on `master` and pass with the change. `packages/actor-init-query` is green (270 tests).

https://claude.ai/code/session_0198qd95HYRdxku1Ahfp7zLu

---
_Generated by [Claude Code](https://claude.ai/code/session_0198qd95HYRdxku1Ahfp7zLu)_

### DIFF
--- a/packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
+++ b/packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
@@ -624,9 +624,6 @@ export class HttpServiceSparqlEndpoint {
     process.send?.({ type: 'start', queryId });
 
     // Send message to master process to indicate the end of an execution.
-    // This is attached before the query runs, because the response can close at any point after this:
-    // the client may disconnect, or the query may fail. A start that is never followed by an end leaves
-    // the master with a timeout that fires later and kills a worker that is serving unrelated queries.
     response.on('close', () => {
       process.send?.({ type: 'end', queryId });
     });
--- a/packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
+++ b/packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
@@ -234,8 +234,16 @@ export class HttpServiceSparqlEndpoint {
 
     // Attach listeners to each new worker
     cluster.on('listening', (worker) => {
+      // Handle worker timeouts
+      const workerTimeouts: Record<number, NodeJS.Timeout> = {};
+
       // Respawn crashed workers
       worker.once('exit', (code, signal) => {
+        // Drop the timeouts of the queries this worker was running, as they can only fire on a dead worker
+        for (const workerTimeout of Object.values(workerTimeouts)) {
+          clearTimeout(workerTimeout);
+        }
+
         if (!worker.exitedAfterDisconnect) {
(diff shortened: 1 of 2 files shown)


packages/actor-init-query/lib/HttpServiceSparqlEndpoint.ts
```diff
@@ -617,6 +623,14 @@ export class HttpServiceSparqlEndpoint {
     // Send message to master process to indicate the start of an execution
     process.send?.({ type: 'start', queryId });
 
+    // Send message to master process to indicate the end of an execution.
+    // This is attached before the query runs, because the response can close at any point after this:
+    // the client may disconnect, or the query may fail. A start that is never followed by an end leaves
+    // the master with a timeout that fires later and kills a worker that is serving unrelated queries.
```
rubensworks:
```suggestion
    // Send message to master process to indicate the end of an execution.
```

Issue #1394, opened 2024-07-10
Improve cost model to determine when Bind Join is used


rubensworks:
#### Issue type:

- :snail: Performance issue 

____
#### Description:

In order to resolve the problem where doing direct joins is cheaper than doing bind joins (#548 and #1196),
we introduced a heuristic (cb8d276a779194b3f422f2b3b979342f9e7d7a55) that only uses bind joins if the smallest stream is much smaller than the largest stream.

Currently, this heuristic uses the parameter 60 (2543e868991fe5ff0de7abc7723e1a6976baae57), which seems to optimize most queries. However, for BSBM queries 2, 5, and 10, results showed that parameter 100 performed significantly better. But for other queries, this 100 value slowed down things again. This shows that this simplistic cost is missing some parameters. (WatDiv C1 requires <= 61, Ruben's query requires > 55)

Some time later, after adding `isRemoteAccess` (https://github.com/comunica/comunica/commit/db38811ab069aefd5a055e5f192d83ab887527dd), I managed to make BSBM 2 and 5 faster. But that made WatDiv S3 and S4 slightly slower. But given the overall positive impact, it was worth it.

The goal of this issue is to come up with a new cost model that includes these hidden parameters. This could include checking the number of join entries, the number of overlapping variables, the join selectivity, ...

____

rubensworks:
Both cost sources this heuristic was standing in for are now expressed in the model itself:

#1798 aligns the multi-bind cost with the other join actors and charged for the bound sub-query.
#1805 removed the assumption that a join variable is a key.

PR #1805, opened 2026-09-09
Estimate join cardinality from distinct values where sources know them

rubensworks:
Towards #1394.

`ActorRdfJoin.getSharedVariableJoinCardinality` estimates a join as `Π card / max(V)^(n-1)`, the textbook `|R||S| / max(V_R, V_S)`, where `V` is the number of distinct values the shared variable takes. Comunica has no `V`, so it substitutes the cardinality, which assumes every binding carries a different value for the join variable. That holds for a one-to-many join and breaks for a many-to-many one, where it understates the result by the average number of duplicates per value.

BSBM Q5 is the case in point: at `-pc 1000` the estimate for the `?product` sub-star is 25 against an actual 682. That is small enough for `minMaxCardinalityRatio` to open a bind join, and Comunica re-executes three patterns once per binding over 279 bindings instead of hashing them once.

This adds an optional `distinctValues` to `MetadataVariable`, filled in by `QuerySourceRdfJs` when the source can answer it from an index, and uses it in place of the cardinality when every entry binding the shared variable reports it. Sources that cannot answer cheaply report nothing and the cardinality still stands in, so nothing regresses on sources without the capability.

Only cases that cost an index lookup are counted:

- a pattern with a single variable left has one value per quad, so the cardinality is the exact answer;
- otherwise only the object of a bound predicate is counted, via `countDistinctTerms(['graph', 'predicate', 'object'], ...)`, because the store's indexes place graph and predicate before the object and its nested map has one entry per distinct object. Counting subjects would mean walking that map, so it is left undefined;
- a variable occupying more than one position of the pattern is left undefined, as its values are constrained by all of its positions at once.

Counts are cached per query execution, scoped on `KeysInitQuery.queryExecutionScope` like the existing cardinality cache. On WatDiv C2 at scale 50 the breakdown per query is 19135 calls into this code, of which 11947 return before reaching the store (single-variable patterns and non-object positions), 7179 are cache hits and 9 are real index lookups. See the note on the cache below.

## Benchmarks

Interleaved old/new rounds on the same machine, medians. WatDiv is the jbr file benchmark; BSBM is the explore mix through a file endpoint, mirroring `benchmark-bsbm-file`.

| benchmark | scale | rounds | total | notable |
| --- | --- | --- | --- | --- |
| WatDiv | 10 (1.09M triples) | 4 | +1.1% | C3 +0.6% |
| WatDiv | 50 (5.52M triples) | 3 | -1.8% | C3 -4%, C1 +9%, C2 +13% |
| BSBM | pc 1000 (350k triples) | 3 | **-31.6%** | Q5 **-70.4%** |
| BSBM | pc 10000 (3.56M triples) | 3 | -1.0% | Q5 -4.3% |

Timings alone cannot separate a plan change from machine noise at these magnitudes, so I also compared physical plans directly, interleaving the two builds three times each so that run-to-run instability separates from build effects:

- **WatDiv scale 10: 100/100 queries plan identically.** Scale 50: 97/100 identical, and the other three differ only in a printed cardinality estimate (598 vs 619) with the same operators.
- **BSBM pc 1000: only Q5 changes**, stably across all rounds.
- **BSBM pc 10000: nothing changes.**

So every WatDiv number above is noise measuring identical plans, and the change is confined to the case it targets. On Q5 at pc 1000 the plan goes from a bind join over 279 bindings (468ms) to symmetric hash joins (168ms).

Worth being explicit about the limit: at pc 10000 master already picks the hash plan, because the 10x-larger cardinalities close the `minMaxCardinalityRatio` gate on their own. This corrects one misestimate that the gate was papering over; it does not replace the gate, and #1394 stays open.

## Notes for review

- `IRdfJsSourceExtended.countDistinctTerms` gains an optional `filters` argument, matching what rdf-stores 2.5 accepts. **Passing the full `['graph', 'predicate', 'object']` term-name list is load-bearing.** `RdfStore.countDistinctTerms` picks its index with `getBestIndexTerms(componentOrders, terms)`, which scores on `terms` alone and never looks at `filters`. On WatDiv scale 10, both forms return the same answer (1173 distinct objects of `schema:actor`), but `countDistinctTerms(['object'], filters)` takes **410 ms** while `countDistinctTerms(['graph', 'predicate', 'object'], filters)` takes **3.34 µs**. The natural call is thus a five-order-of-magnitude trap, and arguably rdf-stores should make index selection filter-aware so callers need not know this.
- **The per-query cache may not be worth keeping.** Of that 3.34 µs, only 0.04 µs is the actual count (`map.size` after two `Map.get`s); the rest is index selection, term ordering, match-path construction and dictionary encoding, all of which are a pure function of the terms array and the store's index configuration and could be memoized inside rdf-stores. Skipping the cache here would cost 7179 extra lookups on WatDiv C2 at scale 50, or ~24 ms on a 1456 ms query (1.6%), and under 0.5 ms on every other query I measured. For comparison, the existing `cardinalityCache` guards `countQuads`, which measures 25.8 µs on the same pattern, roughly 8x more, so the precedent for caching does not transfer as directly as it first appears. I am happy to drop the cache if you would rather keep the diff small.

https://claude.ai/code/session_0198qd95HYRdxku1Ahfp7zLu

### DIFF
--- a/packages/actor-query-source-identify-rdfjs/lib/IRdfJsSourceExtended.ts
+++ b/packages/actor-query-source-identify-rdfjs/lib/IRdfJsSourceExtended.ts
@@ -88,6 +88,9 @@ export interface IRdfJsSourceExtended extends RDF.Source {
    * Returns the number of distinct combinations of the specified terms.
    *
    * This will only be used if `features.indexDistinctTerms` is true.
+   *
+   * @param termNames The term names to count combinations of.
+   * @param filters Optional term names that are bound to specific terms, acting as filters.
    */
-  countDistinctTerms?: (termNames: QuadTermName[]) => number;
+  countDistinctTerms?: (termNames: QuadTermName[], filters?: (RDF.Term | undefined)[]) => number;
 }
--- a/packages/actor-query-source-identify-rdfjs/lib/QuerySourceRdfJs.ts
+++ b/packages/actor-query-source-identify-rdfjs/lib/QuerySourceRdfJs.ts
@@ -6,6 +6,7 @@ import type {
   FragmentSelectorShape,
   IActionContext,
   IQuerySource,
+  MetadataVariable,
   QuerySourceReference,
 } from '@comunica/types';
 import { Algebra, AlgebraFactory, isKnownOperation, TypesComunica } from '@comunica/utils-algebra';
@@ -29,6 +30,10 @@ export class QuerySourceRdfJs implements IQuerySource {
    * execution is garbage-collected, and are never reused across query executions.
    */
   private readonly cardinalityCache = new WeakMap<object, Map<string, number>>();
+  /**
+   * Distinct value counts already determined during a query execution, scoped like the cardinality cache.
+   */
+  private readonly distinctValuesCache = new WeakMap<object, Map<string, number>>();
 
   public constructor(
     source: RDF.Source | RDF.DatasetCore,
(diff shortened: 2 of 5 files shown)


rubensworks:
The CI benchmark table above reads as if this change regresses 9 of the 12 BSBM-File queries. It does not, and there is a control available: #1804 is a SPARQL endpoint IPC fix that cannot touch query planning or execution, and it ran against the same base commit `258c64e`.

Totals, both PRs versus that same base:

| suite | base | #1804 (no-op for execution) | #1805 (this PR) |
| --- | --- | --- | --- |
| WatDiv-File | 2242 ms | 2305 ms (1.03) | 1824 ms (0.81) |
| WatDiv-TPF | 24598 ms | 22263 ms (0.91) | 23164 ms (0.94) |
| BSBM-File | 222 ms | 271 ms (1.22) | 194 ms (0.87) |
| BSBM-TPF | 1590 ms | 1649 ms (1.04) | 1765 ms (1.11) |

Two suites here are inert by construction. TPF sources do not implement `countDistinctTerms`, so `distinctValues` is never set on their metadata and `getSharedVariableJoinCardinality` takes exactly the path it takes on master. Yet WatDiv-TPF reads 0.94 and BSBM-TPF reads 1.11 on this PR, in opposite directions, and the no-op PR lands at 0.91 and 1.04 on the same two suites. That spread is drift between single CI runs on different runners, not signal.

WatDiv-File is the same story for a different reason: I compared physical plans for all 100 benchmark queries under both builds, interleaving the two builds three times each so that run-to-run instability separates from build effects, and every plan is identical (at scale 50, 97 identical and 3 differing only in a printed cardinality estimate). A suite whose plans do not change cannot legitimately read 0.81, and the no-op control read 1.03 on it.

The one number that actually separates the two PRs is BSBM-File Q5:

| | base | #1804 (no-op) | #1805 (this PR) |
| --- | --- | --- | --- |
| BSBM-File Q5 | 115 ms | 146 ms | 52 ms |

That is the query this change targets, and it is the only per-query result that moves against the drift rather than with it. It matches what I measured locally over three interleaved rounds at `-pc 1000`: Q5 -70.4%, whole mix -31.6%.

One thing I do not want to paper over: summing the non-Q5 BSBM-File queries gives 107 ms on base, 125 ms on the no-op control, and 142 ms on this PR. So this PR sits about 14% above the control there. With one run per PR and per-query values between 2 ms and 28 ms I cannot call that either way, and my repeated local runs show no consistent direction on those queries (Q1 +19.5%, Q2 +1.1%, Q3 -1.7%, Q4 -11.4%, Q10 +8.5% across three interleaved rounds). The measured cost of the added work is 9 index lookups per query totalling 0.1-0.5 ms, against 19135 cache hits, so I would not expect a systematic effect at that size, but it is a single CI run and worth a second look if anyone sees it persist.

Caveat on the control itself: #1804 is a different commit on a different runner, and the benchmarks do run through the endpoint it touches. It cannot change which plan is chosen or how a join executes, which is what matters here, but it is not a same-machine A/B.

---

rubensworks:
`Jest tests (macos-latest, 24.x)` is failing on `60b4a41`, twice, and it is not this PR's failure. Recording the evidence rather than leaving it red without explanation.

Both failures are a jest worker dying, not a test failing:

```
● Test suite failed to run
  A jest worker process (pid=6381) was terminated by another process:
  signal=SIGSEGV, exitCode=null.
```

The coverage-threshold errors reported alongside it are downstream of that: the killed worker's suite contributes no coverage, which drags the global figure just under 100%.

What rules this PR out as the cause:

- **The same commit is simultaneously green.** The push-triggered run on `60b4a41` ([34485468875](https://github.com/comunica/comunica/actions/runs/34485468875)) passed in full, including `Jest tests (macos-latest, 24.x)`. The pull_request-triggered run on the identical SHA ([34485478400](https://github.com/comunica/comunica/actions/runs/34485478400)) failed. Identical code, opposite outcomes.
- **A different suite dies each time.** First attempt: `actor-query-operation-update-drop`. Re-run: `actor-function-factory-term-str-langdir`. Neither has anything to do with this diff, and a code-caused failure would not wander between unrelated packages. Jest runs suites across parallel workers, so a crashing worker takes down whichever suite it happened to hold.
- **Only that one job is affected.** On the same commit, `Jest tests` passed on ubuntu 22/24/26, windows 22/24/26, macos-latest 22.x and macos-latest 26.x.
- **The full suite passes locally** on `60b4a41`: 482 suites, 6865 tests, no coverage threshold violation. CI counted 6857 in the first failure, short by exactly the 8 tests of the suite whose worker died.

This diff touches three files (`QuerySourceRdfJs.ts`, `ActorRdfJoin.ts`, `IMetadata.ts`) and none is reachable from either crashed suite. A SIGSEGV is a native process death rather than a failure mode that cardinality arithmetic produces.

I have used my one re-run on it, and it reproduced with the same signature, so there is no fix for me to port; the job would need re-running again or the macOS runner looking at. Everything else on the PR is green.

---

PR #1806, opened 2026-09-09
Compare two IRIs directly in orderTypes

rubensworks:
`TermComparatorExpressionEvaluator.orderTypes` transformed both terms into expression-evaluator terms and evaluated the SPARQL `<` operator twice, regardless of their type.

For two IRIs all of that reduces to comparing their values: the general path reaches them through the `term`/`term` overload, which orders differing term types by priority and otherwise compares `str()`. This adds a short-circuit for that case.

## Measurements

Over 19999 comparisons of IRIs: 0.605µs per comparison before, 0.027µs after, a 22x reduction. That also puts it below a hand-written comparison that reads `.value` off both terms through a closure.

The comparator sits on the path of `ORDER BY`, `MIN` and `MAX`, and of any join that has to order terms. It was the dominant cost in a merge join over sorted bindings: joining two sorted scans of 5006 and 6079 bindings took 13.2ms against a hash join's 4.8ms with the old comparator, and 3.6ms against 5.4ms with this one. Across five input shapes the same join went from 1.61x-2.76x slower than a hash join to 0.66x-0.98x.

## Scope

Only IRIs take the fast path. Literals deliberately keep the general path:

- `xsd:string` literals are ordered with `localeCompare`
- numerics are ordered by their typed value
- language strings are ordered by string, then language tag

None of those are reproduced by a plain value comparison.

## Testing

The general path is kept as a private `orderTypesGeneral` method so a test can assert that both paths agree. The added test does so over every ordered pair of a 25-term corpus covering IRIs, blank nodes, plain and typed literals, language strings, quoted triples and the default graph. The file is at 100% coverage and the full suite passes.

https://claude.ai/code/session_016PucrLg3zGkxDXsQfoVL2w

---
_Generated by [Claude Code](https://claude.ai/code/session_016PucrLg3zGkxDXsQfoVL2w)_

### DIFF
--- a/packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
+++ b/packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
@@ -41,9 +41,6 @@ export class TermFunctionLesserThan extends TermFunctionBase {
    *
    * SPARQL specifies that blankNode < namedNode < literal. Sparql star expands with < quads and we say
    * < defaultGraph: https://www.w3.org/TR/sparql11-query/#modOrderBy
-   *
-   * Public so that a comparator that short-circuits this operator orders differing types by the same
-   * table rather than by its own copy of it.
    */
   public static readonly TERM_ORDERING_PRIORITY = {
     blankNode: 0,
--- a/packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
+++ b/packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
@@ -36,6 +36,23 @@ import type {
 type Tuple<T> = readonly [T, T];
 
 export class TermFunctionLesserThan extends TermFunctionBase {
+  /**
+   * The order this operator puts terms of different types in.
+   *
+   * SPARQL specifies that blankNode < namedNode < literal. Sparql star expands with < quads and we say
+   * < defaultGraph: https://www.w3.org/TR/sparql11-query/#modOrderBy
+   *
+   * Public so that a comparator that short-circuits this operator orders differing types by the same
+   * table rather than by its own copy of it.
+   */
+  public static readonly TERM_ORDERING_PRIORITY = {
+    blankNode: 0,
+    namedNode: 1,
+    literal: 2,
+    quad: 3,
+    defaultGraph: 4,
+  };
+
   // TODO: remove in next major, as it's unused
   public constructor(private readonly equalityFunction: ITermFunction) {
     super({
@@ -199,7 +216,8 @@ To enable comparison, set the ${KeysExpressionEvaluator.fullTermComparison.name}
 
     // Order terms with different types according to a priority mapping
     if (termA.termType !== termB.termType) {
-      return this._TERM_ORDERING_PRIORITY[termA.termType] < this._TERM_ORDERING_PRIORITY[termB.termType];
+      return TermFunctionLesserThan.TERM_ORDERING_PRIORITY[termA.termType] <
+        TermFunctionLesserThan.TERM_ORDERING_PRIORITY[termB.termType];
     }
 
     // Comparison of literals with different or unknown data types: handle non-lexicals and first check dataType
@@ -230,14 +248,4 @@ To enable comparison, set the ${KeysExpressionEvaluator.fullTermComparison.name}
   private comparePrimitives<T>(valueA: T, valueB: T): -1 | 0 | 1 {
     return valueA === valueB ? 0 : (valueA < valueB ? -1 : 1);
   }
-
-  // SPARQL specifies that blankNode < namedNode < literal. Sparql star expands with < quads and we say < defaultGraph:
-  // https://www.w3.org/TR/sparql11-query/#modOrderBy
-  private readonly _TERM_ORDERING_PRIORITY = {
-    blankNode: 0,
-    namedNode: 1,
-    literal: 2,
-    quad: 3,
-    defaultGraph: 4,
-  };
 }
--- a/packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts
+++ b/packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts
@@ -1,8 +1,27 @@
 import type { InternalEvaluator } from '@comunica/actor-expression-evaluator-factory-default';
+import { TermFunctionLesserThan } from '@comunica/actor-function-factory-term-lesser-than';
 import type { ITermFunction } from '@comunica/bus-function-factory';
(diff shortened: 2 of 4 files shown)


packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts
```diff
@@ -29,6 +29,26 @@ export class TermComparatorExpressionEvaluator implements ITermComparator {
       return 1;
     }
 
+    // Two IRIs are ordered by their value, which is exactly what the general path below computes for them,
+    // but reaching that conclusion there costs transforming both terms and evaluating the SPARQL `<`
+    // operator twice. Terms of any other type keep the general path: notably `xsd:string` literals are
+    // compared with `localeCompare`, which a value comparison would not reproduce.
```
rubensworks:
```suggestion
    // Fast path for the common case of IRI comparison.
```

packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts
```diff
@@ -29,6 +29,23 @@ export class TermComparatorExpressionEvaluator implements ITermComparator {
...
+    // Fast path for the common case of IRI comparison.
+    if (termA.termType === 'NamedNode' && termB.termType === 'NamedNode') {
+      if (termA.value === termB.value) {
+        return 0;
+      }
+      return termA.value < termB.value ? -1 : 1;
+    }
+
+    return this.orderTypesGeneral(termA, termB);
+  }
```
jitsedesmet: This introduces a replication of code which could provide issues with keeping the code consistent. Since the speedup is significant however, I would still accept it, but let's also add code links that 'harden' the connected nature of `<` and this thourgh code.

Is suggest we add a test here that checks whether the termTypes are different.
If yes, we access this order object instead: https://github.com/comunica/comunica/blob/7a43c0e008487b6eb3f50c20d1a856fe2c8dd7f5/packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts#L236
(We should make it public static), and order as such.

If not (we know same termType) we can check if termA is either a blankNode or iri, in check case we string-compare. Otherwise, we pipe to the expression evaluator like we do now.

jitsedesmet (changes requested):

packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts
rubensworks: Done in 8544bc6, structured as you suggested.

`_TERM_ORDERING_PRIORITY` is now `public static readonly TERM_ORDERING_PRIORITY`, and `orderTypes` splits three ways:

- **Differing term types** are ordered by that table. Since the table is keyed by the evaluator's names (`namedNode`) rather than RDF/JS's (`NamedNode`), there is a small `EVALUATOR_TERM_TYPES` map to look a term up in it.
- **Same type, IRI or blank node** compares values directly.
- **Anything else** keeps the operator, `xsd:string` above all, which is ordered with `localeCompare`.

Two things worth flagging:

The differing-types branch skips `fullTermComparisonCheck` and `nonLexicalCheck`, which the operator runs before it reaches the priority table. That is safe only because `ActorTermComparatorFactoryExpressionEvaluator` sets both `fullTermComparison` and `nonLexicalComparison` to true, so neither can throw for this comparator. If that ever stops being true, this branch changes behaviour rather than just speeding it up.

A variable has no entry in the table, so it falls through to the operator rather than being guessed at. There is a test pinning that against the general path.

On hardening the connection: the all-pairs differential test now also asserts that each of the three short-circuits is actually reached by the corpus, so the agreement cannot quietly go vacuous if the corpus changes, and that every term type in the corpus is one the table names. Two further tests pin differing-type ordering to the table itself and blank-node ordering to the label.

Measured over 19999 comparisons, general path against short-circuit: two IRIs 3.160µs → 0.042µs, two blank nodes 3.072µs → 0.018µs, differing types 4.856µs → 0.037µs.

---

jitsedesmet: The error throwing discrepancy was also there in your original suggested code. Also know that variables will never enter here within a complete system. That being said, we should indeed handle it here because the package can be used outside of a complete engine.

jitsedesmet (approved):

packages/actor-function-factory-term-lesser-than/lib/TermFunctionLesserThan.ts
```diff
@@ -36,6 +36,23 @@ import type {
...
 
 export class TermFunctionLesserThan extends TermFunctionBase {
+  /**
+   * The order this operator puts terms of different types in.
+   *
+   * SPARQL specifies that blankNode < namedNode < literal. Sparql star expands with < quads and we say
+   * < defaultGraph: https://www.w3.org/TR/sparql11-query/#modOrderBy
+   *
+   * Public so that a comparator that short-circuits this operator orders differing types by the same
+   * table rather than by its own copy of it.
```
rubensworks:
```suggestion
```

packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts
```diff
@@ -29,6 +48,39 @@ export class TermComparatorExpressionEvaluator implements ITermComparator {
       return 1;
     }
 
+    if (termA.termType === termB.termType) {
+      // Two IRIs, or two blank nodes, are ordered by their value, which is exactly what the general path
+      // below computes for them, at the cost of transforming both terms and evaluating `<` twice. Every
+      // other type keeps that path: notably `xsd:string` literals are compared with `localeCompare`,
+      // which a value comparison would not reproduce.
```
rubensworks:
```suggestion
      // Fast path to order two IRIs or two blank nodes by their value.
```

packages/actor-term-comparator-factory-expression-evaluator/lib/TermComparatorExpressionEvaluator.ts
```diff
@@ -29,6 +48,39 @@ export class TermComparatorExpressionEvaluator implements ITermComparator {
...
+      if (termA.termType === 'NamedNode' || termA.termType === 'BlankNode') {
+        if (termA.value === termB.value) {
+          return 0;
+        }
+        return termA.value < termB.value ? -1 : 1;
+      }
+    } else {
+      // Terms of differing types are ordered by type alone, by the same table the `<` implementation
+      // orders them with, so that the two cannot drift apart.
```
rubensworks:
```suggestion
      // Fast path to order terms of differing types.
```

PR #1808, opened 2026-09-11
Disallow SERVICE clauses from targeting local files by default

rubensworks:
Before this change, any engine that can dereference local files would happily resolve a `SERVICE <file:///...>` target from the local file system. For `comunica-sparql-file` (the only engine bundling `@comunica/actor-dereference-file`), and especially for `comunica-sparql-file-http`, this means that a query from an untrusted party can read arbitrary local files:

```sparql
SELECT * WHERE { SERVICE <file:///etc/some-secret.ttl> { ?s ?p ?o } }
```

## Approach

Rather than pattern-matching on the SERVICE IRI, the restriction is enforced at the place that actually grants local file access:

* `ActorOptimizeQueryOperationQuerySourceIdentify` now identifies SERVICE targets with a source context in which the new `@comunica/bus-dereference:blockFileAccess` entry is enabled. This context follows the source into query execution, so it also covers any links that are traversed from the SERVICE target.
* `ActorDereferenceFile` refuses to dereference within such a scope, via `handleDereferenceErrors`, so `SERVICE SILENT` and lenient mode keep their usual semantics (a warning and an empty result, instead of a hard error).
* The new `serviceAllowFileTargets` context entry (`@comunica/actor-init-query:serviceAllowFileTargets`) turns it back on for those who need it:

  ```bash
  $ comunica-sparql-file file.ttl -c '{ "serviceAllowFileTargets": true }' "SELECT * WHERE { ... }"
  ```

Engines without a file dereference actor are unaffected, as blocking only applies to local file dereferencing.

Since SERVICE sources are now identified with a different source context than regular sources, they are cached under a separate cache key, so that a blocked SERVICE source can never be reused for a source that the user explicitly passed (and vice versa). Cache invalidation covers both keys.

## Notes

* This is a behavioural break for anyone relying on `SERVICE <file://...>` in `comunica-sparql-file`.
* `LOAD <file:///...>` is a comparable vector, but it is only reachable when updates are explicitly enabled (`comunica-sparql-file-http -u`), so it is left out of scope here.

## Tests

* Unit tests for the blocked and lenient paths in `ActorDereferenceFile`, and for the SERVICE source context, the `serviceAllowFileTargets` entry, and the separate cache keys in `ActorOptimizeQueryOperationQuerySourceIdentify`.
* System tests in `query-sparql-file` covering the denied, silent, and explicitly allowed cases.
* Full test suite passes (6879 tests), both changed source files remain at 100% coverage, and lint is clean.

https://claude.ai/code/session_019SCVUhg7mCAq7CPZZxuCvQ

---
_Generated by [Claude Code](https://claude.ai/code/session_019SCVUhg7mCAq7CPZZxuCvQ)_

### DIFF
--- a/packages/actor-dereference-file/lib/ActorDereferenceFile.ts
+++ b/packages/actor-dereference-file/lib/ActorDereferenceFile.ts
@@ -35,8 +35,7 @@ export class ActorDereferenceFile extends ActorDereference {
     // Dereferencing local files can be blocked within certain scopes, such as SERVICE targets.
     if (context.get(KeysDereference.blockFileAccess)) {
       return this.handleDereferenceErrors(action, new Error(
-        `Dereferencing the local file '${url}' is not allowed within this scope. ` +
-        `Within SERVICE clauses, this can be allowed via the 'serviceAllowFileTargets' context entry.`,
+        `Dereferencing the local file '${url}' is not allowed within this scope. `,
       ));
     }
 
--- a/packages/actor-dereference-file/lib/ActorDereferenceFile.ts
+++ b/packages/actor-dereference-file/lib/ActorDereferenceFile.ts
@@ -2,7 +2,7 @@ import { accessSync, createReadStream, constants } from 'node:fs';
 import { fileURLToPath, pathToFileURL } from 'node:url';
 import type { IActionDereference, IActorDereferenceArgs, IActorDereferenceOutput } from '@comunica/bus-dereference';
 import { ActorDereference } from '@comunica/bus-dereference';
-import { KeysInitQuery } from '@comunica/context-entries';
+import { KeysDereference, KeysInitQuery } from '@comunica/context-entries';
 import type { IActorTest, TestResult } from '@comunica/core';
 import { failTest, passTestVoid } from '@comunica/core';
 
@@ -29,7 +29,17 @@ export class ActorDereferenceFile extends ActorDereference {
     return URIRegex.exec(str) !== null;
   }
 
-  public async run({ url, context }: IActionDereference): Promise<IActorDereferenceOutput> {
+  public async run(action: IActionDereference): Promise<IActorDereferenceOutput> {
+    const { url, context } = action;
+
+    // Dereferencing local files can be blocked within certain scopes, such as SERVICE targets.
+    if (context.get(KeysDereference.blockFileAccess)) {
+      return this.handleDereferenceErrors(action, new Error(
+        `Dereferencing the local file '${url}' is not allowed within this scope. ` +
+        `Within SERVICE clauses, this can be allowed via the 'serviceAllowFileTargets' context entry.`,
+      ));
+    }
+
     const requestTimeStart = Date.now();
     return {
       data: createReadStream(getPath(url)),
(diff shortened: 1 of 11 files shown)


packages/actor-dereference-file/lib/ActorDereferenceFile.ts
```diff
@@ -29,7 +29,17 @@ export class ActorDereferenceFile extends ActorDereference {
     return URIRegex.exec(str) !== null;
   }
 
-  public async run({ url, context }: IActionDereference): Promise<IActorDereferenceOutput> {
+  public async run(action: IActionDereference): Promise<IActorDereferenceOutput> {
+    const { url, context } = action;
+
+    // Dereferencing local files can be blocked within certain scopes, such as SERVICE targets.
+    if (context.get(KeysDereference.blockFileAccess)) {
```
rubensworks, line 36: Are you sure this doesn't block too much? We still want comunica-sparql-file from the CLI to still be able to query `file@` sources when that source is specified from the CLI as CLI argument.

rubensworks: Good question, but no: sources passed via the CLI (or the `sources` context entry) are unaffected.

The `blockFileAccess` entry is never set on the query context. It is only set on the *source context* of sources identified as SERVICE targets, and that source context is merged in by `ActorQueryOperationSource` solely for operations assigned to that one source. Since `assignExhaustive` assigns the operations inside a SERVICE clause exclusively to the SERVICE source, the entry never reaches a source the user passed in.

The one place the two could have collided was the source cache in `ActorOptimizeQueryOperationQuerySourceIdentify`, which is keyed by URL and shared across query executions: a blocked SERVICE source could otherwise have been handed back for the same file when passed as a query source later on. That is why SERVICE targets are cached under a separate `service:` key prefix (with invalidation covering both keys).

I added a system test in 1474644 for exactly this case, where the file source keeps returning its bindings in the same query in which a SERVICE file target is blocked. Verified against the real CLI too:

```console
$ node engines/query-sparql-file/bin/query.js engines/query-sparql-file/test/assets/dummy.ttl "SELECT * WHERE { ?s ?p ?o }"
[ ... 5 results ... ]
$ node engines/query-sparql-file/bin/query.js "file:///.../dummy.ttl" "SELECT * WHERE { ?s ?p ?o } LIMIT 2"
[ ... 2 results ... ]
```

Both still work, with and without the `file://` scheme.

---

packages/actor-dereference-file/lib/ActorDereferenceFile.ts
```diff
@@ -29,7 +29,17 @@ export class ActorDereferenceFile extends ActorDereference {
...
+      return this.handleDereferenceErrors(action, new Error(
+        `Dereferencing the local file '${url}' is not allowed within this scope. ` +
+        `Within SERVICE clauses, this can be allowed via the 'serviceAllowFileTargets' context entry.`,
+      ));
```
rubensworks, line 39:
```suggestion
      return this.handleDereferenceErrors(action, new Error(
        `Dereferencing the local file '${url}' is not allowed within this scope. `,
      ));
```

PR #1814, opened 2026-09-11
Add a standalone SPARQL expression evaluation engine

### LINKED ISSUE #1809: Standalone Expression Evaluator unavailable
#### Issue type:

- :bug: Bug 

____
#### Description:

The content of package [utils-expression-evaluator](https://github.com/comunica/comunica/tree/master/packages/utils-expression-evaluator) does not reflect its description (README).
Specifically, the mentioned classes (e.g., evaluator classes) are not available.
A part from the issue of the outdated documentation, it seems that is now not possible to use the evaluator outside of the Comunica framework, for example to do synchronous evaluation.
This would be a pity, especially considering that the utils-expression-evaluator package came from the now archived independent module [sparqlee](https://github.com/comunica/sparqlee).

____

jitsedesmet:
Aims to close https://github.com/comunica/comunica/issues/1809

The utils-expression-evaluator README documented evaluator classes that no longer exist, and evaluating expressions outside of a query engine was not possible without wiring up every function factory actor by hand.

Add @comunica/expressions-sparql, an engine that reuses the config-query-sparql configuration, importing only the parts expression evaluation needs. It exposes an ExpressionEngine that creates expression evaluators, term comparators and bindings aggregators, and that defaults the context entries evaluation requires.

Update the expression evaluation documentation to match the implementation, and fix the npm badges and install instructions that still used the package names from before the rename.


packages/actor-term-comparator-factory-expression-evaluator/README.md
```diff
@@ -1,9 +1,9 @@
-# Comunica Inequality Functions Based Term Comparator Factory Actor
+# Comunica Expression Evaluator Term Comparator Factory Actor
 
 [![npm version](https://badge.fury.io/js/%40comunica%2Factor-term-comparator-factory-expression-evaluator.svg)](https://www.npmjs.com/package/@comunica/actor-term-comparator-factory-expression-evaluator)
 
-A [term comparator factory](https://github.com/comunica/comunica/tree/master/packages/bus-rdf-update-quads) actor
-that uses the evaluation of the inequality functions `<` and `=` of the expression evaluator.
+A [term comparator factory](https://github.com/comunica/comunica/tree/master/packages/bus-term-comparator-factory) actor
+that orders terms by evaluating the `<` and `=` functions of the expression evaluator.
```
jitsedesmet: I do not think this is right since we no longer depend on the `=` function?

engines/config-query-sparql/config/config-expressions.json
```diff
@@ -0,0 +1,18 @@
+{
```
rubensworks, line 1: Shall we add a dedicated config package for this?

engines/expressions-sparql/lib/ExpressionEngine.ts
```diff
@@ -0,0 +1,80 @@
...
+ * This engine configures no query operations, so it cannot evaluate the sub-query of an EXISTS itself.
+ */
+const unsupportedExistence: ExistenceResolver = () => {
+  throw new Error(`Evaluating EXISTS requires a ${KeysExpressionEvaluator.existenceResolver.name} in the context`);
+};
+
+/**
+ * A Comunica engine for evaluating SPARQL expressions, without executing queries.
+ */
+export class ExpressionEngine {
```
rubensworks, line 31: I highly recommend making an `IExpressionEngine` interface, within the types package, and using that interface whenever possible.

engines/expressions-sparql/lib/ActorInitExpressions.ts
```diff
@@ -0,0 +1,47 @@
...
+import { ActorInit } from '@comunica/bus-init';
+import type { MediatorTermComparatorFactory } from '@comunica/bus-term-comparator-factory';
+import type { IActorTest, TestResult } from '@comunica/core';
+import { passTestVoid } from '@comunica/core';
+
+/**
+ * The entrypoint of a compiled expression evaluation config.
+ * It only holds the mediators needed for expression evaluation, and is wrapped by an ExpressionEngine.
+ */
+export class ActorInitExpressions extends ActorInit {
```
rubensworks, line 13: I would create an actor-init-expressions actor package for this, similar to actor-init-query. Useful for when we'd want variants of `@comunica/expressions-sparql`.

engines/expressions-sparql/README.md
```diff
@@ -0,0 +1,125 @@
...
+// Parse a query, and take the expression of its FILTER clause
+const query: any = toAlgebra(new Parser().parse(`
+  SELECT * WHERE { ?s ?p ?o FILTER(langMatches(lang(?o), "FR")) }
+`));
+const evaluator = await engine.createEvaluator(query.input.expression);
+
+// Evaluate bindings as a term, ...
+const term = await evaluator.evaluate(BF.fromRecord({ o: DF.literal('Ceci n\'est pas une pipe', 'fr') }));
+// ... or as an effective boolean value, as FILTER does
+const ebv = await evaluator.evaluateAsEBV(BF.fromRecord({ o: DF.literal('This is not a pipe', 'en') }));
```
jitsedesmet, line 46: Beside this example I would also add an example where the user parses an expression directly. The Traqula parser allows any access point for parsing, so that would be nice (although we'd need to check whether algebra transformation also allows this.)

report.md
```diff
@@ -0,0 +1,14 @@
+# Review findings left open
```
rubensworks: This should probably not be here.

engines/expressions-sparql/README.md
```diff
@@ -0,0 +1,152 @@
+# Comunica SPARQL Expressions
+
+[![npm version](https://badge.fury.io/js/%40comunica%2Fexpressions-sparql.svg)](https://www.npmjs.com/package/@comunica/expressions-sparql)
+
+Comunica SPARQL Expressions evaluates SPARQL expressions, aggregates and term ordering
+without setting up a query engine.
```
rubensworks, line 6: Just like we do for the lite query sparql engine, I would mention bundle size here somewhere as well (as it's probably going to be quite small, compared to query-sparql).

engines/expressions-sparql/README.md
```diff
@@ -0,0 +1,152 @@
...
+import { BindingsFactory } from '@comunica/utils-bindings-factory';
+import { toAlgebra } from '@traqula/algebra-sparql-1-2';
+import { Parser } from '@traqula/parser-sparql-1-2';
+import { DataFactory } from 'rdf-data-factory';
+
+const DF = new DataFactory();
+const BF = new BindingsFactory(DF);
+const engine = new ExpressionEngine();
+
+// Parse a query, and take the expression of its FILTER clause
```
rubensworks: It may also be valuable to (also) give a short example on how traqula's algebra factory can be used to manually construct a simple expression (in case people don't want to parse a query).

packages/actor-init-expressions/lib/ExpressionEngineBase.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  public async createTermComparator(context?: IActionContext): Promise<ITermComparator> {
+    return this.actorInitExpressions.mediatorTermComparatorFactory.mediate({ context: this.prepareContext(context) });
+  }
+
+  public async createAggregator(expression: Algebra.AggregateExpression, context?: IActionContext):
+  Promise<IBindingsAggregator> {
+    return this.actorInitExpressions.mediatorBindingsAggregatorFactory
+      .mediate({ expr: expression, context: this.prepareContext(context) });
+  }
```
rubensworks, line 42: Just wanted to say that this is really nice!
I'm sure many people will benefit from this functionality.

packages/utils-monorepo/lib/depCheck.ts
```diff
@@ -64,6 +64,17 @@ async function depInfo(pckg: any): Promise<any> {
       dependency: configPackage,
       dependant: join(pckg.dir, 'engine-default.js'),
     });
+    // An engine can also build upon other config packages, which are only referenced by its config
```
rubensworks: Are you sure this additional check is needed?
Wouldn't this be handled already through the compiled js config file?

rubensworks (changes requested):
Very nice! No major comments.

packages/utils-monorepo/lib/depCheck.ts
jitsedesmet: According to claude (sorry):

> It is needed, and I confirmed it empirically. The compiled engine-default.js contains no reference to any config package at all — it requires only actor and bus packages. With the new block disabled, depcheck reports @comunica/config-expressions-sparql as an extra dependency of engines/expressions-sparql. That's the same reason the pre-existing hard-coded ensureDependency(configPackage) line exists; it just only covers @comunica/config-query-sparql, which is what yarn depcheck passes as argv[2]. The expressions engine is the first engine built on a different config package, so it needed the generic path. I kept the code and rewrote the comment to say all of this, since the old one-liner didn't explain why the compiled config doesn't cover it. Full depcheckTask run passes.

jitsedesmet: I looked into it some more, and it did seem like it was needed. Prompting a bit, I found a new solution that might  be cleaner. Please have a look at that one. It relies on the package naming scheme but I think that is okay since Comunica does have a strict naming scheme.

jitsedesmet:
@rubensworks Looking at the bundle size, half of it is Traqula.
I would accept the PR here as is, but we should really look at exposing a types package for Traqula since it currently depends on `@comunica/utils-algebra` which depends on Traqula (for the types) and therefore on Chevrotain. When Traqula exposes a types package, algebra utils could depend on that and therefore no longer depend on Chevrorain.  

image


For now, I investigated, and it does not seem like there is anything I can do at this point. Tree shaking would help but it seems to be limited by CJS. ESM should help us in that regards it seems.

jitsedesmet:
@rubensworks this PR is ready for review, note that the CI fails without pinning undici. Since this is a master issue not related to this PR, I have reverted that change, but as you can see, CI would pass when pinning the version (second to last commit is green)

jitsedesmet:
I also added webpack bundling :)

rubensworks:
> @rubensworks this PR is ready for review, note that the CI fails without pinning undici. Since this is a master issue not related to this PR, I have reverted that change, but as you can see, CI would pass when pinning the version (second to last commit is green)

CI on master is green though. Can you investigate why it breaks in this PR?

jitsedesmet:
@rubensworks The breaking release of that package has been made after the CI on master last run. The release was this morning https://www.npmjs.com/package/undici

rubensworks:
@jitsedesmet Is it an intentional change on their end? Has an issue already been opened by other consumers that experience such a breaking change? If not, could you open one? (if you believe it's a problem on their end)

jitsedesmet:
@rubensworks They are fixing it: https://github.com/nodejs/undici/pull/5860 it seems like that should be enough. Since they broke response headers all together:

>  lose response headers and stop following redirects

rubensworks:
Aha nice, thanks for checking.
I suspect they will soon release a patch.
Let's wait with merging this one until that is done, as to not make master red...

jitsedesmet:
@rubensworks green :partying_face:

rubensworks:
Hurray!

PR #1815, opened 2026-09-11
Support variables as the target of SERVICE clauses

### LINKED ISSUE #1384: Allow variables as the argument of SERVICE
#### Issue type:

____
#### Description:

I would like to use the SERVICE clause to perform a kind of "user-driven link traversal" (since Comunica can deal with simple RDF resources in the SERVICE clause, not just SPARQL endpoints).

This would look like that:
```
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
PREFIX foaf: <http://xmlns.com/foaf/0.1/>
SELECT *
WHERE {
  ?person a foaf:Person; rdfs:seeAlso ?sa.
  
  SERVICE ?sa {
    ?person foaf:knows ?other.
  }
}
LIMIT 10
```
Unfortunately, currently Comunica does not accept a variable after the SERVICE keyword. I wish it did...

rubensworks:
Enables the `service5` SPARQL 1.1 Federated Query spec test and removes the last `--skip` from `spec:fed-1-1`. Contributes to https://github.com/comunica/comunica/issues/721 and closes https://github.com/comunica/comunica/issues/1384.

Follow-up to https://github.com/comunica/comunica/pull/1779, which handled `service3`, `service6` and `service7`.

## The problem

`assign-sources-exhaustive` rewrites SERVICE clauses into source-annotated operations, but it can only do so for an IRI target. With a variable target the optimizer has nothing to resolve, as the target is only known once the variable has been bound, so the clause is left as a `service` operation that no actor can execute:

```
none of the configured actors were able to handle the operation type service
```

This is the one remaining case; it has never been supported.

## Disabled by default

With a variable target, the queried data decides what gets dereferenced. For an engine answering queries from untrusted parties — a public endpoint over a fixed dataset, say — that would let a query reach arbitrary sources. Variable targets are therefore **disabled by default**, alongside the file-target restriction from https://github.com/comunica/comunica/pull/1808, and produce a clear error:

```
SERVICE clauses with a variable target are not allowed by default, as they let the queried data
determine what is dereferenced. Enable them via the serviceAllowVariableTargets context entry.
```

They are enabled through the `serviceAllowVariableTargets` context entry, or `--serviceAllowVariableTargets` on the command line. Clauses with an IRI target are unaffected. This deserves a mention on the website once merged.

## Changes

The new `@comunica/actor-query-operation-service` handles these clauses at execution time.

Without a bound target it returns a placeholder output whose metadata sets `operationRequired`, the same concept `IJoinEntry.operationRequired` already expresses for join entries. `ActorRdfJoin` reads it while resolving entry metadata it needs anyway, so recognising such an entry is a property read rather than an operation traversal. Because single-input operations pass metadata on — `ActorQueryOperationFilter` forwards `output.metadata` directly, project and slice spread it — the flag survives when the clause ends up wrapped.

The bind-join then materializes the target into an IRI, for which `materializeOperation` now also maps the target term of a SERVICE clause — it is otherwise skipped, since that term is not a pattern position. The actor identifies that IRI as a query source and executes the clause body against it; for `SERVICE SILENT` it falls back to a single empty solution, reusing the `KeysQueryOperation.silent` flag from #1779.

Following #1808, the actor identifies its target with `blockFileAccess` set unless `serviceAllowFileTargets` is given, so a variable target cannot be used to read arbitrary local files either.

The body is annotated with the source as a whole when the source's selector shape accepts it, and only at its leaves otherwise, so that targets which are plain RDF documents rather than SPARQL endpoints work too.

### Query planning

Deferring these clauses to a bind-join only worked when the clause directly followed the operation binding its target. Two parts of planning got in the way, both found by the new system tests:

* **Bind-join entry order.** `ActorRdfJoinMultiBind` sorts its entries and binds the others with the first one. That sorting is by selectivity, so an entry requiring a bind-join could land first, where by definition it cannot be bound. It previously rejected the join in that case, leaving no usable actor. Such entries are now moved out of the first position instead, leaving the order untouched when the first entry does not require pushdown, and the join is only rejected when *every* entry requires it.
* **Filter pushdown.** `ActorOptimizeQueryOperationFilterPushdown` could wrap such an entry in a filter, and `ActorRdfJoinMultiBind.canBindWithOperation` deliberately refuses to bind on any filter that is not a hoisted left-join filter — again leaving no usable actor. Filters are therefore no longer pushed into a join or union holding such an entry.

## Testing

Federated query suite goes from 9/9 with `service5` skipped to **10/10**. SPARQL 1.0 467/476, 1.1 query 326/328, update 157/157, csv/tsv 6/6, json 4/4, service description 3/3, 1.2 263/263 — all matching the baseline on `88ed6e9`. 689 unit tests pass across the touched packages, with full coverage of the new actor, the planning changes and the SERVICE case in `materializeOperation`.

System tests in `engines/query-sparql/test/QuerySparql-service-variable-test.ts` place the clause in 15 situations against two mocked endpoints, asserting both the results and which endpoints were queried:

| | |
|---|---|
| after the pattern binding its target | ✔ |
| **before** the pattern binding its target | ✔ |
| in a group before the pattern binding its target | ✔ |
| within `OPTIONAL` | ✔ |
| within a `UNION` branch | ✔ |
| target bound by `VALUES`, before and after | ✔ |
| target bound by `BIND` | ✔ |
| next to a `FILTER` over its results | ✔ |
| two clauses with a variable target | ✔ |
| next to a clause with an IRI target | ✔ |
| without `serviceAllowVariableTargets` | errors, no endpoint contacted |
| IRI target without `serviceAllowVariableTargets` | ✔ |
| nothing binds the target, plain and `SILENT` | errors |

Five of these fail without the two planning changes above.

The query from https://github.com/comunica/comunica/issues/1384 was also verified against plain Turtle documents:

```sparql
SELECT * WHERE {
  ?person a foaf:Person ; rdfs:seeAlso ?sa .
  SERVICE ?sa { ?person foaf:knows ?other }
} LIMIT 10
```
```
{ "person": "http://ex.org/alice", "sa": "http://docs.ex.org/a.ttl", "other": "http://ex.org/carol" }
{ "person": "http://ex.org/bob",   "sa": "http://docs.ex.org/b.ttl", "other": "http://ex.org/dave" }
```

A variable target bound to a `file://` URL is rejected by default.

## Notes for review

- The spec test harness enables `serviceAllowVariableTargets`, since the suite covers `service5`.
- The target must be bound by something joined with the clause. When nothing binds it the query errors rather than silently returning nothing, which matches the spec requiring the variable to be in scope. This also holds for `SERVICE SILENT`: an unbound target, like a disabled feature, is a problem with the query rather than a failure of the service, so neither is swallowed.
- Not pushing filters into joins holding such an entry costs that optimisation for those queries. Relaxing `canBindWithOperation` to accept filters would avoid it, but that looks deliberate and affects every query, so I left it alone.
- A target bound to a literal or blank node fails `testOperation` with "can only query services by IRI or variable". That produces a clear error, but a silent clause arguably ought to swallow it instead.
- The new actor is registered in `query-operation/actors-v5-3-0.json`, joining the versioned config generation introduced by https://github.com/comunica/comunica/pull/1778. The `query-sparql` and `query-sparql-file` engines resolve through it; the rdfjs engines still resolve through the v5-1-3 chain and so do not get variable SERVICE support.

https://claude.ai/code/session_01X4QghLG8yMWpYmoYHdckv9


packages/actor-query-operation-service/README.md
```diff
@@ -0,0 +1,55 @@
...
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-query-operation-service.svg)](https://www.npmjs.com/package/@comunica/actor-query-operation-service)
+
+A [Query Operation](https://github.com/comunica/comunica/tree/master/packages/bus-query-operation) actor
+that handles SPARQL `SERVICE` operations that survive the optimization phase.
+
+Most `SERVICE` clauses are rewritten into source-annotated operations by
+[`@comunica/actor-optimize-query-operation-assign-sources-exhaustive`](https://github.com/comunica/comunica/tree/master/packages/actor-optimize-query-operation-assign-sources-exhaustive).
+This actor handles the remaining cases:
+
+* `SERVICE SILENT`, where errors must be swallowed and replaced by a single empty solution,
```
rubensworks: Is this line still accurate?

packages/bus-rdf-join/lib/ActorRdfJoin.ts
```diff
@@ -104,6 +105,32 @@ TS
...
+   * Next to entries that explicitly request this, this holds for SERVICE clauses with a variable target,
+   * as those can only be evaluated once that target has been bound by another entry.
+   * @param entry A join entry.
+   */
+  public static isOperationRequired(entry: IJoinEntry): boolean {
+    if (entry.operationRequired) {
+      return true;
+    }
+    let unboundServiceTarget = false;
+    algebraUtils.visitOperation(entry.operation, {
```
rubensworks: This seems quite expensive to do on every join. Is it possible to somehow incorporate this into an existing or a new optimize actor instead? Or maybe elsewhere?

packages/actor-query-operation-service/README.md
rubensworks: No, good catch — that bullet was accurate when this actor also handled `SERVICE SILENT`, but #1779 moved silencing to `ActorQueryOperationSource` via `KeysQueryOperation.silent`, so this actor only deals with variable targets now. Rewritten:

> A Query Operation actor that handles SPARQL `SERVICE` clauses of which the target is a variable.
>
> `SERVICE` clauses with an IRI target are rewritten into source-annotated operations by `@comunica/actor-optimize-query-operation-assign-sources-exhaustive`. That is not possible for a variable target, as it is only known once the variable has been bound, so such clauses are handled at execution time by this actor instead. Their evaluation is deferred to a bind-join, which binds the target before this actor queries it.

---

packages/bus-rdf-join/lib/ActorRdfJoin.ts
rubensworks: Agreed, the traversal is gone. It is now an O(1) property read, with no optimize actor needed.

Rather than inspecting the operation, the actor announces it on the **metadata of its placeholder output**, which the join bus already resolves for every entry anyway:

```ts
// in ActorQueryOperationService.unboundPlaceholder
metadata: async() => ({
  state: new MetadataValidationState(),
  cardinality: { type: 'estimate', value: Number.POSITIVE_INFINITY },
  variables,
  // Join actors must bind our target before this operation can be evaluated.
  operationRequired: true,
}),
```
```ts
public static isOperationRequired(entry: IJoinEntry, metadata: MetadataBindings): boolean {
  return Boolean(entry.operationRequired ?? metadata.operationRequired);
}
```

I reused the `operationRequired` name so the entry flag and its metadata counterpart read as the same concept, and documented the new optional entry on `IMetadata` (which already extends `Record<string, any>`, so this is only about discoverability).

Two reasons this beats the alternatives I tried:

- An optimize actor would have to mark the operation *and every ancestor within the join entry*, and would have to run dead last, since later rewrites create new nodes without metadata — `createJoin`, `createFilter` and friends do not carry it over.
- I also considered just checking whether the entry's own operation is a variable-target SERVICE, which is O(1) too, but `filter-pushdown` can wrap an unhandled operand in a filter (`ActorOptimizeQueryOperationFilterPushdown.ts:465`), which would silently defeat it. Metadata does not have that problem: `ActorQueryOperationFilter` passes `output.metadata` straight through, and project/slice spread it, so the flag survives wrapping for free.

The check also moved below `getMetadatas`, so it runs after the entry-count and entry-type validation rather than before.

Verified: federated 10/10, other suites at baseline, 553 unit tests passing, and `bus-rdf-join` no longer needs the `utils-algebra` dependency I had added.

---

packages/actor-query-operation-service/README.md
```diff
@@ -0,0 +1,52 @@
...
+[![npm version](https://badge.fury.io/js/%40comunica%2Factor-query-operation-service.svg)](https://www.npmjs.com/package/@comunica/actor-query-operation-service)
+
+A [Query Operation](https://github.com/comunica/comunica/tree/master/packages/bus-query-operation) actor
+that handles SPARQL `SERVICE` clauses of which the target is a variable.
+
+`SERVICE` clauses with an IRI target are rewritten into source-annotated operations by
+[`@comunica/actor-optimize-query-operation-assign-sources-exhaustive`](https://github.com/comunica/comunica/tree/master/packages/actor-optimize-query-operation-assign-sources-exhaustive).
+That is not possible for a variable target, as it is only known once the variable has been bound,
+so such clauses are handled at execution time by this actor instead.
+Their evaluation is deferred to a bind-join, which binds the target before this actor queries it.
```
rubensworks:
```suggestion
```

rubensworks: Applied.

---

packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
```diff
@@ -277,6 +278,19 @@ export class ActorRdfJoinMultiBind extends ActorRdfJoin<IActorRdfJoinMultiBindTe
...
     }
     const entriesSorted = entriesTest.get();
+
+    // The first entry is the one whose bindings are pushed into the others,
+    // so entries that require their operation to be pushed down can not take that position.
+    // The sorting above does not take this into account, so we move such an entry out of it here.
+    const bindableIndex = entriesSorted
+      .findIndex(entry => !ActorRdfJoin.isOperationRequired(entry, entry.metadata));
+    if (bindableIndex < 0) {
+      return failTest(`Actor ${this.name} requires an entry of which the operation does not need to be pushed down`);
```
rubensworks:
```suggestion
      return failTest(`Actor ${this.name} requires at least one entry of which the operation does not need to be pushed down`);
```

packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
```diff
@@ -277,6 +278,19 @@ export class ActorRdfJoinMultiBind extends ActorRdfJoin<IActorRdfJoinMultiBindTe
       return entriesTest;
     }
     const entriesSorted = entriesTest.get();
+
+    // The first entry is the one whose bindings are pushed into the others,
+    // so entries that require their operation to be pushed down can not take that position.
+    // The sorting above does not take this into account, so we move such an entry out of it here.
```
rubensworks:
```suggestion
    // Find the first entry that does not require operation pushdown.
```

packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
```diff
@@ -277,6 +278,19 @@ export class ActorRdfJoinMultiBind extends ActorRdfJoin<IActorRdfJoinMultiBindTe
...
+    const bindableIndex = entriesSorted
+      .findIndex(entry => !ActorRdfJoin.isOperationRequired(entry, entry.metadata));
```
rubensworks, line 284: Are you sure this won't modify query plans that do not require modification?

For example, if entry 0 does not require operation, but entry 1 does, no changes are needed yet, right? (please check to confirm)

rubensworks: Confirmed — no change is made in that case, and I have added a test pinning it rather than leaving it to inspection.

`findIndex` returns `0` when entry 0 does not require pushdown, and the splice is guarded by `bindableIndex > 0`, so the sorted order is left exactly as the sorter produced it. Plans without any `operationRequired` entry therefore never reach the reordering at all.

The two new tests assert the resulting order in `sideData.entriesSorted` in both directions:

```ts
it('should move an entry with operationRequired out of the first position', …
  // requiredIndex 0 → [ 1, 0 ]

it('should not reorder entries when the first entry does not require operationRequired', …
  // requiredIndex 1 → [ 0, 1 ]
```

They share a fixture that pins the entry sorting, so the assertion is about this reordering only and not about how the sorter happened to order things.

Both suggestions applied as well.

---

packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
rubensworks: Applied, and the expectation in the corresponding test was updated to match.

---

packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
rubensworks: Applied.

---

engines/query-sparql/README.md
```diff
@@ -17,6 +17,15 @@ It's main distinguishing features are the following:
...
 
+`SERVICE` clauses are not allowed to have a variable as target by default,
+as the queried data then determines what is dereferenced,
+which is problematic when queries originate from untrusted parties.
+If all queries originate from trusted parties, this can be enabled as follows:
+
+```bash
+$ comunica-sparql https://fragments.dbpedia.org/2016-04/en --serviceAllowVariableTargets "SELECT * WHERE { ... }"
+```
+
```
rubensworks: This is not necessary in this README. Should become part of the website later.

rubensworks: Removed — `engines/query-sparql/README.md` is back to exactly what it is on `master`.

I kept the shorter note in `packages/actor-query-operation-service/README.md`, since that is the actor's own documentation and mentions the context entry rather than the CLI flag. Happy to drop that one too if you would rather have it only on the website.

---

engines/query-sparql/test/QuerySparql-service-variable-test.ts
```diff
@@ -0,0 +1,195 @@
...
+const ENDPOINT1 = 'http://endpoint1.example.org/sparql';
+const ENDPOINT2 = 'http://endpoint2.example.org/sparql';
+
+/**
+ * System tests for SERVICE clauses of which the target is a variable.
+ *
+ * Such clauses can only be evaluated once another operation has bound their target,
+ * so they require the query planner to defer them to a bind-join.
+ * These tests place the clause in different positions relative to the operation binding its target,
+ * to make sure the planner handles that deferral no matter where the clause occurs.
```
rubensworks:
```suggestion
```

packages/actor-query-operation-service/lib/ActorQueryOperationService.ts
```diff
@@ -0,0 +1,187 @@
...
+    }
+    return passTestVoid();
+  }
+
+  public async runOperation(operation: Algebra.Service, context: IActionContext): Promise<IQueryOperationResult> {
+    // A target that is still a variable can not be queried yet.
+    // We return a placeholder output, which join actors use to determine that a bind-join is required,
+    // which materializes the target into an IRI before this operation is executed again.
+    if (operation.name.termType !== 'NamedNode') {
+      // Variable targets are determined by the queried data, so they are only allowed when explicitly enabled.
```
rubensworks:
```suggestion
```

packages/actor-query-operation-service/lib/ActorQueryOperationService.ts
```diff
@@ -0,0 +1,187 @@
...
+    if (termType !== 'NamedNode' && termType !== 'Variable') {
+      return failTest(`${this.name} can only query services by IRI or variable, while a ${termType} was given.`);
+    }
+    return passTestVoid();
+  }
+
+  public async runOperation(operation: Algebra.Service, context: IActionContext): Promise<IQueryOperationResult> {
+    // A target that is still a variable can not be queried yet.
+    // We return a placeholder output, which join actors use to determine that a bind-join is required,
+    // which materializes the target into an IRI before this operation is executed again.
```
rubensworks:
```suggestion
    // We replace variable targets with a placeholder that should be bound later in the bind-join.
```

packages/actor-query-operation-service/lib/ActorQueryOperationService.ts
```diff
@@ -0,0 +1,187 @@
...
+    if (operation.name.termType !== 'NamedNode') {
+      // Variable targets are determined by the queried data, so they are only allowed when explicitly enabled.
+      if (!context.get(KeysInitQuery.serviceAllowVariableTargets)) {
+        throw new Error(`SERVICE clauses with a variable target are not allowed by default, as they let the queried data determine what is dereferenced. Enable them via the serviceAllowVariableTargets context entry.`);
+      }
+      return this.unboundPlaceholder(operation, context);
+    }
+
+    try {
```
rubensworks, line 62:
```suggestion
    // Also handle IRIs, which can occur when variable targets are bound at a later stage.
    try {
```

packages/actor-query-operation-service/lib/ActorQueryOperationService.ts
```diff
@@ -0,0 +1,187 @@
...
+      if (!operation.silent) {
+        throw error;
+      }
+      this.logWarn(context, `An error occurred when executing a SERVICE clause: ${(<Error> error).message}`);
+      return await this.emptySolution(context);
+    }
+  }
+
+  /**
+   * Identify the SERVICE target as a query source.
```
rubensworks, line 81:
```suggestion
   * Identify the SERVICE target as a query source.
   * This logic is similar to that in ActorOptimizeQueryOperationAssignSourcesExhaustive#assignExhaustive
```

packages/actor-rdf-join-inner-multi-bind/lib/ActorRdfJoinMultiBind.ts
```diff
@@ -277,6 +278,17 @@ export class ActorRdfJoinMultiBind extends ActorRdfJoin<IActorRdfJoinMultiBindTe
...
     }
     const entriesSorted = entriesTest.get();
+
+    // Find the first entry that does not require operation pushdown.
+    const bindableIndex = entriesSorted
+      .findIndex(entry => !ActorRdfJoin.isOperationRequired(entry, entry.metadata));
+    if (bindableIndex < 0) {
+      return failTest(`Actor ${this.name} requires at least one entry of which the operation does not need to be pushed down`);
+    }
+    if (bindableIndex > 0) {
```
rubensworks, line 289:
```suggestion
    // This case could occur for variable SERVICE targets, which need to be bound later.
    if (bindableIndex > 0) {
```

packages/bus-rdf-join/lib/ActorRdfJoin.ts
```diff
@@ -104,6 +104,18 @@ TS
     this.pushesBindingsToSource = options.pushesBindingsToSource ?? false;
   }
 
+  /**
+   * Check if the operation of the given join entry must be pushed into the join,
+   * instead of using its output stream directly.
+   * Next to entries that explicitly request this, this holds for entries whose metadata requests it,
+   * such as SERVICE clauses of which the target is still an unbound variable.
```
rubensworks:
```suggestion
   * Check if the operation of the given join entry must be pushed into the join.
   * Next to entries that explicitly request this, this holds for entries whose metadata requests it,
   * such as SERVICE clauses of which the target is still an unbound variable.
```

packages/types/lib/IMetadata.ts
```diff
@@ -47,6 +47,14 @@ export interface IMetadata<OrderItemsType extends RDF.Variable | RDF.QuadTermNam
    * All available alternative orders.
    */
   availableOrders?: RDF.QueryOperationOrder<OrderItemsType>[];
+
+  /**
+   * If the operation that produced this stream must be pushed into a join
+   * instead of reading this stream directly, as `IJoinEntry.operationRequired` expresses for join entries.
+   * This is for example set for SERVICE clauses of which the target is still an unbound variable,
+   * which can only be evaluated once a bind-join has bound that target.
```
rubensworks:
```suggestion
   * If the operation that produced this stream must be pushed into a join.
   * This achieves the same as `IJoinEntry.operationRequired`, but can be expressed in metadata.
   * This is for example set for SERVICE clauses of which the target is still an unbound variable,
   * which can only be evaluated once a bind-join has bound that target.
```

PR #1818, opened 2026-09-14
Make the config backwards-compat test read the monorepo configs

rubensworks:
Follow-up to #1817. The `configtest` job did not catch the `sort-limit-pushdown` regression, and would not catch the next one either.

## Problem

`testEngine` installs a previously published engine from npm and then passes the monorepo's config file to `QueryEngineFactory` as `configPath`. That is the only thing it takes from the working tree. Components.js resolves the `ccqs:config/...` imports inside that file through the module state built from `mainModulePath`, which `QueryEngineFactoryBase` sets to the installed engine root, so every sub-config is read from `node_modules/@comunica/config-query-sparql`.

The test therefore reads one file from the working tree and the other 236 from npm. It validates that the top-level `import` list names files that exist in the published config package, and nothing else. Every actor added to a sub-config is invisible to it.

Demonstration on master, before this change. I added an actor referencing a package that exists nowhere to `optimize-query-operation/actors.json`:

```json
"https://linkedsoftwaredependencies.org/bundles/npm/@comunica/actor-totally-nonexistent-xyz/^5.0.0/components/context.jsonld"
```

and the check still passed. The error message when the check does fail names the installed file rather than the working tree one, which is the same symptom:

```
Error while parsing file ".../node_modules/@comunica/config-query-sparql/config/optimize-query-operation/actors.json"
```

## Change

Inject the whole config package from the monorepo into the temporary installation, so the imports resolve to the current configs while the engine keeps its published dependency set. That is the scenario that actually breaks downstream: an engine whose dependencies were fixed at its release, resolving `@comunica/config-query-sparql` to the newest in-range version.

This replaces the step that downgraded the config package to its lowest in-range version. That step no longer has a purpose, and it cost two npm calls per tested version, so the job also gets faster.

Two guards come with it:

* **Config files that are not in the published config package are skipped.** The test derives the engine versions to check from the version in a config file name, but that name is chosen when the config is written, while the file only ships in a later release. A config named after an already published version therefore gets paired with engines that never imported it, and under injection those engines would fail on actors they legitimately do not depend on. #1789 named a config that way, and #1817 tripped over the same thing before #1819 renamed its files, so this is worth guarding rather than leaving to naming discipline.
* **All installed copies are replaced.** `npm` can keep both a hoisted and a nested copy of the config package, and Components.js may pick up either.

The docstring is corrected too. It claimed the config file "will be injected", which was the intent but not what the code did.

## Verification

The check now behaves correctly in both directions, on CI rather than in a local simulation:

| master state | this check |
|---|---|
| before #1817, with the regression present | **fails**, naming the missing `sort-limit-pushdown` context on the unversioned `config-default.json` with `query-sparql@5.0.1` ([log](https://github.com/comunica/comunica/actions/runs/34877583440/job/104092868865)) |
| after #1817 and #1819 | **passes** |

The old version of the check passed in both cases, which is what let the regression reach 5.4.0.

The run log also shows the nested-copy handling doing real work rather than defending against a hypothetical:

```
Injecting the monorepo '@comunica/config-query-sparql' into /tmp/.../node_modules/@comunica/config-query-sparql
Injecting the monorepo '@comunica/config-query-sparql' into /tmp/.../node_modules/rdf-parse/node_modules/@comunica/config-query-sparql
```

Components.js resolved the imports through that second, nested copy. Replacing only the hoisted one would have left the check reading published configs again.

The file also typechecks under `--strict`.

## Possible follow-up

A cheaper static guard would cover the same ground without any npm installs: for every config file that already exists in the latest published `@comunica/config-query-sparql` of the current major, assert that its set of `@context` package entries and actor `@id`s is unchanged. New files are exempt, which is exactly right, and it encodes the rule in `config/README.md` directly. I left it out to keep this PR to one change, but it would fail in seconds rather than four minutes.

A second, narrower gap this PR does not address: during a release, `npm view` can list a version whose tarball is not yet fetchable, and the job hard-fails on the install. That happened on this PR while 5.4.1 was propagating. It predates this change.

https://claude.ai/code/session_01P6dNBhjusUFeMYeQNcVQb3


rubensworks:
`Config backwards-compat tests` is failing on `71551f5a`, but not because of this PR.

The job died during the npm install step, before any config was parsed:

```
Error: Command failed: npm install @comunica/query-sparql-rdfjs@5.4.1
npm error 404 Not Found - GET .../query-sparql-rdfjs-5.4.1.tgz - Not found
```

The 5.4.1 publish is mid-flight, so `npm view` lists 5.4.1 while some tarballs are not fetchable yet. Checking from outside CI right now:

| package @5.4.1 | metadata | tarball |
|---|---|---|
| `@comunica/config-query-sparql` | yes | 200 |
| `@comunica/query-sparql-file` | yes | 200 |
| `@comunica/query-sparql` | yes | missing |
| `@comunica/query-sparql-rdfjs` | yes | missing |
| `@comunica/actor-optimize-query-operation-sort-limit-pushdown` | not yet | missing |

`npm install @comunica/query-sparql-rdfjs@5.4.1` 404s outside CI too, so this reproduces off the runner and is not a property of this branch. The failing line, `npm install ${engineName}@${version}`, is unchanged by this PR and is on master as well, so master's run would hit the same thing at this moment.

I have not spent a re-run on it, since re-running while the tarball is still absent would just fail again. I will re-run once the release has fully propagated, and report the result here.

Worth noting that this is a real, if narrow, robustness gap in the check: during a release, `npm view` can list a version whose tarball is not yet installable, and the job hard-fails on it. It predates this PR and fixing it would widen the diff, so I am leaving it be unless you want it handled.

---

rubensworks:
Closing out my previous comment: the 5.4.1 tarballs are now fetchable, and both affected jobs pass on `f22b9306`.

| job | result |
|---|---|
| `Config backwards-compat tests` | success |
| `Binary tests` | success |

It was propagation lag rather than a stuck publish, and it has cleared. `npm install @comunica/query-sparql-rdfjs@5.4.1` now completes cleanly outside CI too. One thing I flagged earlier and want to correct: `@comunica/actor-optimize-query-operation-sort-limit-pushdown` has no 5.4.1, but that is normal, since lerna only republishes packages whose content changed, and the `^5.4.0` range resolves to the existing 5.4.0.

This also gives the change in this PR its before-and-after evidence, on real CI rather than my local simulation:

* against master **before** #1817, this rewritten check failed on the unversioned `config-default.json` with `query-sparql@5.0.1`, correctly reporting the missing `sort-limit-pushdown` context ([log](https://github.com/comunica/comunica/actions/runs/34877583440/job/104092868865))
* against master **after** #1817 and #1819, it passes

The old version of the check passed in both cases, which is what let the regression reach 5.4.0 in the first place.

The run log also confirms the nested-copy handling is doing real work, not defending against a hypothetical:

```
Injecting the monorepo '@comunica/config-query-sparql' into /tmp/.../node_modules/@comunica/config-query-sparql
Injecting the monorepo '@comunica/config-query-sparql' into /tmp/.../node_modules/rdf-parse/node_modules/@comunica/config-query-sparql
```

Components.js resolved the imports through that second, nested copy. Replacing only the hoisted one would have left the check reading published configs again.

---

rubensworks:
`Binary tests` is failing here, but it is failing on master too, with the same cause.

The `@comunica/query-sparql` binary test queries the live TPF endpoint `https://fragments.dbpedia.org/2015-10/en`, and that endpoint is currently returning a response no parser accepts:

```
Testing @comunica/query-sparql...
RDF parsing failed: none of the configured parsers were able to handle the media type
  ${action.handle.mediaType} for https://fragments.dbpedia.org/2015-10/en
    Unrecognized media type:      (x6)
Unexpected npm bin output
```

Identical output on master at `94a92cbc` ([log](https://github.com/comunica/comunica/actions/runs/35885118839/job/107263170578)) and on this PR ([log](https://github.com/comunica/comunica/actions/runs/35886572214/job/107268115698)). `Config backwards-compat tests` and `Tooling checks` are green on that same master run, so this is specific to `Binary tests` and to the external service, not to the base branch being broken generally.

Nothing in this PR can reach that path: the diff is one new test file plus the script it tests, and `Binary tests` runs published engines against live endpoints. There is no fix to port, since the trigger is outside the repository. I have not spent a re-run on it, because the failure reproduces on master rather than looking like a flake, and because the head it was reported on is already superseded.

Two things that may be worth a separate issue, both visible in that output and both pre-existing:

* the error interpolates nothing: `${action.handle.mediaType}` reaches the user literally, and the six `Unrecognized media type:` lines have an empty type. The template lives in the `@defaultNested` annotations of `ActorRdfParseFixedMediaTypes` and `ActorDereferenceRdf`, so either the substitution is not applied on this path or the media type is genuinely absent, and the message cannot tell you which.
* the binary tests depend on a third-party endpoint staying up and well-behaved, so an outage there turns CI red on every branch at once, as it has now.

---

PR #1820, opened 2026-09-14
Existence callback in context as canonical source for existence check, even when sources can answer

jitsedesmet:
Existence in a SPARQL engine has an optimization that the existence can be sent to sources directly. Since https://github.com/comunica/comunica/pull/1814 adds a context option to override the existing implementation to a callback outside the engine, we need to choose how this callback should work when provided to a SPARQL engine. With the changes here, an engine that receives an EXISTENCE callback will not ask the EXISTS on a source itself.

Depends on https://github.com/comunica/comunica/pull/1814


jitsedesmet:
@rubensworks I did not find how to make a stack PR within the web UI, so I made it a draft PR instead. https://github.com/jitsedesmet/comunica/pull/24 tracks the local changes. Up to you whether you want to land the base version (#1814 ) first or whether you want to grow that one by landing https://github.com/jitsedesmet/comunica/pull/24 first.

PR #1821, opened 2026-09-17
Report no request time when dereferencing local files

rubensworks:
Fixes the sporadic slowdown of BSBM-File queries 7 and 8 that shows up on the continuous performance results and was also seen in jitsedesmet/comunica#24.

## Symptom

Running `jbr run -c 0` in `performance/benchmark-bsbm-file` 12 times on this machine, 3 runs came out in a distinctly slower mode. The two modes are tightly clustered, which is what makes this a plan switch rather than noise:

| run | Q7 | Q8 |
| --- | --- | --- |
| 1 | **109.6 ms** | **203.1 ms** |
| 2 | 47.2 ms | 34.6 ms |
| 3 | 47.6 ms | 34.4 ms |
| 4 | 46.0 ms | 33.4 ms |
| 5 | 49.2 ms | 36.4 ms |
| 6 | **109.1 ms** | **205.1 ms** |
| 7 | 52.2 ms | 36.0 ms |
| 8 | 48.4 ms | 37.6 ms |
| 9 | **116.0 ms** | **210.5 ms** |
| 10-12 | 49-52 ms | 35-37 ms |

The mode is decided once per endpoint process and then holds for all 20 iterations of the query mix, so a whole benchmark run is either fast or slow. Other queries are unaffected beyond a few percent.

## Cause

`ActorDereferenceFile` timed its own `createReadStream` call and reported that as the source's `requestTime`:

```ts
const requestTimeStart = Date.now();
return {
  data: createReadStream(getPath(url)),
  // This should always be after the creation of the read stream
  requestTime: Date.now() - requestTimeStart,
```

`createReadStream` returns before the file is opened and reads nothing, so this measures only whether `Date.now()` happened to tick over during the call. It yields 0 on most runs and 1 on some.

From there:

1. `ActorRdfMetadataExtractRequestTime` puts that value in the source's metadata, and `ActorOptimizeQueryOperationQuerySourceIdentify` caches the identified source for the lifetime of the process, so the one measurement is reused by every query.
2. `LinkedRdfSourcesAsyncRdfIterator` accumulates the source-level metadata into the metadata of every pattern evaluated against the source, which overrides the `requestTime: 0` that `QuerySourceRdfJs` deliberately reports because the store is fully indexed in memory.
3. `ActorRdfJoin.getRequestInitialTimes` returns it unchanged (a file source has no `pageSize`), so the join cost models treat 1 ms as the cost of issuing one request.
4. `ActorRdfJoinOptionalBind` multiplies that by the cardinality of the stream it binds: `requestInitialTimes[0] + metadatas[0].cardinality.value * (... + requestInitialTimes[1] + ...)`.

Queries 7 and 8 are the ones that flip because their chained OPTIONALs over the ratings drive the left-hand cardinality estimate up to about 3.3e12, so 1 ms of noise adds ~3.3e13 to the bind cost with `ioWeight` 10.

The two debug logs below are the same OPTIONAL in Q8, from two processes that differ only in that measurement:

```
requestTime 0 -> 'optional-bind'
  costs: { 'optional-hash-def-nonblocking': 29482235771300, ..., 'optional-bind': 12042611009 }
  coefficients: { 'optional-bind': { iterations: 1204261100, ..., requestTime: 0 } }

requestTime 1 -> 'optional-hash-def-nonblocking'
  costs: { 'optional-hash-def-nonblocking': 29482235771310, ..., 'optional-bind': 32770082294641 }
  coefficients: { 'optional-bind': { iterations: 1204261100, ..., requestTime: 3275803968363 } }
```

Across 8 instrumented endpoint processes the correlation is exact: the one process that measured `requestTime: 1` is the one that chose `optional-hash-def-nonblocking` for 24 of its 36 optional joins, and the one that was slow. Hashing means scanning all ~7000 rating triples per optional instead of binding the handful of reviews that match the product.

## Fix

Report `requestTime: 0` for local files. There is no request latency for a local file, and the value that was being reported did not measure the read anyway.

## Verification

- 14 further benchmark runs after the change, all in the fast mode: Q7 46.5-57.8 ms, Q8 34.1-43.7 ms, no slow run. (Runs 5-9 sit at the top of those ranges because the test suite was running alongside them.)
- 16 runs of a smaller harness that starts the same endpoint and fires Q7/Q8 at it directly: 16/16 fast, against 2/10 slow before the change.
- Full test suite: 7043 passed, 19 skipped, 0 failed.

## Two related issues this does not address

- The one-off dereference time is folded into the per-pattern metadata of **any** document source, not just files. A source given as an HTTP URL to a plain RDF document is also indexed in memory by `ActorQuerySourceIdentifyHypermediaNone`, yet every pattern over it reports the document's fetch latency (tens to hundreds of ms) as its per-request time, which penalises bind joins hard. That is deterministic rather than sporadic, so it does not produce this symptom, but the `requestTime: 0` that `QuerySourceRdfJs` sets is being overridden there too.
- The cardinality estimate for Q8's chained OPTIONALs reaches ~3.3e12 for a query that returns 29 rows. The cost model is only this sensitive to a millisecond because that estimate is so far off.

Happy to follow up on either if you want them handled separately.

https://claude.ai/code/session_01Nby34CczE3cBDmtxNZAZSx

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nby34CczE3cBDmtxNZAZSx)_


packages/actor-dereference-file/lib/ActorDereferenceFile.ts
```diff
@@ -39,11 +39,13 @@ export class ActorDereferenceFile extends ActorDereference {
...
 
-    const requestTimeStart = Date.now();
     return {
       data: createReadStream(getPath(url)),
-      // This should always be after the creation of the read stream
-      requestTime: Date.now() - requestTimeStart,
+      // Local files have no request latency, and `createReadStream` returns before the file is opened,
+      // so timing it only measures clock granularity. The 0 or 1 it yields is then reported as the
+      // per-request time of every pattern over this source, where bind joins multiply it by the
+      // cardinality of the stream they bind, making a plan flip on a millisecond of measurement noise.
```
rubensworks:
```suggestion
      // Local files have no request latency
```

PR #1822, opened 2026-09-18
Keep source annotations of SERVICE clauses when materializing bindings

redreye:
Follow-up to #1815.

`materializeOperation` rebuilt `SERVICE` clauses without their `metadata`. The annotation is now kept, as is already done for patterns and paths.


rubensworks:
Good catch, thanks!

PR #1813, opened 2026-09-11
Add dereferenceFromNamed with configurable named-graph conflict resolution

Bavop:
Further work on [PR #1757](https://github.com/comunica/comunica/pull/1757)

`FROM NAMED` is now differentiated from `FROM`: the `FROM NAMED` IRI's data is placed into an actual named graph instead of being merged into the default graph. When the source itself already contains named graphs, an extra config option (`error` / `overwrite` / `merge`) controls how that conflict is resolved.

PR #1824, opened 2026-09-21
Add dereferenceFromNamed with configurable named-graph conflict resolution

rubensworks:
Continues from https://github.com/comunica/comunica/pull/1813 (by @Bavop), merged with the latest `master` and with the review comments there resolved.

`FROM NAMED` is now differentiated from `FROM`: the `FROM NAMED` IRI's data is placed into an actual named graph instead of being merged into the default graph. When the source itself already contains named graphs, a resolver (`error` / `preferNamed` / `keepSource`) controls how that conflict is resolved.

## Review comments resolved

- **Config versioning**: `actors-v5-3-0.json` is untouched again. The new actor lives in `optimize-query-operation/actors-v5-4-1.json`, with matching `config-default-v5-4-1.json`, `config-file-v5-4-1.json`, `config-file-browser-v5-4-1.json`, `config-rdfjs-v5-4-1.json` and `config-rdfjs-lite-v5-4-1.json`, and the four engines now import those.
- **Context keys**: `dereferenceFromNamed` and `dereferenceFromNamedConflictMode` moved from `KeysQueryOperation` to `KeysInitQuery`, with their context strings renamed to `@comunica/actor-init-query:*`. `sourceAsNamedGraph` stays on `KeysQueryOperation` and uses the suggested doc text; the conflict mode doc now states that it only applies when `dereferenceFromNamed` is enabled.
- **Mode naming**: the modes are now `error`, `preferNamed` and `keepSource`.
- **CLI**: added `--dereferenceFromNamedConflictMode <error|preferNamed|keepSource>`, which is turned into a constant resolver.
- **`ActorOptimizeQueryOperationSetSourcesFromDataset`**: the "are there dataset clauses" check moved into `test`, where it is expressed as `isKnownOperation(action.operation, Algebra.Types.FROM)`, so the actor no longer runs when it has nothing to do. `run` therefore no longer re-checks the operation type, and the README explains what the actor does.
- **`ActorQuerySourceIdentifyHypermediaNone`**: the quad stream is consumed with `on('data')` instead of `for await`, the per-graph mode cache is gone, the resolver receives the full quad, and the conflict error now tells the user which context entry and which modes unblock it.
- **`ActorQuerySourceIdentifyHypermedia`**: the change there was needed because a source's own context is only applied at query time, not while the source is being dereferenced, so `sourceAsNamedGraph` never reached `ActorQuerySourceIdentifyHypermediaNone`. Passing it as the first link's context fixes that; it is now set unconditionally, which drops the extra branch.

## Additional fix

While adding the `FROM NAMED` system tests, the source cache in `ActorOptimizeQueryOperationQuerySourceIdentify` turned out to key purely on the URL. A `FROM NAMED <x>` source therefore reused a cached, untagged source for `<x>` from an earlier query, silently returning no quads for `GRAPH ?g`. Sources tagged with `sourceAsNamedGraph` are no longer cached, since their data differs from the plain source with the same URL.

## Tests

`FROM NAMED` system tests were added in `QuerySparql-test.ts`: exposure as a named graph, absence from the default graph, the conflict error, and `keepSource`. Full suite, lint, depcheck and `test:config-compat` pass locally.

https://claude.ai/code/session_018oXa8URCLqBqcerfapsRkE

---
_Generated by [Claude Code](https://claude.ai/code/session_018oXa8URCLqBqcerfapsRkE)_


packages/actor-optimize-query-operation-set-sources-from-dataset/lib/ActorOptimizeQueryOperationSetSourcesFromDataset.ts
```diff
@@ -0,0 +1,58 @@
...
+      return failTest('This actor can only be used when dereferenceFromNamed is enabled.');
+    }
+    if (!isKnownOperation(action.operation, Algebra.Types.FROM)) {
+      return failTest('This actor can only be used on queries with a FROM or FROM NAMED clause.');
+    }
+    return passTestVoid();
+  }
+
+  public async run(action: IActionOptimizeQueryOperation): Promise<IActorOptimizeQueryOperationOutput> {
+    // The operation type has been asserted in test()
```
rubensworks:
```suggestion
```

rubensworks: Removed.

---

